PR #39, opened 2026-01-29
feat(group): add --allow-unmapped flag for processing fully unmapped reads in fgumi group

nh13:
## Summary

- Adds `--allow-unmapped` flag to `fgumi group` command to support processing fully unmapped reads (both R1 and R2 unmapped)
- Useful for ribosome display and other protocols that produce unmapped reads with UMI sequences
- When enabled, queryname-sorted input is also accepted (in addition to template-coordinate sorted)
- **NEW:** Adds parallel UMI assigner for Edit and Adjacency strategies when processing unmapped data

## Changes

- Add `--allow-unmapped` CLI flag to `GroupReadsByUmi` struct
- Add `allow_unmapped` field to `GroupFilterConfig`  
- Modify `filter_template()` to conditionally allow unmapped templates when flag is set
- Relax sort order validation to accept queryname-sorted input when `--allow-unmapped` is enabled
- Add unit tests for `filter_template` with unmapped templates
- Add integration tests for grouping unmapped reads with Identity and Adjacency strategies
- Add threading mode tests to verify unmapped reads work correctly in all pipeline configurations
- **NEW:** Add `parallel_assigner` module with `ParallelEditAssigner` and `ParallelAdjacencyAssigner`
- **NEW:** Add `BitEnc::with_base_at()` and `base_at()` methods for neighbor UMI generation
- **NEW:** Add runtime warnings when `--allow-unmapped` is used

## Parallel UMI Assigner

When `--allow-unmapped` is enabled, all unmapped reads form a single position group which can contain millions of unique UMIs. The standard O(U²) algorithms become bottlenecks.

The new parallel assigner uses:
- **Parallel edge discovery**: Generate all Hamming-1 neighbors and lookup in hash map
- **Thread-safe union-find**: Concurrent connected component discovery
- **Complexity**: O(U×L/P) where U=unique UMIs, L=UMI length, P=threads

This provides 10-100x speedup for large unmapped datasets.

## Behavior

When `--allow-unmapped` is enabled:
- Templates with both reads unmapped are no longer filtered out
- Queryname-sorted input is accepted
- Unmapped reads are grouped by UMI only within each library/cell barcode
- All unmapped reads form a single position group (since they have no genomic coordinates)
- Edit and Adjacency strategies use parallel assigner automatically
- Identity strategy unchanged (already O(U) with hash lookups)

## Important Notes

[...]

PR #118, opened 2026-02-19
feat(group): add --parallel-group-min-templates for size-based parallel UMI assignment

nh13:
## Summary

- Add `--parallel-group-min-templates N` CLI option that enables the parallel UMI assigner (Edit, Adjacency, Paired) for position groups with at least N templates
- Decouple parallel assigner from `--allow-unmapped` so amplicon and other workflows with large mapped position groups can opt in
- Extract `create_umi_assigner` helper to eliminate duplicated assigner selection logic between pipeline and single-threaded code paths

Stacked on #39 (`feat/allow-unmapped-reads`).

## Details

When `--parallel-group-min-templates` is set, groups meeting the threshold use rayon-based parallel edge discovery. `--allow-unmapped` continues to unconditionally use the parallel path (all unmapped reads form one giant group).

The CLI docs include a warning about thread over-subscription: the parallel assigner uses rayon's thread pool, which is separate from the pipeline's worker threads (`--threads`). For amplicon data (few large groups), pipeline threads are mostly idle so over-subscription is minimal.

## Test plan

- [x] New test: `test_parallel_group_min_templates_activates_for_mapped_data` — verifies parallel path activates with `Some(1)` threshold
- [x] New test: `test_parallel_group_min_templates_none_uses_sequential` — verifies sequential path when threshold is `None`
- [x] All existing `allow_unmapped` tests pass unchanged
- [x] Full CI: `cargo ci-test && cargo ci-fmt && cargo ci-lint` (1840 tests passed)

### DIFF
--- a/src/commands/group.rs
+++ b/src/commands/group.rs
@@ -24,8 +24,7 @@ use fgumi_lib::read_info::{LibraryIndex, compute_group_key};
 use fgumi_lib::sam::{is_sorted, is_template_coordinate_sorted, unclipped_five_prime_position};
 use fgumi_lib::template::{MoleculeId, Template};
 use fgumi_lib::umi::parallel_assigner::{
-    ParallelAdjacencyAssigner, ParallelEditAssigner, ParallelIdentityAssigner,
-    ParallelPairedAssigner,
+    ParallelAdjacencyAssigner, ParallelEditAssigner, ParallelPairedAssigner,
 };
 use fgumi_lib::umi::{UmiValidation, validate_umi};
 use fgumi_lib::unified_pipeline::DecodedRecord;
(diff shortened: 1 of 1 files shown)


nh13:
Superseded by #371, which adopts the same design (decouple from --allow-unmapped, extract create_umi_assigner, over-subscription warning) but replaces the single user-supplied integer threshold with . The new `auto` mode uses per-strategy thresholds (Identity: never, Edit: 1536, Adjacency: 3072, Paired: 128) justified by a new microbench (`benches/umi_assigner_threshold.rs`). Bench data showed a flat threshold was the wrong shape for three of four strategies.

nh13:
(Correction to my previous comment — shell quoting ate the flag name.)

Superseded by #371, which keeps the same design from this PR (decouple from `--allow-unmapped`, extract `create_umi_assigner`, over-subscription warning) but takes the flag value as `N|auto` instead of a bare integer.

The new `auto` mode uses per-strategy thresholds (Identity: `usize::MAX`, Edit: 1536, Adjacency: 3072, Paired: 128) justified by a new microbench (`benches/umi_assigner_threshold.rs`). Bench data showed a flat threshold was the wrong shape for three of four strategies — Paired wins parallel at ~128 templates while Edit/Adjacency need 1.5k–3k and Identity never.

PR #198, opened 2026-03-28
feat: add --includelist option to simulate fastq-reads

nh13:
## Summary

- Adds `--includelist` / `-i` option to `fgumi simulate fastq-reads` that samples UMIs from a whitelist file instead of generating them randomly
- UMI length is inferred from the includelist, overriding `--umi-length` (with a warning if they differ)
- Validates includelist format: one UMI per line, all same length, ACGT-only, blank lines skipped, lowercase uppercased

This enables generating synthetic FASTQ data that exercises the full `correct → extract → ... → consensus` pipeline in fgumi-benchmarks, since reads contain UMIs from a known whitelist that `fgumi correct` can match against.

## Test plan

- [x] 7 unit tests for `load_includelist` (valid, lowercase, blank lines, invalid bases, mismatched lengths, empty file)
- [x] 1 integration test verifying molecule UMIs are sampled exclusively from the includelist
- [x] Manual smoke test: generated 100 molecules with 5-UMI whitelist, verified all truth UMIs are whitelist combinations
- [x] Pre-commit hooks pass (ci-fmt, ci-lint)
- [ ] CI

### DIFF
--- a/src/commands/simulate/fastq_reads.rs
+++ b/src/commands/simulate/fastq_reads.rs
@@ -10,13 +10,13 @@ use crossbeam_channel::bounded;
 use fgumi_dna::dna::complement_base;
 use fgumi_lib::progress::ProgressTracker;
 use fgumi_lib::simulate::{FastqWriter, create_rng};
-use log::info;
+use log::{info, warn};
 use noodles::fasta;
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

PR #219, opened 2026-04-02
perf(sort): reuse buffers in merge phase via producer-consumer buffer pool

nh13:
## Summary

- Add a **buffer return channel** to `GenericKeyedChunkReader` so the consumer
  can return spent `Vec<u8>` buffers to the producer thread for reuse. The
  producer calls `try_recv` (non-blocking) to grab a recycled buffer before
  each record read, falling back to a fresh allocation only when the pool is
  empty. This eliminates per-record heap allocation in the merge phase once the
  pipeline warms up.
- Changed `next_record` on all `ChunkSource` wrappers to accept a `&mut Vec<u8>`
  buffer parameter. Merge loops now write directly from each heap entry's record
  buffer and refill it in-place.
- Removed the intermediate `output_buffer: Vec<Vec<u8>>` batching layer that
  previously collected 2048 owned `Vec`s before flushing.
- Updated all four merge sites: consolidation, template-coordinate, generic, and
  indexed merge.

## Test plan

- [x] `cargo ci-fmt` passes
- [x] `cargo ci-lint` passes
- [x] `cargo nextest run --no-fail-fast` — all 1735 tests pass
- [ ] Benchmark on a representative BAM to confirm merge-phase allocation reduction

### DIFF
--- a/src/lib/sort/raw.rs
+++ b/src/lib/sort/raw.rs
@@ -289,6 +289,9 @@ impl<K: RawSortKey> GenericKeyedChunkWriter<K> {
 /// Auto-detects BGZF compression via magic bytes.
 pub struct GenericKeyedChunkReader<K: RawSortKey + 'static> {
     receiver: Receiver<Option<(K, Vec<u8>)>>,
+    /// Return channel for empty buffers — the consumer sends its old buffer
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

PR #247, opened 2026-04-08
perf(sort): redesign phase 2 work stealing and bound rayon to --threads

nh13:
## Summary

Fixes two independent problems hurting the N+2 sort worker-pool path on large inputs, plus a post-review cleanup sweep.

### Phase 2 redesign: per-file work-stealing state

Phase 2 regressed badly on WES (1kg-wes-HG00100.bam, 13 GB, 219M records): 450s+ wall and ~42 GB peak RSS under `--threads 8 --max-memory 768M`, driven by an over-coupled producer/consumer with global queues that deadlocked under backpressure and left workers idle while OOMing.

Replace it with per-file state:

```rust
struct Phase2FileState {
    reader: Mutex<Phase2Reader>,           // raw BGZF block reads
    reader_eof: AtomicBool,                // hot-path fast check
    raw_blocks: Mutex<VecDeque<...>>,      // bounded FIFO
    decompressed: Mutex<ReorderBuffer>,    // in-order decompressed
    decomp_in_flight: AtomicUsize,         // race-safe drain
}
```

Workers scan files with `try_lock` everywhere and steal work across spill files. Lock order is strict: `reader -> raw_blocks -> decompressed`.

- **Deadlock-free admission**: a popped raw block is always admitted when it is the gap-filler the reorder buffer is waiting for, even if the decompressed cap is full. Prevents head-of-line deadlock where every worker holds a stale raw block while the buffer waits on a missing serial.
- **Race fix**: `is_drained()` must observe `decomp_in_flight > 0` to avoid exiting Phase 2 between a worker popping a raw block and inserting the decompressed bytes.
- **FIFO bug fix**: serial assignment and the matching push into `raw_blocks` happen under both the reader and raw_blocks locks held simultaneously. Releasing the reader lock first allowed another worker to interleave a higher serial ahead of the lower one and break the gap-filler invariant.
- **Hot-path fast check**: `is_drained()` fast-paths on a lock-free `reader_eof` atomic; the common case (reader not at EOF) exits after a single atomic load, versus three mutex acquisitions per decompressed block.

### Rayon oversubscription

Rayon's global pool defaults to `num_cpus::get()`, which silently violated `--threads` on machines with more physical cores than the requested thread count. Every `par_sort` / `par_sort_into_chunks` / `par_sort_unstable_by` / `par_chunks_mut` call in the raw sort path now runs under a `rayon::ThreadPool` built with `num_threads = self.threads`, via `rayon_pool.install(|| ...)`. `rayon::current_num_threads()` inside `parallel_radix_sort_*` now returns the requested thread count, so chunk sizing is consistent with the worker-pool sizing.

Oversubscription with the `SortWorkerPool` is not introduced: every call site is preceded by `drain_pending_spill`, which joins the prior chunk's I/O thread, so the sort worker pool is provably idle when rayon fans out.

### Additional fixes

- Use `par_sort` in the three in-memory sort fallbacks (coordinate, coordinate-index, template-coordinate). The single-threaded `buffer.sort()` serialized the sort step onto one core even with `--threads` set high when the entire input fit in memory.
- In-memory split path replaces the O(n²) `drain(..chunk_size)` loop with O(n) `split_off` from the tail. The old loop memmoved the tail on every drain and became a measurable cost once `entries` grew into the millions under tight `--max-memory` budgets.
- Drop `Phase2FileState::source_id` (always equal to the file's index in the shared vector), `SourceParserState::eof`, and the unused `_worker` parameter on `is_phase_complete` — all written or threaded through but never read.
- Refresh stale module, struct, and function docs to describe the per-file work-stealing Phase 2 model.

### Result

WES (`1kg-wes-HG00100.bam`, `--threads 8 --max-memory 4g`):

| | wall | peak RSS | sort violations |
|---|---|---|---|
| before (5df7e17) | 450s+ | ~42 GB | OOM-prone |
| this branch     | 139 s | 37.6 GB | 0 |

## Test plan
- [x] \`cargo ci-fmt\`
- [x] \`cargo ci-lint\`
- [x] \`cargo ci-test\` (2351 tests pass)
- [x] WES end-to-end sort with coordinate order, verified sorted output
- [ ] Reviewer spot-check of Phase 2 lock order and admission rule

### DIFF
--- a/src/lib/sort/inline_buffer.rs
+++ b/src/lib/sort/inline_buffer.rs
@@ -10,7 +10,7 @@
 //! - ~24 bytes overhead per record vs ~110 bytes
 
 use crate::sort::bam_fields;
-use crate::sort::keys::{RawSortKey, SortContext};
+use crate::sort::keys::{RawCoordinateKey, RawSortKey, SortContext};
 use crate::sort::radix::bytes_needed_u64;
 use crate::sort::segmented_buf::SegmentedBuf;
 use std::cmp::Ordering;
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai review

PR #264, opened 2026-04-14
fix(extract): take last `:`-separated field as UMI from read name

nh13:
## Summary

`fgumi extract --extract-umis-from-read-names` previously hard-coded the **8th** `:`-separated field of the read name as the UMI (`parts[7]`). This silently extracts the wrong value for FASTQs whose read names contain 9+ fields — for example, demultiplexers that fold the sample index into the colon-separated portion of the read name and then append the UMI:

```
@LH00620:304:23LLHJLT4:7:1101:2578:1070:CAATCTATAA+rTTACATAGTT:TCNGCG
```

In that shape the **8th** field is the sample index (`CAATCTATAA+rTTACATAGTT`) and the actual UMI is the **9th and last** field (`TCNGCG`). The previous code wrote the sample index to the `RX` tag without erroring.

## Change

Match fgbio's [`Umis.extractUmisFromReadName`](https://github.com/fulcrumgenomics/fgbio/blob/main/src/main/scala/com/fulcrumgenomics/umi/Umis.scala) (used by `CopyUmiFromReadName`) — always return the **last** `:`-separated field as the UMI when the read name has at least 8 fields.

- Names with **fewer than 8 fields** still produce `None` (no UMI), preserving existing fgumi behavior.
- The standard **8-field** Illumina shape (`@<instr>:<run>:<flowcell>:<lane>:<tile>:<x>:<y>:<UMI>`) is unchanged — the 8th field and the last field are the same.
- Names with **9+ fields** now correctly extract the trailing UMI rather than the field at position 8.
- `+` → `-` UMI normalization is preserved.

The doc comment and `--help` text are updated to describe the behavior and cite the Illumina FASTQ spec.

## Test plan

- [x] New unit tests for `Extract::extract_read_name_and_umi` covering 7-, 8-, 9-field names, `+`→`-` normalization, space-comment stripping, and the `extract=false` early return.
- [x] New end-to-end test (`test_extract_umis_from_read_names_nine_fields_via_execute`) reproducing the real CODEC validation FASTQ shape and asserting the `RX` tag is the trailing 6 bp duplex UMI rather than the sample index.
- [x] All existing extract tests continue to pass (8-field names produce identical RX tags before/after).
- [x] `cargo ci-fmt`, `cargo ci-lint`, full `cargo ci-test` (2501 tests) all green.

### DIFF
--- a/src/lib/commands/extract.rs
+++ b/src/lib/commands/extract.rs
@@ -368,10 +368,15 @@ values.
 
 UMIs may be extracted from the read sequences, the read names, or both. If
 `--extract-umis-from-read-names` is specified, any UMIs present in the read names are extracted;
-read names are expected to be `:` -separated with any UMIs present in the 8th field. If UMI
-segments are present in the read structures those will also be extracted. If UMIs are present in
-both, the final UMIs are constructed by first taking the UMIs from the read names, then adding a
-hyphen, then the UMIs extracted from the reads.
+read names are expected to be `:`-separated and the UMI is taken from the **last** field. At
+least 8 fields must be present — the standard Illumina shape
+`@<instrument>:<run>:<flowcell>:<lane>:<tile>:<x>:<y>:<UMI>`. Names with 9+ fields (e.g.
+produced by demultiplexers that fold the sample index into the colon-separated portion) are
+also handled, with the UMI still coming from the last field. Any `+` characters in the
+extracted UMI are normalized to `-`. If UMI segments are present in the read structures those
+will also be extracted. If UMIs are present in both, the final UMIs are constructed by first
+taking the UMIs from the read names, then adding a hyphen, then the UMIs extracted from the
+reads.
 "#
 )]
 #[command(verbatim_doc_comment)]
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

SPPearce:
@nh13 , do you know if this correctly get the UMI from reads like:
```
@LH00182:10:23LJWKLT4:1:1101:3281:1028:CGTGCGCC+rCGCTCGAG 1:N:0:AGATGAGCCA+GAACACGAGC
GTNATCTCCAAGTTATATAACAGGTGTCTAGATATCTTACCATCTGAACAGATGCCAATACATAAATTTGTTTACAGGTATCTCCTTTCTCAGAGACTAAGAGTTTAGCGGGTGGCATCTGGGTGAGTAAAAATACCTTCTGTTATTGTTA
+
II#9IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII9IIIIIIIIIIIIIIIIIII9IIIIIIIIIIIIIIIIIIIII9IIIIIIIIIIIIIIII9IIII9IIIIIIIII*
@LH00182:10:23LJWKLT4:1:1101:3977:1028:GGTTGGTA+rTAAGAAAT 1:N:0:AGATGAGCCA+GAACACGAGC
AANGCAACATCCACGGAGTGCACATCCTGTTTCTTATTATTTGTGATAGAATCCTTGCAATATCGTCTTAAAGAATGATTCTGGACTGAATCTAACCACTTCCCCCTTGTTGAGCTTTTAGATTAGATTACCTACAATTGTTTGGTATTAG
+
II#9IIIIIIIIIIII*IIIIIIIIIIIIIIIIIIIIII9IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII9IIIIII9IIIIIIIIIIIIIIIII
```
This is what bclconvert gives me from the NovaSeqX, where the `1:N:0:AGATGAGCCA+GAACACGAGC` is the sample barcode.

nh13:
@SPPearce yes — that's the standard 8-field Illumina layout, so the UMI is the 8th `:`-separated field and PR #264's change (last field vs. `parts[7]`) is a no-op for these reads. The fix only matters for demultiplexers that append additional `:`-fields after the UMI (e.g. some CODEC pipelines).

Confirmed by running `fgumi extract --extract-umis-from-read-names` on your exact records:

```
LH00182:...:1028:CGTGCGCC+rCGCTCGAG    RX:Z:CGTGCGCC-rCGCTCGAG
LH00182:...:1028:GGTTGGTA+rTAAGAAAT    RX:Z:GGTTGGTA-rTAAGAAAT
```

Notes:

- The `1:N:0:AGATGAGCCA+GAACACGAGC` comment after the space is stripped before parsing, so the sample index never leaks into the UMI.
- `+` → `-` is applied (matches fgbio's `Umis.extractUmisFromReadName`).
- The lowercase `r` prefix on the second UMI (BCL Convert's marker for "this UMI was reverse-complemented during demux") is **preserved verbatim** as part of the UMI string. fgumi doesn't strip it or undo the reverse-complement — same as fgbio. Worth confirming that's what you want downstream: a paired UMI assigner will still match top/bottom strands correctly because both mates carry the same prefix, but if any tool does naive string equality across reads where one mate has an `r`-prefixed half and the other doesn't, you'd want to normalize before extract.

SPPearce:
Good to know. I don't ever have paired UMIs, so that's useful to know but not going to trip me up :)

PR #170, opened 2026-03-15
feat(zipper): add --restore-unconverted-bases flag for EM-seq

nh13:
## Summary
- Adds `--restore-unconverted-bases` flag to `fgumi zipper` for EM-seq post-alignment base restoration
- After bwameth alignment (C→T converted reads), restores bases to unconverted state at reference positions: T→C for top-strand (YD=f), A→G for bottom-strand (YD=r)
- Ensures consensus calling sees original bases so methylation is determined by the consensus caller
- Defines `YD_FORWARD`/`YD_REVERSE` constants (was inline byte literals)
- Uses `reference_length()` utility for ref span computation (was manual CIGAR walk)

## Stack
3/4 — EM-seq support (depends on #169)

## Test plan
- [x] Tests for top-strand and bottom-strand base restoration
- [x] Tests for indels, unmapped reads, missing YD tag, already-unconverted bases
- [x] Existing zipper tests pass unchanged
- [x] `cargo ci-test && cargo ci-fmt && cargo ci-lint` passes

PR #271, opened 2026-04-14
perf(zipper): speed up --restore-unconverted-bases hot path

nh13:
## Summary

- `ReferenceReader::fetch_slice` returns `&[u8]` instead of `Vec<u8>`; eliminates ~2 GB of `Vec` allocation churn over a 13M-template run. `fetch` becomes a thin wrapper for callers that need owned bytes.
- Borrow the contig name from the BAM header instead of `String`-allocating per record.
- Early-exit `restore_unconverted_bases_in_record` via `memchr::memchr2` when the aligned reference span has no candidate base — skips both the SEQ allocation and the CIGAR walk.
- Precompute the lowercase variants of the target/converted bases once per record so the inner loop avoids `to_ascii_uppercase()` per byte.
- Defer the SEQ `Vec<u8>` allocation until the first base actually needs to change. Reads aligned to ref-C positions but already carrying C in the read (common for highly-methylated CpGs in EM-Seq) now return without allocating SEQ at all.

## Why

`fgumi zipper --restore-unconverted-bases` (added in #170) drops the streaming raw-byte merge path used by TAPs/non-methyl in favour of a `RecordBuf` round-trip so it can fetch reference bases per record. Profiling a real 26.8M-record EM-Seq run showed ~46K records/s — about 10× slower than `bwa mem | bwameth.py` can produce. With default `bwa -K 150000000`, bwameth.py accumulates a chunk's worth of `Bam` objects in Python while waiting on zipper, eventually getting OOM-killed mid-pipe.

The five changes here all sit on the per-record hot path and stack cleanly. None of them change observable output for any record that has bases to restore — the existing 9 restore tests still pass byte-for-byte. The new `no_target_in_span` test locks in the early-exit, and `test_restore_unconverted_bases_preserves_already_unconverted` already exercises the deferred-SEQ-allocation path (it asserts the no-change case, which now also incurs no allocation).

End-to-end speedup will be measured in the next fgumi-benchmarks methylation run; the alloc/copy reductions are mechanical and the early-exit + deferred-allocation are unconditional skips for any read that doesn't need work.

## Test plan

- [x] New unit test `test_fetch_slice_returns_borrowed_bytes` covers `&[u8]` return, byte-equality with `fetch`, and bounds errors.
- [x] New unit test `test_restore_unconverted_bases_no_target_in_span` covers the `memchr2` early-exit fast path.
- [x] Existing 9 `test_restore_unconverted_bases_*` tests pass byte-for-byte (no behavioral regressions).
- [x] Full workspace `cargo ci-test`: 2495 passed, 19 skipped.
- [x] `cargo ci-fmt` and `cargo ci-lint` (clippy pedantic + `-D warnings`) clean.

### DIFF
--- a/crates/fgumi-sam/src/tag.rs
+++ b/crates/fgumi-sam/src/tag.rs
@@ -100,6 +100,8 @@ impl SamTag {
     pub const RG: SamTag = SamTag::new(b'R', b'G');
     /// Number of mismatches (edit distance) to the reference (SAM spec).
     pub const NM: SamTag = SamTag::new(b'N', b'M');
+    /// String encoding of the reference-relative alignment (SAM spec).
+    pub const MD: SamTag = SamTag::new(b'M', b'D');
     /// Mate's mapping quality (SAM spec).
     pub const MQ: SamTag = SamTag::new(b'M', b'Q');
     /// Mate's CIGAR string (SAM spec).
--- a/src/lib/commands/zipper.rs
+++ b/src/lib/commands/zipper.rs
@@ -1058,7 +1058,9 @@ fn restore_unconverted_bases_in_record(
         .reference_sequences()
         .get_index(ref_id)
         .context("reference sequence ID not found in header")?;
-    let ref_name = ref_name.to_string();
+    // Borrow the contig name as &str rather than allocating a fresh String per
+    // record; `fetch_slice` only needs &str for its HashMap lookup.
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #273, opened 2026-04-14
fix(group): emit consecutive MI integers 0..N-1

### LINKED ISSUE #269: `group` assigns non-consecutive MI integers; downstream consensus read names diverge from fgbio
`fgumi group` assigns MI family IDs that span a noticeably larger integer
range than the number of distinct families produced (e.g. on a 1 M-pair test:
751,096 distinct families, max MI value 920,444 — 22% wasted IDs), while
`fgbio GroupReadsByUmi` produces consecutive IDs (max == count − 1).
Groupings are correctness-equivalent (`fgumi compare bams --mode grouping`
reports 0 mismatches) but the gap pattern propagates to `fgumi codec`, whose
consensus read names inherit the MI value (`smoke:N`). Cross-tool
`compare bams` then fails on read-name mismatches even when the underlying
biology is identical.

Suggested fix: emit consecutive MI integers (0..N-1) by either (a) renumbering
once at the end of `group`, or (b) using a contention-free per-output counter
rather than per-thread block allocation.

nh13:
## Summary

- `fgumi group` and `fgumi dedup` reserved MI counter blocks sized by template count, but each position group only emits one MI per distinct UMI family (and paired A/B share a numeric id), so the block was almost always larger than needed. Unused slots became gaps in the global MI space (~22% wasted IDs on a 1 M-pair test) and propagated to downstream consensus read names, breaking cross-tool comparison with fgbio.
- Track a `distinct_mi_count` on `ProcessedPositionGroup` / `ProcessedDedupGroup` computed as `max(MoleculeId::id()) + 1` across assigned templates, and advance the global MI counter by that value instead of `templates.len()`. Fix applies to both the unified parallel pipeline and the single-threaded streaming path.
- Adds two integration tests (single-threaded and multi-threaded pipeline paths) asserting `max(MI) == distinct_count - 1`, matching fgbio's `GroupReadsByUmi`.

Fixes #269

## Test plan

- [x] New failing test `test_group_command_assigns_consecutive_mi_values` reproduces the bug: MIs `{0, 4, 5, 9}`, max 9 vs expected 3.
- [x] Same test passes after the fix.
- [x] New `test_group_command_assigns_consecutive_mi_values_multi_threaded` covers the parallel pipeline path with `--threads 4` and 20 position groups.
- [x] `cargo ci-test`: 2495 passed, 19 skipped.
- [x] `cargo ci-fmt` clean.
- [x] `cargo ci-lint` clean.

## Scope note

This removes the **gaps** in the MI integer space (the literal issue). It does not guarantee MI values appear in monotonically increasing order in the output BAM — parallel serialize workers may reserve their (now correctly-sized) blocks in any interleaved order. If byte-for-byte output-order monotonicity to match fgbio is also required, that is a follow-up (option (a) from the issue: end-of-pipeline renumber, or serial MI assignment during output write).

### DIFF
--- a/src/lib/commands/dedup.rs
+++ b/src/lib/commands/dedup.rs
@@ -177,6 +177,10 @@ pub struct ProcessedDedupGroup {
     pub dedup_metrics: DedupMetrics,
     /// Total input records processed (for progress tracking).
     pub input_record_count: u64,
+    /// Number of distinct numeric molecule IDs assigned in this group. See
+    /// [`fgumi_lib::grouper::ProcessedPositionGroup::distinct_mi_count`] for
+    /// rationale; ensures emitted MI tag integers are consecutive 0..N-1.
+    pub distinct_mi_count: u64,
 }
 
 impl BatchWeight for ProcessedDedupGroup {
@@ -883,6 +887,7 @@ fn process_position_group(
             family_sizes: AHashMap::new(),
             dedup_metrics,
             input_record_count,
+            distinct_mi_count: 0,
         });
     }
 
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #272, opened 2026-04-14
Complete the raw-byte BAM API in fgumi-raw-bam


nh13:
## Summary

`crates/fgumi-raw-bam` lets us mutate BAM records without inflating to `RecordBuf`, which keeps `fgumi zipper`'s streaming raw-byte path on the fast track for non-methyl workflows. The BAM-spec coverage is mostly there but uneven — some fields have full get/edit/append/remove parity, others have only accessors. Closing the gaps would let more zipper code paths (and other commands) stay on the raw-byte path instead of round-tripping through `RecordBuf`.

This issue catalogues the current support matrix and the missing pieces.

## Current support matrix

Based on `crates/fgumi-raw-bam/src/{fields,sequence,cigar,tags}.rs`. "✓" = supported on raw bytes; "✗" = absent; "n/a" = not applicable.

| BAM field | Get/Access | Edit/Set in-place | Append/Resize | Remove |
|---|---|---|---|---|
| `block_size` (u32) | ✓ implicit (`len()`) | ✓ implicit on resize | n/a | n/a |
| `refID` (i32) | ✓ `ref_id()` | ✓ `set_ref_id()` | n/a | n/a |
| `pos` (i32) | ✓ `pos()` | ✓ `set_pos()` | n/a | n/a |
| `l_read_name` (u8) | ✓ `l_read_name()` | ✗ length-changing | ✗ | n/a |
| `mapq` (u8) | ✓ `mapq()` | ✓ `set_mapq()` | n/a | n/a |
| `bin` (u16) | ✗ no accessor | ✗ no setter | n/a | n/a |
| `n_cigar_op` (u16) | ✓ `n_cigar_op()` | ✗ length-changing | ✗ | n/a |
| `flag` (u16) | ✓ `flags()` + `Flags::from_flag` | ✓ `set_flags()` | n/a | n/a |
| `l_seq` (u32) | ✓ `l_seq()` | ✗ length-changing | ✗ | n/a |
| `next_refID` (i32) | ✓ `mate_ref_id()` | ✓ `set_mate_ref_id()` | n/a | n/a |
| `next_pos` (i32) | ✓ `mate_pos()` | ✓ `set_mate_pos()` | n/a | n/a |
| `tlen` (i32) | ✓ `template_length()` | ✓ `set_template_length()` | n/a | n/a |
| `read_name` (NUL-term) | ✓ `read_name()` slice | ✗ length-changing | ✗ | n/a |
[...]

PR #274, opened 2026-04-14
perf(zipper): restore unconverted bases on raw BAM bytes

nh13:
## Summary

- New `restore_unconverted_bases_in_raw_record` operates on packed BAM bytes via `fgumi-raw-bam` primitives — `flags` / `pos` / `ref_id` / `get_cigar_ops` / `find_string_tag_in_record` (`YD`) / `seq_offset` / `get_base` / `set_base` / `remove_tag` (`NM`, `MD`). SEQ is rewritten in place at the 4-bit nibble level; record length is unchanged.
- New `restore_unconverted_bases_in_raw_template` walks `template.raw_records` and applies the per-record function.
- `process_raw` now uses a uniform `convert_template_to_raw` → `merge_raw` → `[restore_raw if methyl]` dispatch — methylation just adds one step to the same fast path TAPs uses, instead of branching into a separate `RecordBuf` round-trip.
- Carries the `fetch_slice`, lowercase precompute, and `memchr2` early-exit wins from #271 into the new function.
- The `RecordBuf` reference impl (`restore_unconverted_bases_in_template` / `_in_record` + `YD_TAG`) is kept under `cfg(test)` as an oracle for the existing tests.

## Why

`fgumi zipper --restore-unconverted-bases` (added in #170) was the only zipper code path that bypassed the streaming raw-byte merge in favour of a `RecordBuf` round-trip — full BAM parse, per-field allocation, mutation, then re-encode every byte. That round-trip dominates the per-record cost on the methylation path.

This PR is the structural fix — it removes the round-trip entirely so the methylation path runs at the same per-record cost as TAPs (which is already on this path). The capability inventory in #272 confirmed the `fgumi-raw-bam` API already exposes everything needed (no length-changing edits required for SEQ restoration).

## Test plan

- [x] 6 new `test_restore_unconverted_bases_in_raw_record_*` tests cover top strand, bottom strand, reverse top strand, no-`YD` skip, unmapped skip, and `NM` / `MD` stale-tag clearing — equivalent to the existing `RecordBuf` tests.
- [x] All 9 existing `test_restore_unconverted_bases_*` tests still pass against the `cfg(test)` `RecordBuf` reference impl, providing parity coverage.
- [x] All other zipper tests (73 total in `commands::zipper`) pass.
- [x] Full workspace `cargo ci-test` passes.
- [x] `cargo ci-fmt` and `cargo ci-lint` (clippy pedantic + `-D warnings`) clean.

## Stacking

Based on #271 to pick up `ReferenceReader::fetch_slice`. GitHub will auto-rebase the base from `nh/zipper-restore-perf` to `main` once #271 lands.

### DIFF
--- a/crates/fgumi-sam/src/tag.rs
+++ b/crates/fgumi-sam/src/tag.rs
@@ -116,6 +116,10 @@ impl SamTag {
     pub const PG: SamTag = SamTag::new(b'P', b'G');
     /// Adapter clipping position (e.g. from Picard `MarkIlluminaAdapters`).
     pub const XT: SamTag = SamTag::new(b'X', b'T');
+    /// Bisulfite strand (written by `bwameth` and related tools): `Z:f` for the
+    /// forward/top strand, `Z:r` for the reverse/bottom strand. Used by `fgumi
+    /// zipper --restore-unconverted-bases` to decide which bases to restore.
+    pub const YD: SamTag = SamTag::new(b'Y', b'D');
 
     // ── fgumi-internal tags ─────────────────────────────────────────────────
 
--- a/src/lib/commands/zipper.rs
+++ b/src/lib/commands/zipper.rs
@@ -68,6 +68,7 @@ use clap::Parser;
 use log::{debug, info, warn};
 use noodles::core::Position;
 use noodles::sam::Header;
+#[cfg(test)]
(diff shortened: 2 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #276, opened 2026-04-15
feat(compare): add --command preset for per-stage compare defaults

### LINKED ISSUE #275: `compare bams`: add `--command <stage>` preset to encode per-stage expected differences
When comparing BAMs from the same pipeline stage produced by different tools
(e.g., fgbio vs fgumi), each stage has a stable set of "expected differences"
— tool-specific tags one side emits, MI-value vs MI-grouping semantics, tie-
break ordering, etc. Every consumer currently has to encode this knowledge
themselves (per-stage `--mode` choices, per-stage tag skip lists, when to pass
`--ignore-order`).

Suggested fix: a `--command <extract|zipper|sort|group|simplex|duplex|codec|filter>`
flag that expands to the canonical `--mode ... --ignore-tags ...
[--ignore-order]` combination for that stage, so callers can write
`fgumi compare bams --command zipper a.bam b.bam` and get the right defaults.
Explicit flags continue to work and override the preset.

nh13:
## Summary

Redesigns `fgumi compare bams` around two orthogonal boolean flags that compose cleanly, fixes a long-standing gap where grouped-stage output couldn't be fully validated, and adds a `--command <stage>` preset.

## CLI changes (breaking)

- **Remove** `--mode {full,content,grouping}` and the `CompareMode` enum
- **Add** `--check-grouping`: verify MI grouping equivalence and exclude MI from the content comparison, so records with different MI values can still be compared byte-for-byte elsewhere
- **Keep** `--ignore-order`: now gated on `--check-grouping` (errors if used alone)
- **Add** `--command <stage>`: preset for a pipeline stage. Supported: extract, zipper, sort, correct, dedup, group, simplex, duplex, codec, filter
- Explicit flags override the preset

## Preset defaults

| Stage | `--check-grouping` | `--ignore-order` |
|-------|-------------------|-----------------|
| extract, zipper, sort, correct, dedup | — | — |
| simplex, duplex, codec, filter | ✓ | — |
| group | ✓ | ✓ |

## Bug fix

Previously, `--mode full` (the old default) compared all tags including MI, which meant it could not be used for any stage where MI values legitimately differ between tools (group, and potentially downstream). The new `--check-grouping` path excludes MI from tag comparison while still verifying grouping equivalence, giving the strongest check that works across all stages.

## Test plan

- [x] 16 integration tests covering default mode, `--check-grouping`, `--check-grouping --ignore-order`, preset usage, and preset override
- [x] Key regression test (`test_check_grouping_different_mi_values_same_grouping_ok`) validates the bug fix: records with different MI values but equivalent grouping now pass `--check-grouping`
- [x] Full CI suite passes (2155 tests, ci-fmt, ci-lint)

Closes #275

### DIFF
--- a/src/lib/commands/compare/bams.rs
+++ b/src/lib/commands/compare/bams.rs
@@ -41,25 +41,48 @@ use std::thread;
 use crate::commands::command::Command;
 use crate::commands::common::parse_bool;
 
-use super::raw_compare::{raw_compare_structured, raw_records_byte_equal};
+use super::raw_compare::{raw_compare_structured_except, raw_records_byte_equal};
 
-/// Comparison mode for BAM files
-#[derive(Debug, Clone, Copy, Default, ValueEnum)]
-pub enum CompareMode {
-    /// Full comparison: verify MI groupings AND compare all BAM content
-    /// Both files must be in the same order (e.g., query-name sorted).
-    /// This combines grouping equivalence check with full content comparison.
-    #[default]
-    Full,
-    /// Content comparison: all fields and tags must match exactly
-    /// This is a pure record-by-record comparison without MI grouping analysis.
-    Content,
-    /// Grouping comparison: verify MI groupings are equivalent (for grouped BAMs)
-    /// Both files must be in the same order (e.g., query-name sorted).
-    /// Validates read names and R1/R2 flags match, then verifies that reads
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #284, opened 2026-04-17
feat(sort): distribute spill files across multiple temp dirs

nh13:
## Summary

- `fgumi sort -T/--tmp-dir` is now repeatable; pass more than one path to spread spill chunks across multiple filesystems in free-space-aware round-robin order.
- `FGUMI_TMP_DIRS` (PATH-style list) is honoured as a fallback when no `-T` flags are passed on the command line. CLI flags take precedence.
- New `TmpDirAllocator` owns the rotation and re-checks free space every 8 spills; dirs below 1 GiB free are dropped from rotation automatically.
- Behaviour with zero or one `-T` flag is unchanged.

## Usage

```bash
# Spread spill across multiple dirs
fgumi sort -i in.bam -o out.bam -T /mnt/ssd1 -T /mnt/ssd2

# Same via env var (useful for site defaults on HPC)
FGUMI_TMP_DIRS=/mnt/ssd1:/mnt/ssd2 fgumi sort -i in.bam -o out.bam
```

## Design notes

- **Distribution policy**: free-space-aware round-robin. At startup, each supplied dir is probed for free space; dirs below the 1 GiB threshold are dropped. In-rotation dirs receive chunks in strict round-robin order. Every 8 calls, the allocator re-probes and prunes dirs whose free space has fallen below the threshold.
- **Hot path**: allocator work happens only at spill boundaries (typically 10s-100s of times per sort), never per-record. Per-record loops in the four `sort_*` fns are untouched.
- **Scope**: only `fgumi sort` spills today, so the option lives in `src/lib/commands/sort.rs` rather than in `CompressionOptions`-style shared common-options.

## Known limitation

Mid-write `ENOSPC` is not auto-recovered — the startup filter and periodic recheck catch most cases *before* a write begins, but if a pooled spill writer hits `ENOSPC` mid-flight the sort fails with a clean error. `TmpDirAllocator::mark_full()` is wired and ready for a follow-up that intercepts the async `PendingSpill` error path.

## New dependency

- `fs4 = "0.13"` (default-features off) — cross-platform `available_space()` query. Pulls in rustix + libc + errno + bitflags only; no async runtime.

## Test plan

- [x] `cargo ci-fmt` clean
- [x] `cargo ci-lint` clean
- [x] `cargo ci-test` all green (2559/2559)
- [x] 18 new tests: 7 `TmpDirAllocator` unit tests (round-robin, `mark_full`, startup filter, periodic recheck, exhaustion), 4 env-var resolver tests, 1 rstest-parameterized clap parse test (4 cases), 1 e2e multi-tmpdir sort test (byte-equal to single-dir)
- [x] `fgumi sort --help` renders new `-T` doc and `EXAMPLES` entries cleanly

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -105,7 +105,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -116,7 +116,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -615,7 +615,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai review

Issue #285, opened 2026-04-17
`group` OOMs at ~30 GB RSS on ~500 M-pair real data, regardless of `--queue-memory`


nh13:
On a real paired-end CODEC dataset (~500 M read pairs, GRCh38 alignment,
template-coordinate-sorted fgumi zipper output, ~38 GB BAM) running on a
c7g.4xlarge (16 vCPU Graviton / **30 GiB RAM**), `fgumi group --strategy
adjacency --edits 1 --min-map-q 1` is OOM-killed at ~30 GB RSS after
processing ~170–180 M of the ~440 M input records. Reproduced 3× with
different memory configs:

| Config                                                              | OOM at (records) | Peak RSS |
|---------------------------------------------------------------------|------------------|----------|
| default (`--threads 16`, 768 MB/thread queue → 12 GB queue)         | 172 M            | ~30 GB   |
| `--threads 16 --queue-memory 16GB --queue-memory-per-thread false`  | 181 M            | ~30 GB   |
| `--threads 8  --queue-memory 4GB  --queue-memory-per-thread false`  | 171 M            | ~30 GB   |

Dropping the queue 12 GB → 4 GB moved the crash point by ~10 M records —
the dominant consumer isn't the pipeline queue, it's the per-position-group
molecule/family state being accumulated. Halving threads (16 → 8) likewise
didn't shift peak RSS, so it's shared state, not per-worker.

`fgbio GroupReadsByUmi` on the same input completes successfully with
`-Xmx8g` (~4× less memory).

Suggested mitigations — any one materially helps:
- Cap in-flight family state (e.g. `--max-family-memory`) and spill
  finalized position-groups to disk under pressure.
- Emit records from a position-group as soon as its reads are all seen
  rather than buffering across many positions.
- Lower the default per-thread queue (768 MB × threads is a lot by default).
- At minimum, document an empirical "GB per 100 M records" rule so users
  can size instances.

Unblocks real-data validation of `fgumi group` against `fgbio
GroupReadsByUmi` at scale — currently the cross-tool compare can only be
run at smoke (~1 M-pair) scale.

PR #287, opened 2026-04-17
fix(group): cap metric memory with per-thread accumulators (#285)

nh13:
## Summary

- Replaces the `Arc<SegQueue<CollectedMetrics>>` in `fgumi group` with `Arc<Vec<Mutex<GroupMetricsAccumulator>>>`, sized to `num_threads`.
- Each worker claims a slot via a TLS-cached index from a process-wide atomic counter and merges per-position-group counts into its slot immediately, so retained memory is O(threads × distinct sizes) instead of O(position groups).
- Eliminates the unbounded ~1 KB-per-position-group `AHashMap::with_capacity(50)` allocation that was the dominant consumer at ~10 GB on a 500 M-pair input (issue #285).

## Why

On a 500 M-pair CODEC BAM (~38 GB) on a 30 GiB box, `fgumi group --strategy adjacency` OOMed at ~30 GB RSS after ~170–180 M records regardless of `--queue-memory` or `--threads` settings. Static analysis pinned the dominant consumer to the `SegQueue` of `CollectedMetrics`, which held one `AHashMap<usize, u64>` (~1 KB backing table even when empty) per position group and was only drained at pipeline end. Downstream `write_all_metrics` only needs the aggregated counters — the per-group objects were pure scratch.

The single-threaded path already merges incrementally; this change brings the pipeline path in line with that.

## What's next

- Heap profile at ~150 M records is pending to confirm peak RSS is flat (±working set) across input sizes, not just "no longer OOMs at 180 M".
- Same pattern (`SegQueue<CollectedXxxMetrics>`) exists in `simplex`, `duplex`, `codec`, `filter`, `clip`, `correct`. Stacked follow-up PR will cover the tractable cases; the consensus commands additionally buffer rejected records as pipeline output, which needs a careful migration.

## Test plan

- [x] `cargo ci-fmt`
- [x] `cargo ci-lint`
- [x] `cargo nextest run` — all 225 group-touching tests pass
- [ ] Heap profile (dhat/heaptrack) on real 500 M-pair CODEC data at 150 M records — confirm `CollectedMetrics` no longer dominates
- [ ] End-to-end run at full scale — confirm peak RSS is flat across 50 M / 150 M / 400 M records on the same input
- [ ] fgbio `GroupReadsByUmi` byte-identity check on a shared-scale input

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Bug Fixes
+
+- Cap `fgumi group` metric memory by merging per-position-group counts into per-thread accumulators instead of a `SegQueue` that grew one `AHashMap` per position group ([#285](https://github.com/fulcrumgenomics/fgumi/issues/285)).
+
 ### Breaking Changes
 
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #290, opened 2026-04-17
refactor(pipeline): PerThreadAccumulator for all per-batch metric collection

nh13:
## Summary

Stacked on #287. Generalizes the group-specific accumulator into a reusable
\`PerThreadAccumulator<A>\` helper and rolls it out across the remaining
pipeline commands that previously used \`Arc<SegQueue<CollectedXxxMetrics>>\`
to buffer one metrics struct per batch for the whole run.

Commits:

1. \`refactor(pipeline)\` — adds \`crate::per_thread_accumulator::PerThreadAccumulator<A>\` with unit tests; migrates \`group\` off its inline implementation.
2. \`fix(filter)\` — drops the \`SegQueue\` in filter's setup; 4×u64 metrics merged per-thread.
3. \`fix(clip)\` — same pattern, 3×u64 metrics.
4. \`fix(correct)\` — per-batch template counts, per-UMI match map, and reject bytes merged into per-thread slots.
5. \`fix(simplex)\` — per-batch consensus stats, overlapping stats, groups, and reject bytes.
6. \`fix(duplex)\` — mirrors simplex.
7. \`fix(codec)\` — mirrors simplex/duplex.
8. \`docs(changelog)\` — rollout note.

## Why

Issue #285 showed that the \`SegQueue\` pattern retained one metrics struct per batch for the life of the run, scaling memory with input size. Group was the acute case (~10 GB at 180 M records via eager \`AHashMap\` allocations) and is the sole known OOM so far, but the other six commands share the same unbounded-growth shape and should be audited together — \`filter\` and \`clip\` leak tens-of-MB per billion records, \`correct\` can retain a per-UMI hashmap per batch, and the consensus commands retain reject bytes that scale with both batches and rejection rate.

This PR does **not** fix reject-buffering in \`simplex\`/\`duplex\`/\`codec\` — those still collect rejected raw BAM bytes into per-thread \`Vec<Vec<u8>>\`s and write them after the pipeline. Streaming rejects to the rejects BAM writer during the pipeline is a separate follow-up (PR #3, to be stacked on this branch).

## What's next

- Await the heap profile from #287 to confirm the group fix is complete and to quantify any residual allocators.
- PR #3 (stacked): streaming rejects writes for \`simplex\`/\`duplex\`/\`codec\`/\`correct\` so rejects-byte retention doesn't grow with reject count.

## Test plan

- [x] \`cargo ci-fmt\`, \`cargo ci-lint\`, \`cargo ci-test\` — 2545 tests pass.
- [ ] Memory microbench on a real run of each migrated command to confirm flat peak RSS w.r.t. input size.
- [ ] Byte-identity output check on a shared input for each command (unchanged behavior expected).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 ### Bug Fixes
 
 - Cap `fgumi group` metric memory by merging per-position-group counts into per-thread accumulators instead of a `SegQueue` that grew one `AHashMap` per position group ([#285](https://github.com/fulcrumgenomics/fgumi/issues/285)).
+- Replace unbounded `SegQueue<CollectedXxxMetrics>` buffering in `filter`, `clip`, `correct`, `simplex`, `duplex`, and `codec` with a shared `PerThreadAccumulator` helper, capping retained metric memory at `O(threads × distinct keys)` across all pipeline commands. Rejects buffering in the consensus commands is unchanged and tracked separately.
 
 ### Breaking Changes
 
--- a/src/lib/commands/clip.rs
+++ b/src/lib/commands/clip.rs
@@ -14,6 +14,7 @@ use crate::grouper::TemplateGrouper;
 use crate::logging::OperationTimer;
 use crate::metrics::clip::{ClipCounts, ClippingMetricsCollection};
 use crate::metrics::writer::write_metrics as write_metrics_tsv;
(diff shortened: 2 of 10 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #293, opened 2026-04-17
fix(pipeline): stream rejects to disk in simplex/duplex/codec/correct

nh13:
## Summary

Stacked on #290. Removes the per-thread \`Vec<Vec<u8>>\` reject-buffering pattern from the four commands that keep it (after #290 removed the metric-buffering pattern from all seven). Rejected raw BAM records are now streamed directly to the rejects BAM during \`serialize_fn\` via a shared BGZF writer guarded by a \`parking_lot::Mutex\`. The mutex only serializes the byte append; BGZF compression runs on the writer's own thread pool.

Commits (one per command):

1. \`fix(simplex)\` — open rejects writer up front; write bytes from serialize; finalize writer post-pipeline. Drops \`rejects: Vec<Vec<u8>>\` from the per-thread accumulator.
2. \`fix(duplex)\` — mirrors simplex.
3. \`fix(codec)\` — mirrors simplex/duplex.
4. \`fix(correct)\` — mirrors simplex/duplex/codec but on \`rejected_raw_records\`. Rejects use the input header (correct preserves alignment context).
5. \`docs(changelog)\` — note.

## Why

With #287/#290 landed, metric memory in the consensus commands and \`correct\` is capped at O(threads × distinct keys). The remaining unbounded growth vector is reject buffering: every rejected raw BAM record was held in per-thread Vecs for the life of the run, then written after the pipeline completed. On protocols with high rejection rates (e.g. duplex calls that fail \`/A\`/\`/B\` partnering, correct runs where most UMIs don't match the whitelist), this can retain multiple GB of raw BAM bytes.

Streaming during the pipeline caps reject memory at roughly the BGZF block size per worker, independent of reject count.

## Why not the pipeline's \`with_secondary\` API

\`run_bam_pipeline_from_reader_with_secondary\` exists and \`filter\` uses it, but it writes the secondary output using \`output_header\`. simplex/duplex/codec build an unmapped-consensus output header that differs from the input header — rejects need the input header since they're original mapped reads. Extending the pipeline API to accept a separate secondary header is a larger surgery; the shared-Mutex pattern here is surgical and keeps the pipeline internals untouched. Worth revisiting once all four commands are in the same shape and a generalization has a clear target.

## Test plan

- [x] \`cargo ci-fmt\`, \`cargo ci-lint\`, \`cargo ci-test\` — 2545 tests pass.
- [x] Each command's integration tests (simplex/duplex/codec/correct) pass, including the \`*_rejects_has_bgzf_eof\` BGZF-EOF sanity tests.
- [ ] A/B memory comparison on a run with a high-rejection-rate input — confirm flat RSS across reject count.
- [ ] Byte-identity check on rejects BAM output vs. the old buffered path (ordering may differ since writes are no longer post-sorted; behavior already was not order-guaranteed).

## Note

The final changelog commit is unsigned (three 1Password biometric signing attempts failed locally). It should be re-signed before merge.

### DIFF
--- a/crates/fgumi-consensus/src/duplex_caller.rs
+++ b/crates/fgumi-consensus/src/duplex_caller.rs
@@ -1723,15 +1723,31 @@ impl DuplexConsensusCaller {
         read_name_prefix: &str,
         read_group_id: &str,
         cell_tag: Option<Tag>,
-    ) -> Result<(ConsensusOutput, ConsensusCallingStats)> {
+        track_rejects: bool,
+    ) -> Result<(ConsensusOutput, ConsensusCallingStats, Vec<Vec<u8>>)> {
         let mut stats = ConsensusCallingStats::new();
         let mut builder = UnmappedSamBuilder::new();
         let mut output = ConsensusOutput::default();
         let methylation_mode = ss_caller.options.methylation_mode;
 
+        // Helper: move raw input records into the reject payload if rejects are being
+        // tracked. Each rejection site moves `a_records`/`b_records` into the reject
+        // vec so we never hold both the input and reject copies simultaneously.
+        // Single-strand rejections are captured separately by `ss_caller`.
+        let collect_rejected_raw =
+            |track: bool, a: Vec<RawRecord>, b: Vec<RawRecord>| -> Vec<Vec<u8>> {
+                if track {
+                    let mut v: Vec<Vec<u8>> = a.into_iter().map(RawRecord::into_inner).collect();
+                    v.extend(b.into_iter().map(RawRecord::into_inner));
+                    v
+                } else {
+                    Vec::new()
+                }
+            };
+
         // Check if we have both strands
         if a_records.is_empty() && b_records.is_empty() {
-            return Ok((ConsensusOutput::default(), stats));
+            return Ok((ConsensusOutput::default(), stats, Vec::new()));
         }
 
         // Check if we have enough input reads (matching fgbio's hasMinimumNumberOfReads)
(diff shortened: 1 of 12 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed review feedback and rebased:

- Dropped 9 commits already on `main` via #287/#290; dropped both `docs(changelog)` commits since release-plz + git-cliff owns `CHANGELOG.md` (per `release-plz.toml`).
- Fixed stale "Rejects buffering semantics are preserved (see follow-up)" comments in simplex/duplex/codec.
- Finalize the rejects writer even when the pipeline errors out, so a partial rejects BAM still gets a valid BGZF EOF (simplex/duplex/codec/correct).
- Moved reject streaming from `serialize_fn` into `process_fn`, writing per-MI-group (per-template for correct) directly through the shared BGZF writer. `ProcessedBatch` no longer carries any reject payload, so peak reject memory is now the BGZF block size rather than `O(batch_size × rejects_per_group)`.

Branch is rebased onto current `main`, 4 commits (one per pipeline command).

The `per_thread_accumulator.rs` thread is now obsolete: after rebase, that file isn't touched by this PR (it belongs to #290).

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #297, opened 2026-04-19
fix(unified-pipeline): suppress deadlock false positive when all queues empty

nh13:
## Summary

- Guard `check_deadlock_and_restore` so it only fires when at least one of the pipeline's `q1..q7` queues still has an item in it.
- When every queue is empty, treat the lack of queue-event activity as upstream starvation (pipe blocked on a slow producer), not a deadlock. Bump `last_progress_time` so we don't re-fire on every monitor tick and return `None`.
- Add regression test `test_no_deadlock_on_empty_queues_stale_timestamp`.
- Repoint the two pre-existing all-queues-empty tests (`test_recovery_disabled_only_logs`, `test_recovery_updates_stable_limit`) to place a single item in `q3_len` so they continue to exercise the genuine stuck-items deadlock path that the detector *should* fire on.

## Why

`check_deadlock_and_restore` declares a deadlock whenever the pipeline's global `last_progress_time` hasn't advanced for `timeout_secs` (default 10 s). That timestamp is only advanced by `record_qN_push` / `record_qN_pop`, so any 10-second window without a queue event trips the detector.

In practice that fires when the pipeline is *blocked on stdin*, not deadlocked. Observed on a full-depth run of

```bash
fgumi extract | fgumi fastq | bwa mem | fgumi zipper | fgumi sort | fgumi group | fgumi codec | fgumi filter
```

on a 552 M-record input: during `fgumi sort`'s phase-2 merge, upstream output to the `group | codec | filter` tail is sporadic. Inside `filter`, all of `q1..q7` drained for longer than `timeout` while the process waited for the next byte from stdin. The detector interpreted that as a deadlock and aborted with a SIGPIPE cascade even though the pipeline was healthy and would have resumed as soon as sort produced more output. The run was lost after ~4 hours of work.

The previous handling offered no way to distinguish starvation from deadlock: any stall that crossed the timeout fired the same code path. `--deadlock-recover` (which doubles the memory limit) does nothing useful for starvation, and a longer `--deadlock-timeout` just delays the false positive.

## What

The fix adds a single pre-check inside `check_deadlock_and_restore`: if `now - last_progress >= timeout` AND every queue is empty, return `DeadlockAction::None` and update `last_progress_time` to `now` so the detector doesn't re-fire on every monitor tick. If real work resumes, the normal per-queue push/pop sites update timestamps and the detector starts fresh; if the pipeline genuinely hangs forever, the process sits idle until the user interrupts — the same liveness guarantee as the pre-detector behaviour, but without the false-positive kill.

### Test plan

- [x] `cargo ci-fmt`
- [x] `cargo ci-lint`
- [x] `cargo test --release --lib --features compare unified_pipeline::deadlock` — **20 passed**, 0 failed
- [x] `cargo test --release --lib --features compare` — **1,974 passed**, 0 failed
- [x] End-to-end: the patched binary completed a 552 M-record `fgumi extract | ... | fgumi filter` pipe to completion where the previous binary aborted at 12% of sort phase-2 merge.

### What's not included

- Genuine-deadlock detection when queues *do* hold items continues to work as before; the stuck-items path `handle_deadlock` is unchanged.
- The `DeadlockConfig::timeout_secs` default and `--deadlock-recover` behaviour are unchanged.
- Per-queue diagnostic logging in `log_queue_state` is unchanged.

### DIFF
--- a/src/lib/unified_pipeline/deadlock.rs
+++ b/src/lib/unified_pipeline/deadlock.rs
@@ -414,6 +414,51 @@ pub fn check_deadlock_and_restore(
 
     // Check for deadlock (no progress for timeout period)
     if now.saturating_sub(last_progress) >= timeout {
+        // False-positive guard. The per-queue timestamps only update when a
+        // record is pushed into or popped from one of q1..q7. If every
+        // queue is empty, no push/pop happens even when the pipeline is
+        // fundamentally healthy — the common case is a stdin pipe that
+        // intermittently stalls for longer than `timeout` because the
+        // upstream process is slow (e.g. `fgumi sort` in phase-2 merge
+        // feeding `group | codec | filter`). Absence of queue activity in
+        // that case is starvation, not deadlock.
+        //
+        // Only raise "deadlock detected" when at least one queue still has
+        // items stuck in it. With empty queues we instead bump the progress
+        // timer and return None, so we don't re-fire every monitor tick and
+        // so real pipeline progress — when it resumes — naturally updates
+        // the per-queue timestamps. A genuine upstream hang will leave
+        // every queue empty forever and the pipeline will sit idle until
+        // the user interrupts; the previous behaviour killed the process
+        // in that same scenario, so we do not lose liveness signalling
+        // that we had before.
+        // In-flight work includes both the bounded queues AND the BAM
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #305, opened 2026-04-21
perf: migrate RecordBuf hot paths to raw-byte accessors

nh13:
## Summary

- **Fixes the 10× simplex-metrics / duplex-metrics regression** introduced by #294. The per-template round-trip through `raw_record_to_record_buf` was rebuilding a full synthetic BAM stream (prelude + `@SQ` dictionary + framing) for *every* record, so per-record cost scaled with `O(header_size)` instead of `O(record_size)`. Agilent-HS2 (53M records, ~3000 `@SQ` entries) projected 3.4h vs. the 3h AWS batch timeout. Investigation writeup: commit message.
- Same-PR cleanup of the other production `RecordBuf` decoders that missed the 48-commit series migration: **`downsample`** and **`zipper`** (mapped-side BAM and StdinBam arms).
- Header-only `create_bam_reader` callers in **`sort`** and **`merge`** swapped to `create_raw_bam_reader`.
- **`compare::bams::deserialize_raw_record`** comment sharpened to document why its noodles-decode-per-record is intentional (off-hot-path diff rendering only).
- **Deletes the dead `SortKey` trait** and its `CoordinateKey` / `QuerynameKey` impls. Zero production or test callers — the live sort path uses `RawSortKey` / `RawCoordinateKey` / `RawQuerynameKey` exclusively.

The SAM arm of `zipper` retains the existing `RecordBufEncoder` round-trip because text-mode SAM has no raw-byte path.

One commit, 8 files, net **−208 LOC** (+234 / −442; most of the delta is the dead `SortKey` scaffolding).

## Test plan

- [x] `cargo ci-test` — 2451/2451 passing. One unrelated flake (`test_multithreaded_extraction_preserves_order`) under full-suite contention; passes in 31ms in isolation.
- [x] `cargo ci-fmt` clean
- [x] `cargo ci-lint` clean
- [ ] Re-run the hung agilent-hs2 AWS batch job to confirm simplex-metrics / duplex-metrics now complete well under the 3h timeout.
- [ ] Spot-check simplex-metrics and duplex-metrics numeric output against pre-regression baselines on a grouped BAM (behaviour is preserved but the unclipped-5' calculation now also counts H clips, matching fgbio / htsjdk; negligible impact in practice since H clips are rare in aligner output).
- [ ] \`zipper\`: end-to-end test through `bwa mem ... | fgumi zipper ...` (SAM stdin) and a BAM file input to confirm both paths still produce identical output to the pre-refactor code.
- [ ] `downsample`: round-trip test that output BAM records are byte-identical to the pre-refactor version (no RecordBuf re-encode quirks).

### DIFF
--- a/src/lib/commands/compare/bams.rs
+++ b/src/lib/commands/compare/bams.rs
@@ -599,9 +599,13 @@ fn start_raw_batch_reader(
 
 /// Deserialize raw BAM record bytes into a noodles `RecordBuf`.
 ///
-/// This is used only when records differ at the raw byte level and we need
-/// human-readable field values for diff reporting. The raw bytes are the BAM
-/// record body WITHOUT the 4-byte length prefix.
+/// **Intentional non-raw decode.** Other production hot paths avoid
+/// `raw_records_to_record_bufs` because it builds a synthetic BAM stream per call;
+/// here it is acceptable because this function only runs *off* the comparison hot
+/// path — after the byte/structured tiers have already flagged a pair as
+/// mismatched and we need typed field values to render a human-readable diff.
+///
+/// The raw bytes are the BAM record body WITHOUT the 4-byte length prefix.
 ///
 /// # Errors
 ///
--- a/src/lib/commands/downsample.rs
+++ b/src/lib/commands/downsample.rs
@@ -5,18 +5,17 @@
 //!
 //! Requires input BAM to be in template-coordinate order (from group).
 
-use crate::bam_io::{create_bam_reader, create_bam_writer, create_optional_bam_writer};
+use crate::bam_io::{RawBamWriter, create_raw_bam_reader, create_raw_bam_writer};
 use crate::logging::OperationTimer;
 use crate::progress::ProgressTracker;
-use crate::sam::SamTag;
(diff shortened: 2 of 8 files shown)


nh13:
Addressed both CodeRabbit findings:

- `downsample::execute`: explicit `writer.finish()?` and `rejects_writer.finish()?` before returning — matches `zipper`'s pattern and flushes any buffered records before the BGZF EOF.
- `shared_metrics::process_templates_from_bam`: dropped the `r1_tid != r2_tid` early-exit so inter-reference pairs are now retained. For interval overlap, same-ref pairs still use the insert span `(min(r1_start, r2_start), max(r1_end, r2_end))`; inter-ref pairs use R1's own range, matching fgbio `CollectDuplexSeqMetrics`:

    ```scala
    val (start, end) = if (rec.refIndex == rec.mateRefIndex) Bams.insertCoordinates(rec)
                       else (rec.start, rec.end)
    ```

Added `test_inter_reference_pairs_are_retained` as a regression test (verified it fails on the pre-fix `continue;` and passes after). All 2452 tests pass locally, lint + fmt clean.

Force-pushed `d6646d3e` → `44b5ac84`.

PR #307, opened 2026-04-21
feat(compare): restore --command preset for per-stage compare defaults

nh13:
## Summary

Reintroduce the `--command <stage>` preset on `fgumi compare bams`, which was originally added in #276 and unintentionally removed in #296 when that breaking refactor synced `src/lib/commands/compare/bams.rs` from a pre-#276 backup branch. The `compare/bams.rs` edits in #296 were absorbed into an "amends to include" bullet alongside the larger bam_codec/simulate scope and the regression slipped through review.

## Why this matters

Downstream `fgumi-benchmarks` relies on this flag for cross-tool equivalency checks between fgumi and fgbio outputs (`workflow/rules/benchmark.smk` → `get_compare_args()`). At the current fgumi main, every `compare_bams_batch` job that hits a stage in the preset set fails immediately with:

```
error: unexpected argument '--command' found
```

Other batches (e.g. `group.identity`) silently run with `--mode full` and fail on expected cross-tool ordering differences, because the preset that would have set `--mode grouping --ignore-order` was also lost.

## What the preset does

`--command <stage>` applies canonical `--mode` and `--ignore-order` defaults. Adapted to the post-#276 `--mode {full,content,grouping}` + `--ignore-order` surface:

| stage | `--mode` | `--ignore-order` | notes |
|---|---|---|---|
| extract, zipper, sort, correct, dedup, filter | content | false | No MI tags or pass-through; deterministic |
| group | grouping | true | MI values and record order may legitimately differ (cross-tool) |
| simplex, duplex, codec | grouping | true | Non-deterministic with `--threads` |

Explicit `--mode` or `--ignore-order` flags always override the preset, e.g.:

```
fgumi compare bams --command group --mode full --ignore-order=false a.bam b.bam
```

## Implementation

- `CommandPreset` enum reintroduced with a `.defaults() -> (CompareMode, bool)` method.
- `mode` and `ignore_order` struct fields changed to `Option<T>` so the resolver can distinguish "unset" from "explicit value". This is internal — the CLI surface still has `--mode` / `--ignore-order` with no default, and the old default-`full` / default-`false` behavior is preserved by `effective_settings()`.
- New `effective_settings(&self) -> (CompareMode, bool)` resolves precedence: explicit flag > `--command` preset default > built-in default (`full`, `false`).
- `execute()` logs `Using --command <preset> preset: mode=..., ignore-order=... (with explicit overrides)` when `--command` is in effect.
- Help text updated with a `COMMAND PRESETS` section replacing the `RECOMMENDED SETTINGS BY COMMAND` table (same information, self-documenting now that the preset is back).

## Test plan

- [x] `cargo build --release --features compare,simulate` — builds clean
- [x] `cargo ci-test` — **2458 tests pass**, 23 skipped; no regressions
- [x] `cargo ci-fmt` — clean
- [x] `cargo ci-lint` — clean
- [x] 6 new unit tests in `src/lib/commands/compare/bams.rs::tests`:
  - `preset_defaults_content_stages_map_to_content_no_ignore_order`
  - `preset_defaults_grouping_stages_map_to_grouping_with_ignore_order`
  - `effective_settings_fall_back_to_built_in_defaults`
  - `effective_settings_use_preset_when_only_command_given`
  - `effective_settings_explicit_mode_overrides_preset`
  - `effective_settings_explicit_ignore_order_overrides_preset`
- [x] `fgumi compare bams --help` — verified `--command` appears and preset table renders correctly
- [x] All 14 existing `test_compare_bams` integration tests still pass

## Out of scope

- Does not touch the `--mode` / `--ignore-order` semantics or any comparison logic — pure CLI / preset-resolution change.
- Does not re-add `clip`, `downsample`, `review` presets (they were not in #276's original enum and are not used by the benchmarks workflow).

### DIFF
--- a/src/lib/commands/compare/bams.rs
+++ b/src/lib/commands/compare/bams.rs
@@ -60,6 +60,58 @@ pub enum CompareMode {
     Grouping,
 }
 
+/// Preset comparison settings for a specific fgumi pipeline stage.
+///
+/// Each variant encodes canonical `--mode` and `--ignore-order` defaults for
+/// comparing BAM output from that stage, including cases (e.g. cross-tool
+/// `group` comparison against fgbio) where MI values or record order may
+/// legitimately differ. Explicit `--mode` or `--ignore-order` flags override
+/// the preset.
+#[derive(Debug, Clone, Copy, ValueEnum)]
+pub enum CommandPreset {
+    /// Extract output: no MI tags; exact content comparison.
+    Extract,
+    /// Zipper output: preserves MI tags unchanged; exact content comparison.
+    Zipper,
+    /// Sort output: deterministic; exact content comparison.
+    Sort,
+    /// Correct output: modifies RX tag only; exact content comparison.
+    Correct,
+    /// Dedup output: deterministic; exact content comparison.
+    Dedup,
+    /// Group output: MI values and record order may differ between tools or
+    /// runs. Verifies grouping equivalence only (unordered).
+    Group,
+    /// Simplex consensus output: non-deterministic with `--threads`.
+    /// Verifies grouping equivalence only (unordered).
+    Simplex,
+    /// Duplex consensus output: non-deterministic with `--threads`.
+    /// Verifies grouping equivalence only (unordered).
+    Duplex,
+    /// CODEC consensus output: non-deterministic with `--threads`.
+    /// Verifies grouping equivalence only (unordered).
+    Codec,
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #308, opened 2026-04-22
fix(compare): preserve /A /B strand suffix in paired-UMI MI keys

nh13:
## Summary

`fgumi compare bams --mode grouping` incorrectly reported paired-UMI BAMs as disagreeing (`Records matched: 0`, `BAM groupings DIFFER`) even when the two files had identical groupings. The MI reader called `str::parse::<i64>()` on the tag value, which rejects the spec-compliant paired-strand encoding `MI:Z:<id>/<A|B>` emitted by both fgumi and fgbio, so every record was counted as "missing MI".

This PR replaces the `i64`-valued MI map with a small `MiKey` enum that parses all three observed MI forms and preserves the `/A` vs `/B` distinction for grouping-equivalence checks:

- `MI:i:<int>` → `MiKey::Int`
- `MI:Z:<int>` → `MiKey::Int`
- `MI:Z:<int>/A` / `MI:Z:<int>/B` → `MiKey::Strand { base, strand }`
- anything else → `None` (treated as missing, as before)

`Display` round-trips each form back to the original BAM encoding so the existing grouping-mismatch error messages (`MI group '…' in BAM1 (N reads) maps to multiple MIs in BAM2`) read unchanged.

## Why not pack into `i64`?

`MoleculeId::compact_index()` (`base*2`, `base*2+1`) would collide with `Single(n)` if a file mixed single and paired MIs, silently aliasing groups and masking real disagreements. The enum is the minimal type that preserves the A/B invariant without that risk.

## Test plan

New tests (all in `src/lib/commands/compare/bams.rs` and `tests/integration/test_compare_bams.rs`):

- [x] Unit: parse `MI:i:42` → `Int(42)`
- [x] Unit: parse `MI:Z:42` → `Int(42)`
- [x] Unit: parse `MI:Z:0/A` → `Strand { base: 0, strand: b'A' }`
- [x] Unit: parse `MI:Z:7/B` → `Strand { base: 7, strand: b'B' }`
- [x] Unit: `13/A` and `13/B` produce different keys (core invariant)
- [x] Unit: missing MI / non-numeric / unknown-strand all return `None`
- [x] Unit: `Display` round-trips each form
- [x] Integration: ordered grouping mode on identical paired-strand BAMs reports `EQUIVALENT` with zero missing MI
- [x] Integration: `--ignore-order` grouping mode on identical paired-strand BAMs reports `EQUIVALENT` with zero missing MI
- [x] Integration: swapping `/A` ↔ `/B` between the two BAMs is detected as a grouping mismatch
- [x] Verified RED: with the fix stashed, both positive integration tests fail with the exact symptom from the bug report (`Missing MI in BAM1: 4`, `Records matched: 0`, `DIFFER`)
- [x] `cargo ci-fmt`, `cargo ci-lint`, `cargo ci-test` (2494 passed)

### DIFF
--- a/src/lib/commands/compare/bams.rs
+++ b/src/lib/commands/compare/bams.rs
@@ -451,8 +451,8 @@ struct GroupingCompareResult {
     key_hash: ReadKeyHash,
     /// Read name as String, only populated when needed for error reporting
     read_name_for_display: Option<String>,
-    mi1: Option<i64>,
-    mi2: Option<i64>,
+    mi1: Option<MiKey>,
+    mi2: Option<MiKey>,
     name_match: bool,
     flag_match: bool,
     diff_detail: Option<DiffDetail>,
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

PR #313, opened 2026-04-22
fix(consensus): gate methylation-dependent filter fns behind simplex

nh13:
## Summary

Follow-up to #311 (flagged in that PR's body). The `methylation` module in `fgumi-consensus` is gated behind `#[cfg(feature = "simplex")]` in `lib.rs`, but three public functions in `crates/fgumi-consensus/src/filter.rs` -- compiled unconditionally -- take `&dyn crate::methylation::RefBaseProvider`:

- `resolve_ref_bases_for_record`
- `mask_strand_methylation_agreement_raw`
- `check_conversion_fraction_raw`

The `noodles::...::Kind` import used only inside `resolve_ref_bases_for_record`'s CIGAR match was also ungated, and 14 tests in the same file use either those functions or `crate::methylation::tests::TestRef` without gating.

Result: `cargo check -p fgumi-consensus --no-default-features --all-targets` failed to compile with 29 errors. Not publish-blocking (the workspace enables `simplex` by default, so `cargo publish --verify` compiles fine), but it breaks downstream consumers who want to depend on `fgumi-consensus` with `default-features = false` and enable a bespoke feature set.

## Fix (2 commits)

1. **Library:** gate the three functions and the `noodles::...::Kind` import behind `#[cfg(feature = "simplex")]`, matching the pattern already applied to the adjacent `From` impls.
2. **Tests:** add `#[cfg(feature = "simplex")]` to each of the 14 tests that touch the gated functions or `methylation::tests::TestRef`.

Total diff: 18 lines added across one file.

## Scope decisions

- **Root `fgumi` crate under `--no-default-features`:** separately broken; fixed in the stacked follow-up #314 (rebases onto main after this lands).
- **CI coverage:** #314 adds a `cargo check --workspace --no-default-features --all-targets` job to `check.yml` that would have caught both this class of bug and the one from #311.

## Verification

- [x] `cargo check -p fgumi-consensus --no-default-features --lib` passes cleanly (was 29 errors + 1 warning)
- [x] `cargo check -p fgumi-consensus --no-default-features --all-targets` passes cleanly (was 29 errors)
- [x] `cargo check -p fgumi-consensus --all-targets` (default features) unchanged -- passes
- [x] `cargo check --workspace --all-targets` passes
- [x] `cargo test -p fgumi-consensus --no-run` (default features) passes -- all tests still compile

### DIFF
--- a/crates/fgumi-consensus/src/filter.rs
+++ b/crates/fgumi-consensus/src/filter.rs
@@ -1603,6 +1603,7 @@ mod tests {
 
     // -- mask_strand_methylation_agreement_raw tests --
 
+    #[cfg(feature = "simplex")]
     #[test]
     fn test_strand_methylation_agreement_concordant() {
         use crate::methylation::tests::TestRef;
@@ -1633,6 +1634,7 @@ mod tests {
         assert_eq!(masked, 0, "Concordant CpG should not be masked");
     }
 
+    #[cfg(feature = "simplex")]
     #[test]
     fn test_strand_methylation_agreement_discordant() {
         use crate::methylation::tests::TestRef;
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #317, opened 2026-04-24
feat(runall): end-to-end streaming UMI pipeline in a single command

nh13:
## Summary

Adds `fgumi runall`, a single command that runs the
`extract → align → zipper → sort → group → consensus → filter` pipeline as
one streaming invocation — no intermediate BAMs on disk. The standalone
commands remain unchanged. `--start-from` / `--stop-after` select any
contiguous slice; stages run concurrently in one process, streaming
records through in-memory queues with backpressure and Ctrl-C cancellation.

## Engine

`src/lib/runall/engine/` is a general streaming-pipeline substrate:
Stage / SpecialStage traits, work-stealing worker pool, bounded queues
with atomic memory accounting, cancellation, deadlock watchdog. A `Plan`
describes `source + stages + sink` as erased types; the runner dispatches
on source/sink variants and builds concrete types once. Stages reuse
per-command logic so runall and the standalone commands produce identical
output for equivalent CLI args.

## Correctness posture

- **Defaults parity tests** pin runall's resolved configs to the
  standalone commands' resolved configs (filter, group, consensus,
  correct, extract, sort), including duplex 1–3-value vectors for
  `--filter::min-reads`, `--filter::max-read-error-rate`, and
  `--filter::max-base-error-rate`.
- **CIGAR parity test** (10 cases, forward + reverse strand, soft + hard
  clip combinations) pins `unclipped_5prime_from_raw_bam` to fgbio's
  `alignmentEnd + trailing-clip` reference formula.
- **`group → filter` and `extract → sort` byte-for-byte deterministic**
  across runs (`-t 1` to isolate from aligner thread non-determinism).
- **AlignAndMerge zipper** errors on mapped-stream reorder: if the
  aligner emits records out of order, the tail check surfaces
  *"processed all unmapped reads but mapped reads remain"* rather than
  silently dropping output. Aligner-*drops* (fewer mapped than input)
  are treated as orphan-unmapped and skipped, matching the documented
  `fgumi zipper --exclude-missing-reads=true` contract.
- Invalid stage orderings, missing references, corrupt BAMs, and
  malformed read structures produce specific user-facing errors
  (not panics).

## Metrics + observability

- Optional `--metrics <path>` writes a per-stage TSV (stage, wall time,
  records in/out).
- Runall stages adopt the project's `PerThreadAccumulator` pattern for
  consensus, correct, and group metrics — matching every other parallel
  standalone command. No shared-mutex hotspots on the batch hot path.
- `--filter::stats <path>` writes the same 4-row TSV as
  `fgumi filter --stats` (`total_reads`, `passed_reads`, `failed_reads`,
  `pass_rate`), using the shared `write_filter_stats` helper for
  byte-identical output.
- Event-based progress tracking (dashboard / heartbeat / summary).

## Performance-relevant knobs (for fgumi-benchmarks)

- `--threads` sizes the pool; worker count × per-batch work determines
  contention on the PerThreadAccumulators.
- Default queue capacities are auto-tuned from `--threads`; pipeline
  memory is globally bounded by the `MemoryTracker` (default 4 GiB).
- AlignAndMerge in-flight chunks bounded to 256 (~10 MB typical), with
  backpressure into the upstream MemoryTracker-accounted `StageQueue` —
  no unbounded channel growth.
- Deadlock watchdog trips after 120s of zero progress; sized to cover
  cold-I/O bwa reference loads on WGS-scale indices.
- `--filter::min-reads` is required iff the plan reaches filter or
  consensus (matches standalone `fgumi filter`'s fgbio-compatible
  "no default, user must supply" semantics).

## Test plan

- [ ] `cargo ci-fmt && cargo ci-lint && cargo ci-test` clean
      (2831 tests passing, 39 skipped on this branch)
- [ ] Run through fgumi-benchmarks to quantify throughput versus the
      standalone chain at equivalent thread counts
- [ ] Spot-check byte equality on a production BAM between
      `fgumi runall --start-from group` and the standalone
      `fgumi filter` output


nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #318, opened 2026-04-25
fix(group,dedup): deterministic MoleculeId numbering across runs

nh13:
## Summary

- Fix `fgumi group` and `fgumi dedup` `--threads N>1` non-determinism: two runs on the same input were producing identical molecule groupings but mismatched `MI:Z` numbering on ~10–15% of records (verified end-to-end on the SRR6109255 paired-UMI smoke), which then propagates into downstream `simplex`/`duplex`/`codec` consensus QNAMEs and breaks `fgumi compare bams` correspondence.
- Add a serial-ordered Q5 (`SerialOrderedArrayQueue`) plus an `OrderedMiAllocator` to advance the cumulative MI counter in pipeline-input order even when serialize workers run concurrently.
- Add an integration regression suite that reproduces the bug deterministically and now passes after the fix.

## Root cause

The unified BAM pipeline's `Q5` (process → serialize) is an unordered FIFO. Worker threads finish position groups in parallel and push completions in arrival order, not pipeline-serial order. The serialize step then advances a shared `AtomicU64::fetch_add` inside `serialize_fn`, so two runs see different completion orders → different cumulative offsets → different `MI:Z` numbers, even though the per-group molecule sets are identical. The single-threaded fast path was not affected because it only ever has one in-flight serialize.

The user-visible report attributed the bug to `fgumi duplex`, but `simplex`/`duplex`/`codec` themselves were already deterministic — they preserve the input `MI:Z` rather than minting new IDs, so they inherit the upstream non-determinism through the QNAME format `prefix:input_mi`. Fixing `group` and `dedup` removes the non-determinism at the source for the whole consensus pipeline.

## Fix

1. **`SerialOrderedArrayQueue<T>`** (`src/lib/unified_pipeline/serial_ordered_array_queue.rs`) — drop-in replacement for `crossbeam_queue::ArrayQueue<(u64, T)>` used as Q5. Pop returns the smallest buffered serial only when it equals the cursor; push exceeds capacity when the cursor's serial isn't yet buffered to avoid producer/consumer deadlock (mirroring the existing `OrderedQueue` safety net).
2. **`OrderedMiAllocator`** (`src/lib/ordered_mi_allocator.rs`) — hands out contiguous `MoleculeId` blocks in serial order. `allocate(serial, count)` blocks via condvar until the cursor reaches `serial`, then packs onto the per-serial offset; `finalize(serial)` folds the per-serial offset into the cumulative counter and advances the cursor.
3. **`serialize_context`** thread-local (`src/lib/unified_pipeline/serialize_context.rs`) — published by `try_step_serialize` (and the single-threaded fast path) so closures can read `(serial, item_idx, batch_len)` without changing the public `serialize_fn` signature. Used by `group`/`dedup`'s closures to call `allocate` with the right serial and `finalize` on the last item of each batch.

`fgumi group` and `fgumi dedup` swap their `AtomicU64::fetch_add` for the new allocator. No other BAM pipeline command needs changes (none mint MI integers themselves).

## Test plan

- [x] `cargo ci-test` (full nextest suite, 2510 tests, 23 skipped, 0 failures)
- [x] `cargo ci-fmt`
- [x] `cargo ci-lint` (clippy `-D warnings -W clippy::pedantic` clean)
- [x] New `tests/integration/test_group_determinism.rs` — three regression tests for `group --threads 1`, `group --threads 4`, `dedup --threads 4`. Each runs the command 6× on the same multi-orientation paired-UMI BAM and asserts byte-identical per-`(QNAME, flags)` `MI:Z` across all runs. Without the fix the threads-4 cases reproduce the bug deterministically (~30k/40k records mismatching); with the fix all 6 runs are identical.
- [x] Real-data smoke: `fgumi zipper → sort → group --threads 8` on SRR6109255 (1.7M records). Pre-fix: 100% of MI tags differed across two runs. Post-fix: byte-identical SAM body across two runs. Also verified `fgumi duplex --threads 8` on the (now stable) grouped output is byte-identical across runs.
- [x] Existing `test_pipeline_concurrency` suite (13 tests) still passes — `SerialOrderedArrayQueue`'s deadlock-avoidance branch (push exceeds capacity when cursor's serial is missing) keeps the multi-threaded contention tests from deadlocking even with the new ordering invariants.

### DIFF
--- a/src/lib/commands/dedup.rs
+++ b/src/lib/commands/dedup.rs
@@ -541,7 +541,16 @@ fn assign_umi_groups(
     no_umi: bool,
 ) -> Result<()> {
     if assigner.split_templates_by_pair_orientation() {
-        let mut subgroups: AHashMap<(bool, bool), Vec<usize>> = AHashMap::new();
+        // Group by pair orientation. Use BTreeMap (not AHashMap) so subgroup
+        // iteration order is deterministic across processes — `AHashMap`'s
+        // per-process random hasher seed would otherwise make
+        // `assigner.assign()` see orientation subgroups in different orders
+        // between two runs of `fgumi dedup` on the same input, which produces
+        // identical molecule groupings but different `MoleculeId` numbering
+        // (and therefore different `MI:Z` tags). Mirrors the same fix in
+        // `src/lib/commands/group.rs`.
+        let mut subgroups: std::collections::BTreeMap<(bool, bool), Vec<usize>> =
+            std::collections::BTreeMap::new();
         for (idx, template) in templates.iter().enumerate() {
             let orientation = get_pair_orientation(template);
             subgroups.entry(orientation).or_default().push(idx);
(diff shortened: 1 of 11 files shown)


nh13:
@coderabbitai review

nh13:
Closing in favor of a cleaner architecture: instead of the in-place `OrderedMiAllocator` + thread-local `SerializeContext` + `SerialOrderedArrayQueue` approach, we'll add a dedicated serial-ordered "MI assign" stage between Process and Serialize that pops batches in `batch_serial` order and rewrites each template's local MoleculeId into a global one in a single pass. `serialize_fn` then becomes pure (no synchronization). Spec to follow on the same branch.

PR #319, opened 2026-04-25
fix(group,dedup): deterministic MoleculeId numbering via opt-in MI Assign stage

nh13:
## Summary

- Fix `fgumi group --strategy paired --threads N>1` and `fgumi dedup --strategy paired --threads N>1` non-determinism: two runs on the same input were producing identical molecule groupings but mismatched `MI:Z` integers on ~10–15% of records (100% on synthetic dense layouts), which then propagated into downstream `simplex`/`duplex`/`codec` consensus QNAMEs and broke `fgumi compare bams` correspondence.
- Architecture: add an **opt-in serial-ordered MI Assign stage** between Process and Serialize. Only `group` and `dedup` install it; every other BAM-pipeline command keeps exactly today's path.
- Lands the design spec at `docs/design/deterministic-mi-numbering.md` and a six-run integration regression test that reproduces the bug deterministically against `main`.

This supersedes the closed in-place fix in #318 — see the spec's "Discarded alternatives" section for why this design is the cleaner shape.

## Root cause

The unified BAM pipeline's `Q5` (Process step output) is an unordered FIFO. Worker threads finish position groups in parallel and push completions in arrival order, not pipeline-serial order. The Serialize step then advances a shared `AtomicU64::fetch_add` (`next_mi_base`) inside `serialize_fn`, observing batches in arrival order. Same input → different completion orders → different cumulative `MoleculeId` offsets → different `MI:Z` integers, even though the per-batch contents are identical. `group` and `dedup` are the **only** BAM-pipeline commands that mint new sequential identifiers from shared cumulative state inside `serialize_fn` — every other command emits per-batch byte content fully determined by that batch's input and relies on the existing `write_reorder` to reassemble on disk.

## Architecture

Six commits, each independently reviewable:

1. **`afd59b7f`** `feat(fgumi-umi): add MoleculeId::with_offset` — small pure helper, six unit tests.
2. **`0001162d`** `feat(unified-pipeline): add BatchOrdinal + opt-in mi_assign_fn hook on PipelineFunctions` — new public hook surface; defaults to `None`; no behavior change.
3. **`5ff423e8`** `feat(unified-pipeline): wire mi_assign_fn into single-threaded BAM fast path` — establishes the contract in the simpler code path; no behavior change.
4. **`5c232a6d`** `feat(unified-pipeline): multi-threaded MI Assign zone in serialize step` — the actual scaffolding. Adds `mi_assign_reorder: Mutex<ReorderBuffer<Vec<P>>>` to `OutputPipelineQueues`, branches `try_step_serialize` on `fns.mi_assign_fn`, and adds `run_bam_pipeline_from_reader_with_mi_assign`. Still no command-visible change.
5. **`9ce9b041`** `fix(group,dedup): assign MoleculeIds via serial-ordered MI Assign hook` — the actual user-visible fix. `group`/`dedup` install the hook, drop their `next_mi_base.fetch_add(...)` from `serialize_fn`, and pass `base_mi = 0` to the emitter.
6. **`ee81d3b3`** `test(group,dedup): integration regression for MI numbering determinism + design spec` — wires in the regression test and lands the design doc.

## Why this is deterministic

* The MI Assign zone holds `mi_assign_reorder` for the duration of (drain Q5 → reorder buffer) + (pop next-in-pipeline-serial-order batch) + (invoke hook on every item in monotonic `(batch_serial, idx_in_batch)` order). The mutex serializes only the hook's cumulative-state advance; the actual `serialize_fn` work runs in parallel on the pre-mutated batch outside the lock.
* Within a position group, `assigner.assign(...)` already produces deterministic local IDs (templates are walked in input order).
* Adding a deterministic `base` to deterministic local IDs gives deterministic global IDs. `serialize_fn` reads pre-rewritten MIs verbatim — no synchronization needed.

## Why this is opt-in

Only `group` and `dedup` need cumulative ordering. Wiring the new stage in unconditionally would force a single-threaded coordination point on every other BAM-pipeline command (cost without benefit) and would break the existing `test_pipeline_concurrency` invariants — exactly the failure mode that closed PR #318 hit. With the hook gated on `fns.mi_assign_fn.is_some()`, every other command keeps `mi_assign_fn = None` and takes the unchanged Q5-direct serialize path.

## Test plan

- [x] `cargo ci-test` — 2501 tests pass, 23 skipped, 0 failures.
- [x] `cargo ci-fmt` clean.
- [x] `cargo ci-lint` (clippy `-D warnings -W clippy::pedantic`) clean.
- [x] `tests/integration/test_group_determinism.rs` — three new regression tests, six runs each, per-`(QNAME, flags)` `MI:Z` parity. Verified failing at HEAD~1 (~32k/40k records mismatch on `--threads 4`); verified passing at HEAD (all six runs byte-identical).
- [x] Existing `tests/integration/test_pipeline_concurrency.rs` (13 tests) still passes — the no-hook fallback path is byte-equivalent to the pre-PR behavior.
- [x] `crates/fgumi-umi` unit tests — 123 tests pass (including 6 new `MoleculeId::with_offset` tests).
- [x] Real-data smoke (out of CI): `fgumi zipper → sort → group --threads 8` on SRR6109255 (1.7M records). Pre-fix: 100% of MI tags differed across two runs. Post-fix: byte-identical SAM body across two runs.

## See also

- Design spec: `docs/design/deterministic-mi-numbering.md`
- Closed in-place fix attempt: #318 (rejected for tighter coupling between `serialize_fn` and synchronization primitives)

### DIFF
--- a/crates/fgumi-umi/src/lib.rs
+++ b/crates/fgumi-umi/src/lib.rs
@@ -67,19 +67,31 @@ impl MoleculeId {
     /// Write string representation into a reusable buffer, returning the bytes.
     ///
     /// Avoids per-call String allocation by reusing the caller's buffer.
+    ///
+    /// # Panics
+    ///
+    /// Panics consistently in both debug and release builds if `base + id`
+    /// would overflow `u64`. With `id` and `base` both bounded by the number
+    /// of molecules in the input (which fits comfortably under `u64::MAX` for
+    /// every realistic sequencing run), this is unreachable in practice; the
+    /// explicit `checked_add` keeps debug and release behavior identical
+    /// rather than wrapping silently in release.
     pub fn write_with_offset<'a>(&self, base: u64, buf: &'a mut String) -> &'a [u8] {
         use std::fmt::Write;
         buf.clear();
         match self {
             MoleculeId::None => {}
             MoleculeId::Single(id) => {
-                write!(buf, "{}", base + id).expect("write to String is infallible");
+                let total = id.checked_add(base).expect("MoleculeId offset overflow");
+                write!(buf, "{total}").expect("write to String is infallible");
             }
             MoleculeId::PairedA(id) => {
-                write!(buf, "{}/A", base + id).expect("write to String is infallible");
+                let total = id.checked_add(base).expect("MoleculeId offset overflow");
+                write!(buf, "{total}/A").expect("write to String is infallible");
             }
             MoleculeId::PairedB(id) => {
-                write!(buf, "{}/B", base + id).expect("write to String is infallible");
+                let total = id.checked_add(base).expect("MoleculeId offset overflow");
+                write!(buf, "{total}/B").expect("write to String is infallible");
             }
         }
         buf.as_bytes()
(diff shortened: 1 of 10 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #330, opened 2026-05-03
refactor(unified_pipeline): unify bam.rs and fastq.rs around shared worker/state infrastructure


nh13:
## Background

`src/lib/unified_pipeline/` currently contains:

| File | Lines | Role |
|---|---:|---|
| `base.rs` | 6440 | Shared infrastructure (`OutputPipelineQueues<I,P>`, `PipelineConfig`, `PipelineStats`, traits: `OutputPipelineState`, `OutputQueueDepths`, `MonitorableState`, `PipelineLifecycle`) |
| `bam.rs` | 4813 | BAM-input-specific state, worker loop, step contexts, configs, six entry-point variants |
| `fastq.rs` | 5462 | FASTQ-input-specific state, worker loop, step contexts, config, one entry point |

Both `BamPipelineState` and `FastqPipelineState` already wrap a shared `OutputPipelineQueues<I,P>` (`bam.rs:599`, `fastq.rs:1395`), so the output-side queues, secondary-output support, error tracking, and progress reporting are already unified.

What's still duplicated between `bam.rs` and `fastq.rs`:

- Worker loop drivers (work-stealing, scheduling integration, drain coordination)
- Pipeline state struct outer shape (each wraps `OutputPipelineQueues` plus input-specific state, but the wrapping logic is duplicated)
- Step context structs (`BamStepContext` ↔ `FastqStepContext`)
- Pipeline config structs (`BamPipelineConfig` ↔ `FastqPipelineConfig`) extending `PipelineConfig`
- Error/progress/cancel/draining state-machine glue
- Multiple convenience entry points per pipeline (`run_bam_pipeline_with_grouper`, `_with_header`, `_from_reader`, `_from_reader_with_mi_assign`, `_from_reader_with_secondary` vs the single `run_fastq_pipeline`)
- `PipelineFunctions<G,P>` is BAM-only today; FASTQ has no analog despite needing the same builder pattern (process_fn, serialize_fn, optional secondary_serialize)

What genuinely differs (and shouldn't be merged):

- Input layer: BGZF block reading + boundary finding for binary records (BAM) vs text-line parsing + paired-stream synchronization (FASTQ)
- Grouping: alignment-position-based via `Grouper` trait + `GroupKeyConfig` (BAM) vs read-name template matching across paired streams (FASTQ)
- Intermediate parse types (`FastqDecompressedBatch`, `FastqBoundaryBatch`, `FastqParsedBatch` are FASTQ-only by nature)

[...]

Issue #329, opened 2026-05-03
refactor(consensus,correct): migrate hand-rolled rejects writers to first-class secondary_output


nh13:
## Background

The unified BAM pipeline supports a first-class secondary output via `OutputPipelineQueues::secondary_output` (`src/lib/unified_pipeline/base.rs:1678`) and the convenience entry point `run_bam_pipeline_from_reader_with_secondary` (`src/lib/unified_pipeline/bam.rs:4242`). The first-class path writes secondary records in the same serial order as primary records (see `bam.rs:2986-2998`).

`commands/filter.rs:495` already uses this pattern for failed-filter records.

Four commands still hand-roll an `Arc<Mutex<Option<RawBamWriter>>>` for their rejects BAM, capturing it in their pipeline closure rather than going through the unified pipeline:

- `src/lib/commands/correct.rs:1127`
- `src/lib/commands/simplex.rs:537`
- `src/lib/commands/duplex.rs:615`
- `src/lib/commands/codec.rs:540`

## Why this matters

1. Four near-identical copies of mutex-management boilerplate across consensus commands.
2. Hand-rolled rejects records land in arbitrary order across worker threads (whoever wins the mutex first); first-class secondary preserves the same serial order as the primary BAM. Order matters for any consumer that expects rejects to align with the source order or for byte-identical reproducibility.
3. `fgumi runall`'s `CorrectChainRunner` calls `run_correct_pipeline` (the standalone-shared body that does the hand-rolling). For runall to honor `--correct::rejects` cleanly, `run_correct_pipeline` should use the unified pipeline's first-class secondary. The other three consensus commands should follow for consistency.

## Validation: per-command perf regression check

Each migrated command must be perf-benched before/after the migration on a representative fixture. The hand-rolled `Arc<Mutex<RawBamWriter>>` and the first-class `set_secondary_output` paths have different locking shapes (per-record mutex vs batched serial reorder); this check confirms the new path is no slower.

### Per-command bench inputs

For each migrated command, bench against an input representative of typical use:

- `correct`: an unmapped BAM with UMI tags (`RX`) at workload-realistic scale (e.g. ≥10M templates), with a non-trivial fraction of UMIs that fall outside the allowed list so `--rejects` exercises the secondary output path.
- `simplex` / `duplex` / `codec`: a template-coordinate-sorted BAM after grouping, with `--rejects rejects.bam` enabled.

[...]

PR #328, opened 2026-05-03
feat(runall): add correct and extract→correct chains

nh13:
**Stacked on #327, which is stacked on #326. Review and merge those first.**

PR 3 of 3 in the runall-on-`unified_pipeline` stack. Adds two new `ChainRunner` implementations and registers them ahead of the `NotImplementedRunner` fallback.

## Summary

- **`CorrectChainRunner`** — `(Correct, Correct)` BAM-linear chain. Forwards to a new `crate::commands::correct::run_correct_pipeline` free function (extracted from `CorrectUmis::execute_threads_mode`) that owns the multi-threaded UMI-correction pipeline body.
- **`ExtractCorrectChainRunner`** — `(Extract, Correct)` fused chain on `unified_pipeline::fastq`. Each `FastqTemplate` is converted to BAM bytes via `make_raw_records_static` and per-template UMI correction is applied in-place to those bytes in one closure. Per-worker LRU cache lives in `thread_local!` storage so the closure body stays `Fn` (not `FnMut`).
- Both runners gate themselves off `--metrics`, `--correct::rejects`, `--correct::metrics`, `--correct::min-corrected`, and require a UMI source (`--correct::umis` or `--correct::umi-files`). Anything outside that envelope falls through to `NotImplementedRunner`.

## Enabling refactor in `crate::commands::correct`

Behaviour-preserving — gated by the existing standalone-correct integration tests, all of which still pass:

- Extracts `pub(crate) fn run_correct_pipeline(params: CorrectPipelineParams<'_>) -> Result<u64>` and `pub(crate) fn finalize_correct_metrics(...)` as free functions; `CorrectUmis::execute_threads_mode` and `CorrectUmis::finalize_metrics` become thin wrappers.
- Promotes `compute_template_correction`, `extract_and_validate_template_umi_raw`, `apply_correction_to_raw` to `pub(crate)`; `RejectionReason`, `TemplateCorrection` (with fields), and `EncodedUmiSet::strings()` likewise.

## Parity tests

Both new tests build a synthetic paired-end gzip FASTQ with a UMI cycle that hits both the perfect-match path (4 of 5 UMIs are in the allowed list) and the dropped-record path (`NNNNNNNN` per 5 templates), use `--max-mismatches 0 --min-distance 1`, and diff via `fgumi compare bams --command correct`:

[...]

PR #331, opened 2026-05-03
feat(runall): honor --correct::metrics and --correct::min-corrected in correct chains

nh13:
Stacked on #328. Wires two of the three remaining `--correct::*` flags currently gated off in PR #328's correct chain runners; the third (`--correct::rejects`) stays gated pending upstream architectural work.

## What changes
- `CorrectChainRunner` (BAM-linear) and `ExtractCorrectChainRunner` (fused FASTQ→BAM) honor `--correct::metrics` (per-UMI metrics TSV) and `--correct::min-corrected` (post-pipeline threshold check).
- Both flags use the same machinery as standalone `fgumi correct`: `PerThreadAccumulator<CollectedCorrectMetrics>` + `finalize_correct_metrics` for metrics, accumulator-based count + threshold for min-corrected.
- Drops the `metrics_unset` and `min_corrected_unset` gates in both runners' `supports()`.
- `--correct::rejects` remains gated, but now fails with a specific error pointing at the two upstream tracking issues (rather than the generic "not yet implemented" fallthrough).

## What stays gated and why
`--correct::rejects` would require either (a) migrating standalone correct's hand-rolled `Arc<Mutex<Option<RawBamWriter>>>` to the unified pipeline's first-class `secondary_output` (issue #329) and (b) adding secondary_output support to the FASTQ pipeline for the fused `ExtractCorrectChainRunner` (issue #330's phase 1+). Both are upstream concerns, not runall-scoped. Once those land, a small follow-up runall PR adds rejects support.

## Tests
- 2 perturbation-verified parity tests vs standalone `fgumi correct` (metrics.tsv text-diff; min-corrected exit-code parity at high+low thresholds, plus BAM parity in the success case).
- Unit tests covering the flag-honoring `supports()` flips and the rejects-set specific error.

## Audit
Every public flag on standalone `fgumi correct` is now either wired through runall or routed via global flags. The single remaining gap (`--correct::rejects`) has a specific error pointer to issues #329 and #330; tests cover that error path.

### DIFF
--- a/src/lib/commands/correct.rs
+++ b/src/lib/commands/correct.rs
@@ -329,18 +329,23 @@ impl MemoryEstimate for CorrectProcessedBatch {
 }
 
 /// Metrics collected from UMI correction processing, aggregated post-pipeline.
+///
+/// `pub(crate)` so the runall correct chain runners can build the same
+/// per-thread accumulator the standalone command uses, then call
+/// [`finalize_correct_metrics`] / replicate the `--min-corrected` threshold
+/// check without diverging from this command's behaviour.
 #[derive(Default)]
-struct CollectedCorrectMetrics {
+pub(crate) struct CollectedCorrectMetrics {
     /// Total templates processed.
-    templates_processed: u64,
+    pub(crate) templates_processed: u64,
     /// Records with missing UMI tag.
-    missing_umis: u64,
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #337, opened 2026-05-15
fix(group,dedup,downsample)!: require SS:template-coordinate

nh13:
## Summary

`is_template_coordinate_sorted` (`crates/fgumi-sam/src/lib.rs`) previously treated the SAM `SS` sub-sort tag as optional, so any header advertising `SO:unsorted GO:query` — notably the queryname-grouped output of `fgumi extract`, which has no `SS` — was silently accepted as template-coordinate sorted. Three commands rely on this check: `group`, `dedup`, and `downsample`. They all feed input through a streaming position grouper (`RecordPositionGrouper`) that requires records sharing a position key to be consecutive — a guarantee that template-coordinate sort provides and that queryname-grouped / FASTQ-order / coordinate-sorted input does not.

Two consequences in the wild, reported by a user this week:

- **Performance.** `fgumi group --threads 4 --allow-unmapped` on a queryname-grouped BAM with diverse `CB` tags: the streaming grouper produces one position group per template (~10M tiny groups). Each group spins up its own `rayon::ThreadPool` inside the parallel UMI assigner. On a 3M-record subset, `--threads 4` took **135 s wall with 856 s of sys time**, vs **34 s** for the same input with `--threads 1` — i.e. multi-threaded was *slower* than single-threaded.
- **Correctness risk.** On mapped data, the same input shape would split a real molecule across multiple position groups and assign distinct `MI` integers to reads that should share one — silent wrong groupings, undetectable from logs and unsafe to feed into downstream consensus.

This PR tightens `is_template_coordinate_sorted` to **require** `SS:template-coordinate` (with or without the `<so>:` prefix), matching fgbio's `SamOrder.apply`. It also removes the `--allow-unmapped` queryname-sort shortcut from `fgumi group` (same underlying issue), updates error messages, the clap doc for `--allow-unmapped`, the UMI grouping user-guide, and the fgbio migration guide. `fgumi group` does **not** sort internally (unlike fgbio's `GroupReadsByUmi`); users must pipe through `fgumi sort --order template-coordinate` first. The CHANGELOG entry is flagged `[**breaking**]`.

This PR is correctness-only. The follow-up rayon-pool-per-group performance bug (which still bites legitimate single large-group workloads at `--threads N`) is a separate change.

### Migration

Insert `fgumi sort --order template-coordinate` between any non-template-coordinate source (`fgumi extract`, `samtools sort -n`, `fgumi merge --order queryname`, etc.) and `fgumi group`, `fgumi dedup`, or `fgumi downsample`. The new error message names this remediation directly.

### Other touch-ups bundled in

- Comment in `is_template_coordinate_sorted` claimed "find the last colon" but the code uses `.position()` (first colon); behavior matches fgbio's `indexOf` parsing, so the comment is corrected and a cross-reference added.
- `src/lib/sam/mod.rs` no longer re-exports `is_sorted` (it had no non-test caller after this PR). The function itself stays `pub` in the published `fgumi-sam` crate; `check_sort` still uses it internally.

## Test plan

- [x] `cargo ci-fmt` clean
- [x] `cargo ci-lint` clean
- [x] `cargo ci-test`: 2078 tests pass, 23 skipped — 11 net new test cases:
  - 10-case parametric `test_group_rejects_non_template_coordinate_input` (5 non-TC header shapes × 2 `--allow-unmapped` modes); the `extract_style_no_ss` case is the direct regression for the reported bug
  - `test_group_rejects_extract_output_without_sort_step` spawns the actual `fgumi` binary and asserts exit status + stderr (CLI-level pin, not just in-process)
- [x] Flipped `test_is_template_coordinate_sorted_valid_minimal` (which previously asserted the lenient behavior) to `test_..._missing_ss_is_rejected`
- [x] Updated `test_full_pipeline_extract_to_filter` to insert `fgumi sort --order template-coordinate` between extract and group (the old test was inadvertently exercising the buggy path)
- [x] Manual repro against the user's BAM: extract-style header now produces the expected error pointing at `fgumi sort`; happy path `sort → group --threads 4` succeeds end-to-end

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Bug Fixes
+
+- [**breaking**] `fgumi group`, `fgumi dedup`, and `fgumi downsample` now strictly require template-coordinate sorted input — the header must advertise `SO:unsorted`, `GO:query`, **and** `SS:template-coordinate`. The previous implementation accepted any `SO:unsorted GO:query` header even without `SS:template-coordinate`, which caused `fgumi extract` output (FASTQ-order, no `SS`) and other queryname-grouped BAMs to be silently treated as template-coordinate sorted. Because the streaming `RecordPositionGrouper` requires records sharing a position key to be consecutive, this could split a single true molecule across multiple position groups and assign distinct `MI` tags to reads that should share one. The `--allow-unmapped` queryname-sort shortcut has also been removed for the same reason. To migrate: insert `fgumi sort --order template-coordinate` between extract (or any non-TC source) and `fgumi group`/`dedup`/`downsample`.
+
 ### Refactor
 
 - Extracted the sort engine into the new `fgumi-sort` crate and the BAM-pipeline I/O layer into the new `fgumi-bam-io` crate. The main `fgumi` binary now consumes both as workspace dependencies; behavior is unchanged.
--- a/crates/fgumi-sam/src/lib.rs
+++ b/crates/fgumi-sam/src/lib.rs
@@ -148,9 +148,19 @@ pub fn is_sorted(header: &Header, sort_order: &[u8]) -> bool {
 /// This sort order is indicated in the SAM header by:
 /// - SO:unsorted (not coordinate sorted)
 /// - GO:query (grouped by query name)
-/// - SS:template-coordinate (optional, but must match if present)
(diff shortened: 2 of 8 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #340, opened 2026-05-16
fix(bgzf): honor compression level 0 in InlineBgzfCompressor

nh13:
## Summary

- `InlineBgzfCompressor::new` clamped its input to `1..=12`, so `--temp-compression 0` was silently identical to `--temp-compression 1`. The `sort --temp-compression` help text and `RawExternalSorter::temp_compression` doc both advertise level 0 as *"disables compression (fastest, uses most disk space)"*, so the implementation was the outlier.
- libdeflate (via the `bgzf` crate) already accepts level 0 — it emits stored deflate blocks, producing valid BGZF whose payload is at least as large as the input. Removing the clamp lets the documented behavior actually take effect.
- All call sites that pass a `compression_level: u32` through to `InlineBgzfCompressor::new` (`SortWorkerPool` for both spill and output, and the `unified_pipeline` writers) inherit the fix without further changes.

## Test plan

- [x] New unit test `writer::tests::test_compression_level_zero_is_uncompressed` compresses 32 KiB of zeros at level 0 and level 1. Sanity-asserts that level 1 collapses the input to < 1 KiB, then asserts the level-0 block is at least as large as the input (stored blocks, not RLE'd to a few bytes), then round-trips the level-0 block back to its original bytes.
- [x] `cargo ci-test` — 2067 tests pass.
- [x] `cargo ci-fmt`, `cargo ci-lint` — clean.

## Notes

- An orphaned level-0 raw-write branch in `GenericKeyedChunkWriter::create` (`crates/fgumi-sort/src/external.rs:425`) still exists but has no callers; cleanup is intentionally out of scope for this PR.

### DIFF
--- a/crates/fgumi-bgzf/src/writer.rs
+++ b/crates/fgumi-bgzf/src/writer.rs
@@ -69,16 +69,17 @@ impl InlineBgzfCompressor {
     ///
     /// # Arguments
     ///
-    /// * `compression_level` - Compression level (1-12, higher = smaller but slower).
-    ///   Level 1 is fastest compression, level 6 is a good balance.
+    /// * `compression_level` - Compression level (0-12, higher = smaller but slower).
+    ///   Level 0 disables compression (stored deflate blocks), level 1 is the
+    ///   fastest compression, and level 6 is a good balance. Values outside
+    ///   `0..=12` fall back to level 6.
     ///
     /// # Panics
     ///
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #341, opened 2026-05-16
feat(sort): zstd as the default temp-spill codec

nh13:
## Summary

Adds **zstd** as a temp-spill codec alongside BGZF and flips the default to zstd. Output BAMs are unchanged: always BGZF, mandated by the BAM spec.

Why zstd for spill: zstd-1 is ~3× faster than libdeflate-1 to compress and ~6× faster to decompress at a comparable ratio on BAM-record data, and spill files are internal to the sort (no other tool reads them), so the on-disk format is ours to evolve.

## What changed, by commit

1. `feat(sort): add SpillCodec enum for temp-spill codec selection` — purely additive: enum + magic-byte constants in a new `crates/fgumi-sort/src/codec.rs`, re-exported from `fgumi_sort`. No call sites use it yet.

2. `feat(sort): zstd as a spill codec, end-to-end` — the substantive change:
   - Per-worker `zstd::bulk::Compressor` + `zstd::bulk::Decompressor` on `SortWorkerState`, reused across jobs, mirroring how `InlineBgzfCompressor` and `libdeflater::Decompressor` are handled today. **No zstd internal multithreading** — parallelism comes from the existing N worker pool threads, exactly as for BGZF. Total thread count is unchanged.
   - `CompressJob.codec` dispatches inside `handle_compress_job`: BGZF takes the existing path; zstd emits one self-contained frame per ~64 KiB input, prefixed by a `u32 LE` length.
   - `Phase2FileState.codec` detected via magic bytes in `set_phase2_files`. `raw_blocks` carries `Vec<u8>` instead of `RawBgzfBlock` so both codecs share the queue; the decompress dispatch picks the right per-worker context.
   - **Per-worker zstd decompress scratch buffer** (`zstd_decompress_buf`, 256 KiB) is the key perf detail — using `decompress_to_buffer` instead of `decompress` avoids per-frame `Vec` allocation. Without this, the merge phase regressed ~24 % on the 1-spill case.
   - New `crates/fgumi-sort/src/zspill_stream.rs::ZspillStreamReader<R>` for the legacy `GenericKeyedChunkReader::open` path (consolidation merge + indexing path).
   - `PooledChunkWriter::new_with_codec(...)` writes the 4-byte file magic `"ZSP1"` when codec=Zstd, then a sequence of `[u32 LE compressed-len][zstd frame]` records.
   - `RawExternalSorter::spill_codec(SpillCodec) -> Self` builder added; default remains `Bgzf` after this commit (flag flip lives in commit 4).

[...]

PR #344, opened 2026-05-16
perf+ux: follow-ups from the agilent sort profile

nh13:
## Summary

Two small, independent improvements surfaced by the perf investigation that motivated #341 (zstd spill codec). Each is its own commit so it can be reverted in isolation if needed.

Items deferred to a later PR (require bench validation):
- **B**: `ProgressTracker::log_if_needed` showed up at 1.18 % of total CPU. The function design is fine (cheap atomic + interval check), but at ~32M calls/s aggregate the per-call cost adds up. Needs an EC2-backed bench to confirm the diagnosis before fixing — likely a per-thread accumulator that touches the global atomic less frequently.
- **C**: per-worker BGZF decompress scratch buffer (mirror of the zstd scratch buffer in #341). Largely subsumed by commit 1 once `decompress_block` pre-sizes — the remaining win is the buffer-pool reuse, which is structural enough to deserve its own PR.

### 1. `perf(bgzf): pre-size decompress_block output buffer`

`decompress_block` returned a fresh `Vec::new()` and let `decompress_and_verify` resize it to the block's uncompressed size. Profiles on a 30-second BAM sort showed **~1.9 % of total CPU in `RawVec::finish_grow` / `mi_theap_malloc_zero_aligned_at_overalloc`**, attributed to that resize growing through several capacity classes before reaching ~64 KiB on every block.

`Vec::with_capacity(BGZF_MAX_BLOCK_SIZE)` lands the first allocation at the final size, eliminating the grow steps. Callers that already use `decompress_block_into` / `decompress_block_slice_into` are unaffected.

### 2. `feat(sort): warn when a -T temp dir is on tmpfs`

Many container images and cloud AMIs (notably Amazon Linux 2023) ship `/tmp` as a tmpfs sized at 50 % of RAM. Spilling to tmpfs makes every byte spilled come back out of RAM, so peak RSS climbs as if `--max-memory` were ignored and the process OOMs on small hosts. The misbehavior is invisible from the inside: free space looks fine, writes succeed, only the working-set numbers reveal the trap.

`TmpDirAllocator::with_probe` now calls `statfs(2)` on each configured temp directory (via the already-vendored `nix` crate on Linux) and emits a `log::warn!` when it detects tmpfs. The sort continues normally; this is informational only.

Non-Linux targets fall through to a stub that always returns false, since `f_type == TMPFS_MAGIC` is a Linux concept.

## Test plan

- [x] `cargo ci-lint` passes (no `--no-verify` needed).
- [x] `cargo ci-fmt` passes.
- [x] `cargo test --release -p fgumi-sort` — 342 unit + 11 integration tests pass.
- [x] `cargo test --release -p fgumi-bgzf` — 18 tests pass.

## Notes

- Independent of #340 and #341; the three can land in any order.
- The pre-size change is one line of substantive code; the tmpfs check is ~30 lines plus a helper.

### DIFF
--- a/crates/fgumi-bgzf/src/reader.rs
+++ b/crates/fgumi-bgzf/src/reader.rs
@@ -320,7 +320,13 @@ pub fn decompress_block(
     block: &RawBgzfBlock,
     decompressor: &mut Decompressor,
 ) -> io::Result<Vec<u8>> {
-    let mut output = Vec::new();
+    // Pre-size to one BGZF block. Avoids the multi-step `RawVec::finish_grow`
+    // we see in profiles (~1.9% of decompress CPU on the merge hot path):
+    // starting from capacity 0, the resize inside `decompress_and_verify` has
+    // to grow through several capacity classes before reaching the final
+    // ~64 KiB. A single `with_capacity` lands at the right size on the first
+    // allocation.
+    let mut output = Vec::with_capacity(crate::BGZF_MAX_BLOCK_SIZE);
     decompress_block_into(block, decompressor, &mut output)?;
     Ok(output)
 }
--- a/crates/fgumi-sort/src/tmp_dir_alloc.rs
+++ b/crates/fgumi-sort/src/tmp_dir_alloc.rs
@@ -10,7 +10,32 @@
 //!   below a minimum threshold are dropped from rotation automatically.
 
 use anyhow::{Result, anyhow, bail};
+use log::warn;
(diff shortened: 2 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #346, opened 2026-05-16
perf(sort): recycle BGZF decompress output buffers via a pre-warmed pool

nh13:
## Summary

Picks up item **C** that PR #344 explicitly deferred: a per-worker BGZF decompress scratch pool. PR #344 commit 1 pre-sizes `decompress_block`'s output `Vec`; this PR makes that buffer pooled and reused across blocks instead of freshly allocated per call.

Stacked on `nh/perf-followups` (PR #344). When #344 merges, this PR's base auto-retargets to `main`.

## Motivation

The fgumi sort pool's Phase 1 (`worker_pool::try_decompress_input`) and Phase 2 (`worker_pool::try_phase2_file_work`) both call `decompress_block(&block, &mut decompressor)`, which returns a fresh `Vec<u8>` per BGZF block. PR #344 commit 1 sized the first allocation correctly via `Vec::with_capacity(BGZF_MAX_BLOCK_SIZE)`, but every block still pays:

- one `malloc` per block (~25 K blocks for a 1.7 GB BAM),
- one `clear_page_erms` first-touch fault when the OS hands us a fresh anonymous page (~2 % of total CPU in the past agilent-hs2 sort profile),
- a matching `mi_free` when the consumer drops the Vec on a different thread (the cross-thread page-ownership-transfer case mimalloc handles via `mi_abandoned_page_try_reclaim`).

## Change

A second `BufferPool` (`SortWorkerPool::decompress_buffer_pool`) alongside the existing compress staging pool, pre-warmed at `BGZF_MAX_BLOCK_SIZE`. Workers checkout before `decompress_block_into(&mut buf)`; consumers (`PooledInputStream` and `MainThreadChunkConsumer`) check the previous block's buffer back in when they advance.

- Pool size: `num_workers * 32` slots. Sized for the Phase 1 working set (`N (workers) + N*8 (decompressed_input cap) + N*16 (reorder buf soft cap) + 1 (current_buf) = N*25 + 1`) with ~25 % margin.
- Phase 2 with K > ~`N*3` spill files: excess decompressions fall back to the empty-Vec path — same behavior as before this PR, graceful degradation.
- Memory budget: `pool_capacity × BGZF_MAX_BLOCK_SIZE` ≈ 2 MiB / worker. 16 MiB at `--threads 8`, 64 MiB at `--threads 32`.
- Error paths return the buffer to the pool before propagating.
[...]

PR #347, opened 2026-05-16
perf(sort): share SegmentedBuf across in-memory sort chunks via Arc

nh13:
## Summary

Eliminates the per-record `RawRecord::from(buffer.get_record(r).to_vec())` allocation when materializing in-memory sort chunks. For a 16 M-record sort that was ~16 M `mi_malloc`/memcpy/`mi_free` cycles and a transient ~2× peak in materialization memory.

Stacked on `nh/perf-decompress-buffer-pool` (PR #346). When #346 merges, this PR's base auto-retargets to `main`.

## Mechanism

A new `InMemoryChunk<K>` whose records borrow their bytes from a single `Arc<SegmentedBuf>` shared across all sibling chunks produced by one `par_sort_into_chunks` call. Each chunk stores `(sort_key, byte_offset_in_data, len)` triples; the merge consumer copies bytes via `extend_from_slice` (one memcpy per record, same as before, but without the matching `malloc`/`free`).

- The source `RecordBuffer`'s `SegmentedBuf` is moved into an `Arc` via `mem::take(&mut self.data)`; `self.refs` is also cleared so any stale `get_record` / `iter_sorted` call fails-fast at the index bound rather than silently slicing an empty buffer.
- Sibling chunks share one `Arc` (verified by a test asserting `Arc::strong_count` and `Arc::as_ptr` equality).
- The queryname streaming path is **unchanged**: it accumulates records upstream as individual `RawRecord`s (per-record alloc is sunk cost), so a second `ChunkSource::MemoryOwned` variant preserves the original zero-copy `mem::swap` merge bridge. An earlier prototype converted queryname to bulk-pack and showed a regression — hence the separate variant.
- A `MemorySources<K>` enum wraps both shapes at the merge-call boundary; `build_chunk_sources` dispatches.

## Benchmarks

`twist-umi.mapped.bam` (1.7 GB / 16.1 M records), `--threads 8`, hyperfine.

**In-memory sort** (no spill, 5 runs):

| Variant | Wall time | σ |
| --- | --- | --- |
| Parent (PR #346) | 4.471 s | ± 0.023 s |
| This PR | **4.266 s** | **± 0.012 s** |

→ 1.05 ± 0.01× faster; σ tightens 2× (allocation jitter removed).

**Smaller workload** (1 M records, 7 runs):
- Parent: 296.4 ms ± 3.3 ms
- This PR: 285.7 ms ± 7.8 ms → 1.04× faster

## Memory peak

Materialization peak drops by ~1×: the pre-PR design transiently held both the source `SegmentedBuf` AND per-record `Vec<u8>` copies. The shared `Arc<SegmentedBuf>` keeps the buffer resident for the entire merge phase instead of allowing per-record incremental free, so steady-state in-merge RSS is slightly higher than the pre-PR end-of-merge RSS; the materialization saving dominates.

## Notes

- `InMemoryChunk` is `pub(crate)` — used only inside `fgumi-sort`.
- Offsets are `u64` (a single `SegmentedBuf` can exceed 4 GiB at high `--max-memory` × `--threads`); lengths remain `u32`.
- Dead `merge_chunks_for_test` helper (no callers) removed in the same commit.
- The signing key was unavailable at commit time; the commit is unsigned. Will re-sign / amend on rebase if needed.

## Test plan

- [x] `cargo ci-fmt` passes.
- [x] `cargo ci-test` — 2067 / 2067 pass.
- [x] Hyperfine bench on `twist-umi.mapped.bam` (in-memory).

Six new unit tests in `crates/fgumi-sort/src/inline.rs`:
- `test_in_memory_chunk_from_parts_reads_records` — basic round-trip.
- `test_in_memory_chunk_take_key_replaces_with_default` — merge consumer semantics.
- `test_in_memory_chunk_shared_arc_across_sibling_chunks` — `Arc::strong_count` decrements as chunks drop.
- `test_in_memory_chunk_empty_default` — empty/default behavior.
- `test_drain_into_single_chunk_round_trips_record_bytes` — end-to-end: insert records into a real `RecordBuffer`, sort, drain, assert `record_bytes(i)` round-trips the input multiset (catches HEADER_SIZE off-by-one).
- `test_par_sort_into_chunks_shares_arc_across_siblings` — 4-thread path returns chunks that all share one `Arc<SegmentedBuf>`.

### DIFF
--- a/crates/fgumi-sort/src/external.rs
+++ b/crates/fgumi-sort/src/external.rs
@@ -23,7 +23,9 @@
 //! 3. Sort by keys while keeping raw records
 //! 4. Write raw record bytes to output
 
-use crate::inline::{ProbeableBuffer, RecordBuffer, TemplateKey, TemplateRecordBuffer};
+use crate::inline::{
+    InMemoryChunk, ProbeableBuffer, RecordBuffer, TemplateKey, TemplateRecordBuffer,
+};
 use crate::keys::{QuerynameComparator, RawSortKey, SortOrder};
 use crate::memory_probe::{
     BufferProbeStats, ConsumerProbeStats, MergeProbe, SpillProbe, force_mi_collect, log_snapshot,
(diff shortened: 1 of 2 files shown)


nh13:
Superseded by #350, which contains the identical Arc<SegmentedBuf> change branched directly off `main` (no dependency on #346). Closing this stacked version in favor of the standalone PR.

PR #348, opened 2026-05-16
perf(sort): pool raw BGZF block buffers on input read

nh13:
## Summary

Adds a recycling pool for the *raw (compressed)* BGZF block buffers on input read, symmetric to PR #346's pool for decompressed-output buffers. Previously every `read_raw_blocks` call allocated one fresh `Vec<u8>` per block; for the twist-umi sort that's ~27 K `mi_malloc` + matching `mi_free` cycles on the input side, mirror-image of the decompress side PR #346 already addressed.

Stacked on `nh/perf-no-record-tovec-chunks` (PR #347). When #347 merges, this PR's base auto-retargets to `nh/perf-decompress-buffer-pool` (PR #346); when #346 merges, to `main`.

### `fgumi-bgzf` additions

- `pub fn read_raw_block_into<R, F>(reader, get_buf: F)` — buffer-recycling variant. `get_buf` is invoked **only after** the 18-byte block header is successfully read; clean EOF returns `Ok(None)` without ever calling `get_buf`, so the pool is never charged for a wasted checkout.
- `pub fn read_raw_blocks_into<R, F, G>(reader, max_blocks, get_buf: F, recycle: G)` — batched variant. The extra `recycle` callback handles the BGZF EOF-marker block at end-of-stream: its buffer is consumed via `get_buf` (to read the trailer) but is **not** returned in the result `Vec`, so without `recycle` the pool would silently lose one slot per stream. Across Phase 2's K spill files that's a K-buffer drain per sort.
- `pub fn RawBgzfBlock::take_data(&mut self) -> Vec<u8>` — extracts the block's owned buffer for the consumer-side checkin.
- `read_raw_block` (no-pool single-block read) is now `#[cfg(test)]` — only test callers remain; production goes through `read_raw_block_into`.

### `fgumi-sort` changes

- New `SharedPipelineState::raw_block_buffer_pool` (pre-warmed at `BGZF_MAX_BLOCK_SIZE`, size `num_workers * 32` — same sizing as the decompress pool).
- Phase 1 `try_read_input_blocks` and Phase 2's chunk-file read step both call `read_raw_blocks_into` with `|| pool.checkout()` and `|buf| pool.checkin(buf)`.
- The decompression workers (`try_decompress_input` and `try_phase2_file_work`) check the raw buffer back in via `pool.checkin(block.take_data())` after `decompress_block_into` completes (and on the error path before propagating the error flag).
- The `raw_block_buffer_pool` field lives **only** on `SharedPipelineState` (not on `SortWorkerPool`), because no external consumer needs the pool — workers access via shared state. This is the only asymmetry with `decompress_buffer_pool`, which `PooledInputStream` and `MainThreadChunkConsumer` access via a public-crate getter.

## Benchmark

`twist-umi.mapped.bam` (1.7 GB / 16.1 M records), `--threads 8`, in-memory sort, 7 runs each via hyperfine:

| Variant | Wall time | σ |
| --- | --- | --- |
| PR #347 base | 4.480 s | ± 0.367 s |
| This PR | **4.390 s** | **± 0.070 s** |

→ 1.02× faster; **σ tightens 5×**.

The wall-time delta is small (input-side allocations are not on the critical path for an in-memory sort) but the variance reduction is the cleanest structural signal — removing input-side allocations eliminates a source of jitter independent of PR #346.

## Memory

Combined PR #346 + this PR pre-warm: `num_workers * 64 * BGZF_MAX_BLOCK_SIZE` ≈ 32 MiB at `--threads 8`, 128 MiB at `--threads 32`. Documented at the field definitions.

## Edge cases

- Mid-block I/O errors (truncated input) drop the checked-out buffer (not routed through `recycle`) — acceptable because such errors abort the entire sort. Documented in the `read_raw_blocks_into` docstring.
- The BGZF EOF marker is detected via `block.is_eof()` after construction (28-byte fixed marker). The buffer is recycled before the block is dropped.
- `block.take_data()` leaves the block with `Vec::new()` (capacity 0); no callsite reads `RawBgzfBlock::data` after `take_data`. Verified by trace.

## Tests

- `cargo ci-fmt`, `cargo ci-lint`, `cargo ci-test` — 2067 / 2067 pass.
- Four new tests in `crates/fgumi-bgzf/src/reader.rs`:
  - `test_read_raw_block_into_reuses_buffer`: pointer-equality proof that the checked-out `Vec` becomes the block's `data` field.
  - `test_read_raw_block_into_skips_callback_on_eof`: `get_buf` is not invoked on clean EOF.
  - `test_take_data_extracts_owned_buffer`: `take_data` semantics.
  - `test_read_raw_blocks_into_recycles_eof_marker_and_iterates_get_buf`: end-to-end with 3 real blocks + EOF marker; asserts `get_buf` called once per block, `recycle` called once for the EOF marker, returned `Vec` has 3 payload blocks, recycled buffer preserves capacity, payloads decompress to original bytes.

## Test plan

- [x] `cargo ci-fmt` passes.
- [x] `cargo ci-test` — 2067 / 2067 pass.
- [x] Hyperfine bench on twist-umi (in-memory).

## Notes

- The signing key was unavailable at commit time; the commit is unsigned. Will re-sign / amend on rebase if needed.

### DIFF
--- a/crates/fgumi-bgzf/src/lib.rs
+++ b/crates/fgumi-bgzf/src/lib.rs
@@ -12,6 +12,6 @@ pub mod writer;
 // Re-export commonly used types
 pub use reader::{
     BGZF_EOF, BGZF_FOOTER_SIZE, BGZF_HEADER_SIZE, RawBgzfBlock, decompress_block,
-    decompress_block_into, decompress_block_slice_into, read_raw_blocks,
+    decompress_block_into, decompress_block_slice_into, read_raw_blocks, read_raw_blocks_into,
 };
 pub use writer::{BGZF_MAX_BLOCK_SIZE, CompressedBlock, InlineBgzfCompressor};
--- a/crates/fgumi-bgzf/src/reader.rs
+++ b/crates/fgumi-bgzf/src/reader.rs
@@ -132,22 +132,41 @@ impl RawBgzfBlock {
     pub fn crc32(&self) -> u32 {
         crc32_from_slice(&self.data)
     }
+
+    /// Take ownership of the block's raw byte buffer, leaving the
+    /// block with an empty `Vec<u8>`.
+    ///
+    /// Use this to return the underlying allocation to a recycling
+    /// buffer pool once the block has been decompressed.
+    #[must_use]
+    pub fn take_data(&mut self) -> Vec<u8> {
+        std::mem::take(&mut self.data)
+    }
 }
 
 // ============================================================================
 // Reading Functions
 // ============================================================================
 
-/// Read a single raw BGZF block from the input.
+/// Read a single raw BGZF block from the input, sourcing its byte
+/// storage from `get_buf` (only invoked once a block header is
+/// successfully read).
 ///
-/// Returns `Ok(Some(block))` if a block was read, `Ok(None)` at EOF,
-/// or an error if reading failed or the data is invalid.
+/// This is the buffer-recycling variant of [`read_raw_block`]: callers
+/// can pass a closure that pulls from a `Vec<u8>` recycling pool to
+/// avoid the per-block `mi_malloc` that the convenience entry point
+/// pays. On clean EOF (`Ok(None)`) the closure is *not* called, so the
+/// pool is never charged for a wasted checkout.
 ///
 /// # Errors
 ///
 /// Returns an error if the block header is invalid or reading fails.
-fn read_raw_block<R: Read + ?Sized>(reader: &mut R) -> io::Result<Option<RawBgzfBlock>> {
-    // Read the 18-byte header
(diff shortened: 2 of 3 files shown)


nh13:
Closing. EC2 c7i.4xlarge bench (`agilent-hs2.aligned.bam` and `wgs-5M.bam` × in-memory & spilling, 5–7 runs each) showed no measurable improvement over its parent on any of 4 workloads — all deltas inside ±0.5 % wall time, consistently within σ. The variance-reduction story from macOS did not reproduce on Linux either (σ comparable to parent across the matrix).

If a future profile shows the input-side raw-block allocations (`mi_malloc` for the compressed bytes Vec) as a meaningful slice of CPU on a real workload, this can be reopened. As-is, no signal.

PR #350, opened 2026-05-16
perf(sort): share SegmentedBuf across in-memory sort chunks via Arc

nh13:
## Summary

Eliminates the per-record `RawRecord::from(buffer.get_record(r).to_vec())` allocation when materializing in-memory sort chunks. For an agilent-hs2 sort (~46 M records) that was ~46 M `mi_malloc` + memcpy + matching `mi_free` cycles plus a transient ~2× memory peak during materialization.

Independent of the rest of the perf stack (#341 / #344 / #346 / #347 / #348) — this PR is branched directly off `main` so it can land standalone. (The same change is also commit-2 of #347 stacked on #346; if the stack lands, this PR becomes redundant and can be closed.)

## Mechanism

A new `InMemoryChunk<K>` whose records borrow their bytes from a single `Arc<SegmentedBuf>` shared across all sibling chunks produced by one `par_sort_into_chunks` call. Each chunk stores `(sort_key, byte_offset_in_data, len)` triples; the merge consumer copies bytes via `extend_from_slice` (one memcpy per record, same as before, but without the matching `malloc`/`free`).

- The source `RecordBuffer`'s `SegmentedBuf` is moved into an `Arc` via `mem::take(&mut self.data)`; `self.refs` is also cleared so any stale `get_record` / `iter_sorted` call fails-fast at the index bound rather than silently slicing an empty buffer.
- Sibling chunks share one `Arc` (verified by a test asserting `Arc::strong_count` and `Arc::as_ptr` equality).
- The queryname streaming path is **unchanged**: it accumulates records upstream as individual `RawRecord`s (per-record alloc is sunk cost), so a second `ChunkSource::MemoryOwned` variant preserves the original zero-copy `mem::swap` merge bridge. An earlier prototype converted queryname to bulk-pack and showed a regression — hence the separate variant.
- A `MemorySources<K>` enum wraps both shapes at the merge-call boundary; `build_chunk_sources` dispatches.

## Benchmark

Bench host: EC2 `c7i.4xlarge` (Intel Xeon Platinum 8488C / Sapphire Rapids, AVX-512), Amazon Linux 2023, us-east-1b. Workload: `agilent-hs2.aligned.bam` (3.85 GiB / ~46 M records). `--order coordinate --threads 16`. Hyperfine 3 warmup + 7 runs.

| Variant | Mean | σ | Min | Max |
| --- | ---:| ---:| ---:| ---:|
| `main` | 23.655 s | ± 0.251 s | 23.313 s | 24.093 s |
| **this PR** | **22.908 s** | **± 0.187 s** | 22.675 s | 23.205 s |

→ **1.03 ± 0.01× faster (3.2 %).**

**The ranges do not overlap** — PR's max (23.205 s) is below main's min (23.313 s) across 7 runs. This is reproducible, not noise.

Variance also tightens (σ 0.187 vs 0.251), consistent with the removal of per-record allocator jitter.

## Memory

Materialization-peak memory drops by ~1× — the pre-PR design transiently held both the source `SegmentedBuf` AND per-record `Vec<u8>` copies. The shared `Arc<SegmentedBuf>` keeps the buffer resident for the entire merge phase (steady-state in-merge RSS is slightly higher than pre-PR end-of-merge RSS), but the materialization saving dominates.

## Tests

- `cargo ci-fmt`, `cargo ci-lint`, `cargo ci-test` — 2092 / 2092 pass.
- Six new unit tests in `crates/fgumi-sort/src/inline.rs`:
  - `test_in_memory_chunk_from_parts_reads_records` — basic round-trip.
  - `test_in_memory_chunk_take_key_replaces_with_default` — merge consumer semantics.
  - `test_in_memory_chunk_shared_arc_across_sibling_chunks` — `Arc::strong_count` decrements as chunks drop.
  - `test_in_memory_chunk_empty_default` — empty/default behavior.
  - `test_drain_into_single_chunk_round_trips_record_bytes` — end-to-end: insert records into a real `RecordBuffer`, sort, drain, assert `record_bytes(i)` round-trips the input multiset (catches `HEADER_SIZE` off-by-one).
  - `test_par_sort_into_chunks_shares_arc_across_siblings` — 4-thread path returns chunks that all share one `Arc<SegmentedBuf>`.

## Test plan

- [x] `cargo ci-fmt` passes.
- [x] `cargo ci-test` — 2092 / 2092 pass.
- [x] Hyperfine bench on EC2 c7i.4xlarge against `main`.

## Notes

- `InMemoryChunk` is `pub(crate)` (only used inside `fgumi-sort`).
- Offsets are `u64` (a single `SegmentedBuf` can exceed 4 GiB at high `--max-memory` × `--threads`); lengths remain `u32`.
- After `par_sort_into_chunks` / `drain_into_single_chunk` returns, both `self.data` and `self.refs` are empty; the buffer is intended to be dropped immediately by the caller — `get_record` and `iter_sorted` will index-panic if called.
- The dead `merge_chunks_for_test` helper (no callers) is removed in the same commit.

### DIFF
--- a/crates/fgumi-sort/src/external.rs
+++ b/crates/fgumi-sort/src/external.rs
@@ -23,7 +23,9 @@
 //! 3. Sort by keys while keeping raw records
 //! 4. Write raw record bytes to output
 
-use crate::inline::{ProbeableBuffer, RecordBuffer, TemplateKey, TemplateRecordBuffer};
+use crate::inline::{
+    InMemoryChunk, ProbeableBuffer, RecordBuffer, TemplateKey, TemplateRecordBuffer,
+};
 use crate::keys::{QuerynameComparator, RawSortKey, SortOrder};
 use crate::memory_probe::{
     BufferProbeStats, ConsumerProbeStats, MergeProbe, SpillProbe, force_mi_collect, log_snapshot,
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #343, opened 2026-05-16
perf(group): cache UMI position to drop redundant aux scan (#334)

nh13:
## Summary

Resolves #334. `fgumi_raw_bam::find_tag_position` accounts for ~5% of CPU time on `fgumi group` because the same record's aux data is scanned independently by the Decode step (RG/CB/MC), the filter step (MQ/UMI), and the UMI-assignment step (UMI again). This change caches the UMI tag's value position on each `DecodedRecord` during the existing parallel Decode pass and slices the value out of that cache in the assignment step, eliminating the second `find_tag_position` walk per template.

- New `fgumi_raw_bam::find_string_tag_position(aux, tag) -> Option<(u32, u16)>` returns the offset (within aux data) of a Z-tag's value bytes and its length, without the trailing NUL — useful to anything that wants to cache a Z-tag position rather than borrow its slice.
- `GroupKeyConfig` gains an opt-in `umi_tag: Option<[u8; 2]>` and a `with_umi_tag` builder.
- `DecodedRecord` carries `umi_value_offset: u32` (sentinel `UMI_OFFSET_UNCACHED = u32::MAX`) and `umi_value_len: u16`, populated by `decode_records` when `umi_tag` is set.
- `Template::from_decoded_records` picks the cached UMI from the primary read (R1, falling back to R2; secondary / supplementary records ignored) and stores it in `cached_umi_position`. `Template::cached_umi()` slices the primary record bytes to recover the UMI without rescanning. `Template::from_records` (used by tests and non-group callers) leaves the cache `None`, so `assign_umi_groups_for_indices_impl` falls back to `find_string_tag` exactly as before.
- `fgumi group` opts in via `GroupKeyConfig::new(..).with_umi_tag(raw_tag)`.

## Test plan

- [x] `cargo ci-fmt`
- [x] `cargo ci-lint`
- [x] `cargo ci-test` — 2078 tests pass, including the existing `test_paired_mi_deterministic` matrix that exercises `fgumi group` end-to-end with multiple thread configurations
- [x] New unit tests:
  - `tags::tests::test_find_string_tag_position_*` (6 cases: first/subsequent tag, empty value, absent, non-Z, missing NUL)
  - `unified_pipeline::base::tests::test_decoded_record_cached_umi_{default_absent,round_trip,out_of_range_returns_none}`
[...]

PR #351, opened 2026-05-18
perf(decode,dedup): single-pass aux scan captures UMI position (#334)

### LINKED ISSUE #334: perf(tags): cache UMI position in DecodedRecord to avoid per-record duplicate find_tag_position scan
`fgumi_raw_bam::tags::find_tag_position` accounts for ~5% of CPU time on `fgumi group` (4.94% measured on a CODEC 8M-record samply profile, threads=4). The function itself is well-optimized (`u16` tag compare, early-exit on missing length sentinels), and the existing single-pass extractors (`extract_aux_string_tags`, `extract_template_aux_tags`, `filter_template_raw`'s inline loop) already amortize multi-tag lookups within a single function call.

The remaining cost is structural: the **same record's aux data is scanned multiple times across the chain** because each step looks up tags independently:

| Site | Tags looked up |
|---|---|
| `compute_group_key_from_raw` (`bam.rs:448`) — Decode | RG, CB, MC (single-pass via `extract_aux_string_tags`) |
| `filter_template_raw` (`group.rs:170-233`) — Process | MQ, RX (single-pass inline loop) |
| `assign_umi_groups_for_indices_impl` (`group.rs:467`) — Process | RX (separate `find_string_tag` call) |
| `update_string_tag` via `emit_templates_raw_with_mi` (`group.rs:2057`) — Serialize | MI (read-modify-write) |

For an 8M-record `fgumi group` run that's ~3 separate aux-data scans per record on top of the single-pass extractors. RX in particular is scanned twice (once during filter, once during UMI assignment).

**Proposed fix.** Add `umi_offset: u32` and `umi_len: u16` to `DecodedRecord` (or the equivalent intermediate carrier), populated once in the Decode step alongside the `GroupKey`. Downstream `assign_umi_groups_*` reads the UMI bytes by slice into the cached offset/len without a second `find_tag_position` call. Estimated saving: ≈100 samples on `fgumi group`-like workloads.

**Verification approach.** Profile-diff `target/profiling/fgumi group` with samply at 999 Hz before/after on `agilent-hs2.mapped.bam` (53M records); expect `find_tag_position` (and its callers in `assign_umi_groups_for_indices_impl`) to drop by roughly half their current contribution.

nh13:
## Summary

Resolves #334. Two improvements landed together because the smaller one revealed the larger one:

1. **`fgumi dedup` opts in to the per-record UMI position cache** (#343 infrastructure) and reads via `template.cached_umi()` in its UMI-assignment loop with a fallback to `find_string_tag`. Skipped in `--no-umi` mode (the assignment site emits `String::new()` and never reads the cache there).

2. **The cache populate is folded into `extract_aux_string_tags`**, so a single linear walk over each record's aux bytes captures RG + cell + MC + UMI position together. The previous separate `cache_umi_position` walk and helper function are deleted. `aux_data_offset_from_record` is now called exactly once per record on the decode hot path.

Both improvements affect every command that consumes the UMI cache via the parallel pipeline — `fgumi group` (already opted in via #343) and `fgumi dedup` (opted in here).

Output BAM is byte-identical pre/post on dedup: `diff <(samtools view) <(samtools view) | wc -l` = 0 across all 53.1M records on the test workload.

## Mechanics

- `AuxStringTags` gains `umi_position: Option<(u32, u16)>` (aux-relative value offset + length, NUL excluded).
- `extract_aux_string_tags` takes `umi_tag: Option<[u8; 2]>`; the inner loop captures the UMI position in the same branch ladder as RG/cell/MC; early-exit mask widens from 7 to 15 when `umi_tag` is requested.
- `compute_group_key_from_raw` returns `(GroupKey, Option<(u32, u16)>)`; `decode_records` converts aux-relative to record-relative and stores via `decoded.set_cached_umi(...)`.
- Single-threaded `group` fast path (`commands/group.rs:1408`) passes `None` for `umi_tag` — it doesn't populate the cache and didn't before.

## Measurements

Workload: 2.0 GB template-coordinate-sorted grouped BAM (53.1M records, 26.6M templates), `--threads 4`, `profiling` profile binary, warm-cache tricord runs.

### `fgumi dedup` (3-run median; runs within 0.02s wall)

| Metric | Parent (#343 tip) | This PR | Δ |
|---|---|---|---|
| wall | 17.82s | **17.67s** | **–0.8%** |
| cpu_time | 68.83s | **67.84s** | **–1.4%** |
| max_rss | 1435 MB | 1435 MB | flat |

### `fgumi group` (8-run statistics)

| Metric | Parent (#343 tip) | This PR | Δ |
|---|---|---|---|
| wall min | 17.67s | 17.66s | flat |
| wall median | 17.92s | 18.91s | (noise — bimodal due to system interference) |
| cpu_time median | 68.36s | 69.31s | (within run-to-run variance) |

`fgumi group`'s improvement is below the noise floor on this workload. Best-case runs match parent (~17.66s). The per-record CPU saving from eliminating the redundant aux walk remains architecturally real (one fewer `aux_data_offset_from_record` call + one fewer inline scan per record) and may be detectable on larger or less-noisy environments.

## Test plan

- [x] `cargo ci-fmt` / `cargo ci-lint` / `cargo ci-test` (2101/2101 pass, 23 skipped)
- [x] Output BAM byte-identical (above)
- [x] `aux_data_offset_from_record` called exactly once per record on the decode hot path (previously twice)

### DIFF
--- a/benches/raw_bam.rs
+++ b/benches/raw_bam.rs
@@ -209,7 +209,7 @@ fn bench_tags_read(c: &mut Criterion) {
     c.bench_function("tags::extract_string_batch", |b| {
         b.iter(|| {
             let v = RawRecordView::new(black_box(&bytes));
-            black_box(v.tags().extract_string_batch(b"CR"))
+            black_box(v.tags().extract_string_batch(b"CR", None))
         });
     });
 }
--- a/crates/fgumi-raw-bam/src/tags.rs
+++ b/crates/fgumi-raw-bam/src/tags.rs
@@ -256,15 +256,28 @@ pub struct AuxStringTags<'a> {
     pub rg: Option<&'a [u8]>,
     pub cell: Option<&'a [u8]>,
     pub mc: Option<&'a str>,
+    /// Aux-relative `(value_offset, value_len)` of the UMI tag's value bytes
+    /// (excluding NUL terminator), when the caller supplied a `umi_tag` to look
+    /// for. `None` when no `umi_tag` was supplied, or when the tag is absent
+    /// or non-Z-typed. Aux-relative — to convert to record-relative for use
+    /// with `update_string_tag_at_position` etc., add the record's aux offset.
+    pub umi_position: Option<(u32, u16)>,
 }
 
-/// Extract RG, cell barcode, and MC tags in a single pass over the aux data.
+/// Extract RG, cell barcode, MC, and optionally UMI tag position in a single
+/// pass over the aux data.
 #[inline]
 #[must_use]
-pub fn extract_aux_string_tags(aux_data: &[u8], cell_tag: impl AsTagBytes) -> AuxStringTags<'_> {
+pub fn extract_aux_string_tags(
+    aux_data: &[u8],
+    cell_tag: impl AsTagBytes,
+    umi_tag: Option<[u8; 2]>,
+) -> AuxStringTags<'_> {
     let cell_tag = cell_tag.as_tag_bytes();
-    let mut result = AuxStringTags { rg: None, cell: None, mc: None };
-    let mut found = 0u8; // bit 0=RG, bit 1=cell, bit 2=MC
+    let mut result = AuxStringTags { rg: None, cell: None, mc: None, umi_position: None };
+    let mut found = 0u8; // bit 0=RG, bit 1=cell, bit 2=MC, bit 3=UMI
+    // Early-exit mask: only require UMI if the caller asked for it.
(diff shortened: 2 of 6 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

PR #359, opened 2026-05-20
fix(duplex-metrics): drop dense (max_ab × max_ba) grid that OOMs on cfDNA hot spots

nh13:
## What this fixes

`fgumi duplex-metrics` OOMs on real cfDNA-class inputs (`vendor`-profile samples — `idt-cfdna`, `kapa-umi`, `agilent-hs2`, `schmitt-abl1`, `twist-umi`) on a 30 GiB `c7g.4xlarge`. The same machine processes the simplex-metrics counterpart of the same files without issue.

The OOM happens **after** the streaming pass: every affected job logs `Processed N templates` and then gets OOM-killed during the per-collector yield-metric pass, at the exact same template count run after run.

### Real cause

`DuplexMetricsCollector::duplex_family_size_metrics()` computed the 2D cumulative `fraction_gt_or_eq_size` by materializing a dense `(max_ab + 1) × (max_ba + 1) × usize` grid and running 2D suffix sums over it. The underlying `duplex_family_sizes: HashMap<(usize, usize), usize>` is **sparse** — typically well under a thousand distinct `(ab, ba)` pairs even on cfDNA-scale inputs — but the grid is sized by the *maxima* of the two dimensions. On a cfDNA sample whose hottest duplex family lands in the tens of thousands of reads per strand, this allocates billions of cells (`max_ab = max_ba = 50_000` → ~20 GiB scratch) before any metric is computed. With 20 collectors and `generate_yield_metric` calling `duplex_family_size_metrics()` per collector, you OOM before producing any output.

## Changes

* **`crates/fgumi-metrics/src/duplex.rs::duplex_family_size_metrics`** — replace the dense-grid 2D suffix-sum with a sparse `O(N²)` iteration over the actual `(key, count)` HashMap entries, where `N` is the number of distinct `(ab, ba)` pairs (typically <1000, dwarfed by upstream I/O cost). No more `max_ab × max_ba` allocation.
* **`src/lib/commands/duplex_metrics.rs::generate_yield_metric` + new `calculate_ideal_duplex_fraction_per_size`** — stop expanding `family_size_metrics` into a flat `Vec<usize>` of `m.family_size` repeated `m.ds_count` times. Iterate per-size buckets and weight each size's binomial probability by its `ds_count`. Mathematically equivalent; eliminates the flat-Vec allocation *and* the `O(ds_count)` redundant `Binomial::cdf` calls per distinct family size.
* **`src/lib/commands/duplex_metrics.rs::process_coordinate_group`** — hoist `ss_groups` and `ds_groups` scratch HashMaps out of the 20-fraction loop and reuse via `.clear()`; skip `entry.2.push(...)` on non-full fractions; switch `update_umi_metrics` to `&[(&str, &str)]` to drop the owned-`String` clone at the call site. These are the original commits in this PR — they're real micro-perf improvements (mirroring the existing pattern in `simplex_metrics::process_coordinate_group`) but did NOT move the needle on the OOM, because the OOM is downstream of the streaming pass. They stay because they reduce allocator churn on the hot path.

## Behaviour

No behavioural change in any output column. Verified by:

- 20 existing `duplex_metrics::tests` pass on the post-fix tree.
- 107 existing `fgumi-metrics::tests` pass (was 106; the new regression test below adds one).
- 7 existing `simplex_metrics::tests` pass (unchanged file, sanity check on the shared metrics path).
- `cargo ci-fmt`, `cargo ci-lint` clean.

## Regression test

`test_duplex_family_size_metrics_sparse_hot_spot` in `crates/fgumi-metrics/src/duplex.rs` constructs a collector with one duplex family at `(50_000, 50_000)` alongside two small entries. Under the previous dense-grid implementation this would have allocated ~20 GiB before any assertion ran; under the sparse implementation it completes in microseconds. Pins the invariant against any future refactor that might silently reintroduce an `O(max_ab × max_ba)` allocation.

## Note on simplex-metrics

Reviewer asked whether simplex-metrics has the analogous bug. **No** — simplex-metrics has no equivalent 2D `(ab, ba)` cumulative structure (its `family_size_metrics` is 1D over `1..=max_size`, which is bounded by `max_size × ~80B` per call). The 1D densification pattern is shared between duplex and simplex `family_size_metrics`, but at realistic `max_size` it does not OOM. Left untouched in this PR.

## Test plan

- [x] `cargo test --release --lib duplex_metrics` — 20 passed
- [x] `cargo test --release --lib simplex_metrics` — 7 passed
- [x] `cargo test --release -p fgumi-metrics` — 107 passed (incl. new sparse-hot-spot regression)
- [x] `cargo ci-fmt`, `cargo ci-lint` clean
- [x] Real `vendor` benchmark on cfDNA-class inputs against the prior broken build reproduced the OOM at the same template count; re-running with this PR's bundled into `2fdcc7b` reproduced the OOM at the same point (confirming the original hoist-only fix was insufficient) — that's the empirical evidence behind this follow-up fix
- [ ] Re-run `vendor` profile against this PR head after a fresh submodule bump to confirm `duplex_metrics_fgumi` no longer OOMs on any of the 5 duplex-capable vendor samples

## Commit sequence

| Commit | What |
|---|---|
| `4649c11` | `perf(duplex-metrics): hoist scratch maps and drop unnecessary string clones` — micro-perf; not the OOM fix |
| `2fdcc7b` | `docs(duplex-metrics): clarify what HashMap::clear actually reuses` — comment-accuracy nit from initial review |
| `f6a1174` | `fix(duplex-metrics): drop dense (max_ab × max_ba) grid that OOMs on cfDNA hot spots` — **the actual OOM fix** |
| `7460059` | `test(duplex-metrics): pin the sparse-hot-spot regression` — guards against re-introducing dense allocation |

### DIFF
--- a/crates/fgumi-metrics/src/duplex.rs
+++ b/crates/fgumi-metrics/src/duplex.rs
@@ -718,6 +718,43 @@ mod tests {
         assert_eq!(metrics[2].ba_size, 3);
     }
 
+    /// Regression test for the sparse-vs-dense bug that OOM'd `fgumi
+    /// duplex-metrics` on cfDNA inputs. With a single duplex family at
+    /// `(50_000, 50_000)`, the previous dense-grid implementation would
+    /// allocate a `(50_001 × 50_001) × usize` scratch grid (~20 GiB) to
+    /// compute the cumulative fractions. The sparse implementation
+    /// iterates only the three observed `(ab, ba)` entries and computes
+    /// `fraction_gt_or_eq_size` correctly in microseconds.
+    ///
+    /// The assertions verify behaviour; the implicit assertion is that
+    /// the test completes without OOM, which guards against any future
+    /// refactor reintroducing an `O(max_ab × max_ba)` allocation.
+    #[test]
+    fn test_duplex_family_size_metrics_sparse_hot_spot() {
+        let mut collector = DuplexMetricsCollector::new(false);
+        collector.record_duplex_family(2, 1);
+        collector.record_duplex_family(5, 3);
+        collector.record_duplex_family(50_000, 50_000);
+
+        let metrics = collector.duplex_family_size_metrics();
+        assert_eq!(metrics.len(), 3, "exactly 3 sparse entries, not max_ab × max_ba");
+
+        // Sorted ascending by (ab, ba): (2,1) → (5,3) → (50000, 50000).
+        let total = 3.0_f64;
+        // (2, 1): all three entries have ab≥2 ∧ ba≥1.
+        assert_eq!(metrics[0].ab_size, 2);
+        assert_eq!(metrics[0].ba_size, 1);
+        assert!((metrics[0].fraction_gt_or_eq_size - 3.0 / total).abs() < 1e-9);
+        // (5, 3): only (5, 3) and (50000, 50000) satisfy.
+        assert_eq!(metrics[1].ab_size, 5);
+        assert_eq!(metrics[1].ba_size, 3);
+        assert!((metrics[1].fraction_gt_or_eq_size - 2.0 / total).abs() < 1e-9);
+        // (50000, 50000): only itself satisfies.
+        assert_eq!(metrics[2].ab_size, 50_000);
+        assert_eq!(metrics[2].ba_size, 50_000);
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

PR #386, opened 2026-06-10
fix(sort): carry library & MI in the completely-unmapped template-coordinate key

nh13:
## Summary

`fgumi sort --order template-coordinate` with auto `--key-types` aborts with `record <name> carries a library value absent from the input's first record; re-run with --key-types library` on **single-library** inputs that contain completely-unmapped read pairs — even though every read belongs to the same library. This breaks template-coordinate sort (and the dedup that chains off it) on real WGS/WES BAMs, which carry unmapped pairs at the tail.

## Root cause

`extract_template_key_inline` packs the read's library ordinal (and MI) into the full `TemplateKey`'s `tertiary` lane for mapped reads and for unmapped-reads-with-a-mapped-mate, but the completely-unmapped path returned `TemplateKey::unmapped(...)`, which hardcoded `tertiary: 0`. So a fully-unmapped read carrying a valid `RG` realized library ordinal `0` while its mapped, same-library peers realized ordinal `N`.

With one library the auto path drops the provably-constant library lane and relies on the decode-time dropped-lane verify to confirm it really is constant. The verify then compared the first (mapped) record's `N` against an unmapped read's `0` and hard-errored — a false positive.

## Fix

`TemplateKey::unmapped` now accepts and packs `library` and `mi` into `tertiary` exactly as `TemplateKey::new` does, so a fully-unmapped read realizes the same library/MI lanes as its mapped same-template peers (and its completely-unmapped sort order becomes consistent with the rest of the key).

## Tests

- `auto_single_library_unmapped_read_sorts_without_false_violation` — single-library input (one `@RG` with `LB`) with a mapped pair + a completely-unmapped pair; template-coordinate sort under auto `--key-types` must succeed. Fails before the fix with the exact production error; passes after.
- `test_template_key_unmapped_packs_library_and_mi` — `TemplateKey::unmapped` packs `library`/`mi` into `tertiary` identically to `TemplateKey::new`.
- `auto_single_library_multi_readgroup_sorts_without_false_violation` — guards the related multi-read-group/one-library case.

Reproduced against a 1000 Genomes WGS BAM (`HG00096`; 12 read groups, one library `HG00096`): the offending record was a completely-unmapped pair near the tail (~record 779.7M).

### DIFF
--- a/crates/fgumi-sort/src/external.rs
+++ b/crates/fgumi-sort/src/external.rs
@@ -3255,9 +3255,12 @@ pub fn extract_template_key_inline(
             );
         }
 
-        // Completely unmapped - sort to end
+        // Completely unmapped - sort to end. Still carry the read's library and
+        // MI lanes so a fully-unmapped read realizes the same tertiary as its
+        // mapped, same-library peers (otherwise the dropped-lane verify treats a
+        // single-library file as if the library varied; see #375).
         let is_read2 = (flag & 0x80) != 0; // is_last_segment flag
-        return TemplateKey::unmapped(name_hash, cb_hash, is_read2);
+        return TemplateKey::unmapped(name_hash, cb_hash, library, mi, is_read2);
     }
 
     // Calculate unclipped 5' position for this read (zero-alloc: reads cigar directly)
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

PR #393, opened 2026-06-14
refactor(sort): drop dead in-sort BAM indexer; --write-index via post-write hook

nh13:
**Stacked PR #2 of the #330 unified-pipeline landing.** Targets the long-lived integration branch `feat-runall` (not `main`); #389 (the reconciliation base) is its merged predecessor. One commit — the diff below is just this change.

Reverts the first-pass reconciliation graft that re-introduced main's `IndexingBamWriter`. In `feat-runall`, `--write-index` is owned by #330's decoupled `IndexBamFinalizeHook` (a post-pipeline pass keyed off `SinkSpec::BamWithIndex`), so the engine's incremental indexer is dead in the live path.

- **fgumi-bam-io:** remove `IndexingBamWriter` / `create_indexing_bam_writer` / `CachedIndexEntry` / `compute_end_vpos` and their exports; keep the #361 "0–12; 0 = uncompressed" compression-level doc note.
- **fgumi-sort:** remove the `write_index` field/setter, `sort_coordinate_with_index`, and `merge_chunks_with_index`; collapse `sort_coordinate` into the dispatch's direct `sort_coordinate_optimized` call; drop the orphaned engine tests.
- Add a **byte-identity regression guard** (`--write-index` on vs off → byte-equal BAM) plus a **fgumi-vs-samtools BAI region-query cross-check** across thread counts to `test_sort_write_index.rs`.
- Correct the CHANGELOG `--write-index` note: the historical byte divergence was the writer backend (noodles `MultithreadedWriter` vs `PooledBamWriter`), not single- vs multi-threaded compression.

Net −809 lines (dead-code removal).

PR #395, opened 2026-06-14
feat(sort): port #330 streaming-sort API onto main's engine for runall fusion

nh13:
**Stacked PR #3 of the #330 unified-pipeline landing.** Targets the long-lived integration branch `feat-runall` (not `main`); #393 (drop dead in-sort indexer) is its merged predecessor. One commit — the diff below is just this change.

Re-implements #330's typed-step streaming sort (`SortAndSpill → SortSpillDecompress → SortMerge`) on top of main's faster sort engine, unblocking the fused in-pipeline `runall` sort. The standalone `fgumi sort` path is untouched (keeps main's engine + zstd spill codec).

**fgumi-sort:**
- Add a slimmed, `from_slots`-only `MergeDriver` (+ `MergeDriverDyn`/`MergeStep`) and its non-blocking slot reader (`SlotMergeSource`/`SlotParserState`, resumable cross-block parser). The pipeline only ever calls `from_slots` over already-decompressed slot blocks, so #330's disk-reading/pool-phase2 machinery is not ported — the merge path carries no codec risk.
- Port the three phase-1 streams (`Coordinate`/`Queryname`/`TemplateCoordinate`) + `into_*_stream` constructors + `into_slot_setup`/`SlotSetup`, routing chunk writes through main's codec-aware `PooledChunkWriter`. The inline non-pipeline `merge_into`/`into_merge_driver` paths are intentionally dropped.
- Add `inmem_chunks_to_keyed` to materialize main's `InMemoryChunk` into the pipeline protocol's `Vec<(K, RawRecord)>` shape; add `SortWorkerPool::set_main_thread` so the streaming path can re-target unpark from a framework worker; wire `merge_slots`, export the streaming types, make `create_output_header` pub.

**pipeline:** pin the fused streaming spill codec to BGZF at the single chokepoint `build_stream`. `SortSpillDecompress` reads spill chunks as raw BGZF blocks; main's default zstd uses incompatible framing. Standalone sort keeps zstd; teaching the decompress step to stream-decode zstd frames is a tracked perf follow-up.

**Validation:** the three-step streaming chain matches legacy sort output across all sort orders (in-memory + multi-spill, t1–t8); 10 ported `from_slots` merge-driver unit tests cover cross-block, one-byte-at-a-time, and stall/resume paths.

**Note:** this is the commit that *introduces* the `fgumi_sort` streaming API that #389/#393 reference, so its tree should compile and test green standalone — it clears the cumulative-base build failures those earlier stacked PRs carried.

**Integration tests:** folds in the trivial repair of two latent integration-test compile errors (`test_correct_command.rs`, `test_e2e_regression.rs`) that this lib change *exposes* — they don't compile against the newly-linking `fgumi_sort` API otherwise. Kept in this commit so the PR's `test`/`lint`/`coverage` jobs are green standalone rather than red-until-the-next-PR.

### DIFF
--- a/crates/fgumi-sort/src/external.rs
+++ b/crates/fgumi-sort/src/external.rs
@@ -32,6 +32,7 @@ use crate::keys::{QuerynameComparator, RawSortKey, SortOrder};
 use crate::memory_probe::{
     BufferProbeStats, ConsumerProbeStats, MergeProbe, SpillProbe, force_mi_collect, log_snapshot,
 };
+use crate::merge_slots::SortMergeSlot;
 use crate::pooled_chunk_writer::PooledChunkWriter;
 use crate::read_ahead::{PooledInputStream, RawReadAheadReader, RecordSource};
 use crate::tmp_dir_alloc::TmpDirAllocator;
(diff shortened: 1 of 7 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #399, opened 2026-06-15
perf(sort): codec-aware streaming spill decompress; use main's zstd codec

nh13:
**Stacked PR (stack position 08) of the #330 unified-pipeline landing.** Targets the long-lived integration branch `feat-runall` (not `main`); #398 (consensus rejects fan-out) is its merged predecessor. One commit.

The fused streaming sort previously pinned its temp spill to BGZF because `SortSpillDecompress` could only inflate raw BGZF blocks, while main's standalone sort defaults to the ~3–6×-faster zstd spill codec (#341). Pinning BGZF left the fused path slower than standalone for no good reason.

Make the streaming decompress codec-aware so the fused path can use zstd too:
- Add `SpillBlockDecompressor` (fgumi-sort): a per-worker codec-aware block reader. BGZF chunks decode via libdeflate; zstd "ZSP1" chunks decode frame-by-frame (`[u32 LE len][zstd frame]`).
- Detect each spill chunk's codec from its file magic (`SpillCodec::from_magic`) in `slots_for_chunk_files`, store it on `SortMergeSlot.codec`.
- `SortSpillDecompress` reads `slot.codec` and dispatches via `SpillBlockDecompressor`.
- Remove the BGZF pin in `build_stream`; the streaming sorter now uses main's default (zstd). Standalone sort is unchanged.

**Reconciliation note (important):** stack-03 (#395) had hardened the three `*SortStream` constructors to force `SpillCodec::Bgzf` (a CodeRabbit-CLI hardening — correct while the decoder was BGZF-only). That premise is now gone (the decoder is codec-aware), and leaving the force in place would *silently defeat* this PR's whole purpose — `build_stream` would hand a zstd sorter to a constructor that overrode it back to BGZF, so the spill would still be BGZF and tests would stay green. This PR therefore also **reverts that constructor force** (restores `self.spill_codec`), which both enables zstd and removes the now-unnecessary foot-gun (the codec-aware decoder reads any codec from magic). This interaction was caught in a pre-PR local review, not by the textual rebase.

**Validation:** round-trip unit tests for both codecs; the three-step streaming chain matches legacy sort output with **zstd** spill across all orders (in-memory + forced multi-spill + large-spill 60k records, t1–t8). Local: `fgumi-sort` 469/469, `three_step_chain` 14/14. clippy + fmt clean.

### DIFF
--- a/crates/fgumi-sort/src/external.rs
+++ b/crates/fgumi-sort/src/external.rs
@@ -1393,16 +1393,18 @@ impl RawExternalSorter {
              queryname / template-coordinate streaming entry points are forthcoming",
             self.sort_order,
         );
+        self.ensure_spill_codec_compat()?;
 
-        // Force BGZF: the streaming merge's `SortSpillDecompress` reads spill
-        // chunks as raw BGZF blocks, so the spill codec must be BGZF regardless
-        // of `self.spill_codec` (which defaults to zstd). Enforced here so this
-        // public constructor can't hand the decoder slots it can't read.
+        // Spill codec from config (zstd by default, the same fast codec
+        // standalone sort uses). `SortSpillDecompress` is codec-aware — it
+        // detects each spill chunk's codec from its file magic (see
+        // `slots_for_chunk_files`) and dispatches accordingly — so the streaming
+        // path no longer has to pin BGZF.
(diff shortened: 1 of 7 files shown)


nh13:
@coderabbitai review

PR #411, opened 2026-06-15
test(stdin): cover single- and multi-threaded stdin input for pipeline commands

nh13:
## Summary

Adds file-vs-stdin parity tests (`-` and `/dev/stdin`) for every command that accepts a BAM on stdin: **group, simplex, codec, duplex, clip, correct, dedup, filter**. This closes the stdin coverage gap that let the codec/simplex single-threaded double-open ship (only multi-threaded stdin was tested before — see #410).

## What it tests

A reusable `assert_stdin_input_parity` helper runs each command three ways — reading the same BAM from a file, from piped stdin (`-`), and from `/dev/stdin` — and asserts the outputs are byte-identical (full `RecordBuf` equality). Commands with a single-threaded fast path (codec/duplex/clip) are exercised both single-threaded (no `--threads`) and multi-threaded (`--threads 2`); the rest run at `--threads 1` and `--threads 2`. Tests are parameterized with `rstest` `#[case]`s so each (command, threading) pair is an independently-named test.

**Non-vacuous by construction:** the helper asserts the file baseline emits ≥1 record before comparing — otherwise a command that silently dropped stdin and produced empty output would pass a parity check trivially. (Writing this guard caught three initially-vacuous cases — codec/duplex/filter were fed input that produced no consensus output — which are now fed proper paired/consensus input via the commands' own proven test builders.)

## sort

`sort` is intentionally **not** covered: it does not currently support stdin. Its terminal-sort path uses a file-to-file `RawExternalSorter` that reads the input by path and double-reads the header (so a stdin stream is drained empty by the second read), and input validation rejects `-`/`/dev/stdin` outright. This is a **pre-existing limitation** (present on `main`, not introduced by the #330 landing). A follow-up stacked PR will teach the sorter to stream stdin once and add sort to this matrix.

## Notes

Made a few existing per-command test builders `pub(crate)` (`create_codec_read_pair`/`create_codec_test_bam`, `create_duplex_molecule`/`create_duplex_bam`) and added `write_filter_consensus_bam`, so the stdin tests reuse the same proven, output-producing inputs rather than duplicating BAM construction.

## Stack context

Test-only follow-up to the issue #330 unified-pipeline landing onto `feat-runall`. Dual-reviewed before push (CLI + skill: clean; CLI's rstest-convention finding applied).

### DIFF
--- a/tests/integration/test_codec_command.rs
+++ b/tests/integration/test_codec_command.rs
@@ -25,7 +25,7 @@ use crate::helpers::bam_generator::create_minimal_header;
 /// In CODEC sequencing, R1 and R2 come from opposite strands of the same molecule,
 /// so a single read pair can produce duplex consensus.
 #[allow(clippy::cast_possible_truncation, clippy::cast_possible_wrap, clippy::too_many_arguments)]
-fn create_codec_read_pair(
+pub(crate) fn create_codec_read_pair(
     name: &str,
     r1_seq: &[u8],
     r2_seq: &[u8],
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai review

PR #414, opened 2026-06-16
refactor(bam-io): collapse cached-UMI sentinel handling at the owning layer

nh13:
## Summary

`primary_read_cached_umi` (template.rs) duplicated the same `cached_umi_position()` → `UMI_OFFSET_UNCACHED`-sentinel → `Option` conversion in both its R1 and R2 arms, leaking the sentinel detail to the caller. Add `DecodedRecord::cached_umi_position_opt() -> Option<(u32, u16)>` — which does the sentinel check once at the type that owns it — and have both arms return it directly. Also drops a redundant function-body `use fgumi_raw_bam;` (the crate is already nameable; the fully-qualified `fgumi_raw_bam::flags::*` paths resolve without it).

## Behavior preservation

Identical: an R1 (cached or not) still wins over R2, and an uncached primary still yields `None` (caller falls back to `find_string_tag`). Cached-UMI, decode, and group-determinism tests pass unchanged.

`DecodedRecord::cached_umi()` is intentionally **kept** — though production reads the cache via `Template::cached_umi`, the `DecodedRecord` accessor is the clean expression of the load-bearing "cached slice == `find_string_tag`" invariant test in `decode.rs`; removing it would force that test to re-implement the slice it verifies.

## Context

Second of the post-#330 subsystem-simplification cleanups (discovery bundle "bam-io/template"). Dual-reviewed before push (CLI: no findings; skill: 0 actionable). The discovery's "delete dead `DecodedRecord::cached_umi()`" item was assessed and declined for the reason above.

### DIFF
--- a/crates/fgumi-bam-io/src/grouping.rs
+++ b/crates/fgumi-bam-io/src/grouping.rs
@@ -242,6 +242,19 @@ impl DecodedRecord {
         (self.umi_value_offset, self.umi_value_len)
     }
 
+    /// Returns the cached UMI `(offset, len)` position, or `None` when no
+    /// position was recorded during decode (the [`Self::UMI_OFFSET_UNCACHED`]
+    /// sentinel). Keeps the sentinel check at the owning layer so callers can
+    /// branch on a plain `Option` instead of comparing against the sentinel.
+    #[must_use]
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

PR #415, opened 2026-06-16
feat(sort): support reading BAM input from stdin

nh13:
## Summary

`fgumi sort -i -` failed on `main` with `Invalid Input BAM file '-': File does not exist`. Every other input-accepting command (`group`, `dedup`, `clip`, …) reads BAM from stdin; `sort` was the gap. This blocks streaming pipelines that pipe a BAM straight into a sort without an intermediate file.

The good news: the sort **engine** on `main` already streams stdin in a single pass. `create_raw_bam_reader_pool_integrated` branches on `is_stdin_path` — a `TeeReader` reads the header, then a `ChainedReader` replays it and streams the rest. The only thing blocking `fgumi sort -i -` was the input-validation gate in `Sort::execute`, which ran `validate_file_exists` with no stdin exemption.

(Note: the gate uses `Path::exists()`, which is `false` for `-` but `true` for `/dev/stdin` — so only `-` actually failed on the sort path today. The fix handles both uniformly via `is_stdin_path`.)

## Fix

- **Exempt stdin** (`-` / `/dev/stdin`) from the file-existence check on the sort path, matching the stdin exemption in `group`/`dedup`/`clip`. No engine changes, no change to the file-input reader path.
- **Reject `sort --verify` from stdin** with a clear message. `execute_verify` opens the input twice (header probe via `create_raw_bam_reader`, then a fresh `File::open` record re-scan), which a non-seekable pipe can't satisfy — so it's rejected up front rather than failing deep in a re-open. The reject keys off `is_stdin_path`, covering both `-` and `/dev/stdin`.

## Tests (TDD)

Backported the file-vs-stdin parity harness and strengthened `compare_bam_records` to full `RecordBuf` identity (bases, qualities, flags, CIGAR, positions, and all aux tags) so a record-level corruption from mishandled input fails the assertion, not just a count/name mismatch. Verified the existing `group`/`simplex` parity tests still pass under the stronger comparator.

Added:
- `test_sort_reads_stdin_once` — file-vs-`-`-vs-`/dev/stdin` byte-for-byte record parity, single- and multi-threaded, with a guard against a vacuous (zero-record) pass.
- `test_sort_verify_rejects_stdin_with_clear_message` — parameterized over `-` and `/dev/stdin`, asserting a non-zero exit and a stderr message mentioning stdin.

Full streaming-input integration suite and the `fgumi-sort` crate pass; `cargo ci-fmt` and `cargo ci-lint` are clean.

## Relationship to #412

This is the `main`-targeting, standalone version of the sort-stdin fix in #412, so it can land independently of the runall stack. #412 is stacked on #411 and targets `feat-runall`, and it additionally patches `src/lib/pipeline/chains/builder.rs` — that file is part of the runall pipeline-chains architecture and does not exist on `main`, so the builder half does not apply here. This PR is the minimal change `main` needs.

## Release Notes

* **Bug Fixes**
  * Sort command now properly validates stdin input. When using the `--verify` flag with non-seekable stdin (`-` or `/dev/stdin`), the command fails early with a clear error message.

* **Tests**
  * Expanded streaming input test coverage to verify sorting consistency between stdin and file input.
  * Enhanced output validation for improved record integrity checking.

### DIFF
--- a/src/lib/commands/sort.rs
+++ b/src/lib/commands/sort.rs
@@ -386,8 +386,24 @@ impl Command for Sort {
             bail!("--write-index cannot be used with --verify");
         }
 
-        // Validate inputs
-        validate_file_exists(&self.input, "Input BAM")?;
+        // Validate inputs. Exempt stdin paths (`-` / `/dev/stdin`): the sort
+        // engine reads stdin in a single pass (a `TeeReader` reads the header,
+        // then a `ChainedReader` replays it and streams the rest), so a
+        // file-existence check would spuriously reject it — matching the stdin
+        // exemption in group/dedup/clip. `--verify` is the exception: it
+        // re-scans the input (header probe + a fresh `File::open` record pass),
+        // which a non-seekable stdin can't satisfy, so reject stdin there up
+        // front with a clear message.
+        if fgumi_bam_io::is_stdin_path(&self.input) {
+            if self.verify {
+                bail!(
+                    "fgumi sort --verify cannot read from stdin (it re-scans the input); \
+                     provide a file path instead"
+                );
+            }
+        } else {
+            validate_file_exists(&self.input, "Input BAM")?;
+        }
 
         // Either --output or --verify must be specified
         if !self.verify && self.output.is_none() {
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #416, opened 2026-06-16
refactor(pipeline): add PipelineError::reconstruct to de-duplicate outcome rematch

nh13:
## Summary

`Pipeline::run` (`core/builder.rs`) and the fused single-thread driver (`core/runtime/fused.rs`) each open-coded the same per-variant match to turn the borrowed `signal.outcome(): Option<&PipelineError>` into the owned error they return — necessary because `PipelineError` is not `Clone` (its `Io` variant holds a non-`Clone` `io::Error`). Extract that into `PipelineError::reconstruct(&self) -> PipelineError`; both call sites collapse to `Some(err) => Err(err.reconstruct())`.

## Behavior preservation

Identical — the `Io` source is still rebuilt from `kind()` + display string, all other variants copied. A future error variant now needs handling in one place instead of two. Fused-pipeline and pipeline error-path tests pass unchanged.

## Context

Third of the post-#330 subsystem-simplification cleanups (discovery bundle "pipeline misc"). Dual-reviewed before push (CLI: no findings; skill: 0 actionable).

### DIFF
--- a/src/lib/pipeline/core/builder.rs
+++ b/src/lib/pipeline/core/builder.rs
@@ -751,7 +751,6 @@ impl Pipeline {
             build_chain_contexts, build_worker_storage, is_fusible_chain, run_fused_single_thread,
             run_worker_loop,
         };
-        use super::signal::PipelineError;
         use super::step::StepKind;
         use super::topology::StepIdx;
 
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai full review

PR #417, opened 2026-06-16
docs(group): trim duplicated rayon-pool caveat in the command long-about

nh13:
## Summary

The "each parallel assigner spawns its own rayon pool of size `--threads` … not an upper bound …" caveat was spelled out in full in two places in `group.rs`: the `--parallel-group-min-templates` field doc (canonical) and the command-level `long_about`, whose copy even ended with "See --parallel-group-min-templates for details." Shorten the command-level copy to the essential point (peak OS-thread count can exceed `--threads`) plus that pointer, so the full explanation lives in one place and the two can't drift.

Help-text only; no code or behavior change.

## Context

Last of the post-#330 subsystem-simplification cleanups (discovery bundle "commands prose"). Dual-reviewed before push (CLI: no findings; skill: trivial docs change, 0 actionable).


nh13:
@coderabbitai review

PR #418, opened 2026-06-16
fix(bam-io): read process stdin directly in open_bgzf_reader

nh13:
## Summary

`open_bgzf_reader` in `fgumi-bam-io` opened its input with `File::open(path)` unconditionally. For a stdin input that breaks the single-threaded reader path: `File::open("-")` fails outright, and `File::open("/dev/stdin")` re-opens fd 0 (losing any bytes a caller already consumed). This made single-threaded `fgumi <cmd> -i -` fail for the commands whose single-threaded fast path routes through `create_raw_bam_reader_with_opts` → `open_bgzf_reader` (simplex, clip, codec, duplex, correct).

## Fix

Add an `is_stdin_path` branch to `open_bgzf_reader` that reads `io::stdin()` directly, so this BGZF reader is the sole consumer of the pipe. The branch honors `--async-reader` for parity with the file path (wrapping stdin in `PrefetchReader`). This is the single, shared change that closes the single-threaded `-` gap for every command that uses this reader — no per-command code is needed for the reader itself.

## Tests (TDD)

Single-threaded `simplex -i -` is the ideal regression: simplex's input gate is already stdin-exempt, so its single-threaded `-` path was broken *only* by `open_bgzf_reader`. Added two tests (written first, red on the `File::open("-")` failure, green after the fix):

- `test_simplex_single_threaded_reads_stdin_once` — file-vs-piped byte-identical consensus records via the single-threaded path.
- `test_simplex_single_threaded_async_reader_over_stdin` — same, with `--async-reader`, exercising the prefetch-on-stdin sub-branch.

Full streaming-input integration suite (9 tests) and the `fgumi-bam-io` crate (106 tests) pass; `cargo ci-fmt` / `cargo ci-lint` clean.

## Stack

This is the foundational piece of porting stdin support across all BAM-input commands. #415 (sort stdin + the file-vs-stdin parity harness these tests use) has merged to `main`, so this targets `main` directly. A follow-up PR adds the per-command validation-gate exemptions (correct/codec/duplex/filter) and the full parity-test matrix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed stdin handling to avoid reopening stdin and prevent data loss during single-threaded processing.
  * Improved behavior for piped input when using the async reader for consistent BGZF decompression.

* **Tests**
  * Strengthened integration coverage for stdin/piped streaming, including a non-vacuous baseline check and corrected grouping tag typing.
  * Added a parameterized test to confirm single-threaded runs read from stdin correctly in both standard and async-reader modes with byte-identical output.

### DIFF
--- a/crates/fgumi-bam-io/src/reader.rs
+++ b/crates/fgumi-bam-io/src/reader.rs
@@ -163,6 +163,22 @@ fn open_bgzf_reader(
     threads: usize,
     label: &str,
 ) -> Result<BgzfReaderEnum> {
+    // stdin can't be re-opened: `File::open("-")` fails outright, and
+    // `File::open("/dev/stdin")` re-reads fd 0 — losing any bytes a caller
+    // already consumed. Read the process stdin directly so this BGZF reader is
+    // the sole consumer (fixes the single-threaded double-open data loss).
+    if is_stdin_path(path) {
+        // Honor --async-reader for stdin too (parity with the file path below):
+        // a pipe benefits from prefetch overlapping upstream production with us.
+        let reader: Box<dyn Read + Send> = if opts.async_reader {
+            log::info!("async {label} enabled: spawning fgumi-prefetch thread for stdin");
+            Box::new(crate::prefetch_reader::PrefetchReader::new(io::stdin()))
+        } else {
+            Box::new(io::stdin())
+        };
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

PR #419, opened 2026-06-16
feat: support reading BAM input from stdin for correct, codec, duplex, filter

nh13:
## Summary

Extends stdin BAM input to the remaining BAM-input commands: `correct`, `codec`, `duplex`, and `filter`. (`group`, `dedup`, `clip`, `sort`, and the simplex single-threaded path already read stdin on `main`.) After this, every BAM-input command accepts `-i -` and `-i /dev/stdin`.

## What was blocking each command

These four called `validate_file_exists(&self.io.input, ...)` unconditionally, which rejected `-`/`/dev/stdin` before the reader ran. Their readers are already stdin-safe (multi-threaded via `create_bam_reader_for_pipeline*`; single-threaded via the `open_bgzf_reader` stdin branch in the parent PR). So the fix is a one-line gate exemption per command, matching the existing pattern in `group`/`dedup`/`clip`:

```rust
if !fgumi_bam_io::is_stdin_path(&self.io.input) {
    validate_file_exists(&self.io.input, "...")?;
}
```

## `correct` single-threaded double-open fix

`correct` additionally had a latent bug: `execute()` opened the input reader unconditionally, then the single-threaded path (`--threads` omitted) dropped it and re-opened the input — fatal for a pipe (the second open gets a drained stream). Fixed by reusing the already-opened reader in single-threaded mode, mirroring `group::execute_single_threaded` (skip the header on the existing reader, wrap in `RawBamReader`). `codec`/`duplex` already open inside their thread branches; `filter`/`dedup` open once — none of them had this issue.

## Tests

File-vs-`-`-vs-`/dev/stdin` byte-identical record parity (`assert_stdin_input_parity`) for `group`, `dedup`, `correct`, `codec`, `duplex`, `filter`, `clip`, each with a non-vacuous (`> 0` records) guard. Commands with a single-threaded fast path (`codec`, `duplex`, `clip`, and now `correct`) are parameterized over both single-threaded (no `--threads`) and multi-threaded — the single-threaded case is what surfaces reader-reuse bugs like the `correct` double-open. Cross-module test builders (`create_codec_*`, `create_duplex_*`) were made `pub(crate)`, and `write_filter_consensus_bam` was factored out of the existing filter test for reuse.

Full streaming-input integration suite passes; `cargo ci-fmt` / `cargo ci-lint` clean.

## Stack

Stacked on #418 (`open_bgzf_reader` stdin fix), which underpins the single-threaded `-` paths here. Base will retarget to `main` once #418 merges.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced stdin streaming consistency across multiple commands (correct, codec, duplex, filter).

* **Tests**
  * Added comprehensive integration tests verifying stdin streaming parity across group, dedup, correct, codec, duplex, filter, and clip commands.

### DIFF
--- a/src/lib/commands/codec.rs
+++ b/src/lib/commands/codec.rs
@@ -46,7 +46,6 @@ use fgumi_bam_io::ProgressTracker;
 use fgumi_raw_bam::RawRecord;
 // RejectionTracker now used via ConsensusStatsOps trait in consensus_runner
 use crate::sam::SamTag;
-use crate::validation::validate_file_exists;
 use log::info;
 use noodles::sam::Header;
 use noodles::sam::alignment::record::data::field::Tag;
@@ -264,7 +263,9 @@ impl Command for Codec {
     fn execute(&self, command_line: &str) -> Result<()> {
         // Validate inputs
         self.validate()?;
-        validate_file_exists(&self.io.input, "input BAM file")?;
+        // Validate the input exists (stdin paths are exempt — the reader
+        // streams them in a single pass).
+        self.io.validate()?;
 
         let timer = OperationTimer::new("Calling CODEC consensus");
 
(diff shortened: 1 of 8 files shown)


nh13:
@coderabbitai review

PR #421, opened 2026-06-16
refactor(runall): drop dead execute_sort_only/execute_group_only helpers

nh13:
First of a small series of verified cleanups/perf opportunities found while reviewing the `feat-runall` stack.

## What

`RunAll::execute` already routes every chain shape — `--stop-after sort`, `--stop-after group`, and everything else — through `chains::build_for`, which derives the stage list and runs the unified chain builder. The standalone `execute_sort_only` / `execute_group_only` helpers were left behind carrying `#[allow(dead_code)]` and have no callers. This removes them (and the two doc/comment references that named them).

It also corrects a stale note in `CLAUDE.md`: there is no `execute_*_then_*` file-to-file fallback dispatcher in `commands::runall`. `ChainBuilder::add_sort` performs the intermediate sort in-pipeline (streaming), so chains containing an intermediate Sort run through `build_for` like every other shape.

## Why

Pure dead-code removal + doc accuracy — no behavior change. 112 lines deleted, 6 added.

## Verification

- `cargo build` / `cargo ci-fmt` / `cargo ci-lint` (clippy `-D warnings -W clippy::pedantic`) all clean — the build itself confirms the helpers had no callers.
- `cargo nextest run -E 'test(runall)'` → 143 passed, 0 failed (covers the `--stop-after sort` and `--stop-after group` paths, e.g. `rejects_out_with_stop_after_group_does_not_create_file`).

PR #422, opened 2026-06-16
perf(group): byte-compare MI tag for boundary detection, label once per group

nh13:
Wave 1 perf cleanup (#2 of the feat-runall review series; follows #421).

## What

`MiGrouper` and `MiGroupIterator` materialized an owned `String` per record (`String::from_utf8_lossy(tag).into_owned()`, plus a second allocation for the cell-tag composite) purely to detect MI-group boundaries by comparing consecutive keys. MI grouping streams over the entire post-group BAM, so that is one or two heap allocations on every record.

This stores the current run's grouping key as its raw comparison bytes (`MiKey`) and compares the next record's tag value(s) by byte-equality — allocation-free on the common no-transform path (and for the cell tag). The display `String` label handed to `MiGroup` / the iterator output is built from those bytes only once per group, at the boundary. The transform path still allocates per record to evaluate the transform (inherent and rare), and there it correctly compares *transformed* values.

## Behavior

Unchanged. Preserved: the composite label format (`"MI\tCELL"`), the missing-cell case (`"MI\t"`), and the skip-records-without-an-MI-tag policy. The `get_mi` unit tests still assert the label output (the method is retained behind `#[cfg(test)]`).

## Verification

- `cargo ci-fmt` / `cargo ci-lint` (clippy `-D warnings -W clippy::pedantic`) clean.
- `cargo nextest` — 688 tests in the mi/group/runall/streaming set pass, including `test_grouper_cell_tag_composite_grouping`, `test_cell_tag_*` (composite, none, missing-cell, with-transform), `test_with_transform*`, and `test_group_determinism::*`.

### DIFF
--- a/src/lib/mi_group.rs
+++ b/src/lib/mi_group.rs
@@ -105,8 +105,90 @@ impl MemoryEstimate for MiGroupBatch {
 /// Type alias for raw-byte MI tag transformation function.
 type MiTransformFn = Box<dyn Fn(&[u8]) -> String + Send + Sync>;
 
+/// Borrowed MI tag transformation, shared by `MiGrouper` and `MiGroupIterator`
+/// (whose stored transforms differ only in `Send + Sync` bounds).
+type MiTransform<'a> = Option<&'a dyn Fn(&[u8]) -> String>;
+
 /// Type alias for raw-byte record filter function.
 type RecordFilterFn = Box<dyn Fn(&[u8]) -> bool + Send + Sync>;
+
+/// The grouping key for a run of consecutive records, stored as the raw
+/// comparison bytes so that boundary detection stays allocation-free on the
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
Addressed the nitpick: the duplicated `extract_key`/`matches_key` logic is now consolidated into `MiKey::from_record` / `MiKey::matches_record` (plus the existing `label`). Each of `MiGrouper` and `MiGroupIterator` now keeps only a small typed `transform()` accessor to bridge the one real difference between them (the grouper's stored transform is `Send + Sync`, the iterator's is not). Behavior is unchanged; the full mi/group/consensus/runall test set passes.

nh13:
@coderabbitai review

PR #423, opened 2026-06-16
fix(simulate): emit per-base arrays and float error rates in consensus-reads (filter-consumable)

nh13:
## Summary

`fgumi simulate consensus-reads` documents its output as "suitable for input to `fgumi filter`", but `build_consensus_record` only wrote read-level integer tags — so `fgumi filter` could not actually consume it (it rejected 100% of the reads). Three related gaps:

1. **Missing per-base `cd`/`ce` arrays** (`CD_BASES`/`CE_BASES`). filter masks bases using these per-base arrays; with them absent every base reads as zero-depth → masked → the read is rejected.
2. **Missing per-base duplex arrays** (`AD_BASES`/`AE_BASES`/`BD_BASES`/`BE_BASES`) for `--duplex` output — filter's duplex masking path has the same dependency, so `--duplex` output was also rejected wholesale.
3. **Read-level `cE`/`aE`/`bE` written as integer counts**, but filter reads them as **float error rates** (`find_float_tag`), matching what `fgumi simplex`/`duplex` emit (`vanilla_caller`). The integer tags were silently skipped, so filter's read-level error filter never applied to simulated data.

## Fix

`build_consensus_record` now emits, for both the single-strand and duplex paths:

- per-base depth/error arrays (i16, one entry per base), consistent with the read-level summary tags — depth spans `[min, max]` (`min == cM/aM/bM`, `max == cD/aD/bD`) and the error counts are spread one-per-base;
- `cE`/`aE`/`bE` as **float** error rates derived from those arrays (`sum(errors)/sum(depths)`).

`cD`/`cM`/`aD`/`bD`/`aM`/`bM` remain integers (depths), matching production output.

## Tests

- `test_build_consensus_record_emits_per_base_cd_ce_arrays`: per-base arrays present, length == read length, `max == cD`, `min == cM`, errors sum to `cE`, and `cE` is a float.
- `test_build_consensus_record_duplex_emits_per_base_strand_arrays`: per-base `aD/aE/bD/bE` arrays present and consistent with the per-strand summary tags; `aE`/`bE` are floats.
- End-to-end: `fgumi filter` now keeps reads from both **simplex (514/1000)** and **`--duplex` (846/1000)** simulate output instead of rejecting all.

Full `build_consensus_record` suite (17 tests) passes; `cargo ci-fmt` / `cargo ci-lint` clean.

## Scope note

The `ac`/`bc` per-base base-call strings (`AC`/`BC_BASES`) are only consulted by filter's opt-in single-strand-agreement check (off by default), so they're left as a follow-up rather than blocking this fix.

## Context

Found while manually verifying stdin support across the BAM-input commands (#418/#419). Independent of the stdin work — targets `main` directly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consensus record output to be compatible with `fgumi filter` by adding per-base depth/error arrays (simplex: `CD_BASES`/`CE_BASES`; duplex: `AD_BASES`/`AE_BASES`/`BD_BASES`/`BE_BASES`) alongside the existing summary tags.
  * Consensus error rates are now calculated and stored as float values for simplex and duplex modes, with per-base errors distributed consistently.

* **Tests**
  * Added regression coverage for simplex and duplex layouts, including array lengths, depth min/max matching, error totals, and float error-rate storage.

### DIFF
--- a/src/lib/commands/simulate/consensus_reads.rs
+++ b/src/lib/commands/simulate/consensus_reads.rs
@@ -695,17 +695,62 @@ fn build_consensus_record(
         .sequence(seq)
         .qualities(quals);
 
-    // Add consensus tags (note: cE is error count in this file, not error rate).
-    b.add_int_tag(SamTag::CD, cd).add_int_tag(SamTag::CM, cm).add_int_tag(SamTag::CE, ce);
+    // `fgumi filter` masks bases using the PER-BASE depth/error arrays and reads
+    // the read-level error tags (cE / aE / bE) as FLOAT rates — exactly what
+    // `fgumi simplex`/`duplex` emit (see `vanilla_caller`). Without the per-base
+    // arrays every base reads as zero depth and the read is rejected outright;
+    // with the error tags written as integer counts, filter's `find_float_tag`
+    // skips them. Emit both faithfully so the output honors its documented
+    // "suitable for input to `fgumi filter`" contract.
+    //
+    // Build per-base arrays consistent with the read-level summary tags: depth
+    // spans [min, max] (min == cM/aM/bM, max == cD/aD/bD) and the total error
+    // count is spread one-per-base. The read-level error tag is the resulting
+    // rate (sum of per-base errors / sum of per-base depths).
+    let read_len = seq.len();
+    let per_base_arrays =
+        |max_depth: i32, min_depth: i32, error_count: i32| -> (Vec<i16>, Vec<i16>) {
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #430, opened 2026-06-18
perf(pipeline): hash reorder buffer with ahash instead of SipHash

nh13:
Wave 2 perf cleanup (#7 of the feat-runall review series).

## What

`ReorderState::buffer` keys items by their framework-internal `u64` ordinal but used the std `HashMap` default hasher (SipHash-1-3) — DoS-resistant but slow, and pointless for monotonic, adversary-free integer keys. Every ordered (`ByOrdinal` / `ByItemOrdinal`) edge funnels through this buffer (insert + `contains_key` + remove per released item, under the state lock), so it sits on the per-item path of every consensus pipeline.

Switches the buffer to `ahash::AHashMap`, the fast hasher already used across the codebase (`umi`, `filter`, `grouper`).

## Behavior

Unchanged — same map API, same ordering and backpressure invariants. (A `VecDeque` ring keyed by `ordinal - base` is the maximal version since release is strictly monotonic; `AHashMap` is the zero-risk drop-in and was chosen for that reason.)

## Verification

- `cargo nextest -E 'test(reorder) or test(pipeline)'` — 450 tests pass, including the in-module reorder unit tests and the `pipeline_core_compile_fail` trybuild suite.

### DIFF
--- a/src/lib/pipeline/core/reorder.rs
+++ b/src/lib/pipeline/core/reorder.rs
@@ -26,7 +26,8 @@
 //!     transport (`CountBounded` / `Unbounded` — see PR 1 caveat below).
 //!   - "Stashed" items (must-accept overflow when transport rejected, or
 //!     items pulled but not yet at their turn) live in `state.buffer:
-//!     HashMap<u64, T>`.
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

PR #431, opened 2026-06-18
perf(pipeline): dedupe the TemplatesToMiGroups bridge and stop cloning records

nh13:
Wave 2 perf cleanup (#8 of the feat-runall review series) — the highest-value item: it hits the primary runall workload (fused sort→group→consensus / group→consensus). Two commits, read in order.

## 1. `refactor(pipeline)`: extract the bridge into one factory

`add_simplex`, `add_duplex`, and `add_codec` each inlined a ~70-line copy of the `TemplatesToMiGroups` bridge (the fused group→consensus path that splices each template's MI into the `MI` tag and runs records into per-MI groups). The three copies were identical except for one flag — `duplex_strip_strand` (true for duplex, false for simplex/codec). Hoisted into a single `templates_to_mi_step(limit_bytes, duplex_strip_strand)` factory called from all three sites. Pure dedup, no behavior change.

## 2. `perf(pipeline)`: move tagged bytes into records instead of cloning

The bridge tags each record's MI in a reused `scratch` buffer, then did `RawRecord::from(state.scratch.clone())` — a full per-record heap copy on top of the `extend_from_slice`, once per read on the main runall workload. Now moves the buffer into the record with `std::mem::replace`, reinstalling a fresh `Vec` pre-sized to the previous capacity so the next record reuses it without a realloc (the per-record `clear()` keeps it warm). Same allocation count as the clone, minus the record-body memcpy. Output unchanged.

## Verification

- `cargo nextest -E 'test(runall) or test(group_determinism)'` — 149 tests pass, covering all three fused group→{simplex,duplex,codec} shapes and MI determinism.
- Net −132 lines.

### DIFF
--- a/src/lib/pipeline/chains/builder.rs
+++ b/src/lib/pipeline/chains/builder.rs
@@ -293,7 +293,13 @@ fn templates_to_mi_step(
                             assign_tag_bytes,
                             state.mi_buf.as_bytes(),
                         );
-                        current_records.push(RawRecord::from(state.scratch.clone()));
+                        // Move the tagged bytes into the record instead of cloning
+                        // them; reinstall a fresh scratch buffer pre-sized to the
+                        // same capacity so the next record reuses it without a
+                        // realloc (the per-record `clear()` above keeps it warm).
+                        let cap = state.scratch.capacity();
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

PR #432, opened 2026-06-18
perf(sort): size radix passes from the max mapped key, not the unmapped sentinel

nh13:
Wave-4 / sort-hotspot work (tasks 0 + 1 from `2026-06-18-sort-hotspot-tasks.md`). Supersedes the original "skip the scan" framing of this PR — profiling showed the real win is the unmapped sentinel, not the scan.

## What

`RecordBuffer` tracks the running maximum **mapped** `sort_key` as records are pushed and hands it to the radix sort. This both removes the standalone O(n) max-finding scan *and* — the bigger win — excludes the unmapped sentinel from the max.

Coordinate keys pack `(tid << 34) | ((pos+1) << 1) | reverse`, so mapped keys span ~5 bytes (tid ≤ 24) to ~6 (ALT/decoy contigs). But unmapped reads (`tid < 0`) get key `u64::MAX`; a single one dragged the radix's `bytes_needed` to the full 8 bytes. Coordinate BAMs almost always have an unmapped tail, so this fired on ~every input — 3 wasted LSD passes over all records. Sizing `bytes_needed` from the mapped max only; unmapped keys still sort to the tail because `u64::MAX` truncates to all-`0xFF` under any width and stays stable.

## Result (added `coordinate_radix_sort` criterion bench)

| n | scan-all (8 passes) | mapped-max (5 passes) |
|---|---|---|
| 1M | 47 M rec/s | **108 M rec/s** |
| 8M | 45 M rec/s | **104 M rec/s** |

~2.3× faster radix; the radix bucket was ~14% of sort CPU.

## Correctness

Two debug-only guards: the radix precondition `all(key <= max_key || key == u64::MAX)`, and the buffer-level streaming-vs-scan equivalence over non-sentinel keys. New unit test mixes mapped + 5% unmapped above the radix threshold and asserts the tracked max excludes the sentinel and the output is fully ordered with unmapped at the tail. The scanning wrappers are retained unchanged for external callers / the template path.

## Commits

1. `perf(sort)`: the mapped-max tracking + sentinel exclusion + asserts + test + CLAUDE.md unsafe-ledger update + `RecordRef::new`.
2. `bench(sort)`: the `coordinate_radix_sort` bench + radix-API re-exports (mirrors the existing `natural_compare` bench export).

## Verification

- `cargo nextest -p fgumi-sort` 474 pass; `-E 'test(sort) or test(runall)'` 269 pass.
- CodeRabbit CLI (`--agent`) + CodeRabbit-style self-review: 0 findings.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -150,7 +150,8 @@ because the hot path runs once per record (BAM workloads are millions to billion
 of records) and a safe rewrite measurably regresses sort throughput.
 
 - **`crates/fgumi-sort/src/inline.rs`** — three `#[allow(unsafe_code)]` regions:
-  the `radix_sort_record_refs` (coordinate) and `radix_sort_template_refs` /
+  the `radix_sort_record_refs_with_max` (coordinate; `radix_sort_record_refs`
+  delegates to it after scanning for the max) and `radix_sort_template_refs` /
   `radix_sort_template_field` (template) LSD radix sorts use `Vec::set_len` to
   skip per-element initialization on the auxiliary scratch buffer, plus
   raw-pointer slice swaps to avoid double-borrow restrictions across the
--- a/benches/core_functions.rs
+++ b/benches/core_functions.rs
@@ -1305,6 +1305,82 @@ fn bench_queryname_sort_strategies(c: &mut Criterion) {
     group.finish();
 }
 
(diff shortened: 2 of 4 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #438, opened 2026-06-19
feat(simulate): add `simulate aligner` replay subcommand for runall benchmarking

nh13:
## Summary

Adds `fgumi simulate aligner`, a feature-gated fake streaming-aligner subprocess for `fgumi runall --aligner::command`. It drains the interleaved FASTQ on stdin (discarding it) while concurrently streaming a pre-captured aligned BAM verbatim to stdout, so a benchmark can measure `runall`'s pipeline overhead without a real aligner's compute dominating every run.

It exists to replace the fragile `replay-aligner.sh` shell script in fgumi-benchmarks, which read stdin and wrote stdout in lockstep on a single thread and thereby **deadlocked `AlignAndMerge`** on any input larger than one in-flight window. Confirmed: with the deadlock detector disabled the shell replay hangs indefinitely (>600s, ~1.6% CPU, zero output) on the CODEC sample, while real `bwa mem` on the same input completes — because a real streaming aligner overlaps read/align/write and never withholds output waiting for future input.

## How it works

Two workers that run concurrently and never block each other while streaming:

- a **drainer** that reads stdin to EOF and discards it, and
- an **emitter** that copies the replay BAM verbatim to stdout.

Because the replay file already *is* BGZF/BAM on disk — exactly the byte stream `AlignAndMerge`'s reader expects from a BAM-emitting aligner — the emitter is a raw byte copy: the BAM header is emitted first, and every record (including secondary/supplementary) is passed through in stored order, all for free. `AlignAndMerge` regroups records into templates by queryname, so the replay BAM must be in input (queryname-grouped) order, which is how `bwa mem` emits it.

Backpressure comes for free from the OS pipes plus `AlignAndMerge`'s own in-flight gate; the threads never wait on each other, so it cannot deadlock the caller.

## Details

- New subcommand `src/lib/commands/simulate/aligner.rs`, behind the existing `simulate` feature gate, registered in `simulate/mod.rs` alongside the other `simulate` subcommands.
- On an emitter error (e.g. a corrupt/truncated replay BAM), stdout is closed before joining the drainer so the caller sees EOF and winds down, rather than the drainer join hanging on a caller still feeding FASTQ.
- A `BrokenPipe` on stdout (the caller went away) is a clean exit, the way a real aligner exits on SIGPIPE.
- Trailing positional tokens (the `{ref}` / `{threads}` a command template substitutes) are accepted and ignored, so the command is a drop-in for real-aligner templates.

## Tests

- Unit tests over generic `Read`/`Write`: verbatim passthrough, extra-token tolerance, `BrokenPipe`-clean-exit, an emitter-error-doesn't-hang regression, and a single-threaded write-then-read caller that deadlocks an emit-before-drain implementation (verified the liveness tests fail against the buggy shapes).
- Integration test driving the real binary as a subprocess through OS pipes (the exact `--aligner::command` shape), feeding >64 KiB stdin against a >64 KiB replay and asserting byte-identical stdout + exit 0 within a watchdog, plus a clean non-zero exit on a missing replay BAM.
- End-to-end: the fgumi-benchmarks `runall_end_to_end` CODEC threads-2 case that previously deadlocked now completes in ~14s with the deadlock detector on.

## Docs

- Design doc at `docs/design/2026-06-18-simulate-aligner.md`.
- `docs/simulate-cli.md` documents the new subcommand.

## Notes

- The fgumi-benchmarks `runall_end_to_end` rule is updated separately to invoke `fgumi simulate aligner` and drop `replay-aligner.sh`; that change pins `tools/fgumi` to a commit carrying this subcommand.
- Targets `feat-runall` because that is the branch the runall benchmark uses; the subcommand is independent of the runall pipeline and can be upstreamed to `main` separately.

## Summary by CodeRabbit

* **New Features**
  * Added `fgumi simulate aligner` to benchmark the align stage by streaming a pre-captured BAM to stdout while draining stdin.
  * Accepts extra trailing command tokens (ignored) to match invocation templates.
* **Documentation**
  * Documented the new subcommand and the expected `--replay-bam` ordering/streaming behavior, plus a supporting design note.
* **Bug Fixes**
  * Improved liveness and shutdown behavior to avoid deadlocks and handle broken-pipe scenarios cleanly.
* **Tests**
  * Added integration tests for byte-for-byte BAM passthrough and for failing when `--replay-bam` is missing.

### DIFF
--- a/docs/design/2026-06-18-simulate-aligner.md
+++ b/docs/design/2026-06-18-simulate-aligner.md
@@ -1,0 +1,65 @@
+# Design: `fgumi simulate aligner`
+
+Date: 2026-06-18
+Branch: `nh/feat-simulate-aligner-fr` (off `feat-runall`)
+
+## Problem
+
+fgumi-benchmarks' `runall_end_to_end` rule fakes the aligner by replaying a
+pre-captured aligned BAM (so the benchmark measures fgumi's pipeline, not bwa's
+compute). Today that replay is `workflow/scripts/replay-aligner.sh`, which reads
+stdin and writes stdout in **lockstep on a single thread** — and that deadlocks
+fgumi's AlignAndMerge (verified: hangs indefinitely; real bwa completes). Replace
+it with a small, tested fgumi subcommand.
+
+## Design
+
+`fgumi simulate aligner --replay-bam <bam> <ref>` — a fake aligner subprocess
+invoked via `--aligner::command`. Two independent threads:
+
+- **Drainer:** read stdin to EOF, discard.
+- **Emitter:** copy the replay BAM bytes verbatim (header + all records in stored
+  order) directly to stdout (no BAM parsing/re-encoding, no samtools).
+
+The threads never wait on each other, so it cannot deadlock. A real streaming
+aligner reads and writes concurrently; two threads capture that. Backpressure
+comes for free from the OS pipes plus fgumi's own in-flight gate.
+
+This is deliberately the opposite of the shell script's one-thread lockstep —
+that coupling was the entire bug.
+
+## Details
+
+- Lives at `src/lib/commands/simulate/aligner.rs`, behind the existing
+  `simulate` feature gate, registered in `simulate/mod.rs` like the other
+  `simulate` subcommands.
+- Emit the header before/independently of draining (the emitter is its own
+  thread, so this is automatic).
+- Stream records **verbatim including secondary/supplementary** — fgumi's
+  reader regroups by queryname; filtering would break its template counts.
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #440, opened 2026-06-19
refactor(pipeline): extract pipeline::core into fgumi-pipeline-core crate (120s → 6s compile-fail test)

nh13:
## Summary

Extracts the `pipeline::core` typed-step engine from the `fgumi` crate into a new `crates/fgumi-pipeline-core` crate, re-exported as `pipeline::core` so every existing `crate::pipeline::core::…` path is unchanged.

## Why

The `trybuild` compile-fail test (`pipeline_core_compile_fail`) runs >120s in CI (flagged `SLOW [>120.000s]`). The cases only exercise `pipeline::core` trait/type bounds, but trybuild compiles each in an isolated sandbox that links the **entire `fgumi` dependency graph** — `noodles`-bam, `fgumi-sort`, `fgumi-consensus`, `mimalloc`, etc. — from scratch, none of which the cases need.

`pipeline::core` is actually dependency-light: `ahash`, `crossbeam-queue`, `parking_lot`, `log`, and a single `noodles::sam::Header` type. Moving it to its own crate lets the compile-fail tests build against just that small crate.

**Result: `pipeline_core_compile_fail` drops from ~120s to ~6s.**

## What changed

- New `crates/fgumi-pipeline-core` (24 files, the former `src/lib/pipeline/core` subtree). `git` tracks the moves as renames.
- `fgumi` re-exports it: `pub use fgumi_pipeline_core as core;` in `pipeline/mod.rs`. No call sites change.
- `PipelineBuilder::{append_source, append_step, append_step2}` change `pub(crate)` → `pub` — the `chains` layer (still in `fgumi`) drives them across the new crate boundary. No other visibility changes were needed.
- Two `process2` pipeline-run tests that reached into `pipeline::steps` move from the builder test module to `steps/process.rs` (their natural home — the core crate can't depend on the steps layer).
- The compile-fail test + cases move into the new crate; `.stderr` snapshots regenerated; case imports rewritten to `fgumi_pipeline_core::…`.
- New crate carries `#![deny(unsafe_code)]`; the documented `erased.rs` `#[allow(unsafe_code)]` sites and the CLAUDE.md unsafe-allowlist paths / framework reference are updated to the new location.

## Verification

- 157 `fgumi-pipeline-core` unit tests pass; the 2 relocated `process2` tests pass.
- All workspace test targets compile (`cargo test --workspace --no-run`).
- `cargo fmt --check` and `cargo ci-lint` (clippy pedantic, `-D warnings`) clean.
[...]

PR #441, opened 2026-06-19
docs(pipeline): pin the deliberate Step2 reorder-stage asymmetry

nh13:
## Summary

Resolves a review question on `TypedStep2::build_output_set` (raised against the `fgumi-pipeline-core` extraction, #440): why doesn't it apply the Serial/Exclusive reorder-collapse that the single-input `TypedStep::build_output_set` has?

**Because applying it would be a correctness bug.** This PR documents the deliberate asymmetry and adds a test that pins it.

## Why the asymmetry is intentional

The single-input collapse rests on a *universal* property: a Serial single-input step consumes one already-ordered stream, so any input ordinal it propagates onto an output is emitted in push order — the reorder stage is redundant, and collapsing `ByItemOrdinal` → `None` is a free optimization.

A two-input **merge** has no such guarantee. It interleaves two branches, so a `ByItemOrdinal` output that propagates an *input's* ordinal can be pushed out of ordinal order even under serial (mutex-serialized, single-dispatcher) execution — e.g. when branch B's next-needed ordinal hasn't arrived but branch A's later one has. For such a step the reorder stage is load-bearing; dropping it would silently deliver records out of order.

Today's production `Step2`s happen not to need it — `PairRawFastq` declares `BranchOrdering::None`, and `ZipperMergeStep` assigns fresh *sequential* output ordinals (push order already equals ordinal order) — but the framework cannot assume that for an arbitrary merge, so it conservatively keeps the reorder for every ordered `Step2` output.

## Changes

- A doc comment on `TypedStep2::build_output_set` explaining the deliberate difference from the single-input path.
- `step2_serial_byitemordinal_output_is_reordered_not_collapsed`: pushes a Serial `Step2`'s `ByItemOrdinal` output out of ordinal order (2, 0, 1) and asserts the consumer receives it in ordinal order (0, 1, 2). It fails if the single-input collapse is ever added to the `Step2` path (verified by temporarily applying the collapse).

## Note

Stacked on #440 (the `fgumi-pipeline-core` extraction) so it lives in the new crate location; targets that branch and rebases onto `feat-runall` once #440 merges. No behavior change — documentation + a pinning test.

## Summary by CodeRabbit

* **Tests**
  * Added regression test to ensure ordered output delivery is properly reordered rather than collapsed during serial execution, improving reliability and correctness validation.

### DIFF
--- a/crates/fgumi-pipeline-core/src/erased.rs
+++ b/crates/fgumi-pipeline-core/src/erased.rs
@@ -556,6 +556,31 @@ impl<S: Step2> ErasedStep for TypedStep2<S> {
     }
 
     fn build_output_set(&self) -> (OutputQueueSet, OutputsViewAny) {
+        // Unlike `TypedStep::build_output_set`, a `Step2` does NOT collapse a
+        // `ByOrdinal` / `ByItemOrdinal` output to `None` for `Serial` /
+        // `Exclusive` kinds — the ordering is passed through verbatim and the
+        // reorder stage is kept.
+        //
+        // The single-input collapse rests on a universal property: a Serial
+        // single-input step consumes one already-ordered stream, so any input
+        // ordinal it propagates onto an output is emitted in push order, making
+        // the reorder stage redundant. A two-input MERGE has no such guarantee.
+        // It interleaves two branches, so a `ByItemOrdinal` output that
+        // propagates an *input's* ordinal can be pushed out of ordinal order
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

PR #448, opened 2026-06-20
fix(consensus): clamp consensus depth at i16::MAX instead of wrapping u16

nh13:
## Summary

On very deep UMI families, fgumi silently overflowed the consensus depth counter, corrupting the `cD`/`cM`/`cE` consensus tags and — because the error rate is `errors / depth` — changing a real filter keep/drop decision relative to fgbio. This makes fgumi clamp consensus depth at `i16::MAX` (32_767) exactly as fgbio does.

### How it was found

Benchmarking fgumi against fgbio 4.0.0 on a deep cfDNA panel (`idt-cfdna`), the filter-stage BAM equivalency checks differed for every consensus strategy (simplex edit/identity/adjacency and duplex). The molecule `MI:40915` has **277,312 raw reads** (~138,656 per read-end). fgumi reported `cD = 7_584`, which is exactly `138_656 mod 65_536` — a `u16` wraparound. fgbio reported `cD = 32_767` (`Short.MaxValue`).

### Root cause

`ConsensusBaseBuilder::observations` was `[u16; 4]`, incremented with `+= 1`, and the per-base depth/error vectors were `Vec<u16>`. A single position deeper than 65_535 wrapped the counter (a debug build panics with "attempt to add with overflow"; the release build that runs in production wraps silently). The wrapped depth then:

- corrupts the per-read `cD`/`cM` (max/min depth) and the per-base `cd` array;
- corrupts `cE` (`errors / depth`), which in the `MI:40915` case inflated the error rate past `--max-read-error-rate`, so fgumi **dropped** a consensus read that fgbio **kept** (the 52_756 vs 52_758 record-count difference observed in the benchmark).

This is not a v0.3.1 regression — the `u16` types predate v0.3.0; the path was simply never exercised until a family exceeded 65_535 reads.

### Fix

fgbio's `VanillaUmiConsensusCaller` counts depth in `Int` and clamps the per-base depth/error to `Short.MaxValue` when storing, deriving `cD`/`cM`/`cE` from the clamped arrays (`VanillaUmiConsensusCaller.scala:324-340,380-382`). This mirrors it:

- Widen `observations` / `contributions()` / `observations_for_base()` / `all_observations()` from `u16` to `u32` so the counter never wraps.
- Clamp the per-base depth and error to `i16::MAX` when storing, so `cD`/`cM`/`cE` and the per-base `cd`/`ce` arrays all match fgbio. The **unclamped** depth is still used for the min-reads threshold, exactly as fgbio does (`VanillaUmiConsensusCaller.scala:329`).

**Duplex needs no change.** It builds per-strand consensus through the (now-corrected) vanilla caller and sums the two clamped strand depths as `i32` without re-clamping — which is exactly what fgbio does (`DuplexConsensusCaller.scala:480-489`). Its error accumulation already clamps to `i16::MAX`.

### Test-helper fix

The change also completes the test-only aux-tag scanner `find_i16_array_tag_in_aux` (in `fgumi-raw-bam/testutil`) to skip the unsigned `C`/`S`/`I` (and `A`/`H`) BAM aux types. It previously handled only `c`/`s`/`i`/`f`/`Z`, so when `cD` serializes as `S` (which a value of 32_767 now does) the scanner hit `_ => break` and never reached the per-base array tag. The production BAM reader handles all aux types; this only affected the test helper, and the new test surfaced it.

## Tests

- `base_builder`: 70_000 single-base observations no longer wrap `contributions()` (was `7_584` via `mod 65_536`).
- `vanilla_caller`: a 40_000-read family (one disagreeing observation) reports `cD = cM = 32_767` and `cE = 1 / (32_767 * read_len)` — i.e. computed from the clamped depth, matching fgbio.

`cargo ci-test` (2,214 tests), `cargo ci-fmt`, and `cargo ci-lint` (clippy pedantic, `-D warnings`) all pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented wraparound when counting observations in very deep UMI families.
  * Updated consensus per-base depth/error handling to saturate at the expected i16-compatible limit, matching fgbio semantics.
* **Changes**
  * Widened per-base observation counters returned by the consensus base builder from `u16` to `u32`.
* **Tests**
  * Added coverage for saturation/overflow boundaries.
  * Expanded test utilities to handle more array tag element types.

### DIFF
--- a/crates/fgumi-consensus/src/base_builder.rs
+++ b/crates/fgumi-consensus/src/base_builder.rs
@@ -229,8 +229,13 @@ pub struct ConsensusBaseBuilder {
     /// Kahan summation compensation terms for each base - SIMD vectorized
     compensations: f64x4,
 
-    /// Count of observations for each base
-    observations: [u16; DNA_BASE_COUNT],
+    /// Count of observations for each base.
+    ///
+    /// `u32` (not `u16`) so very deep UMI families — a single position can exceed
+    /// `65_535` reads in high-duplication libraries — accumulate without wrapping.
+    /// The per-base depth is later clamped to `i16::MAX` only at tag emission, to
+    /// match fgbio's `Short` consensus-depth representation.
+    observations: [u32; DNA_BASE_COUNT],
 
     /// Pre-computed correct probabilities adjusted for post-UMI errors
     adjusted_correct_table: Vec<LogProbability>,
@@ -461,7 +466,7 @@ impl ConsensusBaseBuilder {
     ///
     /// This is the sum of observations across all four bases
     #[must_use]
-    pub fn contributions(&self) -> u16 {
+    pub fn contributions(&self) -> u32 {
         self.observations.iter().sum()
     }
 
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #455, opened 2026-06-23
fix(umi): make sequential and parallel assigners agree on UMI case

nh13:
## Summary

Each UMI assignment strategy has a sequential implementation (`crates/fgumi-umi/src/assigner.rs`) and a parallel one (`src/lib/umi/parallel_assigner.rs`), and both are documented to "produce identical results". The parallel implementations fold UMI case (they count, match, and tie-break on the uppercased UMI), but the sequential implementations did not. On mixed-case UMIs the two could therefore diverge:

| Strategy | Sequential (before) | Parallel | Reachable via CLI? |
|---|---|---|---|
| **Adjacency** | equal-count tie-break compared the **raw first-seen** string | uppercased string | No — `group`/`dedup` pre-uppercase non-paired UMIs |
| **Edit** | matching was case-**sensitive** | uppercased | No — same CLI masking as adjacency |
| **Paired** | counting, matching, **and strand A/B assignment** all case-**sensitive** | fully case-insensitive (canonicalizes to uppercase) | **Yes** — the sequential paired path keeps raw case (`group.rs` builds `prefix:part0-prefix:part1` from the raw segments) |

The paired case is the only one reachable through the CLI today: `fgumi group`/`dedup` selects the parallel paired assigner when `--threads > 1` and the sequential one otherwise, so on mixed-case paired/duplex UMIs the two thread settings could produce different groupings *and* different strand assignments. The adjacency and edit divergences are library-only, because the CLI uppercases non-paired UMIs before `assign()`.

## Behavior comparison: fgbio main vs fgumi main vs this PR

How each tool/implementation handles UMI **case**, per strategy. "case-insensitive" = `acgt` and `ACGT` are treated as the same UMI; "case-sensitive" = they are treated as different.

| Strategy | fgbio main | fgumi main — sequential (`--threads 1`) | fgumi main — parallel (`--threads N`) | This PR (both) |
|---|---|---|---|---|
| Identity | case-insensitive | case-insensitive | case-insensitive | case-insensitive (unchanged) |
| Edit | case-sensitive | case-**sensitive** | case-insensitive | **case-insensitive** |
| Adjacency | case-sensitive | counting case-insensitive, **tie-break case-sensitive** | case-insensitive | **case-insensitive** |
| Paired | case-sensitive | case-**sensitive** | case-insensitive | **case-insensitive** |

Verified against fgbio `main` (`GroupReadsByUmi.scala`): its `IdentityUmiAssigner.assign` uppercases, but the edit/adjacency/paired assigners key on the raw (case-sensitive) UMI and the tool feeds them the raw `RX` tag — so fgbio is itself internally inconsistent on case across strategies.

Takeaways:

- **On standard (uppercase ACGT) UMIs, all three columns agree** — every difference above manifests only on non-standard mixed-case UMIs.
- **fgbio main** is internally inconsistent (identity case-insensitive; edit/adjacency/paired case-sensitive).
- **fgumi main** is inconsistent between its own sequential and parallel implementations for edit/adjacency/paired (the bug this PR fixes).
- **This PR** makes fgumi uniformly case-insensitive across all strategies and both implementations.
- This does **not** change fgumi-vs-fgbio behavior on uppercase data. fgumi deliberately normalizes UMIs to uppercase (at the CLI boundary and now uniformly in the assigners), so on mixed-case input fgumi intentionally differs from fgbio's mostly-case-sensitive behavior — and cannot be byte-identical to fgbio there regardless, because fgumi's adjacency counting case-folds via `BitEnc` by design.

## Fix

Fold case in all three sequential assigners so they match their parallel counterparts:

- **Adjacency** — tie-break on the UMI bytes uppercased on the fly (`Iterator::cmp`, no per-comparison allocation).
- **Edit** and **Paired** — uppercase the inputs once at the top of `assign()` and run the existing logic over the uppercased UMIs (strand-assignment logic is unchanged; it now simply operates on the uppercased forms).

This is a no-op for the uppercase UMIs the CLI emits, so it does not change CLI output on standard data.

## Tests

- Sequential-vs-parallel parity on mixed-case input for all three strategies (`test_sequential_and_parallel_{adjacency,edit,paired}_agree_on_mixed_case`), comparing grouping structure (and, for paired, strand). Each fails on `main` and passes here.
- A sequential-only case-folded tie-break test for the adjacency assigner (`test_adjacency_equal_count_tiebreak_is_case_insensitive`).

`cargo ci-fmt`, `cargo ci-lint`, and `cargo ci-test` (2210 tests) all pass.

## Reading order

1. `crates/fgumi-umi/src/assigner.rs` — the three sequential fixes + docs.
2. `src/lib/umi/parallel_assigner.rs` — the four new tests (and a corrected stale "matches the sequential assigner exactly" comment).

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated UMI assignment to treat UMIs case-insensitively across error-correction strategies, including deterministic equal-count tie-breaking for mixed-case inputs.
  * Ensured sequential and parallel assignment produce matching grouping topology for adjacency, edit, and paired/duplex strategies.

* **Tests**
  * Added regression coverage for case-insensitive behavior and parity between sequential and parallel implementations.
  * Included additional assertions covering paired strand/base relationships and consistent base linkage across orientations.

### DIFF
--- a/crates/fgumi-umi/src/assigner.rs
+++ b/crates/fgumi-umi/src/assigner.rs
@@ -897,6 +897,11 @@ impl UmiAssigner for IdentityUmiAssigner {
 /// 3. If one match, add to that group
 /// 4. If multiple matches, merge all matching groups
 ///
+/// # Case sensitivity
+///
+/// UMI matching is case-insensitive: bases are folded to uppercase before comparison, so
+/// this agrees with the parallel edit assigner on mixed-case input.
+///
 /// # Thread Safety
 ///
 /// Uses atomic counter for ID generation, making it safe to use across threads.
@@ -938,12 +943,17 @@ impl UmiAssigner for SimpleErrorUmiAssigner {
             return Vec::new();
         }
 
+        // Match case-insensitively, consistent with the parallel edit assigner (which
+        // uppercases before encoding). Folding case here keeps the two implementations in
+        // agreement on mixed-case input; it is a no-op for the uppercase UMIs the CLI emits.
+        let upper_umis: Vec<Umi> = raw_umis.iter().map(|umi| umi.to_uppercase()).collect();
+
         let mut umi_sets: Vec<BTreeSet<Umi>> = Vec::new();
         // Track seen UMIs to skip duplicate processing
         let mut seen: std::collections::HashSet<&str> =
-            std::collections::HashSet::with_capacity(raw_umis.len());
+            std::collections::HashSet::with_capacity(upper_umis.len());
 
-        for umi in raw_umis {
+        for umi in &upper_umis {
             // Skip duplicate UMIs - they're already in a set
             if !seen.insert(umi.as_str()) {
                 continue;
@@ -1004,7 +1014,7 @@ impl UmiAssigner for SimpleErrorUmiAssigner {
         }
 
         // Build result Vec indexed by input position
-        raw_umis.iter().map(|umi| umi_to_id[umi]).collect()
+        upper_umis.iter().map(|umi| umi_to_id[umi]).collect()
     }
 
     fn as_any(&self) -> &dyn Any {
@@ -1047,6 +1057,21 @@ struct Node {
 ///    - Search for unassigned UMIs with `count < root_count / 2 + 1`
 ///    - If within edit distance, add as child and continue recursively
 /// 4. Assign same molecule ID to root and all its descendants
+///
+/// # Case sensitivity
+///
+/// UMI matching, counting, and the equal-count tie-break are all case-insensitive: bases
+/// are compared after folding to uppercase. This mirrors the parallel implementation so the
+/// two agree on mixed-case input.
+///
+/// This matches fgbio's user-facing output: fgbio's `GroupReadsByUmi` uppercases every UMI
+/// before assignment (`canonicalize(rawTag.toUpperCase)`), so its assigner only ever sees
+/// uppercase UMIs. fgumi folds case inside the assigner instead of relying on the caller,
+/// which is a no-op on the uppercase data both tools group in practice and keeps the library
+/// API from being misused case-sensitively. (fgbio's raw `AdjacencyUmiAssigner` *class*
+/// counts and tie-breaks on the case-sensitive string — its `countMismatches` folds case but
+/// its counting and ordering do not — but that path is never reached with mixed case because
+/// the CLI pre-uppercases.)
 ///
 /// # Multi-threading
 ///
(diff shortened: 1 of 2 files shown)


nh13:
## fgbio baseline validation for the case-folding change

Following up on the request for fgbio-baseline coverage: I drove fgbio's actual `UmiAssigner` classes (`SimpleErrorUmiAssigner`, `AdjacencyUmiAssigner`, `PairedUmiAssigner`) on the exact inputs the new parity tests use, to confirm empirically what fgbio does with mixed-case UMIs rather than asserting it from memory.

### What fgbio does

- **Matching is case-insensitive.** `Sequences.countMismatches` uppercases each base (`Character.toUpperCase`), so the raw edit assigner groups `ACGT` and `acgt` together.
- **Adjacency counting and the equal-count tie-break are case-sensitive.** Counts come from `SimpleCounter` over the raw strings, and ordering is `sortBy((-count, umi))` on the raw UMI string.
- **The `GroupReadsByUmi` CLI uppercases every UMI before assignment** (`canonicalize(rawTag.toUpperCase)`), so fgbio's assigner only ever sees uppercase UMIs in practice.

### Result

With CLI-style uppercasing (what fgbio actually feeds its assigners), fgbio and fgumi produce identical groupings on mixed-case input:

| input | fgbio raw class | fgbio CLI (uppercases first) | fgumi (this PR) |
|---|---|---|---|
| edit `[ACGT, acgt]` | `{ACGT, acgt}` | `{ACGT}` | `{ACGT, acgt}` → one molecule |
| adjacency `[AAAaAA×2, AAAGAC×2, AAAAAC]` | `{AAAGAC, AAAAAC}`, `{AAAaAA}` | `{AAAAAA, AAAAAC}`, `{AAAGAC}` | same as fgbio CLI |
| paired `[ACGT-TGCA, acgt-tgca, TGCA-ACGT]` | `acgt-tgca` split off | one molecule, fwd → strand A, rev → strand B | same as fgbio CLI |

So this PR makes fgumi **converge** with fgbio's user-facing output on mixed-case input — it does not diverge. The only difference is *where* the case fold happens: fgbio folds in the CLI before calling the assigner, while fgumi folds inside the assigner itself, which keeps the library API from being misused case-sensitively. On uppercase input (all real data) the fold is a no-op, so byte-for-byte parity with fgbio is unchanged there.

The parity tests now pin these fgbio-derived groupings (not just sequential-vs-parallel agreement), and the `# Case sensitivity` doc comments describe the relationship accurately.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #456, opened 2026-06-24
fix(docs): write tool/metric index as index.md to fix broken readthedocs link

znorgaard:
## Summary

Fixes the broken https://fgumi.readthedocs.io/en/latest/tools/README.html link (and the equivalent `metrics/README.html`).

**Root cause:** mdBook renders a source file named `README.md` to `index.html` (GitHub-style), but it does **not** rewrite links that *target* `README.md` — it only swaps the extension. So a link to `tools/README.md` became `tools/README.html`, which never exists; the real page is `tools/index.html`.

**Fix:** the doc generators now write the tool/metric index pages as `index.md` instead of `README.md`, and all links point at `index.md`. This removes the `README` special-casing entirely, keeping every link in agreement with the rendered filename.

## Changes

- `generate_tools.rs` / `generate_metrics.rs` — write the index page as `index.md` (with a comment at each write site explaining why `README.md` is unsafe).
- `generate_summary.rs` — sidebar index links → `tools/index.md`, `metrics/index.md`.
- `docs/src/index.md` & `docs/src/guide/working-with-metrics.md` — hand-written links updated to match.
- Added `test_generate_writes_index_not_readme` (xtask) asserting `generate()` writes `tools/index.md` and never `tools/README.md`, as a regression guard.

## Verification

- Rebuilt the book: the home-page link now renders as `tools/index.html` (which exists), a full-book grep for `README.html` is empty, and mdBook reports no broken-link warnings.
- `cargo ci-fmt`, `cargo ci-lint`, and `cargo test -p xtask` (23 tests) all pass.

## Potential follow-up (not in this PR)

The underlying reason this shipped silently is that nothing fails the build on a dead internal link — Read the Docs runs `mdbook build`, but mdBook only *warns* on broken internal links. Adding [`mdbook-linkcheck`](https://github.com/Michael-F-Bryan/mdbook-linkcheck) as an mdBook backend would catch this class of bug across all pages going forward. It's a new tool dependency that both `.readthedocs.yaml` and the local `docs-build` flow would need to install, so I've left it out of this fix — happy to do it in a separate PR if we want the defense-in-depth.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Updated command and metrics reference links to use `index.md` pages.
  * Refreshed guide and landing-page links so reference pages open correctly.

* **Bug Fixes**
  * Fixed broken navigation for generated command and metrics documentation by standardizing the reference page filenames.

### DIFF
--- a/docs/src/guide/working-with-metrics.md
+++ b/docs/src/guide/working-with-metrics.md
@@ -15,7 +15,7 @@ fgumi commands produce structured metrics files for quality control and analysis
 | `simplex-metrics` | Comprehensive simplex QC metrics | `--output` (prefix) |
 | `group` | Family sizes, grouping metrics, position group sizes | `--metrics` (prefix), `--family-size-histogram`, `--grouping-metrics` |
 
-See the [Metrics Reference](../metrics/README.md) for field-level documentation of each metric type.
+See the [Metrics Reference](../metrics/index.md) for field-level documentation of each metric type.
 
 ## File Formats
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2739,6 +2739,7 @@ dependencies = [
  "fgumi",
  "quote",
  "syn 2.0.118",
+ "tempfile",
 ]
 
 [[package]]
--- a/crates/xtask/Cargo.toml
+++ b/crates/xtask/Cargo.toml
@@ -21,6 +21,9 @@ syn = { version = "2", features = ["full", "extra-traits"] }
 quote = "1"
 fgumi = { path = "../.." }
 
+[dev-dependencies]
+tempfile = "3"
+
 [lints.clippy]
 pedantic = { level = "deny", priority = -1 }
 # Relaxations for build tool code (heavy string formatting)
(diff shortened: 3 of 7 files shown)


nh13:
@coderabbitai review

nh13:
Thanks @znorgaard for finding and fixing this!!!

docs/src/guide/working-with-metrics.md
```diff
@@ -15,7 +15,7 @@ fgumi commands produce structured metrics files for quality control and analysis
 | `simplex-metrics` | Comprehensive simplex QC metrics | `--output` (prefix) |
 | `group` | Family sizes, grouping metrics, position group sizes | `--metrics` (prefix), `--family-size-histogram`, `--grouping-metrics` |
 
-See the [Metrics Reference](../metrics/README.md) for field-level documentation of each metric type.
+See the [Metrics Reference](../metrics/index.md) for field-level documentation of each metric type.
```
znorgaard, line 18: [thought](https://pullpo.io/cc?l=thought) 
Note this will have the link work on the readthedocs site, but not when looking at this markdown file in the repo.

We can't have both modes work with relative paths. Lower in this file we link to `https://github.com/fulcrumgenomics/fgumi/blob/main/docs/compare-cli.md` which will always redirect to the github doc file because that file doesn't exist in the mdBook source tree.

So, two slightly different approaches are being used in the same file which I don't love, but seems slightly outside the scope of this PR (fix broken readthedocs links).

PR #459, opened 2026-06-24
fix(consensus): rejects fan-out deadlock and CODEC disagreement data loss (2/6)

nh13:
**Stack: 2 / 6** · base: `nh/runall-fix-1-foundation`

Two critical-path correctness fixes in the consensus/rejects flow.

- **Rejects fan-out deadlock** — all-clean batches now emit an empty (zero-byte) rejects block instead of taking the `only_a` path, keeping the rejects-branch `ByItemOrdinal` reorder serials dense so the pipeline can't wedge.
- **CODEC disagreement data loss** — `consensus_reads_typed` now captures duplex-disagreement source reads into `rejected_reads` and bumps the previously-dead reject counter before propagating the error, on both the threaded and single-threaded codec paths.

Both land with regression tests (the deadlock test is verified non-vacuous — it wedges with the fix reverted).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HDD duplex-disagreement rejection handling: disagreement errors now update HDD rejection statistics and, when reject tracking is enabled, capture the raw rejected inputs for downstream output.
  * Ensured `--rejects` fan-out always emits a per-batch rejects output (even when empty) across CODEC, duplex, simplex, and correction, preventing pipeline wedging and keeping batch serials dense.
* **Tests**
  * Expanded integration coverage for CODEC, duplex, simplex, and correction to validate non-wedging behavior and correct, EOF-terminated BGZF `--rejects` output with byte-identical `RecordBuf`s.
  * Added an integration helper to read BAM records into `RecordBuf` values for stronger assertions.

### DIFF
--- a/crates/fgumi-consensus/src/codec_caller.rs
+++ b/crates/fgumi-consensus/src/codec_caller.rs
@@ -1454,8 +1454,26 @@ impl CodecConsensusCaller {
         &mut self,
         records: Vec<RawRecord>,
     ) -> std::result::Result<ConsensusOutput, CodecConsensusError> {
-        let result = self.consensus_reads_raw(&records)?;
-        // When a group fails to produce consensus, all its records are rejected
+        let result = match self.consensus_reads_raw(&records) {
+            Ok(r) => r,
+            // A recoverable high-duplex-disagreement reject. fgbio routes the
+            // source records to `--rejects` (RejectionReason.HighDuplexDisagreement)
+            // and bumps `consensusReadsFilteredHighDisagreement`. Mirror both here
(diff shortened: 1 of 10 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #482, opened 2026-07-07
feat(fastq): UMI-in-read-name output and BAM→FASTQ on the typed-step chain

### LINKED ISSUE #480: feat(fastq): add UMI-in-read-header output to fgumi fastq (replace samtools fastq -U)
## Motivation

Some downstream tools (e.g. Illumina DRAGEN) consume UMIs embedded in the FASTQ read header rather than in a BAM tag. Users currently reach for `umi-tools`, which is slow and additionally requires a separate trimming step and manual reformatting of the header for duplex UMIs.

`fgumi extract` already does the hard parts — it parses UMIs via read structures, trims them off the template, and stores them in the `RX` tag. What's missing is a fast way to emit FASTQs with the UMI formatted into the read header.

Today this is fully achievable by piping `fgumi extract` into `samtools fastq -U`:

```bash
fgumi extract -i R1.fq.gz R2.fq.gz -r 8M+T 8M+T --sample S --library L -o extracted.bam
samtools fastq -U -n -1 out_R1.fq.gz -2 out_R2.fq.gz -0 /dev/null -s /dev/null extracted.bam
```

`samtools fastq -U` reads the `RX`/`OX` tag, appends the UMI to the read name, and (helpfully) joins duplex UMIs with `+` — exactly the format DRAGEN expects:

```
@readname:AAAAAAAA+CCCCCCCC
```

This works well, but it hardcodes the `:` name separator and `+` duplex join, and it requires an external tool in the pipeline. We'd like `fgumi fastq` to cover this natively so the extract → FASTQ path is a single-vendor, fully-configurable recipe.

## Proposal

Add UMI-in-header support to `fgumi fastq` (BAM → FASTQ), mirroring `samtools fastq -U` with configurable delimiters:

- Read the UMI from a configurable tag list (default `RX,OX`) and append it to the read name.
- Configurable separator between the read name and the UMI (default `:`).
- Configurable delimiter joining multiple/duplex UMIs, translating the tag's stored `-` to the chosen character (default `+` for DRAGEN; other platforms such as MGI/Element/ONT may want a different delimiter or none).
- Paired `R1`/`R2` output files (not only interleaved stdout).
- gzip/bgzf output for the paired files.

`fgumi fastq` already supports dropping the `/1 /2` suffix (`--no-read-suffix`) and preserves the full read name, so those need no change. `fgumi extract` needs no changes — its `RX` tag and trimming already provide the clean hand-off.

### Proposed flags (names TBD)

| Flag | Meaning | Default |
|------|---------|---------|
| `--umi-in-header` | Append the UMI from a tag to the read name | off |
| `--umi-tag <TAGS>` | Tag(s) to source the UMI from | `RX,OX` |
| `--umi-name-delim <CHAR>` | Separator between read name and UMI | `:` |
| `--umi-sep <CHAR>` | Delimiter joining multiple/duplex UMIs | `+` |
| `-1 / -2 <FILE>` | Write R1/R2 to separate files | (interleaved stdout) |
| gzip output | Compress the paired output files | as implied by `.gz` extension |

## Out of scope

- Single-pass FASTQ→FASTQ output directly from `fgumi extract` (no BAM intermediate). Tracked separately if the extra I/O pass proves to be a bottleneck.

nh13:
Reworks `fgumi fastq` to (a) emit UMIs in the read name in Illumina DRAGEN format and (b) run all BAM→FASTQ conversion — interleaved and paired split — through the typed-step pipeline chain, sharing one work-stealing compression pool.

**Stacked on #476** (base `nh/sort-6-chains-wiring`). Do not merge before #476. Review this PR's own diff via the 4 commits below.

**What's here (suggested reading order):**
1. `b8aaa812` — UMI-in-read-name (`--annotate-read-names`/`-U`, `--umi-tag`, delimiters) matching `samtools fastq -U` DRAGEN output, plus paired `-1/-2/-0` BGZF split.
2. `899325db` — route interleaved BAM→FASTQ through the chain (new `WriteRawFile` raw/BGZF sink).
3. `3f5911c8` — new 3-output ordered pipeline primitive (`OrderedBytesTuple3` + `Process3WithWorkerState`), the one-branch-wider analog of the existing `Tuple2`/`Process2`.
4. `cd12c109` — route paired `--out1/--out2/--out0` through the chain via a 3-way fan-out; retire the standalone `run_paired` loop + `FastqSink`.

**Why the chain:** the old paired loop opened a separate pool of `--threads` BGZF workers per output (three `.gz` outputs at `--threads N` = 3N compression workers). On the chain the three per-file compressors share one pool, and RSS is bounded by the lean queue-memory default instead of growing per output.

**Correctness:** the paired encode emits a block on all three ordered branches for every input batch (empty where a branch has no records) — a skipped batch serial would wedge that branch's reorder stage; a deterministic unit test pins this. Paired output is byte-parity with `samtools fastq -1/-2/-0` (asserted in-test); interleaved with `samtools fastq -U`.

**Validation:** full suite green; `cargo ci-lint`/`ci-fmt` clean; `--no-default-features` builds. tricorder on 4.2M records: 5.7× wall speedup at `-@8`, RSS bounded ~390 MB (flat `-@4`→`-@8`) vs ~6 GB under the old per-thread default.

Closes #480.

## Summary by CodeRabbit

* **New Features**
  * Added FASTQ output support in both interleaved and paired/split modes.
  * FASTQ read names can now include UMIs, with configurable tag priority and separators.
  * Added support for raw byte output and improved gzip/BGZF writing behavior.

* **Bug Fixes**
  * Prevents FASTQ exports from overwriting the input or reusing the same output path.
  * Keeps paired FASTQ branch ordering stable, even when a branch has no records.

* **Documentation**
  * Expanded guidance for UMI-annotated FASTQ workflows and updated command descriptions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -681,6 +681,7 @@ dependencies = [
  "fgumi-simd-fastq",
  "fgumi-sort",
  "fgumi-sort-cli",
+ "fgumi-tag",
  "fgumi-umi",
  "flate2",
  "fs4",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -148,6 +148,8 @@ rstest = "0"
 proptest = "1.10"
 criterion = { version = "0.8", features = ["html_reports"] }
 fgumi-raw-bam = { workspace = true, features = ["test-utils", "noodles"] }
+fgumi-tag = { workspace = true }
+flate2 = { version = "1.1", features = ["zlib-rs"] }
 noodles = { version = "0.111.0", features = ["bam", "sam"] }
 trybuild = "1.0"
 
--- a/crates/fgumi-bam-io/src/grouping.rs
+++ b/crates/fgumi-bam-io/src/grouping.rs
@@ -282,6 +282,14 @@ impl DecodedRecord {
         self.data.as_ref()
     }
 
+    /// Immutable access to the underlying [`RawRecord`], for read-only
+    /// consumers (e.g. the FASTQ-encode step) that need the typed accessors
+    /// (`flags()`, sequence, qualities, tags) rather than the raw byte slice.
+    #[must_use]
+    pub fn record(&self) -> &RawRecord {
+        &self.data
+    }
+
     /// Mutable access to the underlying [`RawRecord`]. Used by mid-chain
     /// `Parallel` Process steps that need to mutate record bytes (e.g.,
     /// MQ bumping, tag rewriting) without rebuilding the `DecodedRecord`
--- a/crates/fgumi-bam-io/src/lib.rs
+++ b/crates/fgumi-bam-io/src/lib.rs
@@ -40,6 +40,6 @@ pub use reader::{
 pub use reorder::{DrainReady, ReorderBuffer};
 pub use writer::{
     BamWriter, BgzfWriterEnum, RawBamWriter, bai_sidecar_path, create_bam_writer,
-    create_optional_bam_writer, create_raw_bam_writer, write_bai_index, write_bai_sidecar,
-    write_bam_header,
+    create_bgzf_writer, create_optional_bam_writer, create_raw_bam_writer, write_bai_index,
+    write_bai_sidecar, write_bam_header,
 };
--- a/crates/fgumi-bam-io/src/writer.rs
+++ b/crates/fgumi-bam-io/src/writer.rs
@@ -112,6 +112,25 @@ pub(crate) fn make_bgzf_writer(
     }
 }
 
+/// Create a BGZF-compressed writer over an arbitrary `output`, selecting a
+/// multi-threaded encoder when `threads > 1`.
+///
+/// This exposes the same BGZF backend fgumi uses for BAM output so other output
+/// formats (e.g. gzipped FASTQ) get identical, block-gzip (BGZF) framing and
+/// multi-threaded compression rather than a separate implementation.
+///
+/// `compression_level` is the BGZF/DEFLATE level (0–12, where 0 means
+/// uncompressed); values outside the valid range fall back to the backend
+/// default.
+#[must_use]
+pub fn create_bgzf_writer(
+    output: Box<dyn Write + Send>,
+    threads: usize,
+    compression_level: u32,
+) -> BgzfWriterEnum {
+    make_bgzf_writer(output, threads, compression_level)
+}
+
 /// Write a BAM header (magic, SAM text, reference sequences) to any writer.
 ///
 /// Shared implementation used by the BAM writer factories (e.g. [`RawBamWriter`]).
--- a/crates/fgumi-pipeline-core/src/handles.rs
+++ b/crates/fgumi-pipeline-core/src/handles.rs
@@ -1031,6 +1031,32 @@ where
     }
 }
 
+impl<A, B, C> OutputHandles<crate::outputs::OrderedBytesTuple3<A, B, C>>
+where
+    A: Send + HeapSize + Ordered + 'static,
+    B: Send + HeapSize + Ordered + 'static,
+    C: Send + HeapSize + Ordered + 'static,
+{
+    /// Borrow the typed per-branch view for an ordered + byte-bounded 3-tuple
+    /// output. Same view type as plain `(A, B, C)` — the view carries only
+    /// `BranchOutputHandle`s; the ordering is encoded in the queue transport.
+    ///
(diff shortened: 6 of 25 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai On the `-N`/always-suffix point (fastq.rs:481): intentionally not adding a force-suffix flag. In split `--out1`/`--out2` mode the `/1` `/2` suffix is omitted to match `samtools fastq -1/-2` default behavior, which keeps R1/R2 names identical for downstream pairing — the correct convention here. No current use case needs `-N`-style forced suffixes in split files, so leaving it out (YAGNI); happy to add it if a concrete need arises.

On the per-output BGZF thread pools (fastq.rs:206): kept the undivided per-sink pools (benchmarked ~7% faster at low `-@` than dividing the budget) and now log the effective total compression-worker count when multiple gzipped outputs are opened, so oversubscription is visible. A deeper fix — routing all FASTQ compression through the shared work-stealing pool via a typed-step FASTQ sink — is planned as a follow-up once the sort-engine stack lands.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed the review feedback (branch rebased onto the updated base and force-pushed):

**Inline threads (resolved):**
- `docs/src/guide/best-practices.md` — fixed the h2→h4 heading skip (`####` → `###`).
- `src/lib/commands/fastq.rs` — the duplicate-output dedup now canonicalizes the existing **parent directory** and re-joins the file name for not-yet-existing outputs, so aliases that differ only through a symlinked intermediate directory (or a leading `./dir/`) still collide before two sink threads can truncate the same target. Added a `#[cfg(unix)]` regression test (`test_fastq_aliased_nonexistent_output_paths_rejected`) that fails without the fix.
- `tests/integration/test_fastq_command.rs` — `read_fastq_maybe_gz` now decodes `.bgz` as well as `.gz`, and a new `test_fastq_paired_output_bgz_extension_is_bgzf` exercises the `.bgz` output path.

**Comments that failed to post (no thread):**
- `crates/fgumi-bam-io/src/writer.rs` — corrected the `create_bgzf_writer` doc range from `(1–12)` to `(0–12, where 0 means uncompressed)`, matching the sibling doc comments and the `0..=12` convention already used elsewhere in the file.
- `crates/fgumi-bam-io/src/grouping.rs` (rstest for the group-key loop tests) — **not applied here**: those tests are pre-existing on the base branch (`nh/sort-6-chains-wiring`), not part of this PR’s diff (this PR only adds the `record()` accessor to `grouping.rs`). Folding a test refactor of base-branch code into this PR would pollute its diff; the rstest change belongs on the base PR.

All 2318 tests pass; `cargo fmt`/`clippy --pedantic` clean.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #484, opened 2026-07-08
fix(zipper): right-size SAM-parsed records to bound pipeline memory

nh13:
## Summary

Pipeline-mode `fgumi zipper` reading SAM on stdin — the canonical `bwa mem … | fgumi zipper` — buffered essentially its entire decompressed input, reaching **14–18 GiB** on a vendor WES sample (`agilent-hs2`) while the fast-path and BAM-input paths stayed bounded (~0.1–2 GiB). Root-caused with `dhat` to a per-record over-allocation in the SAM parse step.

## Root cause

`parse_sam_chunk_into_decoded` handed every `DecodedRecord` the encoder's **≥4 KiB scratch buffer** (via `mem::replace(encoder.get_mut(), Vec::with_capacity(cap.max(4096)))`), so a ~60-byte record body carried a 4 KiB allocation — a **~20× per-record over-allocation**. Because records are held in flight through the pipeline, that scaled to tens of GiB. The BAM-input path was unaffected (its records are right-sized). dhat confirmed exactly: `703,413 blocks × 4 KiB` live at the memory peak.

Compounding it, `DecodedRecordBatch::new` summed `raw_bytes().len()` (~200 B) rather than allocated capacity (4 KiB) into `total_bytes`, so the byte-backpressure budget under-counted real memory 20× and couldn't bound the queue.

## Fix

- **`parse/sam.rs`**: copy each encoded body into a right-sized buffer (`to_vec`) and reuse the encoder's scratch buffer across records, instead of donating the oversized buffer per record. The encoder still grows once to the largest record and never reallocates; output is unchanged.
- **`pipeline/steps/types.rs`**: `DecodedRecordBatch::new` sums each record's allocated capacity (`MemoryEstimate::estimate_heap_size`), matching `BamTemplateBatch`'s accounting, so byte-backpressure reflects true heap footprint and can't be fooled by an over-allocated record.

## Verification

- New unit tests (TDD, both fail before the fix): records are right-sized, and `heap_size` reflects capacity not len.
- End-to-end: `agilent-hs2` `zipper --threads 2` drops **18 GiB → 2.7 GiB** with **identical output** (58,971,241 records).
- `ci-fmt` / `ci-lint` clean; full `ci-test` green (2261 passed).

Targets `nh/sort-6-chains-wiring` (#476) because the chains SAM-parse path lives on this stack.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory handling for decoded records to avoid retaining oversized internal buffers and to keep per-record storage closer to actual data size.
  * Updated decoded record batch memory accounting to use allocated capacity/heap footprint rather than raw byte length, improving consistency and downstream behavior.
* **Tests**
  * Added regression coverage to ensure decoded records don’t keep excessive encoder capacity.
  * Added checks that batch heap/memory estimates match allocated buffer capacity, preventing capacity/length mismatches.

### DIFF
--- a/src/lib/commands/zipper.rs
+++ b/src/lib/commands/zipper.rs
@@ -989,24 +989,29 @@ pub(crate) mod merge_step {
         pub output_byte_limit: u64,
     }
 
-    /// One batch in flight on a branch + the cursor into it.
+    /// One batch's templates in flight on a branch + the cursor into them.
+    /// Owns the `Vec<Template>` moved out of the batch (via
+    /// [`BamTemplateBatch::into_parts`]) so the drain cursor can replace
+    /// slots in place — the cached `total_bytes` is dropped with the batch,
+    /// so there is no stale accounting to worry about.
     struct PendingBatch {
-        batch: BamTemplateBatch,
+        templates: Vec<Template>,
         cursor: usize,
     }
 
     impl PendingBatch {
         fn new(batch: BamTemplateBatch) -> Self {
-            Self { batch, cursor: 0 }
+            let (_, templates) = batch.into_parts();
+            Self { templates, cursor: 0 }
         }
 
         fn current(&self) -> Option<&Template> {
-            self.batch.templates.get(self.cursor)
+            self.templates.get(self.cursor)
         }
 
         fn take_current(&mut self) -> Option<Template> {
             let i = self.cursor;
-            if i >= self.batch.templates.len() {
+            if i >= self.templates.len() {
                 return None;
             }
             self.cursor += 1;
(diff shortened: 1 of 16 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed the outside-diff finding on `DecodedRecordBatch` cache-safety, and applied the same treatment to its sibling `BamTemplateBatch` for consistency:

- Both structs' fields (`batch_serial`, `records`/`templates`, `total_bytes`) are now **private**. The cached `total_bytes` can no longer drift out of sync with the records/templates it accounts for — there is no in-place mutation path (`records_mut`/`templates_mut` deliberately do not exist), so byte-bounded admission control always sees the true heap footprint.
- Callers construct via `new` (recomputes the cache) or `BamTemplateBatch::from_parts` (the documented escape hatch for a caller that already summed `heap_size`), read via `batch_serial()` / `records()` / `templates()` / `total_bytes()`, and take ownership by consuming the batch via `into_records()` / `into_parts()`.
- Updated every consumer accordingly (group/correct/clip/filter/serialize/templates_to_records/zipper/align_and_merge). The one prior in-place mutation site (`zipper`'s template-drain cursor) now owns the moved-out `Vec<Template>`; the `BumpMq` test mutates then rebuilds via `new` so the cache is recomputed.
- Added round-trip tests for the accessor/consume contract on both types.

PR #485, opened 2026-07-08
fix(group,dedup): pair duplex strands sharing an unclipped 5' coordinate

nh13:
## Summary

fgumi's `paired` grouping strategy can split a single duplex molecule into two when a template's two mates share an unclipped 5' coordinate — i.e. fully-overlapping / short-insert fragments. This diverges from fgbio `GroupReadsByUmi`, which pairs the two strands into one molecule.

## Root cause

Duplex strand pairing prefixes each half of the paired UMI with a lower- vs higher-coordinate-read marker (`lowerReadUmiPrefix` / `higherReadUmiPrefix`) so the top strand's prefixed UMI is the exact reverse of the bottom strand's. That prefix is chosen by `is_r1_genomically_earlier_raw`, which broke the position tie with `r1_pos <= r2_pos`. On a tie (`r1_pos == r2_pos`) that returns `true` for **both** strands (top = R1 forward, bottom = R1 reverse), so both get the lower prefix on R1's half, their prefixed UMIs are no longer reverses of each other, they fail to reverse-match in the assigner, and the molecule is split in two.

fgbio breaks the same tie on strand (`GroupReadsByUmi.umiForRead`): `pos1 < pos2 || (pos1 == pos2 && r1.positiveStrand)`. This PR mirrors that — on a tie, R1 is "earlier" iff it is on the forward strand. The buggy helper was shared by `group` and `dedup`; both are fixed.

## `compare bams` was blind to it

The grouping-mode comparison keyed only on the full MI (`base` + `/A`|`/B`), treating the two strands as independent groups. A strand-pairing split keeps the per-MiKey mapping a consistent bijection (`X/A↔X/A`, `X/B↔Y/A`), so it reported the split as `EQUIVALENT` with zero mismatches — which is why the benchmark's own `group.paired` equivalence check never flagged this. This PR adds a base-level (strand-suffix-stripped) molecule-membership check: reads that share a molecule in one BAM must share a molecule in the other. A genuine strand-pairing split now makes the groupings `DIFFER`; a molecule relabel or an `/A`↔`/B` swap stays `EQUIVALENT`.

## Tests

- New regression test `test_paired_assigner_pairs_overlapping_duplex_strands`: builds the two strands of one duplex molecule at equal unclipped 5' positions and asserts they group into one molecule. Verified RED on the pre-fix code (`["0/A","0/A","1/A","1/A"]`, two bases), GREEN after.
- Repaired `test_paired_assigner_explicit_ab_ba_symmetry`: the port of fgbio's "correctly group reads with the paired assigner when the two UMIs are the same" used an FR-only builder (so it never built the reverse-strand bottom read) and asserted two separate groups, inverting fgbio's contract. It now builds a real reverse-strand bottom read and asserts one molecule with `/A` + `/B`, matching fgbio.
- Unit tests for the compare base-level check (split flagged; molecule relabel, `/A`↔`/B` swap, and single-strand relabel not flagged).

Full suite green (2215 tests); fmt and clippy (pedantic) clean.

## Validation on real data

Reproduced end-to-end on the agilent-hs2 vendor sample (SRR30485713): fgumi's own `group → duplex → filter` produced 1,025,610 filtered records vs fgbio's 1,025,612 — a single degenerate duplex molecule (chr11:119116712; 8 top-strand + 3 bottom-strand templates of a fully-overlapping fragment) that fgumi split and fgbio kept. After the fix fgumi produces 1,025,612, byte-identical to fgbio. The behavior is present on `main` (0.4.0), so it is not a recent regression. The fixed `compare bams` flags exactly that one molecule out of 10,220,662 (`Grouping mismatches: 1`, `DIFFER`) where the old tool reported `EQUIVALENT`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved BAM grouping consistency checks with an additional molecule-level duplex strand-pairing validation to catch split/merge discrepancies.
  * Fixed strand-aware tie-breaking for paired-read processing when mates share identical unclipped 5’ coordinates.
  * Updated pairing/UMI grouping behavior so duplex strands from the same molecule remain together and aren’t split.
* **Tests**
  * Added unit tests for duplex pairing mismatch detection, and expanded regression coverage for label swapping, relabeling, single-strand cases, and fully overlapping short-insert templates.

### DIFF
--- a/src/lib/commands/common.rs
+++ b/src/lib/commands/common.rs
@@ -889,6 +889,36 @@ pub(crate) fn serialize_raw_bam_records<R: AsRef<[u8]>>(
     Ok(records.len() as u64)
 }
 
+/// Checks whether R1 is genomically earlier than R2 from their raw BAM bytes.
+///
+/// Shared by `group` and `dedup` so their paired-UMI assignment can't drift.
+/// Uses zero-allocation CIGAR iteration; unmapped reads return position 0
+/// (matching noodles `unwrap_or(0)` behavior). Ordering is by reference id,
+/// then unclipped 5' coordinate, then strand.
+///
+/// The strand tie-break mirrors fgbio `GroupReadsByUmi.umiForRead`'s
+/// `pos1 == pos2 && r1.positiveStrand`: when both mates share an unclipped 5'
+/// coordinate (fully-overlapping / short-insert pairs), R1 is "earlier" iff it
+/// is on the forward strand. A bare `r1_pos <= r2_pos` returns `true`
+/// unconditionally on a tie, which assigns the lower/higher paired-UMI prefix
+/// inconsistently between the two duplex strands — so the strands fail to
+/// reverse-match and one duplex molecule is incorrectly split into two.
+pub(crate) fn is_r1_genomically_earlier_raw(r1: &[u8], r2: &[u8]) -> bool {
+    use fgumi_raw_bam::RawRecordView;
+
+    let ref1 = fgumi_raw_bam::ref_id(r1);
+    let ref2 = fgumi_raw_bam::ref_id(r2);
+    if ref1 != ref2 {
+        return ref1 < ref2;
+    }
+    let r1_pos = fgumi_raw_bam::unclipped_5prime_from_raw_bam(r1);
+    let r2_pos = fgumi_raw_bam::unclipped_5prime_from_raw_bam(r2);
+    if r1_pos != r2_pos {
+        return r1_pos < r2_pos;
+    }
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #487, opened 2026-07-08
feat(simd-fastq): validate record structure and surface errors instead of panicking

nh13:
Stacked on #486. This closes the FASTQ-reader validation gaps that the #486 investigation surfaced: fgumi reimplemented its FASTQ reader from scratch (`fgumi-simd-fastq`) rather than porting fgbio's `FastqSource`, and its parsing contract diverged from what fgbio enforces (and tests in `FastqIoTest`).

> **Reviewer note:** please review after #486 — the base of this PR is `nh/feat-extract-strip-read-number-suffix`, not `main`. GitHub will retarget it to `main` automatically once #486 merges.

## Problem

`SimdFastqReader` only enforced one of the four structural invariants fgbio's `FastqSource` checks, and did so by panicking:

| fgbio `FastqSource` rejects… | fgumi reader (before) |
|---|---|
| header not starting with `@` | ✅ enforced — but via `assert!` (**panic**) |
| quality header not starting with `+` | ❌ **silently accepted** |
| `seq.len != qual.len` | ❌ **silently produced a corrupt record** |
| not the four expected lines | ⚠️ `assert!` (panic) / dropped as leftover |

The length-mismatch case is the worst: it emitted a record with mismatched sequence and quality with no error at all.

## Change

- Add a fallible parse core, `try_parse_single_record`, returning a typed `FastqParseError` (`MissingAtPrefix`, `TooFewLines`, `MissingPlusPrefix`, `LengthMismatch { seq_len, qual_len }`). It validates all four invariants.
- The streaming `SimdFastqReader` (the production path, used by `extract` via `ReadSetIterator`) now routes failures through its **existing** `io::Result` item channel as an `InvalidData` error instead of panicking. `extract` surfaces a clean message and a non-zero exit.
- The infallible zero-copy `parse_records` scan keeps its panic-on-malformed contract via a thin documented wrapper (it assumes well-formed/trusted input, e.g. benchmarks).

The new checks are **O(1)** beyond the newline scan already required for record framing (a single-byte compare for `+`, a length compare for seq/qual) — negligible against the per-record allocations already on the path, so no hot-path regression.

### End-to-end behavior

Before: a length-mismatched record was written silently. After:

```
$ fgumi extract --inputs bad.fq --output out.bam --sample s --library l --read-structures +T
Error: FASTQ sequence and quality lengths differ (4 bases vs 2 quals)
```

## Tests (TDD)

Written test-first (watched each fail before implementing):

- **Reader-level** (`reader.rs`), mirroring fgbio's `FastqIoTest` throwing cases adapted to fgumi's `io::Result` contract: missing `@`, missing `+`, and seq/qual length mismatch each yield an `InvalidData` error rather than panicking or silently succeeding.
- **Parser-level** (`parser.rs`) for `try_parse_single_record`: the `Ok` case, a **zero-length-read** `Ok` guard (empty seq/qual, `0 == 0`, must not regress), and one case per error variant.

Full suite green (2219 passed); fmt and clippy clean.

## Why this wasn't caught before

The `/1`/`/2` and validation behaviors are all properties of fgbio's `FastqSource` (the reader), tested in `FastqIoTest` — not `FastqToBam`/`FastqToBamTest`. The fgumi port targeted `FastqToBam` → `extract` and wrote a fresh SIMD reader, so neither the reader behaviors nor their tests came across. #486 fixed the `/1`/`/2` half; this PR fixes the validation half.

## Audit of the other FASTQ path (done)

The unified-pipeline FASTQ parser (`src/lib/unified_pipeline/fastq.rs` + `src/lib/fastq_parse.rs::FastqRecord::from_slice`) was audited and is **already clean** — no changes needed:

- `FastqRecord::from_slice` independently validates all four invariants (`@`, four lines, `+`, equal seq/qual lengths) and returns `io::Result` `InvalidData`, with unit tests per error case. Its boundary scanners (`detect_prefix_end` and the reverse scan) re-validate all four per 4-line window, so malformed records can't be mis-framed.
- The `/1`/`/2` QNAME strip from #486 already covers this path: extract's unified route (`run_fastq_pipeline` → `make_raw_records_static`) converges on the same `extract_read_name_and_umi` that #486 fixed. The broad `strip_read_suffix` in `fastq_parse.rs` is only used by `grouper.rs` for name *comparison*, never to write a QNAME, so it does not over-strip output names.

The only real gap was the `fgumi-simd-fastq` reader, which this PR fixes. (Minor: the 4-invariant check now lives in three spots — this reader, `from_slice`, and the boundary scanners — a possible future consolidation, but each has a distinct borrowed/owned/scan context.)

## Summary by CodeRabbit

* **New Features**
  * Exposed `FastqParseError` in the public API for structured, fallible FASTQ parsing.
* **Bug Fixes**
  * Improved FASTQ parsing validation to report specific structural issues (missing prefixes, incomplete records, and sequence/quality length mismatches).
  * While iterating FASTQ input, malformed records now surface as clear invalid-data errors and preserve the original parse error details for troubleshooting.

### DIFF
--- a/crates/fgumi-simd-fastq/src/lib.rs
+++ b/crates/fgumi-simd-fastq/src/lib.rs
@@ -33,5 +33,5 @@ mod reader;
 
 pub use bitmask::FastqBitmask;
 pub use lexer::lex_block_full;
-pub use parser::{FastqRecord, find_record_offsets, parse_records};
+pub use parser::{FastqParseError, FastqRecord, find_record_offsets, parse_records};
 pub use reader::SimdFastqReader;
--- a/crates/fgumi-simd-fastq/src/parser.rs
+++ b/crates/fgumi-simd-fastq/src/parser.rs
@@ -1,5 +1,7 @@
 //! FASTQ parser: uses SIMD-produced newline bitmasks to find record boundaries.
 
+use std::fmt;
+
 use crate::lexer;
 
 /// A borrowed FASTQ record with zero-copy slices into the input buffer.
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed the two outside-diff-range findings from the latest review (both in `crates/fgumi-simd-fastq/src/parser.rs`):

- **🔴 Guard the quality slice before indexing** — `try_parse_single_record` panicked on `@r\nA\n+\n` (record ends right after the `+` line, so `qual_start > qual_end` and `&record[qual_start..qual_end]` panicked). Added a bounds check that returns `FastqParseError::TooFewLines` before slicing, plus a regression case (`plus_line_but_no_quality`) in the malformed-input table.
- **proptest for structural invariants** — added `test_try_parse_single_record_never_panics`, a proptest feeding arbitrary byte slices and asserting the parser only ever returns `Ok`/`FastqParseError`, never panics.

Both fixes are amended into the existing commit.

nh13:
@coderabbitai review

PR #489, opened 2026-07-08
fix(extract): match fgbio strict read-name UMI extraction (EXT-01/03/04)

nh13:
## Summary

`fgumi extract --extract-umis-from-read-names` is fgumi's port of fgbio's `FastqToBam -n`, which calls `Umis.extractUmisFromReadName(name, strict=true)`. fgbio's strict path upper-cases the extracted UMI, reverse-complements `r`-prefixed segments, and **throws** on any character outside `ACGTN-`. fgumi previously wrote the last read-name field to `RX` verbatim — no upper-casing, no `r`-revcomp, and it silently accepted illegal UMI characters.

This closes three findings from the fgbio behavioral-parity tracker:

- **EXT-03** — upper-case the read-name UMI (`Umis.scala:111-117`). Case-sensitive UMIs otherwise mis-group vs fgbio.
- **EXT-04** — reverse-complement `r`-prefixed segments (`Umis.scala:102-115`, `reverseComplementPrefixedUmis` defaults true, as `FastqToBam` uses it). For a `+`-delimited dual UMI only the prefixed segment is reverse-complemented. Reuses the shared `fgumi_dna::dna::reverse_complement` helper rather than a new copy.
- **EXT-01** — reject (return an error) a UMI containing any character outside `ACGTN-`, matching fgbio strict mode (`Umis.scala:121-123`, `UmisTest.scala:110-114`). The surfaced error names the offending read.

**Out of scope by decision:** EXT-02 — fgumi's lenient `>=8`-field / last-field extraction (which also supports demultiplexers that append a sample index, producing 9+ fields) is intentionally kept (tracker Decisions Q2). The field-count logic is unchanged; only the transform + validation of the extracted field now matches fgbio.

The new `is_valid_umi_char` mirrors fgbio's `Umis.isValidUmiCharacter` (`ACGTN-`) and is deliberately distinct from `fgumi_umi::validate_umi`, which implements the more permissive `GroupReadsByUmi` counting rule (only rejects upper-case `N`).

## fgbio ↔ fgumi parity evidence

Adversarial fixture (8-field read names, so fgbio strict and fgumi both select the same last field), run against **fgbio 4.1.0**:

| read-name UMI | fgbio `FastqToBam -n` | fgumi **before** | fgumi **after** |
|---|---|---|---|
| `…:acgtn` | `ACGTN` | `acgtn` | `ACGTN` |
| `…:rGGTTAA` | `TTAACC` | `rGGTTAA` | `TTAACC` |
[...]

PR #491, opened 2026-07-08
fix(dna): IUPAC-aware, case-preserving complement via shared 256-byte LUT

nh13:
## What & why

fgumi's `complement_base` handled only A/C/G/T/N and passed every IUPAC ambiguity code through **un-complemented**, while also uppercasing lowercase bases — diverging from fgbio's `Sequences.complement` / htsjdk `SequenceUtil`. `SequencesTest.scala:161` pins `Sequences.revcomp("NRG") == "CYN"`; fgumi produced `"CRN"`. The same defect lived in two more independently hand-rolled tables: `fgumi-raw-bam`'s per-base tag reverse-complement, and the `fastq` command's BAM→FASTQ complement (the latter folding IUPAC codes to `N`).

This replaces all three with a single shared `const [u8; 256]` table, `fgumi_dna::COMPLEMENT`:

- full IUPAC complement (`R↔Y`, `K↔M`, `B↔V`, `D↔H`, `S`, `W`, `U→A`, `N`)
- case-preserving (`a→t`), matching fgbio (CODEC lowercase-`n` round trips)
- unknown bytes pass through unchanged (identity); the `simulate` command relies on this
- branch-free single load (a `u8` index into a 256-entry table elides the bounds check), replacing the previous branchy `match` — no hot-path regression

### Call sites rewired

- `fgumi-dna` — `complement_base` / `reverse_complement{,_in_place,_str}`
- `fgumi-raw-bam` — `reverse_complement_string_tag_in_place` (gains a `fgumi-dna` dep)
- `fastq` command — its `COMPLEMENT` static derives its values from the shared table but keeps the FASTQ-validity policy of folding invalid bytes to `N` (unlike the consensus/tag paths, which pass unknowns through)

Also removes the now-redundant `complement_base_preserve_case` (`complement_base` is itself case-preserving) and points `reverse_complement_str` at it.

## Tracker findings

Addresses **R2-SEQ-01** (IUPAC complement), **R2-SEQ-02** (case preservation), **R2-SEQ-03** (documented unknown-passthrough), **R2-SEQ-04** (the tag-path copy), and **R2-SEQ-05** (the `fastq` command's table) from the round-2 fgbio behavioral-parity audit.

## Parity evidence

The correct outputs are fgbio's own unit-test vectors: `SequencesTest.scala:160-161` asserts `revcomp("NRG") == "CYN"` and `revcomp("acacNNNN") == "NNNNgtgt"`. The new fgumi tests assert exactly these, written RED-first — before the fix `reverse_complement(b"NRG")` returned `CRN`; after, `CYN`. A synthetic CLI fgbio-vs-fgumi run was deliberately skipped for this item: it is a pure primitive whose ground truth *is* those fgbio vectors, and the CLI trigger (IUPAC bases in read sequence or tags) does not occur in real data, so a run would only exercise a hand-injected fixture. The tool-level round-2 items (clip, codec, group/consensus guards, metrics) will carry the full end-to-end fgbio reproduction.

## Testing

- TDD RED→GREEN for each of the three sites (`dna.rs`, `tags.rs`, `fastq.rs`).
- Full workspace suite: 2223 passed, 0 failed. `cargo ci-fmt` and `cargo ci-lint` clean.
- CodeRabbit self-review (`/coderabbitai-review`) run against the branch diff; its one finding — the third, un-unified `fastq.rs` complement table — is folded into this PR as R2-SEQ-05.

## Stacking

Stacked on #489 (`nh/fix-extract-readname-umi-strict`). Both PRs edit `crates/fgumi-dna/src/dna.rs`: #489 adds `U→A` uracil complement for its `r`-prefix UMI reverse-complement, and this PR's shared LUT subsumes that (the table already maps `U→A`/`u→a`; #489's uracil test coverage is preserved). **Merge #489 first** — this PR will retarget to `main` automatically once #489 lands.

## Summary by CodeRabbit

* **New Features**
  * Reverse-complementing now fully supports IUPAC nucleotide codes with case preserved.
  * String reverse-complement leaves non-ASCII characters unchanged.
  * Exposed a public complement lookup for consistent use across components.
* **Bug Fixes**
  * FASTQ and tag reverse-complementing now correctly handles ambiguous bases; invalid/non-alphabet bytes map to a safe default.
* **Breaking Changes**
  * Removed the previously available case-preserving single-base complement helper; switch to the new complement lookup.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -775,6 +775,7 @@ dependencies = [
  "anyhow",
  "bstr 1.12.1",
  "bytemuck",
+ "fgumi-dna",
  "fgumi-tag",
  "noodles",
  "proptest",
--- a/crates/fgumi-dna/src/dna.rs
+++ b/crates/fgumi-dna/src/dna.rs
@@ -2,49 +2,91 @@
 //!
 //! This module provides common DNA sequence operations like reverse complement.
 
-use crate::{NO_CALL_BASE, NO_CALL_BASE_LOWER};
+/// Lookup table mapping each byte to its Watson–Crick / IUPAC complement.
+///
+/// Single source of truth for base complementation, shared by [`complement_base`]
+/// (and thus every consensus reverse-complement path) and by `fgumi-raw-bam`'s
+/// per-base tag reverse-complement. It mirrors fgbio's `Sequences.complement` /
+/// htsjdk `SequenceUtil`:
+///
+/// - **Full IUPAC**: `A<->T`, `C<->G`, `U->A`, `R<->Y`, `K<->M`, `B<->V`,
+///   `D<->H`, and the self-complements `S`, `W`, `N`.
+/// - **Case-preserving**: lowercase input yields lowercase output (`a->t`),
+///   matching fgbio — e.g. CODEC's lowercase `n` padding survives round trips.
+/// - **Unknown bytes pass through unchanged** (identity default). fgbio *throws*
+///   on invalid bytes; fgumi keeps the non-panicking passthrough because such
+///   bytes are unreachable on real ACGTN(+IUPAC) sequence/tag data and the
+///   `simulate` command relies on passthrough (`fastq_reads.rs`). Strict
+///   validation, if ever needed, belongs at a dedicated entry point, not here.
+///
+/// Indexed by a `u8` into a 256-entry table, so the bounds check is elided:
+/// complementation is a single branch-free load.
+pub const COMPLEMENT: [u8; 256] = {
+    let mut table = [0u8; 256];
+    // Identity default so unknown bytes pass through unchanged. Counter is a `u8`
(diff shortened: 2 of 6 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #492, opened 2026-07-08
fix(clip): unmap fully-clipped reads and match fgbio unmap/mate fidelity (CLIP-01)

nh13:
## What & why

`clip` (and every internal caller of `SamRecordClipper` / `RawRecordClipper`) diverged from fgbio when clipping removes all of a read's aligned bases. This PR closes the gap and, in doing so, ports the full fgbio/htsjdk fidelity for the unmap and mate-info paths.

fgbio `SamRecordClipper.clip{Start,End}OfAlignment` computes `numberOfClippableBases` (the sum of alignment `M`/`=`/`X` and insertion `I` CIGAR op lengths) and, when the requested clip is `>=` that count, calls htsjdk `SAMUtils.makeReadUnmapped` and returns — the read becomes **unmapped** with an **empty CIGAR**, keeping its bases/qualities. fgumi instead clipped every aligned base, leaving the read **mapped** at a stale `POS` with an all-clip CIGAR (e.g. `50H`) and an empty sequence — silent corruption. This is tracker finding **CLIP-01**.

## Fix

**1. Unmap when the whole alignment is clipped (CLIP-01).**

- Add the `numberOfClippableBases` guard to **both** `SamRecordClipper` and `RawRecordClipper`, in `clip_start_of_alignment` and `clip_end_of_alignment`: when `clippable <= requested`, unmap the read and return the clippable count, rather than proceeding to clip.
- `make_read_unmapped[_raw]` mirrors htsjdk `SAMUtils.makeReadUnmapped`: clear reference / pos / mapq / CIGAR / TLEN and the duplicate, secondary, supplementary, and proper-pair flags; set `UNMAPPED`; reverse-complement bases + reverse quals (then clear the reverse flag) for reverse-strand reads; bases/quals are otherwise preserved.
- Two fgumi unit tests that pinned the old all-soft-clip behavior (`10S10M` clip 10, `10M` clip 10) are updated to assert the unmap.

**2. Reorient strand-sensitive aux tags on unmap (htsjdk `reverseComplement(true)` fidelity).**

htsjdk `makeReadUnmapped` reverse-complements a reverse-strand read via `reverseComplement(true)`, which does more than `SEQ`/`QUAL`: it also reverse-complements the base-oriented tags in `SAMRecord.TAGS_TO_REVERSE_COMPLEMENT` (`E2`, `SQ`) and reverses the quality-oriented tags in `TAGS_TO_REVERSE` (`OQ`, `U2`). fgumi's `make_read_unmapped[_raw]` previously left these forward-oriented, so a reverse-strand read carrying e.g. `OQ` that clipping unmapped diverged from fgbio — squarely in the CLIP-01 path.

- `make_read_unmapped` (typed) and `make_read_unmapped_raw` now reorient these tags for reverse-strand reads, driven by shared `TAGS_TO_REVERSE_COMPLEMENT` / `TAGS_TO_REVERSE` constants so the two paths stay in lockstep.
- Only the `Z`-string encoding is handled (the sole form these tags take in practice); htsjdk additionally reverses array-encoded forms, but none of these tags is emitted as an array by real tooling. Documented inline.
- Adds `E2` / `U2` / `SQ` to the `SamTag` catalog.

**3. Full mate-info parity, including the one-mapped mate-reverse flag.**

Because fgbio `ClipBam` then calls `SamPairUtil.setMateInfo(r1, r2, true)` on every pair, the clipper unmap surfaced a paired-read gap: fgumi's `clip` only refreshed the MC/MQ tags. This PR replaces that with `set_mate_info_raw`, a faithful port of htsjdk `setMateInfo` (all three branches — both mapped, both unmapped, one of each) covering mate ref/pos/strand, the mate-unmapped flag, MQ/MC tags, and TLEN. This makes paired output match fgbio when a mate unmaps, and also fixes stale mate-pos/TLEN in the ordinary both-mapped case (previously never updated after a clip shifted a start).

- In the one-mapped branch, the mapped read's mate-reverse flag now reflects the unmapped read's **actual** strand (`SamPairUtil.java:267`) rather than an assumed `false`. The clipper's own unmap clears `REVERSE`, but a read that arrived unmapped-on-input may still carry it, and `set_mate_info_raw` runs on every pair, so such a read reaches this branch.

## Parity evidence (fgbio 4.1.0 `ClipBam` vs `fgumi clip`, `compare bams --mode content`)

Reproducer: `reports/fgbio-parity-fixtures/clip-01-unmap-fully-clipped.sh`.

**Scenario A — mapped fragment, `50M`@chr1:100, `--read-one-five-prime 60`:**

| | before | after |
|---|---|---|
| fgumi | `FLAG=0 POS=150 CIGAR=50H` (mapped, corrupt) | `FLAG=4 POS=0 CIGAR=*` (unmapped) |
| vs fgbio | **DIFFER** | **IDENTICAL** |

**Scenario B — FR pair, both clipped past alignment (`--read-one-five-prime 101 --read-two-five-prime 101`, ClipBamTest.scala:621):**

| | before | after |
|---|---|---|
| fgumi | mate-flags stale (`MATE_REVERSE` set, `RNEXT=chr1`) | `FLAG=77/141`, `RNEXT=* PNEXT=0 TLEN=0` |
| vs fgbio | **DIFFER** | **IDENTICAL** |

Also verified **IDENTICAL** to fgbio for normal paired clipping that does *not* unmap (overlap clip, fixed 3′ clip, clip-past-mate), confirming the `setMateInfo` change is safe on the common path.

## Tests / CI

- Clipper unmap cross-checks (`crosscheck_unmap_*`): start/end, boundary (`clip == clippable`), no-unmap regression guard, and negative-strand revcomp — on both `SamRecordClipper` and `RawRecordClipper`.
- New tag-reorientation cross-checks: negative-strand reorients `OQ`/`U2` (reversed) and `E2` (reverse-complemented) identically on the typed and raw paths; positive-strand leaves them untouched.
- `test_set_mate_info_raw_all_branches` (all three htsjdk branches) plus a new one-mapped case asserting the mate-reverse flag follows the unmapped read's actual strand.
- `cargo ci-fmt`, `cargo ci-lint`, `cargo ci-test` (2217 tests) all green.

Refs: CLIP-01.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -794,6 +794,7 @@ dependencies = [
  "fgumi-tag",
  "log",
  "noodles",
+ "rstest",
  "tempfile",
 ]
 
--- a/crates/fgumi-sam/Cargo.toml
+++ b/crates/fgumi-sam/Cargo.toml
@@ -20,6 +20,7 @@ tempfile = "3.3.0"
 
 [dev-dependencies]
 fgumi-raw-bam = { workspace = true, features = ["test-utils"] }
+rstest = "0.26"
 
 [lints.clippy]
 pedantic = { level = "deny", priority = -1 }
--- a/crates/fgumi-sam/src/clipper.rs
+++ b/crates/fgumi-sam/src/clipper.rs
@@ -45,6 +45,45 @@ macro_rules! slice_array {
     };
 }
 
+/// Base-oriented aux tags that are reverse-complemented (with `SEQ`) when a read is
+/// reverse-complemented, mirroring htsjdk `SAMRecord.TAGS_TO_REVERSE_COMPLEMENT`.
+const TAGS_TO_REVERSE_COMPLEMENT: [fgumi_raw_bam::SamTag; 2] =
+    [fgumi_raw_bam::SamTag::E2, fgumi_raw_bam::SamTag::SQ];
+
+/// Quality-oriented aux tags that are reversed (with `QUAL`) when a read is
+/// reverse-complemented, mirroring htsjdk `SAMRecord.TAGS_TO_REVERSE`.
+const TAGS_TO_REVERSE: [fgumi_raw_bam::SamTag; 2] =
+    [fgumi_raw_bam::SamTag::OQ, fgumi_raw_bam::SamTag::U2];
+
+/// Reorients the strand-sensitive aux tags of a `RecordBuf` the way htsjdk
+/// `SAMRecord.reverseComplement` does when unmapping a reverse-strand read:
+/// reverse-complement the base-oriented tags ([`TAGS_TO_REVERSE_COMPLEMENT`]) and
+/// reverse the quality-oriented tags ([`TAGS_TO_REVERSE`]).
+///
+/// Only the `Z`-string encoding is handled — the sole form these tags take in practice.
+/// htsjdk additionally reverses array-encoded (`B:...`) values, but none of these tags
+/// is emitted as an array by real tooling, and skipping them keeps this in lockstep with
+/// the raw path (`RawTagsMut::reverse_string` / `reverse_complement_string`).
+fn reorient_strand_tags(record: &mut RecordBuf) {
+    for tag in TAGS_TO_REVERSE_COMPLEMENT {
+        if let Some(Value::String(s)) = record.data_mut().get_mut(&tag.to_noodles_tag()) {
+            let mut bytes = s.to_vec();
+            bytes.reverse();
+            for b in &mut bytes {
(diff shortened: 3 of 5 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #493, opened 2026-07-08
fix(filter): match fgbio FilterConsensusReads mean-qual, tag guard, and docs (FILT-01/02/03)

nh13:
## What & why

Three parity fixes for `filter` against fgbio `FilterConsensusReads`.

### FILT-01 (S2) — mean base quality is computed prior to masking, over the full read
fgbio evaluates the read-level mean-base-quality filter as `sum(all base qualities) / read length` on the **unmasked** read (`FilterConsensusReads.scala:247`). fgumi computed the mean **after** masking, over **non-N** bases only — so a read whose low-quality bases are masked (and thus dropped from a non-N mean) keeps an inflated mean and passes a threshold fgbio rejects it against.

Fix: a new `mean_base_quality_full_length` helper is captured in `process_record_raw` **before** masking and threaded into the read-level check. The no-call count is still taken **after** masking, exactly as fgbio does (fgbio counts Ns post-masking).

### FILT-02 (S3) — hard-fail when consensus tags (cD/cE) are absent
fgbio fails hard if a read lacks the per-read consensus depth/error tags (`FilterConsensusReads.scala:242`) — `FilterConsensusReads` only operates on consensus reads. fgumi silently treated absent tags as a Pass and kept non-consensus reads.

Fix: `filter_read` now errors when `cD` or `cE` is absent. This covers both simplex and duplex (duplex calls `filter_read` first). A fgumi unit test that pinned the old pass-through behavior is updated to assert rejection.

### FILT-03 (S4) — false `--sort-order` in `--help`
The `filter --help` long description advertised a `--sort-order` flag that does not exist. Removed the claim (output is emitted in input order; pipe through `fgumi sort` for coordinate output). The `--min-mean-base-quality` arg/field docs are corrected from "after masking" to "prior to masking, over the full read length".

Reproducer: `reports/fgbio-parity-fixtures/filt-01-02-03-consensus-filter.sh`.

| scenario | before | after |
|---|---|---|
| **F1** mean-qual: read `AAAAAAAAAA` q40×5+q10×5, `--min-base-quality 30 --min-mean-base-quality 30` (pre-mask mean 25, non-N mean 40) | fgbio drops / fgumi **keeps** → DIFFER | both **drop** → MATCH |
| **F2** read with no cD/cE | fgbio exits 1 / fgumi exits 0 → DIFFER | both **exit 1** → MATCH |
| **F3** `filter --help` | mentions `--sort-order` (no such flag) | claim removed |

## Tests / CI

- New unit tests: `test_filter_read_requires_consensus_tags` (all of {missing both, missing cD, missing cE} reject; both present pass) and `test_mean_base_quality_full_length_includes_all_bases` (full-length 21.0 vs non-N 40.0).
- Consensus-read test fixtures across the filter tests gain the per-read `cD`/`cE` tags they should always carry; the pinned pass-through test is flipped to assert rejection.
- `is_duplex_consensus` is unchanged (out of scope).
- `cargo ci-fmt`, `cargo ci-lint`, `cargo ci-test` (2215 tests) all green.

Refs: FILT-01, FILT-02, FILT-03.

## Summary by CodeRabbit

* **Bug Fixes**
  * Consensus read filtering now requires both consensus tags to be present before applying read-level thresholds, returning an error when they’re missing.
  * Base-quality filtering now uses the full read length for its average, matching expected consensus behavior.
  * Read filtering checks were updated so masked bases no longer skew quality calculations.
  * Test coverage was expanded to verify missing-tag errors and the updated quality calculation behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -743,6 +743,7 @@ dependencies = [
  "log",
  "noodles",
  "rand 0.10.1",
+ "rstest",
  "thiserror 2.0.18",
  "wide 1.5.0",
 ]
--- a/crates/fgumi-consensus/Cargo.toml
+++ b/crates/fgumi-consensus/Cargo.toml
@@ -24,6 +24,7 @@ fgumi-raw-bam = { workspace = true, features = ["noodles"] }
 
 [dev-dependencies]
 fgumi-raw-bam = { workspace = true, features = ["test-utils"] }
+rstest = "0.26"
 
 [features]
 default = ["simplex", "duplex", "codec"]
--- a/crates/fgumi-consensus/src/filter.rs
+++ b/crates/fgumi-consensus/src/filter.rs
@@ -709,6 +709,12 @@ pub fn mask_bases(
     let ce_vals = bam_fields::find_array_tag(&record[aux_off..], SamTag::CE_BASES)
         .map(|r| bam_fields::array_tag_to_vec_u16(&r));
 
+    // Per-base depth/error masking only applies when BOTH per-base tags are present, matching
+    // fgbio's `pb = depths != null && errors != null` guard (FilterConsensusReads.scala:281,
+    // 287-289). When they are absent we must NOT treat depth as 0 and mask everything —
+    // only the base-quality mask applies.
+    let has_per_base = cd_vals.is_some() && ce_vals.is_some();
+
     let mut masked_count = 0;
     for i in 0..len {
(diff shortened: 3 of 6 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #494, opened 2026-07-08
fix(filter): only mask by base quality when per-base tags are absent (FILT-04)

nh13:
> **Stacked on #493** (base branch `nh/fix-filter-consensus-parity`). Review/merge #493 first; this PR's diff is against that branch.

## What & why (FILT-04)

fgbio's vanilla consensus filter computes `pb = depths != null && errors != null` and guards its per-base **depth** and **error** masks with `pb &&` (`FilterConsensusReads.scala:281,287-289`). When the per-base tags `cd`/`ce` are absent, fgbio masks bases **only** by base quality.

fgumi's `mask_bases` instead read an absent `cd` tag as per-base depth `0`, so `0 < min_reads` masked **every** base — the read became all-N and was dropped on the no-call filter, where fgbio keeps it. This surfaced while building the FILT-01 fixture (#493).

## Fix

Gate the depth and error masks on `has_per_base = cd_vals.is_some() && ce_vals.is_some()`, mirroring fgbio's `pb`. The base-quality mask still applies unconditionally. The duplex path (`mask_duplex_bases`) is intentionally untouched — fgbio *requires* per-base tags there, a separate concern.

Reproducer: `reports/fgbio-parity-fixtures/filt-04-mask-absent-perbase-tags.sh`. Input: unmapped consensus read with per-read `cD/cE` but **no** per-base `cd/ce`, all A, quals 5×q40+5×q10, `--min-base-quality 30 --min-reads 5`.

| | before | after |
|---|---|---|
| fgbio | keeps `AAAAANNNNN` (quality-masked only) | same |
| fgumi | masks all → drops (0 reads) | keeps `AAAAANNNNN` |
| `compare bams --mode content` | **DIFFER** | **IDENTICAL** |

## Tests / CI

- New `test_mask_bases_without_per_base_tags_masks_only_by_quality` (no per-base tags → only the low-quality bases masked, not all).
- A test that exercised depth masking with only the `cd` tag is updated to also carry `ce` (per-base masking requires both).
- `cargo ci-fmt`/`ci-lint`/`ci-test` (2215) all green.

Refs: FILT-04.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed base masking to avoid treating missing per-base consensus tags as zero depth/error.
  * Depth- and error-based masking now runs only when both required per-base tag sets are present, reducing overly aggressive masking.
* **Tests**
  * Strengthened coverage for scenarios where per-base tags are absent or incomplete, ensuring masking is driven only by minimum base quality rather than depth/error logic.

### DIFF
--- a/crates/fgumi-consensus/src/filter.rs
+++ b/crates/fgumi-consensus/src/filter.rs
@@ -709,6 +709,12 @@ pub fn mask_bases(
     let ce_vals = bam_fields::find_array_tag(&record[aux_off..], SamTag::CE_BASES)
         .map(|r| bam_fields::array_tag_to_vec_u16(&r));
 
+    // Per-base depth/error masking only applies when BOTH per-base tags are present, matching
+    // fgbio's `pb = depths != null && errors != null` guard (FilterConsensusReads.scala:281,
+    // 287-289). When they are absent we must NOT treat depth as 0 and mask everything —
+    // only the base-quality mask applies.
+    let has_per_base = cd_vals.is_some() && ce_vals.is_some();
+
     let mut masked_count = 0;
     for i in 0..len {
         let depth = cd_vals.as_ref().map_or(0u16, |v| v.get(i).copied().unwrap_or(0));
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #495, opened 2026-07-08
fix(consensus): require both aD and bD to treat a read as duplex (CONS-02)

nh13:
> **Stacked on #494** (which is stacked on #493). Base branch `nh/fix-filter-mask-absent-perbase-tags`. Merge #493 → #494 → this, in order.

## What & why (CONS-02)

fgbio `Umis.isFgbioDuplexConsensus` (`Umis.scala:144`) is `rec.contains(aD) && rec.contains(bD)` — a read carrying only one of the two AB/BA raw-read-count tags is a **simplex** consensus, routed to `filterVanillaConsensusRead`. fgumi's `is_duplex_consensus` used `aD` **OR** `bD`, so a single-strand-tagged read was routed to the **duplex** filter and rejected by the BA tier (worst-strand depth `0 < min-reads`), dropping a read fgbio keeps.

Surfaced while building the FILT-01 fixture (#493).

## Fix

`is_duplex_consensus`: OR → AND, matching fgbio. This also aligns the shared helper with `shared_metrics.rs`, which already computes `has_ad && has_bd` locally (the shared helper was the outlier).

## Blast radius

`is_duplex_consensus` callers: `filter.rs:768` (masking/filter routing) and `tags.rs:189` (`is_simplex_consensus = has_cd && !is_duplex_consensus`). Well-formed consensus reads carry **both** `aD` and `bD`, so their routing is unchanged — only single-strand-tagged (malformed) reads are affected. Full `cargo ci-test` (2215) passes across filter / duplex-metrics / group / review.

Reproducer: `reports/fgbio-parity-fixtures/cons-02-is-duplex-requires-both.sh`. Input: consensus read with per-read `cD/cE` and only `aD` (no `bD`), per-base `cd/ce`, all high-quality, `--min-reads 5`.

| | before | after |
|---|---|---|
| fgbio | keeps (simplex path) | same |
| fgumi | **drops** (duplex BA tier) | keeps |
| `compare bams --mode content` | **DIFFER** | **IDENTICAL** |

## Tests / CI

- New `test_is_duplex_consensus_requires_both_ad_and_bd` (only-aD / only-bD / neither → not duplex; both → duplex).
- A test asserting single-`aD` → duplex is updated to the fgbio semantics.
- `cargo ci-fmt`/`ci-lint`/`ci-test` (2215) all green.

Refs: CONS-02.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved duplex-consensus detection to require both forward and reverse read-count tags.
  * Records with only one duplex read-count tag are no longer incorrectly classified as duplex.
  * Added coverage for all tag-presence combinations to improve detection reliability.

### DIFF
--- a/crates/fgumi-consensus/src/filter.rs
+++ b/crates/fgumi-consensus/src/filter.rs
@@ -438,11 +438,14 @@ impl MethylationTags {
 
 /// Detects if a raw BAM record is a duplex consensus.
 ///
-/// Checks for presence of `aD` or `bD` tags in the aux data.
+/// Matches fgbio `Umis.isFgbioDuplexConsensus` (`Umis.scala:144`), which requires **both**
+/// the AB and BA raw-read-count tags (`aD` **and** `bD`). A read carrying only one of the two
+/// is treated as a simplex consensus (fgbio routes it to `filterVanillaConsensusRead`); routing
+/// it through the duplex filter would spuriously reject it on the BA tier (worst-strand depth 0).
 #[must_use]
 pub fn is_duplex_consensus(aux_data: &[u8]) -> bool {
     bam_fields::find_tag_type(aux_data, SamTag::AD).is_some()
-        || bam_fields::find_tag_type(aux_data, SamTag::BD).is_some()
+        && bam_fields::find_tag_type(aux_data, SamTag::BD).is_some()
 }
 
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #505, opened 2026-07-09
fix(codec): symmetric per-pair FR classification + NotPrimaryFrPair reject metric (CODEC-01, R2-CODEC-01)

nh13:
Fixes the codec caller's FR-pair handling to match fgbio 4.1.0 `CodecConsensusCaller`. Bundles round-1 **CODEC-01** (classification) and round-2 **R2-CODEC-01** (rejection metric) — they are two halves of one behavioral change in the same code window and fgbio ships them together. Tracker: `reports/2026-07-08-fgbio-behavioral-parity-tracker.md` (PR 9) + round-2 §R2.4.

## Findings

**CODEC-01 (S1) — dovetail FR pairs silently dropped.** fgbio classifies FR membership **symmetrically per template pair** (`isPrimaryFrPair`, `CodecConsensusCaller.scala:424-433`) — it evaluates the orientation on the *reverse-strand record only*, whose branch is CIGAR-derived and thus independent of which mate you pass, deliberately avoiding htsjdk's per-record asymmetry (samtools/htsjdk#1771) on dovetails whose aligned ends coincide. fgumi classified **per record** via `is_fr_pair_raw`, whose forward-strand branch is TLEN-derived and *disagrees* with its CIGAR-derived reverse branch on such dovetails: the forward read was mis-classified NOT-FR and dropped, its mate then became a singleton bucket and was dropped too → the whole dovetail template produced no consensus with zero accounting.

**R2-CODEC-01 (S2) — non-FR templates vanished from the rejection metrics.** fgbio counts every non-primary-FR template under `RejectionReason.NotPrimaryFrPair` (`not_primary_fr_pair`, seeded so codec always emits the row). fgumi dropped non-FR reads with a bare `continue` and no rejection accounting, so they disappeared from `raw_reads_rejected_for_*`.

## Changes

- **`overlap.rs`** — new `is_primary_fr_pair_raw(a, b)`: symmetric per-pair FR test that, after the mapped/mate-mapped/same-ref/opposite-strand preconditions, derives FR orientation from the reverse-strand record only (via `is_fr_pair_raw`'s TLEN-independent CIGAR branch), so it is order-independent and correct on dovetails.
- **`codec_caller.rs`** — classification moved from per-record (Phase 1) to per-pair after grouping; non-primary-FR templates (singleton buckets **or** non-FR pairs) are rejected and counted as `NotPrimaryFrPair`, mirroring fgbio's `partition` + `rejectRecords` (its `case _ => false` arm sweeps singletons in, so those are counted too). Unmapped primaries now flow to the per-pair check and are counted rather than silently dropped, matching fgbio's `isPrimaryFrPair` (which returns false for unmapped).
- **`caller.rs` / `rejection.rs` / `consensus.rs`** — add `NotPrimaryFrPair` to the internal + centralized `RejectionReason` enums and to `ConsensusMetrics`; emit it when non-zero (`OPTIONAL_REJECTIONS`), fixing the vanishing count for codec without making simplex/duplex emit a spurious `=0` row. (fgbio additionally seeds the row to 0 for codec so it is always present even at zero; that always-emit-zero seeding is a separate metric-format concern noted in-code and left to the rejection-row seeding work.)

Both the single-threaded and threaded codec paths go through the shared `consensus_reads_raw`, so both get the fix.

## Verification (§0)

- **fgbio oracle, verified live** (fgbio 4.1.0 `CallCodecConsensusReads`): an RF (non-FR) template is rejected with stats `raw_reads_rejected_for_not_primary_fr_pair  2`; a dovetail FR pair is kept and yields a consensus. This confirms the behavior being matched is real.
- **Unit tests** drive the real `consensus_reads_from_sam_records` path: `is_primary_fr_pair_raw` is symmetric on a dovetail where the per-record check disagrees, plus RF/FF/cross-chrom negatives; `test_not_emit_consensus_for_rf_pair` now asserts `NotPrimaryFrPair == 2`; `test_dovetail_fr_pair_is_not_rejected_as_non_fr` proves the dovetail is kept (not dropped as non-FR).
- `cargo ci-fmt` / `cargo ci-lint` clean; full suite **2215 passed**.

**Note on the fgumi-side E2E:** a full `fgumi codec` CLI run is documented in the fixture but not asserted here — on synthetic 2-molecule poly-A input the codec streaming grouper forms 0 MI groups and passes the reads to `--rejects` before the per-molecule caller runs (identical before/after, a pre-grouping plumbing limitation orthogonal to this fix). The fgumi-side behavior is instead pinned by the unit tests above, which exercise the exact production classification/counting code. Fixture: `reports/fgbio-parity-fixtures/r2-codec-01-not-primary-fr-pair.sh`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved paired-read filtering to correctly retain valid dovetail FR pairs.
  - Rejected templates that do not contain exactly one primary FR pair, preventing ambiguous inputs from being silently skipped.

- **Metrics**
  - Added reporting for templates rejected because they lacked a single primary FR pair.
  - Added human-readable descriptions and metric keys for this rejection reason.

### DIFF
--- a/crates/fgumi-consensus/src/caller.rs
+++ b/crates/fgumi-consensus/src/caller.rs
@@ -377,6 +377,8 @@ pub enum RejectionReason {
     IndelErrorBetweenStrands,
     /// Potential collision
     PotentialCollision,
+    /// Template did not have a single primary FR pair of reads (codec)
+    NotPrimaryFrPair,
     /// Other unspecified reason
     Other,
 }
@@ -402,6 +404,7 @@ impl RejectionReason {
             Self::OrphanConsensus => 'P',
             Self::IndelErrorBetweenStrands => 'D',
             Self::PotentialCollision => 'C',
+            Self::NotPrimaryFrPair => 'R',
             Self::Other => 'O',
         }
     }
@@ -428,6 +431,7 @@ impl RejectionReason {
             Self::PotentialCollision => {
                 "Potential collisions (reads with the same MI but different strands)"
             }
+            Self::NotPrimaryFrPair => "Template did not have a single primary FR pair of reads",
             Self::Other => "Other reason",
         }
     }
@@ -458,6 +462,7 @@ impl RejectionReason {
             Self::ZeroLengthAfterTrimming => CentralReason::ZeroBasesPostTrimming,
             Self::OrphanConsensus => CentralReason::OrphanConsensus,
             Self::PotentialCollision => CentralReason::DuplicateUmi,
+            Self::NotPrimaryFrPair => CentralReason::NotPrimaryFrPair,
         }
     }
 }
--- a/crates/fgumi-consensus/src/codec_caller.rs
+++ b/crates/fgumi-consensus/src/codec_caller.rs
@@ -589,7 +589,10 @@ impl CodecConsensusCaller {
             .find_string(SamTag::MI)
             .map(|b| String::from_utf8_lossy(b).to_string());
 
-        // Phase 1: Filter on raw bytes — keep paired, primary, mapped, FR-pair reads
+        // Phase 1: Filter on raw bytes — keep paired, primary reads. FR classification is done
+        // per *pair* in Phase 2 (see is_primary_fr_pair_raw), NOT per record: the per-record
+        // is_fr_pair_raw check is asymmetric on dovetails whose aligned ends coincide and would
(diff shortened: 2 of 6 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #506, opened 2026-07-09
fix(metrics): count empty molecule-end halves, orient duplex UMIs by R1 strand, guard malformed UMIs (DXM-01/02/03, SIM-01)

nh13:
Makes `duplex-metrics` and `simplex-metrics` UMI counting match fgbio's `CollectDuplexSeqMetrics` and stay internally consistent. Four related fixes, one commit each.

## DXM-01 — count UMIs with an empty molecule-end half (`duplex-metrics`)

fgbio splits `RX` with `split("-", -1)` (keeping a trailing empty field) and records **both** halves; an empty molecule-end half (`-CCC`, `CCC-`, from single-index / single-strand designs) becomes an empty-string (`""`) UMI (`CollectDuplexSeqMetrics.scala:407-408`, `CollectDuplexSeqMetricsTest.scala:362`). fgumi skipped the **whole** read pair when either half was empty, dropping even the present half.

Fix: remove the empty-component skip in `update_umi_metrics`. Rewrote the inverted `test_handle_empty_umi_components` (tracker PIN-01).

**E2E** (`dxm-01-empty-half-umi.sh`): `.umi_counts.txt` **DIFFER → MATCH** (fgumi was missing the `""`, `CCC`, `GGG` rows).

## DXM-02 — orient `--duplex-umi-counts` UMIs by R1 strand (`duplex-metrics`)

The duplex UMI must be oriented to the F1R2 reading of the top strand: fgbio reports `u1-u2` when R1 is on the positive strand and `u2-u1` when R1 is on the negative strand (`CollectDuplexSeqMetrics.scala:419-425`; docstring `AAAA-GGGG`, R1 negative → `GGGG-AAAA`). fgumi oriented by the MI `/A`,`/B` suffix, assuming `/A` = positive strand — not strand-reliable (fgbio's own test has `5/A` on the negative strand), so reverse-strand-first families were recorded in the wrong orientation.

Fix: add `r1_positive` to `TemplateInfo` (from the first-segment read's `REVERSE` flag) and orient by the actual R1 strand. Also unifies the `/A`/`/B` handling.

**E2E** (`dxm-02-duplex-umi-orientation.sh`, reverse-strand-first `ACC-TGG`): `.duplex_umi_counts.txt` **DIFFER → MATCH** (fgumi `ACC-TGG` → `TGG-ACC`).

## DXM-03 — reject a malformed non-2-segment duplex UMI (`duplex-metrics`)

On an RX that is not exactly two `-`-delimited segments (e.g. no dash), fgbio's `split("-", -1)` + `case Array(u1, u2)` throws a `MatchError`, and fgumi's own `group`/`dedup` bail with a clear error. duplex-metrics silently skipped such reads.

Fix: bail with a clear message (threads a `Result` through the metrics collection callback). simplex keeps no such guard (its UMIs are N-component by design).

**E2E** (`dxm-03-malformed-rx-guard.sh`, dashless `AAATTT`): before, fgbio rejects (exit 1) while fgumi silently succeeds (exit 0); after, **both reject (exit 1)**.

## SIM-01 — count UMIs with an empty molecule-end half (`simplex-metrics`)

`simplex-metrics` dropped empty components with the same `.filter(|s| !s.is_empty())`. fgumi's `simplex-metrics` is a fgumi-specific tool (fgbio has no `CollectSimplexSeqMetrics`), so there is no direct fgbio spec — but this keeps it internally consistent with the DXM-01 fix and with fgbio's single-strand UMI counting. Removed the empty filter (new unit test added).

## Testing

`cargo ci-fmt && cargo ci-lint && cargo ci-test` all green (2218 tests). New/updated unit tests, each watched fail against the unfixed code first: `test_handle_empty_umi_components` (DXM-01), `test_duplex_umi_orientation_follows_r1_strand` (DXM-02), `test_malformed_duplex_umi_is_rejected` (DXM-03), `test_count_umis_with_empty_molecule_end_half` (SIM-01).

Adversarial fixtures added under `reports/fgbio-parity-fixtures/` for the compare-hardening capstone.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved duplex UMI orientation using the actual R1 strand, producing more accurate metrics.
  * Correctly preserves empty molecule-end UMI components during consensus calculations.
  * Malformed duplex UMIs are now reported as errors instead of being silently skipped.
  * Processing errors are now surfaced reliably during metrics collection.
* **Tests**
  * Added coverage for malformed UMIs, strand-based orientation, empty UMI components, and updated interval handling.

### DIFF
--- a/src/lib/commands/duplex_metrics.rs
+++ b/src/lib/commands/duplex_metrics.rs
@@ -573,18 +573,20 @@ impl DuplexMetrics {
                 continue;
             }
 
-            // Split the RX tag to get individual UMI parts
+            // Split the RX tag to get individual UMI parts. fgbio uses
+            // `split("-", -1)`, which keeps a trailing empty field, and requires
+            // exactly two parts (`case Array(u1, u2)`); a UMI with a different number
+            // of `-`-separated parts is skipped here rather than panicking as fgbio
+            // does on a `MatchError`.
             let parts: Vec<&str> = rx.split('-').collect();
             if parts.len() != 2 {
                 // Not a valid duplex UMI, skip
                 continue;
             }
 
-            // Check that both components are non-empty
-            if parts[0].is_empty() || parts[1].is_empty() {
-                // Empty component, skip
-                continue;
-            }
+            // Do NOT skip empty molecule-end halves (e.g. `-CCC` or `CCC-`). fgbio
+            // counts them, recording the empty half as an empty-string UMI, so
+            // single-index / single-strand designs are not undercounted (DXM-01).
 
             // Add UMI parts based on strand. The String allocations here are
             // bounded by the consensus caller's `&[String]` signature; only the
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #507, opened 2026-07-09
fix(consensus): panic in aOrNotB/ln_a_minus_b when a < b (R2-NUM-01)

nh13:
fgbio's `aOrNotB` (`NumericTypes.scala:188-193`, pinned by `NumericTypesTest.scala:83`) throws `IllegalArgumentException("Subtraction will be less than zero.")` when `a < b`. fgumi's `ln_a_minus_b` had no such guard and silently returned `-inf`. This adds the guard as a `panic!` (fgbio throws unconditionally, not debug-only).

The near-equal case (within float noise) is still treated as equal and returns `-inf`, so the panic only fires on a genuine `a < b`.

**Reachability:** unreachable from production — the sole caller (`ln_error_prob_two_trials`) always orders its arguments `a >= b`. This is a cheap defensive guard, matching fgbio's own unit-tested behavior.

**Test:** `test_ln_a_minus_b_panics_when_a_less_than_b` ports fgbio's `aOrNotB(q20, q10)` throw case. `cargo ci-fmt`/`ci-lint` clean; full suite green.

Tracker: round-2 §R2.4 PR 10 (R2-NUM-01). This is the first, independent slice of PR 10; the other consensus glue items are tracked separately (see notes below).

---

Scope note — the rest of PR 10 was re-triaged against the live fgbio 4.1.0 oracle:
- **R2-CCI-02 (MI-less throw) → refuted/WONTFIX**: fgbio 4.1.0 `CallMolecularConsensusReads` does **not** throw on a record missing the MI tag — it groups it under a null key and orphan-filters it (verified live). fgumi's silent skip is an acceptable non-throwing divergence.
- **R2-CCI-01 (pre-group filter)**: faithful application drops unmapped-unpaired reads pre-consensus, which fgumi supports via `group --allow-unmapped`. Deferred to a dedicated PR that adds `--allow-unmapped` to simplex/codec/duplex and gates the filter on it.
- **R2-UCC-01/02 + R2-MET-05 seeding**: caller-internal + rejection-metric work, tracked for a follow-up (with oracle verification of the "throws on >1 CB" claim first).

## Summary by CodeRabbit

* **Bug Fixes**
  * Made log-space subtraction more defensive about input ordering and floating-point edge cases.
  * Treats values within floating-point epsilon as equal (returns negative infinity).
  * Triggers an explicit failure when the result would be negative beyond the epsilon tolerance.
* **Tests**
  * Expanded deterministic coverage for the exact epsilon boundary behavior.
  * Added randomized test sweeps to validate correctness across Phred ranges and ensure the guard only triggers as intended.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -742,6 +742,7 @@ dependencies = [
  "fgumi-sam",
  "log",
  "noodles",
+ "proptest",
  "rand 0.10.1",
  "rstest",
  "thiserror 2.0.18",
--- a/crates/fgumi-consensus/Cargo.toml
+++ b/crates/fgumi-consensus/Cargo.toml
@@ -25,6 +25,7 @@ fgumi-raw-bam = { workspace = true, features = ["noodles"] }
 [dev-dependencies]
 fgumi-raw-bam = { workspace = true, features = ["test-utils"] }
 rstest = "0.26"
+proptest = "1.10"
 
 [features]
 default = ["simplex", "duplex", "codec"]
--- a/crates/fgumi-consensus/src/phred.rs
+++ b/crates/fgumi-consensus/src/phred.rs
@@ -180,9 +180,24 @@ fn ln_one_minus_exp(x: f64) -> f64 {
     }
 }
 
-/// Computes log(exp(a) - exp(b)) where a > b.
+/// Computes log(exp(a) - exp(b)) where a >= b.
 ///
-/// Matches fgbio's `aOrNotB` function.
+/// Matches fgbio's `aOrNotB` function (`NumericTypes.scala:188-193`), including its guard:
+/// subtracting a larger probability from a smaller one is a caller error, so `a < b`
+/// panics rather than silently returning `-inf` (R2-NUM-01). The sole production caller
+/// (`ln_error_prob_two_trials`) always orders its arguments `a >= b`, so the guard is
+/// unreachable there — it is a defensive check.
+///
+/// # Divergence from fgbio
+/// fgbio's `aOrNotB` has no equality tolerance. fgumi treats `a` and `b` that differ by less
+/// than `f64::EPSILON` as equal and returns `-inf` (`LnZero`) so the defensive R2-NUM-01 panic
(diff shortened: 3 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed in the latest push. Extended the two `ln_a_minus_b` boundary case-tables:

- `test_ln_a_minus_b_panics_when_a_below_b` now includes an `a < b` gap of **exactly** `f64::EPSILON` at three representative log-probability bases (`0`, `ln(0.5)`, and a negative base).
- `test_ln_a_minus_b_near_equal_returns_neg_inf` gains the sub-`EPSILON` (`EPSILON/2`) absorb case at the same bases.

One correction to the suggestion: the tolerance is a **strict** `(a - b).abs() < f64::EPSILON`, so a gap of *exactly* `EPSILON` is **not** absorbed — it reaches the panic (this is fgbio-parity, since fgbio's tolerance-free `aOrNotB` also throws there). Pinning the `== EPSILON` case to `LnZero` would have been incorrect; it is pinned to the panic. Only gaps strictly inside `EPSILON` become `LnZero`.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #508, opened 2026-07-09
fix(consensus): sort-order + error-rate guards and duplex rejection rows (CONS-01, DUP-01, R2-MET-05, R2-UCC-01)

nh13:
Burn-down PR #10 of the fgbio↔fgumi behavioral-parity tracker. Four findings, all verified end-to-end against fgbio 4.1.0. Stacked on #505 (`nh/fix-codec-fr-pair-rejection`) because the rejection-machinery findings touch the same files (`caller.rs`, `codec_caller.rs`, `consensus.rs`, `rejection.rs`); retarget to `main` once #505 merges.

## CONS-01 (S3) — consensus sort-order guard

fgbio's `UmiConsensusCaller.checkSortOrder` errors unless the input is template-coordinate sorted; `simplex`/`duplex`/`codec` had no such guard (only `group`/`dedup` did), so coordinate-sorted standalone input silently interleaved molecules and split every one of them — corrupting the entire output. Added `check_consensus_sort_order` (mirroring fgbio's three cases exactly: template-coordinate → accept, `SO:unsorted`+`GO:query` → warn, else → error), wired into the already-read header of both the threaded and single-threaded paths of all three callers (stdin-safe; no re-open).

## DUP-01 (S3) — duplex error-rate validation

fgbio's `CallDuplexConsensusReads` requires `errorRate{Pre,Post}Umi > 0` (Q0 ⇒ P(err)=1). `duplex` accepted a raw `u8` with no check. Added `Duplex::validate()` rejecting a zero pre/post-UMI error rate. (codec already guarded this; simplex correctly allows `>= 0`.)

## R2-UCC-01 (S4) — reject fragment reads

fgbio's `DuplexConsensusCaller` partitions unpaired/fragment reads out and rejects them as `NonPairedReads`; fgumi folded them into the strand groups where they were mis-accounted. Added a `NonPairedReads` rejection reason and partition/reject them in `consensus_reads`.

## R2-MET-05 (S4) — seed the duplex KV rejection rows

[...]

PR #511, opened 2026-07-09
fix(consensus): fgbio rejection-reason parity + thread-independent duplex metrics

nh13:
## Summary

Two fgbio behavioral-parity fixes in the consensus callers and their rejection metrics (round-2 findings R2-UCC-01 and R2-MET-05 row-seeding), plus a correction to how caller-side rejection reasons map to the centralized rejection metric.

**Stacked on #504** (`nh/fix-metrics-fgbio-parity`) — it builds on that PR's rejection-metric work. Merge #504 first; this auto-retargets to `main`.

## What changed

**Rejection-reason mapping parity.** fgbio rejects unpaired/fragment reads as `non_paired_reads` and potential MI collisions between independent duplex molecules as `potential_umi_collision` (`UmiConsensusCaller.scala:81,83`). fgumi's caller-side `RejectionReason::FragmentRead` / `PotentialCollision` were mapped to the wrong centralized reasons (`single_strand_only` / `duplicate_umi`), so codec fragment rejections and duplex collision rejections were emitted under the wrong `raw_reads_rejected_for_*` keys. Added `NonPairedReads` / `PotentialUmiCollision` to the metrics `RejectionReason` (fgbio-exact code + description), wired them through `ConsensusMetrics`, and corrected `to_centralized()`.

**R2-UCC-01 — duplex fragment accounting.** fgbio partitions out unpaired/fragment reads and rejects them as `non_paired_reads` (`DuplexConsensusCaller.scala:204-205`). fgumi dropped them from the R1/R2 strand groups but folded them into the input total; they are now counted as `FragmentRead` (→ `non_paired_reads`).

**R2-MET-05 — always-emit duplex rejection rows.** fgbio seeds each caller's `usedBy*` rejection reasons to zero so their rows always appear (`initializeRejectCounts`). Added `ConsensusMetrics::to_kv_metrics_seeded` + `DUPLEX_SEEDED_REJECTIONS`; the duplex command now seeds `non_paired_reads`, `single_strand_only`, and `potential_umi_collision` so those KV rows are always present.

## Deliberately not ported: R2-UCC-02 (distinct cell-barcode `require`)

fgbio's callers `require(barcodes.length <= 1)` and throw when a source molecule carries more than one distinct cell barcode. fgumi does **not** replicate this, on purpose: fgumi's grouper keys on `(MI, cell)`, so a molecule spanning multiple cell barcodes is split into per-cell groups and consensus-called **per cell** (lossless), where fgbio fails fast. Verified end-to-end — a same-MI/two-CB molecule yields two per-cell consensus reads (`commands::simplex::tests::test_end_to_end_single_end_workflow`). Because the caller therefore never sees more than one cell barcode through the CLI, a caller-level distinct-CB guard would be unreachable dead code, so it is not added. This is a deliberate, lossless architectural divergence (GRP-02 class).

## Notes

- Both fgbio "throws" originally in scope were confirmed against the fgbio 4.1.0 source; the distinct-CB `require` is real but moot for fgumi's per-cell architecture (above), while the fragment rejection is real and ported.
- There are two `RejectionReason` enums — a caller-side one (`fgumi-consensus/caller.rs`) mapped via `to_centralized()` to the metrics one (`fgumi-metrics/rejection.rs`); both are updated here.

## Thread-independent duplex metrics

While validating the seeding, the duplex statistics output turned out to depend on `--threads`: the single-threaded path wrote the wide `ConsensusMetrics` table while the threaded path wrote fgbio's KV format, and the threaded path reported `consensus_reads_emitted` from the caller stat (one per molecule) rather than the number of reads actually emitted (so a 6-read run reported 3). Both are fixed — single-threaded duplex now writes the same seeded KV format, and the threaded `consensus_reads` is taken from the pipeline's written-record count. Added single-vs-multi-threaded stats-equality assertions to the simplex, duplex, and codec threading-parity tests to lock the invariant. simplex (KV) and codec (wide) were already self-consistent.

_Separate follow-up (not in this PR): codec writes the wide, non-KV format in both paths — self-consistent but not fgbio's KV shape._

## Testing

- `to_kv_metrics_seeded`: seeded reasons always emitted exactly once even at zero (seeded to `0`), omitted from the un-seeded output at zero, and not duplicated when non-zero.
- Single-vs-multi-threaded stats equality for simplex, duplex, and codec.
- Updated the `to_centralized()` mapping-pinning tests to the corrected reasons.
- Full suite: 2215 tests pass; `ci-fmt` and `ci-lint` clean.

### DIFF
--- a/crates/fgumi-consensus/src/caller.rs
+++ b/crates/fgumi-consensus/src/caller.rs
@@ -448,16 +448,18 @@ impl RejectionReason {
             Self::QualityTooLow | Self::QualityTrimmed => CentralReason::LowBaseQuality,
             Self::FailedQC => CentralReason::NotPassingFilter,
             Self::MissingUmi => CentralReason::MissingUmi,
-            // For reasons that don't have a direct centralized equivalent,
-            // we use the closest semantic match
-            Self::FragmentRead => CentralReason::SameStrandOnly,
+            // fgbio maps unpaired/fragment reads to `non_paired_reads` and MI
+            // collisions to `potential_umi_collision`
+            // (UmiConsensusCaller.scala:81,83), not to `single_strand_only` /
+            // `duplicate_umi` — match fgbio's rejection accounting.
+            Self::FragmentRead => CentralReason::NonPairedReads,
             Self::Unmapped
             | Self::Mapped
             | Self::SecondaryOrSupplementary
             | Self::IndelErrorBetweenStrands => CentralReason::NoValidAlignment,
             Self::ZeroLengthAfterTrimming => CentralReason::ZeroBasesPostTrimming,
             Self::OrphanConsensus => CentralReason::OrphanConsensus,
-            Self::PotentialCollision => CentralReason::DuplicateUmi,
+            Self::PotentialCollision => CentralReason::PotentialUmiCollision,
         }
     }
 }
@@ -1067,7 +1069,7 @@ mod tests {
 
         // Test every variant maps to a centralized reason
         let mappings = [
-            (RejectionReason::FragmentRead, CentralReason::SameStrandOnly),
+            (RejectionReason::FragmentRead, CentralReason::NonPairedReads),
             (RejectionReason::InsufficientReads, CentralReason::InsufficientSupport),
             (RejectionReason::QualityTooLow, CentralReason::LowBaseQuality),
             (RejectionReason::Unmapped, CentralReason::NoValidAlignment),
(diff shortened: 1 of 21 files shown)


nh13:
Superseded by #508.

The rejection-reason parity here (seed duplex KV rows; map fragment/collision rejections to non_paired_reads/potential_umi_collision) is implemented in #508 via the `to_kv_metrics(ConsensusCallerKind)` API. This PR's unique contribution — making duplex's single-threaded stats match the multi-threaded seeded KV format (thread-independent metrics) — has been folded into #508: duplex `execute()` now emits `to_kv_metrics(ConsensusCallerKind::Duplex)`, so both paths (and both simplex paths) write the same seeded key-value format. Closing as absorbed.

PR #501, opened 2026-07-09
fix(review): emit review rows in fgbio coordinate + MI/read-number order (#497)

nh13:
## What & why

`fgumi review` is a Rust port of fgbio `ReviewConsensusVariants`, but it emitted the detail `.txt` rows in a different order than fgbio (issue #497). Two divergences:

1. **Variant order.** fgbio drives a coordinate-ordered `SamLocusIterator` over the variant intervals, zips it with `variants.iterator`, and `require`s the two stay in sync (`ReviewConsensusVariants.scala:241`) — so its rows come out in coordinate (dictionary) order. fgumi used input-file order (`review.rs`), diverging for non-coordinate-sorted input.
2. **Per-read order within a variant.** fgbio sorts each variant's rows by the string `toMi(rec) + readNumberSuffix(rec)` (`ReviewConsensusVariants.scala:258`), e.g. `"10/1"` sorts before `"2/1"`. fgumi emitted them in BAM-query order with no sort.

Anyone diffing fgumi's review output against fgbio's, or consuming it with fgbio's ordering, saw spurious differences. As a port, fgumi should reproduce fgbio's ordering.

## Fix

- Add `Review::order_variants_by_dictionary`, which orders the loaded variants by the consensus BAM's reference sequence dictionary (index, then position) and errors on a contig absent from the dictionary — matching fgbio's `require`. `run()` applies it once, before both extraction and review, so the extracted BAMs and the `.txt` are both coordinate-ordered.
- In `generate_review_file`, buffer each variant's rows and sort them by the `MI-base + read-number` string before appending. A stable sort keeps BAM-query order for ties, matching fgbio's stable `sortBy`.

## Stacking

Stacked on #500 (`nh/fix-review-consensus-parity`, REV-01…05) because both touch `generate_review_file`. #500 is in turn stacked on #488. **Merge order: #488 → #500 → this.** Base retargets automatically as each parent merges.

The changes are orthogonal: #500 fixes row *content* (genotype string, `ref` column, `to_mi` guard, spanning-deletion exclusion); this fixes row *order*.

## Tests

- `test_order_variants_by_dictionary` — scrambled variants sort to dictionary order; an unknown contig errors.
[...]

PR #522, opened 2026-07-09
fix(review): match fgbio ReviewConsensusVariants parity (REV3-01..15)

nh13:
Stacked on #501 (base branch `497/nh/fix-review-output-row-order`).

Drives the residual `review`-command findings from the fgbio final audit to parity with **fgbio 4.1.0** `ReviewConsensusVariants`. Each finding was re-verified against fgbio source and reproduced with a real `fgbio ReviewConsensusVariants` oracle both before and after the fix. After the fixes, the VCF- and interval-list-path `.txt` outputs and the `.consensus.bam` read multiset are byte/name-identical to fgbio on the repro fixture.

## Behavioral fixes

- **REV3-01 — SNP selection checks ALT too.** fgbio `.filter(_.isSNP)` requires REF length 1 **and** at least one ALT with every ALT a single base. fgumi previously checked only REF length, so it included insertions (`A→ATG`), mixed sites, and monomorphic records. Now dropped, matching fgbio.
- **REV3-02 — consensus site base counts dedup overlapping mates.** fgbio's `BaseCounts` counts distinct read names per base; fgumi added one count per queried record, double-counting a short-insert FR pair whose mates both overlap the locus. Now deduplicated by distinct read name.
- **REV3-03 — reference base from the VCF REF.** On the VCF path fgbio uses `v.getReference.getBases()(0)`; fgumi read the FASTA base even on the VCF path. A VCF whose REF disagrees with the FASTA now classifies reads against the VCF REF.
- **REV3-04 — interval-list `filters` default is `PASS`.** fgbio renders `filters.getOrElse("PASS")`; fgumi rendered `NA` for interval-list rows.
- **REV3-05 — `.consensus.bam` writes each read once.** A consensus read non-reference at ≥2 variant loci was written once per locus; fgbio's single `query(variants)` emits it once. Now deduplicated by read identity.
- **REV3-13 — read-number suffix for unpaired reads (new, surfaced by the oracle).** `readNumberSuffix` is `/2` only for paired second-of-pair reads; fgumi keyed off `!is_first_segment()`, giving unpaired reads `/2`. Now `/1`.
- **REV3-14 — MAF filtering was entirely non-functional (new; subsumes REV3-08).** `calculate_maf` parsed sample values via `format!("{:?}")` string-munging that never matches noodles' typed values, so **both** AF- and AD-based gating silently kept every variant. Rewritten to read the typed sample value; the AD path yields `NaN` for AD summing to zero, so such variants are excluded — matching fgbio (`1 - 0/0`).

## Documented divergences (help text)

REV3-06 (interval lists not merged), REV3-07 (CIGAR `N`-skip, RNA), REV3-09 (QC-fail/secondary reads), REV3-11 (multi-valued `AF`), REV3-12 (CLI/output surface), REV3-15 (empty-result output has no header, systemic to fgumi's metrics writer).

## Dropped

REV3-10 (within-locus row order) — already handled on this base by #501 plus the existing MI+read-number sort.

## Tests

Adds real-fgbio-verified tests for each behavioral fix (SNP-drop, mate dedup, VCF REF, interval `PASS`, multi-locus single-write, unpaired `/1`, AF/AD MAF gating) and converts the read-number-suffix test to an rstest case table. `cargo ci-fmt && cargo ci-lint && cargo ci-test` all green (2232 tests).

## Reviewers

Local `/coderabbitai-review` run clean after one doc-drift fix (documented the `NaN` MAF return). `coderabbit --agent` CLI was rate-limited at review time; relying on the PR-level CodeRabbit bot for the second pass.

## Summary by CodeRabbit

* **New Features**
  * Improved variant review compatibility with fgbio behavior, including a documented “Known divergences” section.

* **Bug Fixes**
  * More accurate SNP handling (single-base REF with appropriate ALT).
  * Consensus/TSV output now uses the VCF reference allele and correct per-read/baseline deduplication.
  * Corrected paired/unpaired read naming and ensured coordinate-sorted consensus output.
  * Improved MAF filtering from typed VCF sample AF/AD values.
  * Grouped-read handling no longer fails when required metadata is missing; missing variant filters now default to `PASS`.

* **Tests**
  * Added REV3 regression and parsing semantics coverage.

### DIFF
--- a/src/lib/commands/review.rs
+++ b/src/lib/commands/review.rs
@@ -113,6 +113,36 @@ genotypes will be used.
 The `--maf` parameter controls which variants get detailed per-read information in the output file.
 Only variants with a minor allele frequency at or below this threshold will have detailed information
 written.
+
+## Known divergences from fgbio `ReviewConsensusVariants`
+
+These are intentional, documented behavior differences; the parity-relevant ones
+(SNP selection, base-count dedup, VCF REF, filters default, read-number suffix,
+MAF gating) are matched exactly.
+
+- **Overlapping interval lists are not merged (REV3-06):** fgbio applies
+  `IntervalList.uniqued(false)` (sorts and merges overlapping/abutting intervals);
+  fgumi emits one variant per position per interval, so overlapping intervals in
+  the input produce duplicate rows. Provide a pre-merged interval list to match.
+- **CIGAR `N` (ref-skip) sites (REV3-07):** an RNA read that skips over a variant
+  site with an `N` operator is extracted by fgbio (its `readPosAtRefPos==0` fires
+  for both `D` and `N` gaps) but not by fgumi. RNA-only; irrelevant to DNA
+  consensus review.
+- **Vendor-QC-fail reads (REV3-09):** fgbio's `SamLocusIterator` pileup runs with an
+  emptied filter list (`setSamFilters(Collections.emptyList())`) and a mapping-quality
+  cutoff of 0, so its only hardcoded exclusions are unmapped and vendor-QC-fail reads —
+  secondary/supplementary reads are *not* skipped. fgumi's region queries additionally
+  include vendor-QC-fail reads in counts and rows. Consensus reads essentially never set
+  the QC-fail flag, so this is a no-op in practice.
+- **Multi-valued `AF` (REV3-11):** fgbio parses a genotype `AF` with `toDouble`
+  (which throws on a comma-separated list); fgumi takes the first value. Divergent
+  only for multiallelic `AF` FORMAT fields.
+- **CLI / output surface (REV3-12):** the N-ignoring flag is `--ignore-ns` (`-N`)
+  rather than fgbio's `--ignore-ns-in-consensus-reads`; fgumi always writes
+  `.bam.bai` sidecars for the output BAMs.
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Re: the outside-diff note on \`extract_consensus_reads\` buffering all matching consensus records before the coordinate-sorted write — this is an intentional, accepted trade-off, not something we plan to change here.

The buffering is required for correctness: per-locus indexed queries return a read that is non-reference at multiple variant loci once per locus, and query order is not globally coordinate-sorted, so we dedup by read identity and sort by (ref id, pos) before writing to produce a valid indexed \`.consensus.bam\`. This mirrors fgbio, whose single \`consensusIn.query(variants)\` likewise materializes the matching reads in coordinate order. Memory is bounded by the reads overlapping the reviewed variant set (a targeted list for this tool), not by the full genome. If a genome-wide variant list ever becomes a real use case, a merged multi-region query or external sort would be the follow-up — tracked separately rather than in this parity PR. The rationale is documented inline at the buffer site and the sort site.

PR #523, opened 2026-07-09
fix(filter): count masked bases only for retained primary reads (FILT-05)

nh13:
## What & why (FILT-05)

fgbio's `FilterConsensusReads` accumulates its `maskedBases` statistic **only** inside `if (r1Result.keepRead && r2Result.keepRead)` and **only** over the primary R1/R2 reads (`FilterConsensusReads.scala:207-219`). Secondary/supplementary reads are filtered and written out but never added to the tally, and reads in dropped templates contribute nothing — the log reads "Masked X of Y bases in **retained primary** consensus reads."

fgumi's `filter` instead added every processed record's masked-base count to the `Total bases masked` tally regardless of outcome:

- the template path (`execute_threads_mode_template`) summed masked bases for **every** record, ignoring `template_pass`, per-record pass, and read type;
- the single-read path (`execute_threads_mode_single_read`) emitted a record's masked count even when it was dropped.

So a read that was masked and then dropped (e.g. by the no-call threshold) inflated the statistic where fgbio reports zero. This was noted as a caveat while confirming the FILT-01/02/04 parity work (#493/#494).

## Fix

Add `retained_primary_masked_bases(raw_records, masked_by_record, template_pass)` next to `template_passes` in the consensus crate — it sums masked bases only for the **primary** reads of a **retained** template (returns 0 for a dropped template; skips secondary/supplementary). Both filter paths now route their tally through it; the single-read streaming path treats the record as its own single-read "template".

This is a **statistic-only** change. The mask/keep/drop decisions and the emitted BAM are unchanged (those already match fgbio via FILT-01/02/04) — only the log-only `Total bases masked` number moves to fgbio parity. It is not written to `--stats` and not compared by `compare bams`.

## Parity evidence

Input: one unmapped consensus read, `AAAAAAAAAA`, quals 5×Q40 + 5×Q10, per-read `cD:i:10`/`cE:f:0.01` and per-base `cd`/`ce` present, run with `--min-base-quality 30 --min-reads 1 --max-base-error-rate 0.1`. The 5 Q10 bases mask → 5 Ns.

| scenario | `--max-no-call-fraction` | outcome | before | after |
|---|---|---|---|---|
| masked then **dropped** | `0.2` (0.5 > 0.2) | read dropped | `Total bases masked: 5` | `Total bases masked: 0` |
| masked but **kept** | `0.6` (0.5 < 0.6) | read kept (`AAAAANNNNN`) | `Total bases masked: 5` | `Total bases masked: 5` |

Dropped-read output is identical (0 records) both ways; only the statistic changes. The kept case is unchanged, confirming the fix zeros only dropped reads.

## Tests / CI

- New `#[rstest]` case table `test_retained_primary_masked_bases` (kept primary; dropped template → 0; secondary/supplementary excluded even when kept; primary pair sums; supplementary excluded from an otherwise-kept template; mixed template dropped → 0).
- `cargo ci-fmt` / `ci-lint` / `ci-test` (2241 tests) all green.

Refs: FILT-05.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected “masked bases” counting to match fgbio retained-primary semantics.
  * Masked-base totals now include only retained **primary** reads.
  * Secondary and supplementary reads no longer contribute to masked-base totals.
  * Improved accuracy in both single-read and template modes, including paired/unpaired and empty/kept/dropped templates.
* **Tests**
  * Added coverage to verify correct retained-primary masking behavior and that length mismatches fail fast.

### DIFF
--- a/crates/fgumi-consensus/src/filter.rs
+++ b/crates/fgumi-consensus/src/filter.rs
@@ -393,6 +393,53 @@ pub fn template_passes(raw_records: &[RawRecord], pass_map: &AHashMap<usize, boo
 
     has_primary && all_primary_pass
 }
+
+/// Sums the masked-base counts that fgbio's `maskedBases` statistic includes: only the
+/// **primary** reads of a **retained** template contribute; every read of a dropped
+/// template and every secondary/supplementary read contributes zero.
+///
+/// This mirrors `FilterConsensusReads.scala:207-219`, where fgbio accumulates
+/// `maskedBases += r1Result.maskedBases + r2Result.maskedBases` **only** inside
+/// `if (r1Result.keepRead && r2Result.keepRead)` and **only** over the primary R1/R2 —
+/// supplementary/secondary reads are filtered and written out, but their masked bases are
+/// never added to the tally ("Masked X of Y bases in retained primary consensus reads").
+///
+/// `masked_by_record[i]` is the number of bases masked in `raw_records[i]`; the slices are
+/// parallel and must be the same length. `template_pass` is the result of
+/// [`template_passes`] for this template (for the per-record streaming path, a single-read
+/// "template" whose pass value is the record's own filter result).
+///
+/// # Panics
+///
+/// Panics if `raw_records` and `masked_by_record` differ in length. A release-mode-silent
+/// truncation here would under/over-count the exact "Total bases masked" statistic this
+/// helper exists to compute, so the invariant is enforced unconditionally rather than only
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #537, opened 2026-07-10
fix(umi): fold case in the sequential edit assigner (audit P2)

nh13:
## What

Converges the **Edit** UMI strategy's sequential (`SimpleErrorUmiAssigner`) and parallel (`ParallelEditAssigner`) implementations on case. The parallel path uppercases before encoding; the sequential path matched raw bytes case-sensitively, so mixed-case UMIs could group differently depending on `--threads`. This folds case in `SimpleErrorUmiAssigner::assign` to match the parallel path and fgbio.

Second-wave item **P2** from the feat-runall audit (`findings/09-umi.md`, finding 1).

This is the same class of bug that `main`'s #455 ("make sequential and parallel assigners agree on UMI case") fixed — but **#455 is not in feat-runall's history**. feat-runall converged only the *adjacency* paths (via its own FU-004 raw-first-seen tie-break work) and left *edit* case-divergent. Since `main` already carries the fix and feat-runall does not, this targets **feat-runall** (stacked on `nh/audit-3-parity`). It ports only the edit-path portion of #455's approach; feat-runall's adjacency convergence is intentionally left alone (it converged differently than #455 and will be reconciled at merge time).

## fgbio parity

fgbio's `GroupReadsByUmi` uppercases every UMI before assignment (`canonicalize(rawTag.toUpperCase)`), so case-insensitive grouping is the fgbio-faithful behavior — this converges with fgbio's user-facing output rather than diverging from it. In production this is a no-op: `umi_for_read` already uppercases non-paired UMIs before the assigner sees them, so only direct library callers passing mixed case change behavior.

## Residual divergence (documented, not converged)

After this change the only remaining edit seq/parallel difference is for UMIs that are not `BitEnc`-encodable for a reason **other than** `N` (length > 32 bases, or a non-`ACGTN` character): the parallel path gives each its own molecule id, the sequential path groups them by string edit distance. `N`-containing UMIs never reach either assigner — those templates are discarded up front (`discarded_ns_in_umi`), matching fgbio — so this residual is unreachable via the CLI. The two previously-inaccurate "identical to"/"consistent with sequential" comments on `ParallelEditAssigner` are corrected to state this precisely.

The deeper *Paired*-strategy seq/parallel divergence (`findings/09` finding 2) is a separate, out-of-scope item and is not touched here.

## Tests

- `test_edit_sequential_parallel_agree_on_mixed_case` — sequential vs parallel produce the same grouping on mixed-case input (fails before, passes after).
- `test_simple_error_assigner_is_case_insensitive` — direct case-insensitivity check on the sequential assigner in the `fgumi-umi` crate.

Full suite green: `cargo ci-test` 2333 passed / 8 skipped; `cargo ci-lint` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * UMI assignment now treats mixed-case spellings consistently, grouping equivalent UMIs into the same molecule.
  * Sequential and parallel edit-based assignment now produce consistent molecule groupings for mixed-case UMIs.
  * Invalid or unsupported UMI sequences receive consistent standalone molecule assignments in parallel processing.
* **Tests**
  * Added regression and randomized coverage for case-insensitive UMI grouping and sequential/parallel assignment parity.

### DIFF
--- a/crates/fgumi-umi/src/assigner.rs
+++ b/crates/fgumi-umi/src/assigner.rs
@@ -909,6 +909,12 @@ impl UmiAssigner for IdentityUmiAssigner {
 /// are merged. This can lead to transitive clustering where UMIs are grouped together even
 /// if they differ by more than the threshold.
 ///
+/// UMI matching is case-insensitive: bases are folded to uppercase before comparison, so this
+/// agrees with the parallel edit assigner (`ParallelEditAssigner`, which uppercases before
+/// encoding) on mixed-case input. This matches fgbio's edit-distance assigner, which compares
+/// UMIs case-insensitively inside its mismatch counter (`Sequences.countMismatches`) rather than
+/// uppercasing before assignment — the same grouping result, folded at a different point.
+///
 /// # Algorithm
 ///
 /// 1. For each UMI, find all existing groups that contain at least one UMI within threshold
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #540, opened 2026-07-10
fix(sort): make the queryname sort stable for exact-key ties (audit C2)

nh13:
## What

`fgumi sort --order queryname[::natural]` sorted each in-memory batch and spill chunk with `sort_unstable`, so records whose queryname keys are exactly equal (same name + same flag class — e.g. several secondary alignments of one read) could emerge in an arbitrary order. `samtools sort -n` (stable `ks_mergesort`), fgbio, and fgumi's own arena/runall path all preserve input order for such ties, so the standalone `sort` disagreed with all three and produced non-deterministic-vs-samtools byte output.

Second-wave item **C2** from the feat-runall audit (`findings/05-sort.md`, finding 3).

## Fix

Switch the four in-chunk sorts in `sort_queryname_keyed` from `*sort_unstable_by` to their stable equivalents (`par_sort_by` / `sort_by`). `entries` is in ingest order before the sort, so a stable sort keeps equal-key runs in input order; the loser-tree k-way merge already breaks cross-chunk ties by chunk (ingest) order (`keys[a].cmp(keys[b]).then_with(|| a.cmp(&b))`), so per-chunk stability makes the whole queryname sort stable. Coordinate and template paths are unaffected — they use the stable LSD radix, not this comparison sort.

`crates/fgumi-sort/src/external.rs` and this key-only `sort_unstable_by` exist on `origin/main` (independently of the feat-runall arena path in `ref_sort.rs`, which is feat-runall-only), so this is a **main** bug and the PR targets `main`.

## Tests

`test_sort_queryname_preserves_exact_tie_input_order` — rstest over single-threaded in-memory, multi-threaded `par_chunks_mut`, and spill paths: several names each appearing many times at strictly increasing positions must come out in ingest (position) order per name. The two in-memory cases failed before this change.

`cargo ci-test` 2254 passed / 22 skipped; `fgumi-sort` suite 454 passed; `cargo ci-lint` clean.

> Note: this commit is currently unsigned — the local 1Password signing agent was wedged this session (repeated signed attempts failed). It will be re-signed before merge.

## Summary by CodeRabbit

* **Bug Fixes**
  * Queryname sorting now preserves the original input order when records have identical queryname keys.
  * Stability is maintained across both in-memory and spill-based sorting, including single- and multi-threaded processing.

* **Tests**
  * Added coverage verifying consistent ordering for records with duplicate querynames.

### DIFF
--- a/crates/fgumi-sort/src/external.rs
+++ b/crates/fgumi-sort/src/external.rs
@@ -2382,7 +2382,12 @@ impl RawExternalSorter {
 
                 timer.time_sort(|| {
                     use rayon::prelude::*;
-                    rayon_pool.install(|| entries.par_sort_unstable_by(|a, b| a.0.cmp(&b.0)));
+                    // Stable sort: `entries` is in ingest order, so a stable sort keeps
+                    // exact-queryname-key ties in input order — matching `samtools sort -n`,
+                    // fgbio, and the arena/runall path. The loser-tree merge already breaks
+                    // cross-chunk ties by chunk (ingest) order, so this per-chunk stability
+                    // makes the whole queryname sort stable.
+                    rayon_pool.install(|| entries.par_sort_by(|a, b| a.0.cmp(&b.0)));
                 });
 
                 // Write keyed temp file with parallel BGZF compression via worker pool.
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #539, opened 2026-07-10
fix(cli-common): accept EB/EiB units and make AUTO_RESERVE_CAP 32-bit-safe (audit C3/D3)

nh13:
## What

Two small `fgumi-cli-common` correctness fixes, batched because they touch the same file.

**C3** (`findings/12`, finding 2) — `parse_memory_size` rejected any input containing `e`/`E` as "Scientific notation not supported" *before* handing off to `bytesize`. But `bytesize` accepts `EB`/`EiB` (exabytes), which contain an uppercase `E`, so `parse_memory_size("1EB")` / `"1EiB"` wrongly errored. Replaced the blanket `contains('e') || contains('E')` check with `looks_like_scientific_notation`, which only fires for a genuine mantissa-exponent — a digit or `.`, then `e`/`E`, then a digit or a sign. `1EB`/`1EiB` now parse; `1e5`, `1.5e10`, `2E+3`, `3e-2` are still rejected.

**D3** (`findings/12`, finding 1) — `const AUTO_RESERVE_CAP: usize = 10 * 1024 * 1024 * 1024` evaluates to 10 GiB, which exceeds `u32::MAX`, so on any 32-bit target (`usize == u32`) it is a hard const-evaluation overflow and the crate fails to compile. This directly contradicts `detect_total_memory`, which deliberately saturates for 32-bit. Compute the cap with `10usize.saturating_mul(...)` so it saturates to `usize::MAX` on 32-bit instead of failing const-eval; there `resolve_reserve`'s `cap.min(total / 2)` yields `total / 2`, matching the 32-bit intent.

## Target rationale (main vs feat-runall)

The `fgumi-cli-common` crate does not exist on `origin/main` — it is feat-runall infrastructure. Both are feat-runall-only, so this targets the audit stack (`nh/audit-3-parity`).

## Tests

- `test_parse_memory_size_scientific_notation_guard` — rstest table: `1EB`/`1EiB`/`512MiB` accepted; `1e5`/`1E5`/`1.5e10`/`2E+3`/`3e-2` rejected (the EB/EiB cases failed before this fix).
- D3 is compile-time; `saturating_mul` is a const fn that cannot overflow on any target. No 32-bit CI target exists to exercise it, but the fix is correct by construction.

`cargo ci-test` (cli-common) 33 passed; `cargo ci-lint` clean.

[...]

PR #542, opened 2026-07-10
fix(cli-macros,pipeline-core): harden multi_options + reorder buffer (audit D1/D2/D4)

nh13:
## What

Three latent-robustness guards from the feat-runall audit, batched (all "fail loud instead of silently misbehaving"). **D3** already landed with C3 in #539.

**D1/D2** (`multi_options` macro, `findings/10`) — the Multi struct renames every field to `<prefix>_<field>`, and the attribute classifiers only inspect `key = value` / bare-`key` spellings. Two forms slip through:
- **D1** — cross-field reference attrs (`requires`, `conflicts_with`, `required_if_eq`, `required_unless_present`, …) carry an unprefixed arg-id string that dangles once the field is renamed, so clap panics ("arg id not defined") when it builds the `runall` command.
- **D2** — clap's `key(value)` call form for `long`/`short`/`default_value`/`default_value_t`/`required` arrives as `Meta::List`, unseen by the classifiers → duplicate `long`, mis-detected required-ness, or a field silently flipped to "required".

Added `unsupported_arg_attr_error`, rejected both at macro-expansion time (`assert!`, matching the existing `#[command(flatten)]` rejection), and documented the restriction in the macro rustdoc. No current `#[multi_options]` struct uses either form (`aligner.rs` already enforces its coupling in `resolve()` instead), so this is a fail-loud guard for future stages.

**D4** (`ReorderStage`, `findings/04` Q1) — both `state.buffer.insert(seq.ordinal, …)` sites silently overwrite an existing entry for the same ordinal, dropping the buffered item and leaking its bytes into `buffer_bytes`. `ByOrdinal` serials are unique by construction, but a buggy `ByItemOrdinal` upstream could collide. Added a `debug_assert!` before each insert (mirroring `HeldSlot::put`'s hard assert).

`fgumi-cli-macros` and `fgumi-pipeline-core` do not exist on `origin/main` — both are feat-runall infrastructure. Feat-runall-only, so this targets the audit stack (`nh/audit-3-parity`).

## Tests

- `unsupported_arg_attr_error`: accepts the supported forms; rejects cross-reference attrs (D1) and the `key(value)` call form (D2); leaves unclassified call-form keys (`value_parser(...)`) untouched.
- `duplicate_ordinal_trips_debug_assert`: debug-gated `#[should_panic]` that pushes a duplicate ordinal and confirms the guard fires. All 10 existing macro smoke tests still pass (valid usage unchanged).

`cargo ci-test` 2332 passed / 8 skipped; `cargo ci-lint` clean.

> Note: this commit is currently unsigned — the local 1Password signing agent was wedged this session. It will be re-signed before merge.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for multi-option argument attributes, with clear compile-time errors for unsupported configurations.
  * Prevented silent data corruption when duplicate reorder ordinals are received; conflicting entries now trigger an explicit failure.
  * Added coverage for supported and unsupported argument forms and duplicate reorder scenarios.

* **Documentation**
  * Clarified the macro’s error conditions and supported attribute usage.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -770,6 +770,7 @@ dependencies = [
  "clap",
  "proc-macro2",
  "quote",
+ "rstest",
  "syn 2.0.117",
 ]
 
--- a/crates/fgumi-cli-macros/Cargo.toml
+++ b/crates/fgumi-cli-macros/Cargo.toml
@@ -18,6 +18,7 @@ proc-macro2 = "1"
 [dev-dependencies]
 clap = { version = "4", features = ["derive"] }
 anyhow = "1.0"
+rstest = "0.26"
 
 # cargo-release: don't tag from this crate. The workspace `release.toml`
 # (if/when configured) should create a single tag from the main `fgumi`
--- a/crates/fgumi-cli-macros/src/lib.rs
+++ b/crates/fgumi-cli-macros/src/lib.rs
@@ -45,9 +45,17 @@ use syn::{Lit, Meta};
 ///
 /// # Panics
 ///
-/// Panics if applied to a non-struct item, a struct without named fields,
-/// or a struct containing a `#[command(flatten)]` field (the struct must
-/// be flat — re-exposing nested clap-flattened state is out of scope).
+/// Panics (with a clear build-time message) if applied to:
+/// * a non-struct item, or a struct without named fields;
+/// * a struct containing a `#[command(flatten)]` field (the struct must be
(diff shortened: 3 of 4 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #545, opened 2026-07-10
refactor(group): share the group sort-order precondition across both paths (audit B2)

nh13:
## What

`group` is the only command with **two** orchestrations of the same stage — the single-threaded `Group::execute` path and the chain builder's `add_group` — and each carried a **byte-for-byte** copy of the sort-order precondition: the same `is_template_coordinate_sorted` / `is_sorted(QUERY_NAME)` predicate, the same two bail messages (with/without `--allow-unmapped`), and the same info logging. A change to the accepted orders or the remediation hint had to be made in both, or the two paths would silently accept/reject different inputs for the same command — the standalone-vs-runall divergence class the per-module audit repeatedly hit.

Second-wave item **B2** (`findings/15`, finding 2).

## Fix

Extract the policy into `commands::common::require_group_input_sort(header, allow_unmapped)` (error strings + logging preserved verbatim) and call it from both paths. The predicates were already shared (`crate::sam`); this consolidates the *policy* that wraps them. Added a unit test pinning the accept/reject matrix.

## Scope

- This is finding-2's genuine within-command duplication (group's two paths). The `add_dedup` and `downsample` template-coordinate checks are each **single-site** with intentionally different remediation text (dedup: "merged.bam"; downsample: "output from group / fgbio GroupReadsByUmi"), so they aren't duplicated within a command and are left as-is (consolidating would homogenize distinct messages).
- **B1 (finding 3) is deferred**: unifying the single-threaded group path onto the chain (deleting `execute_single_threaded`) is not a mechanical refactor — it **changes user-visible behavior** (the single-threaded path hard-rejects SAM while the chain accepts it) and touches the allocation-free single-thread fast path. It needs a dedicated output+metrics parity harness and a decision on the SAM-acceptance change before landing. The V3/V4 symptoms it would root-cause are already fixed as targeted bugs (#536).

## Target rationale

`add_group`/`build_for` and `commands/common.rs` are feat-runall infrastructure (absent on `origin/main`), so this targets the audit stack (`nh/audit-3-parity`).

## Tests

`require_group_input_sort_accept_reject_matrix` (new); `cargo ci-test` 2332 passed / 8 skipped; `cargo ci-lint` clean.

> Note: this commit is currently unsigned — the local 1Password signing agent was wedged this session. It will be re-signed before merge.

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized validation of input sort order for read-grouping workflows.
  * Template-coordinate-sorted input is accepted consistently.
  * Queryname-sorted input is accepted when unmapped reads are enabled; otherwise, users receive a clear remediation message.
  * Added informative logging describing the detected input sort order.

### DIFF
--- a/src/lib/commands/common.rs
+++ b/src/lib/commands/common.rs
@@ -121,6 +121,54 @@ pub fn add_pg_to_builder(
     fgumi_bam_io::header::add_pg_to_builder(builder, crate::version::VERSION.as_str(), command_line)
 }
 
+/// Validate that `header` advertises a sort order the group stage accepts, and
+/// emit the accompanying info logging.
+///
+/// Accepts template-coordinate order always, and queryname order when
+/// `allow_unmapped` is set. On rejection, bails with the order-specific
+/// remediation hint. Shared by `Group::execute`'s single-threaded path and the
+/// chain builder's `add_group` so the two orchestrations of the same stage
+/// cannot drift on the accepted orders, the error text, or the info logging —
+/// the standalone-vs-runall divergence class. The predicates themselves are the
+/// shared `crate::sam::{is_template_coordinate_sorted, is_sorted}`.
+pub(crate) fn require_group_input_sort(
+    header: &Header,
+    allow_unmapped: bool,
+) -> anyhow::Result<()> {
+    use crate::sam::{is_sorted, is_template_coordinate_sorted};
+    use anyhow::bail;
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #547, opened 2026-07-10
refactor(group): fold the single-threaded path onto the chain (audit B1)

nh13:
## What

`group` was the **only** command with two orchestrations of the same stage: a bespoke `execute_single_threaded` streaming driver (taken when `--threads` is unset) and the typed-step chain via `build_for` (taken for `--threads N`, including `--threads 1`). Every other multi-threaded command already routes solely through `build_for` — group's inline fast path was vestigial scaffolding left behind by the issue #330 migration, kept in lockstep with the chain by hand. That two-path split was the structural root of the single-threaded-vs-chain divergences the audit found (V3/V4: metrics + help drift, #536).

Second-wave item **B1** (`findings/15`, finding 3), the one deferred from #545.

## How it's safe

`ThreadingOptions::num_threads()` maps `None → 1`, so routing the unset case through `build_for` yields a **one-worker chain that degrades to the sequential assigner** — output-identical to `--threads 1` (which already routed through the chain). Deleted the `threads.is_none()` branch and `execute_single_threaded` + its exclusive helpers (`process_and_write_position_group`, `write_all_metrics`); `execute` now always builds the `ChainSpec`. **Net −308 lines**, one fewer path to keep in sync.

## `--threads` unset now behaves identically to `--threads 1`

That's the whole point. The one place they previously diverged was **SAM input**: the chain (`--threads 1`) already accepted SAM via `ReadSamChunks`/`ParseSamChunk`, while the bespoke unset path was BAM-only and bailed `"SAM input requires --threads N"`. That divergence was a symptom of the two-orchestration problem, not a contract — unifying onto the chain removes the inconsistent restriction, so unset now accepts SAM too. Nothing that previously worked changes; an error case simply starts succeeding to match `--threads 1`. Output, MI numbering, metrics, and logging are otherwise unchanged.

## Tests (TDD)

- **`group_no_threads_matches_threads_1`** — pins that `--threads` unset produces byte-identical grouping (record order + `MI:Z` tags) to `--threads 1`. **It passed on the pre-refactor code**, proving the two paths already agreed for BAM, so the deletion is behavior-preserving (this was the parity gate before touching the source).
- **`test_group_sam_input_without_threads_now_accepted`** — unset + SAM now matches the BAM run (the chain already accepted SAM at `--threads 1`).

Stale "single-threaded fast path" comments corrected across `group.rs`, `common.rs`, the chain's `build_group_chain`, and the determinism test.

[...]

PR #538, opened 2026-07-10
fix(simplex): enforce --min-reads/--max-reads range on the runall path (audit V5/C1)

nh13:
## What

The standalone `Simplex::execute` rejects `--min-reads 0` ("a value of 0 admits empty groups") and `--max-reads < --min-reads` before dispatching, but `runall` never calls `Simplex::execute` — it builds the chain directly via `build_for`. Neither the `Stage::Simplex` arm nor `add_simplex` re-checked the range, so `fgumi runall … --consensus simplex --simplex::min-reads 0` (or a max below the min) was silently accepted, producing output the standalone command refuses and breaking runall's "same as standalone" contract.

Second-wave item **C1 / V5** from the feat-runall audit (`findings/02-unified_pipeline.md`, finding 2).

## Fix

Hoist the two checks into a shared `SimplexOptions::validate_read_bounds` and call it from both `Simplex::execute` and the chain builder's `add_simplex`. This mirrors the pattern already used for the duplex path (V1, `DuplexConsensusCaller::validate_min_reads` called in `add_duplex`) and the codec path (`.validate()`).

## Target rationale (main vs feat-runall)

`runall` and the chain builder (`src/lib/pipeline/chains/builder.rs`) do not exist on `origin/main` — they are the feat-runall typed-step surface (issue #330). This is a feat-runall-only gap, so the PR targets the feat-runall audit stack (`nh/audit-3-parity`).

## Tests

- `test_validate_read_bounds` — rstest table over valid and degenerate (`min_reads==0`, `max<min`) inputs.
- `chain_build_simplex_rejects_degenerate_read_bounds` — build-level regression: `build_for([Group, Simplex])` now errors for both degenerate configs (it built successfully before this fix).

`cargo ci-test` 2340 passed / 8 skipped; `cargo ci-lint` clean.

> Note: this commit is currently unsigned — the local 1Password signing agent was wedged at commit time (three signed attempts failed). It will be re-signed before the audit stack is finalized/merged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added consistent validation for simplex read limits.
  * Invalid configurations, including zero minimum reads or maximum reads below the minimum, are now rejected earlier with clear errors.
  * Pipeline construction now applies the same validation as the standalone simplex command.

* **Tests**
  * Added coverage for valid and invalid read-limit configurations.
  * Added integration tests confirming invalid pipeline configurations fail during build.

PR #549, opened 2026-07-10
refactor(consensus): fold single-threaded simplex/duplex/codec onto the chain

nh13:
## What

The follow-up to the group unification (#547). `simplex`, `duplex`, and `codec` were the **last** commands still branching on `threads.is_none()` for a bespoke single-threaded streaming path (opening input via the shared BAM-only `open_single_threaded_consensus_input`), while `--threads N` — including `--threads 1` — already routed through the typed-step chain via `build_for`. This completes issue #330's goal: **every command routes solely through the chain**.

## How it's safe

`ThreadingOptions::num_threads()` maps `None → 1`, so routing the unset case through `build_for` yields a one-worker chain — output-identical to `--threads 1`. Verified per command by parity tests (`{simplex,duplex,codec}_no_threads_matches_threads_1`) that **pass on the pre-refactor code**, proving the paths already agreed before I touched the source. Deleted each command's `threads.is_none()` streaming path, made the `ChainSpec` build unconditional, and deleted the now-unused `open_single_threaded_consensus_input` helper. **Net −454 lines**, three fewer orchestrations to keep in sync.

## Consistency change (same as group)

`simplex`/`duplex`/`codec` with `--threads` unset now **accept SAM input** (the chain's `ReadSamChunks`/`ParseSamChunk` handle it) instead of bailing `"SAM input requires --threads N"`. `--threads 1` already accepted SAM, so this only makes unset match it — not a breaking change, an error case starts succeeding. The `test_single_threaded_consensus_rejects_sam_with_threads_hint` test (which enforced the old BAM-only restriction for all three) is **replaced** by `test_single_threaded_consensus_accepts_sam` asserting the new behavior. Stale "single-threaded fast path not migrated here" comments corrected in the chain-command modules.

## Tests
- `{simplex,duplex,codec}_no_threads_matches_threads_1` — pin unset ≡ `--threads 1` consensus output.
- `test_simplex_sam_input_without_threads_now_accepted` + `test_single_threaded_consensus_accepts_sam` — the SAM behavior change.

`cargo ci-test` 2336 passed / 8 skipped; `cargo ci-lint` clean.

## Reconcile notes
- Touches `simplex.rs` in a different region than #538 (C1, min-reads validation) — composes cleanly.
- Replaces `test_single_threaded_consensus_rejects_sam_with_threads_hint`, which #543 (E3) rstest-ified. At reconcile the deletion/replacement wins (the test's premise — consensus rejects SAM without `--threads` — is now false).
- Independent of #547 (group); same pattern, sibling commands. Targets the audit stack (`nh/audit-3-parity`).

## Summary by CodeRabbit

* **New Features**
  * Simplex, duplex, and codec now run through a unified single-stage execution flow, yielding consistent results when `--threads` is omitted or set to `--threads 1`.
* **Bug Fixes**
  * Consensus commands no longer fail when `--threads` is unset and SAM input is provided.
  * Improved CLI validation: `--ref` now requires `--methylation-mode`.
* **Tests**
  * Added regression tests ensuring byte-for-byte identical consensus records, BAM headers, and full `--stats` output for omitted `--threads` vs `--threads 1` across codec/duplex/simplex.
  * Hardened SAM/BAM streaming input parity checks.


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed the outside-diff finding on `duplex.rs:344-375` (no thread to resolve, so noting it here).

Most of it was already gone in the current head — `_writer_threads`, `_cell_tag`, and `_track_rejects` had been removed, leaving only `_methylation_mode`. That one is now deleted: `resolve_methylation_mode` is a pure `map_or` with no side effects, and the chain re-derives the mode in `ChainBuilder::add_duplex` (`builder.rs:3338`), so the call was doing nothing.

One correction to the review: `_timer` is **not** fine. `OperationTimer` has no `Drop` impl (`crates/fgumi-cli-common/src/lib.rs:173`) — it logs on `new()` and on an explicit `log_completion(count)`, nothing else. Since `log_completion` is no longer called, `_timer` was logging the start line and silently dropping the "completed in ..." half, and `duplex` was the only one of the three still doing it (`simplex` and `codec` dropped the timer entirely). It is now removed to match, with a comment recording why; the chain's `StageTimingFinalizeHook` already logs `Pipeline ... ran in Xs` for all three.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed the outside-diff finding on `common.rs:578-587` (no thread to resolve, so noting it here).

The `--threads` help text claimed `codec`, `duplex`, `simplex`, and `clip` all keep a bespoke single-threaded path when the flag is omitted. After this PR that is only true of `clip` — it is the sole remaining `if let Some(threads) = self.threading.threads { ... } else { execute_single_threaded(..) }` in `src/lib/commands/`; `codec`/`duplex`/`simplex` now route through `chains::build_for` unconditionally, so an omitted flag resolves to the same one-worker chain as `--threads 1`. Reworded the doc comment to say exactly that and to call out `clip` as the exception.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #552, opened 2026-07-10
fix(consensus)!: clamp codec/duplex scalar depth+error tags to fgbio's Short ceiling

nh13:
## What

Clamp the scalar consensus depth tags (`aD`/`bD`/`cD` and the `aM`/`bM`/`cM` minima) **and** the error-rate tags (`aE`/`bE`/`cE` — both the depth denominators and the error numerators) to fgbio's `Short` ceiling (`i16::MAX` = 32767) in the **codec** and **duplex** callers, matching what the **vanilla** (simplex) caller already does and what fgbio emits. This makes all three consensus callers consistent with each other and bit-identical to fgbio.

## Why

fgbio stores per-base consensus depth *and* per-base error counts as `Array[Short]` capped at `Short.MaxValue` and derives every scalar depth/error-rate tag from those capped arrays. fgumi keeps per-base depth and errors in `u16`. The vanilla caller already matches fgbio by clamping both depths and errors at push (with a test), and `base_builder.rs` documents the intent to clamp "at tag emission" — but the codec and duplex callers emitted the scalar depth *and* error-rate tags **uncapped**, so they diverged from fgbio on families deeper than 32767 reads (e.g. measured `aD 59034 vs 32767`). The three callers were inconsistent with each other and with fgbio, and this had been re-litigated repeatedly (the per-base arrays were hardened before, but the scalars were left uncapped).

## How

- Add a shared `clamp_per_base_to_fgbio_short(u16) -> i32` helper.
- codec/duplex: clamp each per-base depth before `max`/`min`/sum. For the **combined** duplex/codec `cD`/`cM`, cap each strand's per-base value **before** summing, exactly as fgbio's `totalDepths(i) = min(abD_i, 32767) + min(baD_i, 32767)` — so a mixed pair where only one strand saturates yields the correct intermediate value (e.g. 62767), not an uncapped sum.
- Cap the error-rate **numerators** the same way: `aE`/`bE`/`cE` are now summed from per-base errors capped at the ceiling, matching fgbio's capped `errors` `Array[Short]`. Both the numerator and the denominator of every error rate now derive from Short-capped per-base values.
- Only the emitted tags are capped; the unclamped depth is still used for the `--min-reads` threshold, so no behavior is lost.

## Tests

- Helper case table (in-range / at-ceiling / above / `u16::MAX`).
- Duplex + codec depth-saturation tests covering the intermediate `cD` case (`aD`→32767, `bD`→30000, `cD`→62767), alongside the existing vanilla saturation test.
- Duplex + codec error-rate numerator-cap tests: a per-base error above the ceiling yields the capped rate (`aE` = 32767/(32767·4) = 0.25) rather than the uncapped value (0.3052); verified to fail before the numerator fix.

## Notes

- **Breaking** (`!`): changes the emitted depth- and error-rate-tag values for consensus families deeper than 32767 reads (now capped, matching fgbio).
- Pairs with the compare-tool change on `nh/compare-hardening` (#530), which drops the now-unnecessary depth-saturation tolerance and compares these tags exactly. With the error numerators now capped too, the tolerance can be dropped for `aE`/`bE`/`cE` as well, not just the depth tags.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated consensus depth and error calculations to respect fgbio’s Short ceiling before computing SAM scalar and per-base tags.
  - Prevented overflow/wraparound in very deep coverage scenarios, including duplex and strand-level depth/error-rate outputs.
  - Ensured duplex totals and error-rate numerators/denominators are derived from capped per-strand/per-base inputs for consistent tag behavior.

- **Tests**
  - Added regression tests for saturation edge cases (including mixed-saturation) and for overflow-safe error-rate denominator computation.
  - Added offline “fgbio oracle” saturation comparisons against pinned BAM captures.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -736,6 +736,7 @@ dependencies = [
  "anyhow",
  "approx",
  "bstr 1.12.1",
+ "fgumi-bam-io",
  "fgumi-dna",
  "fgumi-metrics",
  "fgumi-raw-bam",
@@ -745,6 +746,7 @@ dependencies = [
  "proptest",
  "rand 0.10.1",
  "rstest",
+ "tempfile",
  "thiserror 2.0.18",
  "wide 1.5.0",
 ]
--- a/crates/fgumi-consensus/Cargo.toml
+++ b/crates/fgumi-consensus/Cargo.toml
@@ -24,6 +24,10 @@ fgumi-raw-bam = { workspace = true, features = ["noodles"] }
 
 [dev-dependencies]
 fgumi-raw-bam = { workspace = true, features = ["test-utils"] }
+# Read fgbio-oracle fixture BAMs back into RawRecords in the saturation-oracle
+# tests. fgumi-bam-io does NOT depend on fgumi-consensus, so this is cycle-free.
+fgumi-bam-io = { workspace = true }
+tempfile = "3.3.0"
 rstest = "0.26"
 proptest = "1.10"
 
--- a/crates/fgumi-consensus/src/caller.rs
+++ b/crates/fgumi-consensus/src/caller.rs
@@ -342,6 +342,23 @@ pub fn calculate_error_rate(depths: &[u16], errors: &[u16]) -> f32 {
     total_errors as f32 / total_depth as f32
 }
 
+/// fgbio's `Short` per-base depth ceiling (`i16::MAX` = 32767).
+///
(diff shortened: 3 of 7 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #553, opened 2026-07-10
fix(umi): fold case in the sequential paired assigner to match the parallel path

nh13:
## What

Converges the **case** dimension of the sequential-vs-parallel divergence in the paired UMI assigner. Stacked on #537 (P2, the sequential *edit* assigner case fold); this is the paired sibling.

The paired strategy has two independent sequential-vs-parallel divergences (documented in the S7-002/FU-004 NOTE in `parallel_assigner.rs`):

1. **Case** — fixed here.
2. **Reverse-orientation adjacency matching** (e.g. `GAAA-CCCC` vs `AAAA-CCCC`, 1 edit apart) — a separate, still-open follow-up tracked by the reverse-orientation-edges fix on `main` (GRP3-01); it arrives via the eventual `feat-runall` ← `main` reconcile and is deliberately **not** ported here to avoid duplicating in-review work.

## Why

`ParallelPairedAssigner::canonicalize` uppercases, so the parallel path's node identity is case-insensitive. The sequential `PairedUmiAssigner` counted, matched, tie-broke, and strand-assigned on the RAW string. On mixed-case paired UMIs — which the CLI does **not** pre-uppercase — `group`/`dedup --threads 1` (sequential) and `--threads N` (parallel) could group and strand-assign the same molecule differently. That is a real, CLI-reachable `--threads`-dependent grouping bug.

Folding case in the sequential path makes the two agree, and also matches fgbio's user-facing output (its `GroupReadsByUmi` uppercases every UMI before assignment via `canonicalize(rawTag.toUpperCase)`). It is a no-op for the uppercase UMIs the CLI emits.

This is intentionally **different** from the single-UMI adjacency tie-break, which stays RAW/case-sensitive on *both* paths (S7-002/FU-004): there the parallel path was converged onto raw, so no case fold is needed; here the parallel path is already uppercase, so the sequential path is folded onto it instead. The adjacency DIRECTION test (`adjacency_tie_break_is_case_sensitive_raw_not_uppercased`) is untouched and still passes.

## Changes

- `PairedUmiAssigner::assign` works on uppercased UMIs; `is_same_umi`/`canonicalize` uppercase before comparing/canonicalizing.
- New `paired_sequential_and_parallel_agree_on_mixed_case` parity test (inputs 0 edits apart, so the still-open reverse-orientation divergence is not exercised) — verified to fail before the fix ("Partitions diverge between assigners").
- New `test_paired_helpers_are_case_insensitive` unit test.
- Updated the S7-002/FU-004 NOTE to document both divergences and which is now converged.

## Verification

`cargo ci-test` (2334 passed) and `cargo ci-lint` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Paired UMI assignment is now case-insensitive, returning consistent results across mixed-case and reverse-strand representations.
  * Sequential and parallel paired assignment now agree on molecule grouping and strand/canonical UMI outcomes, including tie-breaking.
* **Tests**
  * Added new assertions for case-insensitive paired UMI comparison and canonicalization.
  * Extended sequential-versus-parallel verification for mixed-case inputs, including coverage of alternate graph-path behavior and boundary handling.

### DIFF
--- a/crates/fgumi-umi/src/assigner.rs
+++ b/crates/fgumi-umi/src/assigner.rs
@@ -1884,12 +1884,27 @@ impl UmiAssigner for PairedUmiAssigner {
             assert!((umi.split('-').count() == 2), "UMI {umi} is not a paired UMI");
         }
 
+        // Work on uppercased UMIs so counting, matching, tie-break, and strand assignment are
+        // all case-insensitive. Unlike the single-UMI adjacency tie-break (deliberately RAW /
+        // case-sensitive per S7-002 / FU-004), the parallel paired assigner
+        // (`ParallelPairedAssigner::canonicalize`) already uppercases, so its node identity is
+        // case-insensitive; the sequential paired path MUST fold case too, or `group`/`dedup
+        // --threads 1` (sequential) and `--threads N` (parallel) group mixed-case paired UMIs —
+        // which the CLI does NOT pre-uppercase — differently. Folding here converges the two
+        // paired paths on the case dimension. NOTE: this is an intentional DIVERGENCE from
+        // fgbio, whose paired *grouping* comparator (`AdjacencyUmiAssigner.matches`,
+        // `GroupReadsByUmi.scala:248-256`) is case-SENSITIVE — only `countMismatches` (strand
+        // assignment / the edit strategy) folds case. fgumi opts for case-insensitive paired
+        // grouping for robustness and sequential/parallel consistency; it is a no-op for the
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #556, opened 2026-07-10
refactor(pipeline-core): extract the duplicated retry_held into one helper

nh13:
## What

`OutputHandles<Single<T>>::retry_held` and `OutputHandles<OrderedBytesSingle<T>>::retry_held` had byte-identical bodies (take → retry → put-back-on-reject), and *each* doc-comment claimed to be "the single canonical definition" of the re-hold invariant — a contradiction, and exactly the drift risk the comment warns about (a copy that forgets the put-back silently drops a final batch).

Extract the shared logic into a private `retry_held_impl(held, retry)` that takes the shape-specific `retry` as a closure; both methods delegate to it, so there is one real canonical definition.

## Notes

- No behavior change — the existing `single_retry_held_drains_then_reholds` test still passes.
- `fgumi-pipeline-core` is feat-runall-only (not on `main`), so this does not conflict at the eventual reconcile. Targets `nh/audit-3-parity`.

## Verification

`cargo ci-test` 2332 passed; `cargo ci-lint` clean.

## Summary by CodeRabbit

* **Refactor**
  * Improved internal handling of retry operations for held output slots.
  * Preserved existing retry behavior and safeguards when an operation is rejected.

### DIFF
--- a/crates/fgumi-pipeline-core/src/handles.rs
+++ b/crates/fgumi-pipeline-core/src/handles.rs
@@ -849,6 +849,33 @@ pub(crate) struct UnitOutputsView;
 // Typed accessors on OutputHandles<O>
 // ─────────────────────────────────────────────────────────────────────────────
 
+/// The single canonical definition of the held-slot re-hold invariant, shared by
+/// every single-output shape's `retry_held` (the `Single<T>` and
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

PR #560, opened 2026-07-10
fix(fastq): honest missing-quality default + singleton desync warning (W9c: FASTQ3-02/03, +FASTQ3-04 doc)

nh13:
## What / why

`fgumi fastq` writes interleaved FASTQ for `bwa mem -p`. Three findings on that command (recommended-default dispositions for the product calls — override welcome):

- **FASTQ3-03 (fix):** a read with no stored quality (all `0xFF`, the SAM no-quality sentinel) was emitted as `~`×len (**Q93**) via `QUAL_TO_ASCII[0xFF]`, falsely claiming near-perfect quality. Now emits a fixed **Q33 (`B`)** per base, matching `samtools fastq` (the tool this command is modeled on). Extracted the conversion into a testable `encode_quality_into`; a partially-`0xFF` (malformed) record is left to the per-byte map.
- **FASTQ3-02 (warn):** the streaming converter had no missing-mate handling, so a lone mate (pair absent, or dropped by `--exclude-flags`) silently desyncs R1/R2 pairing for every later read under `bwa mem -p`. Now tracks the paired R1/R2 balance and **warns** when the counts differ (non-fatal — single-end input and deliberate mate filtering are legitimate).
- **FASTQ3-04 (doc — audit premise disproven):** the audit said "strip the suffix before appending (samtools does)". Real repro (samtools 1.23.1) **disproves** it — default samtools never appends; `-N` appends *without* stripping (`foo/1`→`foo/1/1`, exactly what fgumi does). No tool strips, and the `/1/1` case only arises on a spec-violating re-imported QNAME. Documented in the NOTES section rather than changed.

## Verification (§0)

Real-tool repro vs samtools 1.23.1: a no-quality read → **both** emit `BBBBBBBB`; dropping a mate (`-F 0x80`) now emits `Interleaved FASTQ paired-read counts differ (R1=1, R2=0): 1 singleton mate(s) will desync…`. RED-first `#[rstest]` table for the quality default (`~~~~`≠`BBBB` before the 0xFF guard). `cargo ci-fmt && ci-lint && ci-test` green (2268).

## Reviewers

Both ran. In-diff fix folded by amend: converted the missing-quality test to an `#[rstest]` `#[case]` table (CodeRabbit-CLI; matches the repo's testing convention).

Main-based. W9c of the extract/FASTQ residuals; completes W9.

## Summary by CodeRabbit

* **New Features**
  * Added clearer FASTQ conversion guidance for read-name suffixes, missing qualities, and interleaved paired-read handling.
  * Missing quality scores are now represented consistently with `B` quality characters.

* **Bug Fixes**
  * Added a warning when interleaved output may become desynchronized because paired-read counts differ.

### DIFF
--- a/src/lib/commands/fastq.rs
+++ b/src/lib/commands/fastq.rs
@@ -12,7 +12,7 @@ use fgumi_bam_io::create_raw_bam_reader;
 use fgumi_raw_bam::{
     RawRecord, extract_sequence_into, quality_scores_slice, read_name as raw_read_name,
 };
-use log::info;
+use log::{info, warn};
 use std::io::{BufWriter, Write, stdout};
 use std::path::PathBuf;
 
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

PR #573, opened 2026-07-11
docs: fix stale doctest crate paths in consensus/umi and gate doctests in CI

nh13:
## Problem

Several compiled doctests referenced `use fgumi_lib::...`, but the `consensus`, `phred`, and `umi` modules were long ago extracted from the monolithic `fgumi_lib` crate into standalone `fgumi-consensus` and `fgumi-umi` crates. Those crates don't depend on `fgumi_lib`, so every such doctest failed to compile with `error[E0433]: unresolved crate fgumi_lib`.

These never turned CI red because **CI runs tests via `cargo nextest`, and nextest does not compile or run doctests** (both the `test` job's `cargo ci-test` and the `coverage` job use nextest). So the doctest rot was invisible.

## Fix

**1. Repoint the stale imports to the crates' own roots** (12 compiled doctests):

| Before | After | File |
|---|---|---|
| `fgumi_lib::phred::X` | `fgumi_consensus::phred::X` | `crates/fgumi-consensus/src/phred.rs` (6) |
| `fgumi_lib::consensus::ConsensusSequence` | `fgumi_consensus::ConsensusSequence` | `crates/fgumi-consensus/src/sequence.rs` (1) |
| `fgumi_lib::umi::assigner::X` | `fgumi_umi::assigner::X` | `crates/fgumi-umi/src/assigner.rs` (5) |

Doctests *inside* the `fgumi_lib` crate (`src/lib/...`, e.g. `logging`, `validation`, `simulate`) correctly reference `fgumi_lib::` as a self-reference and are left unchanged.

**2. Gate doctests in CI so this can't rot again.** Add a `ci-doctest` cargo alias (mirroring the other `ci-*` aliases) and run it as a new "Doctests" step in the `test` job, right after the nextest run, with the same feature set as `ci-test`.

## Verification

```
cargo ci-doctest   # test --doc --workspace --features compare,simulate,profile-adjacency --locked
# 87 doctests passed, 0 failed
```

`cargo ci-fmt` clean.

## Out of scope (noted follow-up)

The remaining `fgumi_lib::...` references live in `,ignore`-fenced module-level examples (the consensus caller-trait walkthroughs in `caller.rs`, `base_builder.rs`, `codec_caller.rs`, `duplex_caller.rs`). They are not compiled, so the new CI gate does not cover them, and some reference since-renamed symbols (e.g. `vanilla_consensus_caller` -> `vanilla_caller`) that need per-example verification. Fixing/​un-ignoring those is a separate docs pass.

## Context

[...]

PR #574, opened 2026-07-11
docs+ci: fix broken intra-doc links and gate rustdoc in CI

nh13:
## Problem

`RUSTDOCFLAGS="-D warnings" cargo doc` fails on ~15 unresolved or private-item intra-doc links spread across four crates. They accumulated silently because **nothing in CI builds the docs** — the test/coverage jobs run `cargo nextest` (which doesn't run doctests), and there was no docs job. Same root cause as the doctest gap fixed in #573, one layer up (doc *links* rather than doctest *code*).

## Changes

**1. Fix the broken intra-doc links** (`docs:` commit) — repoint or demote each:
- **fgumi-raw-bam**: `[`RawRecord`]` → `[`RawRecord`](crate::RawRecord)`; `encode`/`encode_into` → `RecordBufEncoder::`; `query`/`read_header` → `Self::`; `MemoryEstimate` (downstream trait), `append_int_tag` (`pub(crate)`), `BAM_CIGAR_TYPE` (private const) → code spans.
- **fgumi-sort**: a `[`SpillCodec::Zstd`]` occurrence in a doc comment lacking the reference-link definition → inline path.
- **fgumi-consensus**: `rejected_reads`/`take_rejected_reads` → `Self::`.
- **fgumi (main)**: `simulate` model links → `crate::simulate::…`; `[`fetch`]` → `Self::fetch`; private `SLOT_COUNTER`/`THREAD_SLOT` → code spans.

**2. Gate rustdoc in CI** (`ci:` commit) — add a `ci-doc` alias and a `docs` job that runs `cargo doc --no-deps --workspace` with `RUSTDOCFLAGS="-D warnings"`, so any rustdoc warning (broken link, private-item link) fails the build going forward.

## Verification

- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --workspace --features compare,simulate,profile-adjacency` → clean.
- `cargo ci-fmt`, `cargo ci-lint`, `cargo ci-tag-literals` → clean.

## Note

This started as a broader CI-hardening pass. The MSRV half (correcting the declared `rust-version`) is **intentionally not here**: the declared `1.87.0` is false (deps `noodles`/`wide` force ≥1.89), but bumping it to the real floor unlocks ~85 MSRV-gated `collapsible_if` (let-chain) clippy warnings across the workspace that `-D warnings` would turn into errors. That needs its own decision (adopt let-chains vs. `allow` the lint) and PR, so it's decoupled to keep this one focused and green.

## Summary by CodeRabbit

* **Documentation**
  * Improved API documentation clarity and consistency across the project.
  * Corrected Rustdoc links and intra-doc references for easier navigation.
  * Clarified record encoding, buffer reuse, memory estimates, and simulation-related argument mappings.
  * Refined documentation for readers, sorting, tagging, and threading APIs.

* **Quality Improvements**
  * Added a documentation validation job that treats Rustdoc warnings as errors.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -5,6 +5,9 @@ ci-lint = "clippy --workspace --all-targets --features compare,simulate,profile-
 ci-tag-literals = "run --package xtask -- check-tag-literals"
 # Run documentation tests (nextest does not run doctests, so ci-test skips them)
 ci-doctest = "test --doc --workspace --features compare,simulate,profile-adjacency --locked"
+# Build the docs and fail on any rustdoc warning (e.g. broken intra-doc links).
+# RUSTDOCFLAGS="-D warnings" is set by the caller (CI job / pre-push).
+ci-doc = "doc --no-deps --workspace --features compare,simulate,profile-adjacency --locked"
 # Run tests with test-utils feature enabled (allows binary tests to use library test utilities)
 t = "test --features test-utils"
 # Generate and serve documentation locally (runs xtask then mdbook serve)
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -6,6 +6,12 @@ on:
       - main
   pull_request:
 
+# Every job here only checks out the repo and runs read-only build/test/lint
+# steps (the coverage upload authenticates with CODECOV_TOKEN, not the workflow
+# token), so drop the inherited GITHUB_TOKEN down to read-only for all of them.
+permissions:
+  contents: read
+
 env:
   CARGO_TERM_COLOR: always
   CARGO_INCREMENTAL: 0
@@ -18,6 +24,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          persist-credentials: false
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8 # stable
       - name: Set up compilation cache (sccache)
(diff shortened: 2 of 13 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed CodeRabbit feedback from the latest review:

**Token permissions (`.github/workflows/check.yml`)** — resolved. Rather than adding `permissions: contents: read` to only the two flagged jobs (`docs`, `msrv-lockstep`), added a single top-level `permissions: contents: read` block. Every job in this workflow only checks out the repo and runs read-only cargo build/test/lint steps — the `coverage` job authenticates its Codecov upload with `secrets.CODECOV_TOKEN`, not the workflow token — so a workflow-level read-only default covers all jobs uniformly and satisfies the zizmor `excessive-permissions` warning without per-job duplication.

**Outside-diff finding on `crates/fgumi-consensus/src/codec_caller.rs` (`duplex_disagreement_fixture`)** — not a valid issue; no change made. The claim was that the fixture uses matching FR sequences so the HDD tests reach `panic!(...)` instead of the typed disagreement path. That is not the case: the fixture reuses the FR pair from `test_not_emit_consensus_high_disagreement`, where the single-strand overlap positions (10 on each side) count as disagreements, so the fixture produces a non-zero disagreement count and rate. Confirmed by running the tests — all 7 relevant tests (`test_high_duplex_disagreement_counted::{by_count,by_rate}`, `test_high_duplex_disagreement_tracks_rejects`, `test_consensus_reads_typed_disagreement_{count,rate}`) pass, reaching the `Err(...)` HDD path and asserting the HDD counters, never the `panic!`. This PR also only touches an intra-doc link in that file, not the fixture.

PR #572, opened 2026-07-11
refactor(fastq): consolidate read-name suffix stripping and assert cross-parser parity

nh13:
## Summary

Follow-on to #512. Two related pieces of cleanup around FASTQ read-name and record handling:

1. **Consolidate the extract command's read-name suffix stripping onto the shared helper.** `extract.rs` carried two private helpers — `strip_read_number_suffix` (QNAME write path) and `strip_read_suffix_extract` (paired-FASTQ sync validation) — that both implemented the same rule the shared `fastq_parse::strip_read_suffix` now provides (strip a trailing space comment, then a trailing `/` + single ASCII digit; preserve `.`/`_`/`:` separators and multi-digit runs, matching fgbio `FastqSource`). All five call sites now use `strip_read_suffix`, and the two duplicated helpers are deleted. This is the follow-up #512 explicitly deferred.

2. **Pin the two independent FASTQ record validators to identical behavior, and fix a divergence found while doing so.** fgumi has two independent single-record validators: `fastq_parse::FastqRecord::from_slice` (used by `parse_fastq_records` and, transitively, by the `unified_pipeline` parser, whose `parse_single_fastq_record` delegates to it) and the SIMD crate's `try_parse_single_record` (behind `parse_records` and `SimdFastqReader`). Nothing tied them to the same behavior, and they had diverged on one input.

## The divergence (fixed)

A record whose third newline is its final byte — no quality line at all, e.g. `@r\nAC\n+\n` — was handled inconsistently:

- SIMD `try_parse_single_record`: returns a clean `FastqParseError::TooFewLines`.
- `FastqRecord::from_slice`: computed `qual_end - qual_start` unguarded, where `qual_start > qual_end`. That underflows — a **debug-build panic**, and a **misleading length-mismatch error** in release.

`from_slice` documents `# Errors` for malformed input (a panic violates that contract). This PR adds a guard that rejects the record as malformed, matching the SIMD parser. In production, framing never hands `from_slice` such a slice (complete records always have four newlines), so this is a latent-robustness fix, not a data-path behavior change.

## Cross-parser parity test

Adds a shared behavioral-parity test in `fastq_parse`: one `#[rstest]` case table plus a proptest run identical inputs through **both** validators and assert they agree — either both accept with identical name/sequence/quality, or both reject. The table folds in scenarios that previously lived in only one parser's own suite:

- SIMD-only before: single-base read; no-quality-line reject.
[...]

PR #575, opened 2026-07-11
build(msrv)!: honest MSRV in lockstep with the toolchain + let-chain adoption

nh13:
## Why

`rust-version = "1.87.0"` was a false, unverified promise. The dependency floor already requires more (`noodles`/`wide` → 1.89, `sysinfo` → 1.88), and CI builds on the `rust-toolchain.toml` pin (1.93.0). An MSRV that nothing tests and that consumers can't rely on is worse than none — the same "contract with no gate rots" failure as the doctest/doc-link gaps.

The principled fix is one enforced source of truth: **fgumi supports exactly the Rust it develops and tests on.**

## What

**1. `rust-version` → 1.93.0**, matching the `rust-toolchain.toml` channel (`build(msrv)!` commit).

**2. Adopt let-chains (85 sites).** Honestly declaring MSRV ≥ 1.88 means clippy correctly suggests collapsing `if cond { if let PAT = expr { … } }` into `if cond && let PAT = expr { … }` (stabilized 1.88). Rather than suppress the lint to hide the MSRV, the code adopts the idiom the MSRV unlocks — applied mechanically via `cargo clippy --fix`, so the code, the toolchain, and the lints all agree. Rewrites are behavior-preserving (nested-if and `&& let` short-circuit identically).

**3. `msrv-lockstep` CI job** (`ci:` commit) that fails if the two versions drift. Because every other job builds on the rust-toolchain.toml channel, keeping them equal means **all of CI already verifies the MSRV compiles** — no separate MSRV build job needed.

## Marked `!` (semver-relevant)

Raising the published MSRV is a breaking change for the crates.io libraries — hence the `!`. It reflects reality (the crates already didn't build below ~1.89); this makes the manifest honest.

## Verification

- `cargo ci-lint` (`-D warnings -W clippy::pedantic`) at 1.93 → clean (was 85 `collapsible_if` before the migration).
- `cargo ci-test` → **4756 passed, 0 failed** (behavior preserved).
- `cargo ci-fmt` → clean.
- Lockstep check dry-run: `rust-version=1.93.0 == channel=1.93.0` → OK.

## Notes

- The bump and the let-chain migration are one commit on purpose: let-chains don't compile below 1.88, so they can't be split.
- Independent of #574 (docs) and #572/#573; some files overlap on non-adjacent lines and should merge cleanly.

## Summary by CodeRabbit

- **Compatibility**
  - Raised the minimum supported Rust version to 1.93.0.
  - Updated developer setup guidance to reflect the new toolchain requirement.

- **Build Reliability**
  - Added automated checks to ensure the declared minimum Rust version matches the pinned build toolchain, preventing version drift.

- **Maintenance**
  - Streamlined internal processing and validation logic without changing command behavior or public interfaces.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -152,3 +152,26 @@ jobs:
         # (lib + tests + benches) with no default features to surface
         # regressions at PR time.
         run: cargo check --workspace --no-default-features --all-targets
+
+  msrv-lockstep:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+      - name: rust-version matches the pinned toolchain
+        # We support exactly the Rust we develop and test on: `rust-version`
+        # (the published MSRV) is kept identical to the `channel` in
+        # rust-toolchain.toml. This fails if the two drift, so neither can be
+        # bumped without the other. Because every other job builds on the
+        # rust-toolchain.toml channel, keeping them equal means all of CI
+        # already verifies the MSRV builds -- no separate MSRV build job needed.
+        run: |
+          set -euo pipefail
+          msrv=$(grep -m1 -E '^rust-version' Cargo.toml | grep -oE '[0-9]+\.[0-9]+\.[0-9]+')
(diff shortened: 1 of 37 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #579, opened 2026-07-12
docs(consensus): fix stale module-example paths and declare them non-code

nh13:
Closes the last "ungated doc claim" from the audit (task-list T5).

## Problem
The `//!` module-doc walkthroughs in `fgumi-consensus` (`base_builder`, `caller`, `codec_caller`, `duplex_caller`) were fenced ` ```rust,ignore ` — **rendered as Rust but never compiled** — and had rotted: they import from the old monolith path `fgumi_lib::consensus::...` (the crate is now `fgumi_consensus`) and reference the renamed `vanilla_consensus_caller` module (now `vanilla_caller`). Readers were shown import paths that don't exist, with no gate to catch it. #574 explicitly deferred these because its rustdoc gate doesn't cover `ignore` blocks.

## Fix
These are genuine teaching sketches (undefined context vars, elided bodies, trait-shape skeletons), so they can't compile without gutting their clarity. Rather than leave them masquerading as verified Rust:
- **corrected the crate paths** (`fgumi_consensus::…`) and the `vanilla_caller` rename, including the prose "See Also" cross-references, so what's shown is accurate; and
- **changed the fences to ` ```text `**, honestly declaring them illustrative rather than as Rust the gates would be expected to verify.

This is the "declare as non-code" side of the fix — the principled resolution for examples that legitimately can't be compiled. (Full ` # `-hidden compilable doctests are possible but would clutter the teaching examples with boilerplate; deferred as optional.)

## Verification
`RUSTDOCFLAGS="-D warnings" cargo doc -p fgumi-consensus` produces the **identical** error set as pristine main — i.e. this PR introduces **zero** new rustdoc warnings. The crate's remaining broken intra-doc links (`rejected_reads`/`take_rejected_reads`) are fixed by the sibling PR #574; once #573/#574/this all land, the crate doc-builds clean under the rustdoc gate.

## Summary by CodeRabbit

* **Documentation**
  * Updated consensus module examples to use current public crate paths.
  * Corrected references to the vanilla caller in related documentation.
  * Improved example formatting and highlighting for clearer presentation.

### DIFF
--- a/crates/fgumi-consensus/src/base_builder.rs
+++ b/crates/fgumi-consensus/src/base_builder.rs
@@ -105,7 +105,7 @@
 //!
 //! Consensus callers typically use `ConsensusBaseBuilder` as follows:
 //!
-//! ```rust,ignore
+//! ```text
 //! use fgumi_consensus::base_builder::ConsensusBaseBuilder;
 //!
 //! // Create builder with error rates
--- a/crates/fgumi-consensus/src/caller.rs
+++ b/crates/fgumi-consensus/src/caller.rs
@@ -86,7 +86,7 @@
 //!
 //! To implement a custom consensus caller, implement the `ConsensusCaller` trait:
 //!
-//! ```rust,ignore
+//! ```text
 //! use fgumi_consensus::caller::{ConsensusCaller, ConsensusCallingStats, ConsensusOutput};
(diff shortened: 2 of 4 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #585, opened 2026-07-14
test(codec): read stats as fgbio KV metrics in pre-group-filter test

nh13:
`test_codec_allow_unmapped_gates_pregroup_filter` parsed the codec stats file as a wide `ConsensusMetrics` TSV, but `codec` emits fgbio's vertical key-value format (`ConsensusKvMetric`). The header mismatch made `read_tsv` fail with `Error parsing delimited data file header` in all four cases (fast-path/threaded × default/allow-unmapped), failing `test` and `coverage` on `main`.

This was a semantic merge conflict, not a code regression: #513 switched codec stats to the KV format, then #509 (merged after) added this test against the old wide layout. Both merged cleanly as text, so the break only surfaced on the merged `main` tip — which is why the release PR (#454) and branches rebased on it (#519, #527) all inherit the same four failing cases.

The fix reads the rows back as `ConsensusKvMetric` and looks up the counts by key (`raw_reads_considered`, `consensus_reads_emitted`), mirroring the sibling `simplex` tests. The assertions' intent is unchanged: the supplementary alignment is filtered before grouping (2 raw reads considered) and the single FR molecule is emitted (1 consensus read).

Verified locally: all four cases pass; full `codec` module green; `fmt` and `clippy --all-targets -D warnings` clean.

## Summary by CodeRabbit

* **Tests**
  * Updated CODEC integration test coverage to validate the current `--stats` tab-separated output format.
  * Added checks for raw reads considered and consensus reads emitted statistics.

### DIFF
--- a/src/lib/commands/codec.rs
+++ b/src/lib/commands/codec.rs
@@ -886,15 +886,30 @@ mod tests {
         cmd.outer_bases_length = 0;
         cmd.execute("test")?;
 
-        let metrics: Vec<ConsensusMetrics> = DelimFile::default().read_tsv(&stats_path)?;
-        let metrics = metrics.first().expect("stats file should contain one metrics row");
+        // Codec emits fgbio's vertical key-value metrics format, so read the rows
+        // back as `ConsensusKvMetric` and look up the two counts by key: the
+        // supplementary read must be filtered before grouping, so exactly the two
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #550, opened 2026-07-10
fix(extract): pool quality-encoding detection across all input FASTQs (W9a: EXT3-01, S1)

nh13:
## What / why (EXT3-01, S1)

`fgumi extract` sampled quality scores for encoding detection from **`inputs[0]` only**. When the first file's quality range differs from the template reads' — the realistic trigger is an all-high-quality index read in `inputs[0]` — detection is forced to Phred+64 even though the template reads in `inputs[1..]` are Phred+33, silently shifting **every template base quality by ~31**.

fgbio `FastqToBam` detects from `heads.transpose.flatten` — the pooled heads of **all** sources (`FastqToBam.scala:128–130`). This PR matches that: extract now pools the first `QUALITY_DETECTION_SAMPLE_SIZE` records of every input before detecting, via a new testable `sample_detection_quals` helper.

## Verification (§0)

Real-tool repro — R1 = all-Q71 `h`, R2 = low-Q7 `(`:

| | encoding | R1 QUAL | R2 QUAL |
|---|---|---|---|
| fgumi **before** (inputs[0] only) | Illumina (wrong) | `IIIIIIII` (Q40) | `!!!!!!!!` (Q0) |
| fgumi **after** (pool all) | Standard | `hhhhhhhh` (Q71) | `((((((((` (Q7) |
| `fgbio FastqToBam` | Standard | `hhhhhhhh` | `((((((((` |

fgumi-after base qualities are **byte-identical to fgbio** (`fgbio` logged "Detected FASTQ quality encoding: Standard"). RED-first TDD: the new pooling test fails (`Illumina` ≠ `Standard`) against a first-input-only sampler, passes once all inputs are pooled. `cargo ci-fmt && ci-lint && ci-test` green (2273 tests).

## Reviewers

Both ran. In-diff fix folded by amend: dropped the async reader for the bounded sampling head-scan (CodeRabbit-CLI — the ≤400-record peek gains nothing from async; the main extraction still honors `--async-reader`).

## Scope

This PR is **EXT3-01 only** (the S1). The related **EXT3-03** (fgbio's `case Nil => fail(...)` hard-fail on no-compatible-encoding, same detection block) is deferred: it's a stricter error-behavior change that needs its own fgbio-parity repro of the exact "Nil" condition, and is lower severity (S3). Tracked as a W9 follow-up. Stacked on #496 (`nh/fix-extract-read-structure`) — auto-retargets to main on its merge; both edit `extract.rs`.

## Summary by CodeRabbit

[...]

PR #587, opened 2026-07-17
docs(extract): document EXT3-03 encoding-failure parity with fgbio (W9 follow-up to #550)

nh13:
## What / why (EXT3-03, W9 follow-up to #550)

Follow-on to #550 (EXT3-01, now merged). #550 deferred **EXT3-03** — fgbio `FastqToBam`'s `case Nil => fail(...)` hard-fail on the "no compatible encoding" branch — as a separate item. This PR closes it out.

### Finding

Tracing both sides shows fgumi's failure behavior **already matches fgbio's**, so EXT3-03 is a documentation + parity-lock change, not a behavior change.

fgbio derives the failure from its encoding model: it builds the set of encodings whose ASCII range contains every observed quality char and fails with `case Nil => fail("Quality scores in FASTQ files do not match any known encoding.")` when that set is empty (`FastqToBam.scala:131`). Because Standard's ASCII range `[33, 126]` is the **union** of all three fgbio encodings (Solexa `[59, 126]` and Illumina `[64, 126]` are subsets), fgbio's compatible set is empty on *exactly* the condition fgumi already bails on in `QualityEncoding::from_stats`: an observed quality char `< 33` or `> 126`.

I checked the empty-input, all-empty-quals, and ambiguous-range paths too — none is a case where fgumi succeeds but fgbio fails. The failure *sets* are identical.

### fgumi message is a deliberate improvement

The only real difference is the message. fgbio emits a static string; fgumi reports the observed `[min, max]` range and the required range, pointing the user straight at the offending file. This divergence is intentional and now documented as such (not a parity gap).

| | message on no-compatible-encoding |
|---|---|
| fgbio | `Quality scores in FASTQ files do not match any known encoding.` |
| fgumi | `Invalid quality scores detected: range [20, 50]. Quality scores must be in the printable ASCII range (33-126)` |

## Changes

- Document the EXT3-03 fgbio parity at the detection bail site in `QualityEncoding::from_stats`, and in the module `# Quality Encoding` header.
- Record that fgumi's dynamic, range-reporting message is a deliberate improvement over fgbio's static one.
- Consolidate the `invalid_low` / `invalid_high` tests into an `#[rstest]` case table (`detect_reports_observed_range_on_no_compatible_encoding`, cases `below_min_char` / `above_max_char` / `both_out_of_range`) that asserts both the fgbio-equivalent failure set and that the improved range-reporting message survives.

No runtime behavior change; net diff is docs + a test refactor (1 file, +45/−24).

## Verification

- Targeted encoding-detection tests green (16/16, incl. the 3 new parity cases).
- `cargo ci-fmt` and `cargo ci-lint` clean.
- Full `cargo ci-test` is green except one pre-existing, unrelated failure: the workspace-wide `xtask check_tag_literals` lint walks into stray nested agent worktrees under `main/.claude/worktrees/` (foreign branches, untracked by this repo) and flags *their* bench files. This diff contains no bare SAM-tag byte literals and does not touch that check.

## Summary by CodeRabbit

* **Documentation**
  * Clarified how quality encoding is detected across all input FASTQ files.
  * Documented improved error reporting when no compatible encoding is found, including the observed quality range.

* **Bug Fixes**
  * Error messages now report the minimum and maximum observed quality scores for invalid encodings.

* **Tests**
  * Added coverage verifying that invalid quality errors include the observed score range.

### DIFF
--- a/src/lib/commands/extract.rs
+++ b/src/lib/commands/extract.rs
@@ -12,7 +12,10 @@
 //! # Quality Encoding
 //!
 //! Automatically detects and handles both Phred+33 (standard) and Phred+64 (Illumina 1.3-1.7)
-//! quality encodings.
+//! quality encodings. Detection pools the heads of all input FASTQs (EXT3-01), matching fgbio
+//! `FastqToBam`. When no encoding matches the observed qualities, fgumi fails like fgbio's
+//! `case Nil => fail(...)` (EXT3-03) but improves on its static message by reporting the
+//! observed quality range in the error (see [`QualityEncoding::from_stats`]).
 
 use crate::commands::command::Command;
 use crate::commands::common::{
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #591, opened 2026-07-18
fix: recompute the BAM bin field after raw POS/CIGAR mutations (clip, zipper)

nh13:
## Summary

`fgumi clip` (and, in a narrower case, `fgumi zipper`) emitted a stale BAM `bin` field. The raw pipelines mutate an already-encoded BAM record in place and write the bytes verbatim, so — unlike the noodles/htsjdk encoders, which recompute the indexing bin on write — there is no serialization step to refresh `bin` (bytes 10–11) after `POS` or the CIGAR changes. Any clip that moved a read into a different bin shipped the pre-clip value, diverging from fgbio.

### Reproduction

A read placed at 16300 with `200M` straddles the 16 kb bin boundary (level-4 bin 585). A 116-base 5′ soft-clip moves it to `pos 16416`, `116S84M`, whose correct bin is 4682:

| Tool | resulting alignment | `bin` |
|------|--------------------|-------|
| fgbio ClipBam | `pos 16417`, `116S84M` | **4682** ✓ |
| fgumi clip (before) | `pos 16417`, `116S84M` | **585** ✗ (stale) |

The same class of bug affected `zipper`'s mate reconciliation: a read relocated to its mapped mate's coordinate kept the stale unmapped bin (4680) instead of its placed position's bin.

Note: fgumi's own BAI/CSI writer recomputes start/end from `POS`+CIGAR and ignores the record's `bin`, so its own index was unaffected. The stale bin hurt external consumers (htsjdk/samtools-based tools reading fgumi output) and broke byte-parity with fgbio.

## What changed

- **`fgumi-raw-bam::bin`** (new module): the canonical SAM spec §5.3 `reg2bin`, the `UNMAPPED_BIN` constant, and `bin_from_raw_bam` / `set_bin_from_raw_bam`, which compute a record's bin from its `POS` and CIGAR reference span (unmapped `POS < 0` → 4680, keying off the alignment start like htsjdk `computeIndexingBin`, not the unmapped flag). Plus a `RawRecord::recompute_bin()` wrapper.
- **`clip`**: recompute the bin at the three raw mutation primitives (`clip_start_of_alignment`, `clip_end_of_alignment`, `make_read_unmapped_raw`) so a clipped record is always self-consistent, and in `set_mate_info_raw` when a clipped-unmapped read is relocated or a pair is cleared to unmapped. `upgrade_clipping` only converts clip type (no `POS`/span move), so it correctly needs no recompute.
- **`zipper`** (`Template::fix_mate_info`): recompute after relocating an unmapped mate to its mate's coordinate and after clearing a pair to unmapped.
- **`simulate`**: `region_to_bin` now delegates to the shared `reg2bin` instead of duplicating the spec code; the private `UNMAPPED_BIN` in `builder.rs` also shares the constant.

## Testing

- Unit tests at every mutation site (clip start/end/unmap, `set_mate_info_raw` relocation, `fix_mate_info` relocation), each asserting the emitted bin equals `reg2bin` of the post-op coordinates — TDD, each written failing first.
- `reg2bin` is covered by the full authoritative vector table ported from htsjdk `GenomicIndexUtilTest.testRegionToBinDataProvider`, exercising every binning level (16 kb through the whole-window bin 0).
- End-to-end: rebuilt binary re-run on the reproductions above — every previously-stale case now matches `reg2bin` of the post-op coordinates, and **directly matches fgbio ClipBam / ZipperBams** output bins.
- Full workspace suite (5243 tests), `cargo ci-fmt`, and `cargo ci-lint` all pass.

## Reading order

1. `crates/fgumi-raw-bam/src/bin.rs` — the shared helper and its semantics.
2. `crates/fgumi-sam/src/clipper.rs` and `src/lib/commands/clip.rs` — the clip fix.
3. `src/lib/template.rs` — the zipper fix.
4. `src/lib/commands/simulate/mod.rs` — the delegation refactor.

## Summary by CodeRabbit

* **New Features**
  * Added BAM bin calculation and update support based on alignment position and CIGAR span.
  * Added APIs to compute or refresh bins for raw BAM records.
  * Added automatic bin updates when clipping reads or modifying mate information.

* **Bug Fixes**
  * Prevented stale bin values after coordinate, CIGAR, or mapping-status changes.
  * Correctly handles unmapped and zero-reference-span reads.

* **Tests**
  * Added coverage for bin boundaries, clipping, unmapped reads, and coordinate mutations.

### DIFF
--- a/crates/fgumi-raw-bam/src/bin.rs
+++ b/crates/fgumi-raw-bam/src/bin.rs
@@ -1,0 +1,187 @@
+//! BAM bin field (`reg2bin`) computation.
+//!
+//! The BAM binary record carries a 16-bit `bin` field (bytes 10–11) that indexes
+//! the alignment into the UCSC binning scheme (SAM spec §5.3). The bin is a pure
+//! function of the alignment's reference span — its 0-based start (`POS`) and the
+//! reference length consumed by its CIGAR — so any operation that moves `POS` or
+//! rewrites the CIGAR (clipping, mate relocation, unmapping) invalidates it and
+//! must recompute it.
+//!
+//! The noodles BAM *encoder* recomputes the bin from `POS`+CIGAR whenever a
+//! `RecordBuf` is serialized, and htsjdk does the same on write — but fgumi's
+//! raw-byte pipelines mutate the encoded record in place and emit the bytes
+//! verbatim, so there is no encode step to refresh the field. These helpers are
+//! the raw-byte equivalent: call [`set_bin_from_raw_bam`] (or the
+//! `RawRecord::recompute_bin` convenience wrapper) after any in-place `POS`/CIGAR
+//! change so the emitted record carries a correct bin.
+
+use crate::cigar::reference_length_from_raw_bam;
+use crate::fields::pos;
+
+/// The BAM "unmapped bin" — `reg2bin(-1, 0)` per SAM spec §4.2.1. Records with no
+/// alignment position (`POS < 0`) use this value.
+pub const UNMAPPED_BIN: u16 = 4680;
+
+/// Compute the BAM bin for a reference region using the SAM spec §5.3 `reg2bin`
+/// reference implementation.
+///
+/// `beg` is the 0-based leftmost aligned position; `end` is the 0-based position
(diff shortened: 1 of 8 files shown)


nh13:
@coderabbitai review

PR #593, opened 2026-07-19
perf(simulate): stream records into the sort instead of an intermediate BAM

nh13:
Follow-up to #576. That PR made `simulate` delegate ordering to the canonical `fgumi-sort` engine by writing an unsorted temp BAM and handing it to `sort()`. This removes the temp BAM.

## The redundancy

`RawExternalSorter` was reachable only through `sort(input_path, output_path)`, so a producer that already has records in memory had to write a BAM just to hand it straight back. Two costs, and the second is the one that bites:

1. Every run pays a full compress → write → read → decompress round-trip, even when the records fit in memory and the sort's existing zero-temp-file fast path would otherwise touch disk only for the output.
2. Runs larger than the memory limit keep the whole intermediate on disk *alongside* the spill chunks — and it lands in the **output** directory, so `simulate` demanded ~2.1x the output size on the destination volume.

The engine never needed the file: every per-order implementation is already driven by a `RecordSource` iterator. The file-to-file signature was the only thing in the way.

## The change

**`feat(sort)`** — `RawExternalSorter::sort_records(records, header, output)` sorts an in-process record stream. It runs the same accumulate/spill/merge path as `sort`, including the in-memory fast path; both entry points now share `create_worker_pool` (which also enforces the spill-codec invariant, so neither can reach a misconfigured pool) and `sort_from_source`.

The new `RecordSource::Stream` variant carries fallible items. The first `Err` ends the stream and surfaces through `take_error`, which every sort order already checks *before* writing output — so a producer failure aborts rather than leaving a silently truncated BAM.

**`perf(simulate)`** — `mapped-reads` and `grouped-reads` generate on a scoped thread that streams into `sort_records`. Nothing is materialized before the output, and generation overlaps the sort's accumulate/spill/merge work.

Records cross the channel in batches of 256 (the batch size the sort's own read-ahead reader uses). This is load-bearing, not incidental: sending records one at a time costs **more than the intermediate BAM it replaces** — measured ~7% slower end-to-end before batching, ~7% faster after.

## Routing threads and memory

Both commands were running the sort with a hardcoded 512 MB limit, no `initial_capacity` tuning, and no way to steer spill off a tmpfs `/tmp`. They now take the same knobs as `fgumi sort`, with the same names, defaults, and semantics: `--max-memory`, `--memory-reserve`, `--memory-per-thread`, `-T/--tmp-dir` (plus `FGUMI_TMP_DIRS`). `--threads` now drives the sort's thread count and the per-thread memory scaling.

**Behavior change worth flagging:** the default is now `768M` *per thread* rather than `512M` fixed, so `-t 4` gets a 3 GB budget and spills later than before.

## Measured

1M molecules (6,003,722 records, 4 threads), versus the implementation merged in #576:

| | before | after |
|---|---|---|
| wall, spilling (512M budget) | 13.85 s | **13.18 s** |
| wall, no spill (200k mol) | 2.65 s | **2.47 s** |
| peak disk, output directory | 1234 MB | **593 MB** |
| peak disk, output + temp | 1722 MB | **1182 MB** |

Output records are byte-identical to the previous implementation for mapped-reads, grouped-reads, and grouped-reads-duplex (verified against a binary built at 829cebdc; only `@PG` differs, since the two binaries have different paths).

## Tests

`fgumi-sort` gains 12 tests pinning the invariant every caller depends on: `sort_records` output is byte-identical to `sort` over the same records for all four sort orders, on both the fits-in-memory and spill-and-merge paths; an empty stream writes a header-only BAM; and a producer error aborts with no output file left behind, whether it fails on the first record or after a spill.

The `simulate` integration table now runs every case at both a budget that holds the whole run and one that forces spill-and-merge, and asserts the spilled output matches the in-memory output record for record. That comparison is over records rather than whole files because `@PG` embeds the command line, which necessarily differs between the two runs.

## Verification

- `cargo ci-test` -> 5463 passed, 0 failed
- `cargo ci-fmt`, `cargo ci-lint` (`-D warnings -W clippy::pedantic`), `cargo ci-doctest` -> clean

## Reading order

Start with `crates/fgumi-sort/src/external.rs` (`sort` / `sort_records` / `sort_from_source`), then `read_ahead.rs` for the `Stream` variant's error contract, then `simulate/common.rs` for `RecordSink` and `SortResourceArgs`. The two command files are mechanical after that.

## Summary by CodeRabbit

- **New Features**
  - Simulation commands now stream generated reads directly into sorting, reducing intermediate file usage.
  - Added memory, reserve, per-thread, and temporary-directory options for controlling simulation sorting resources.
  - Added an in-process record-streaming API for sorting generated records.

- **Bug Fixes**
  - Sorting now aborts cleanly when record generation fails, preventing incomplete output files.

- **Tests**
  - Added coverage confirming identical results across memory and spill-to-disk configurations.
  - Added validation for empty streams and producer errors.

### DIFF
--- a/crates/fgumi-sort/src/external.rs
+++ b/crates/fgumi-sort/src/external.rs
@@ -1884,6 +1884,71 @@ impl RawExternalSorter {
     ///
     /// Returns an error if reading, sorting, or writing the BAM file fails.
     pub fn sort(&self, input: &Path, output: &Path) -> Result<RawSortStats> {
+        let pool = self.create_worker_pool()?;
+
+        // Open input BAM and create record source
+        // N+2 model: workers do ReadInputBlocks + DecompressInput,
+        // main thread reads records directly from PooledInputStream.
+        debug!(
+            "Phase 1: Pool-integrated input reading ({} workers, N+2 model)",
+            pool.num_workers()
+        );
+        let (record_source, header) = {
+            let (reader, header) =
+                create_raw_bam_reader_pool_integrated(input, &pool, self.async_reader)?;
+            (RecordSource::direct(reader), header)
+        };
+
+        self.sort_from_source(record_source, &header, output, pool)
+    }
+
+    /// Sort records produced in-process, without a BAM input file.
+    ///
+    /// This is the streaming counterpart to [`Self::sort`]: instead of reading
+    /// an input BAM, records are pulled from `records` as the sort's accumulate
+    /// phase consumes them. Producers that would otherwise write an unsorted BAM
+    /// only to hand it straight back for sorting should use this — it removes a
+    /// full compress/write/read/decompress round-trip, and removes the
+    /// intermediate file's disk footprint entirely (which for inputs that exceed
+    /// `memory_limit` would otherwise sit alongside the spill chunks).
+    ///
+    /// Spill and merge behave exactly as in [`Self::sort`]: when everything fits
+    /// within `memory_limit` nothing touches disk except the output; past that
+    /// limit, sorted chunks spill to the configured temp directories and are
+    /// k-way merged.
+    ///
+    /// `header` is the input header — the output header's sort-order tags are
+    /// derived from it the same way [`Self::sort`] derives them from the input
+    /// BAM's header, and `pg_info` (if set) is applied to it.
+    ///
+    /// An `Err` item from `records` aborts the sort and is propagated; the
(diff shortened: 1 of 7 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #568, opened 2026-07-11
perf(compare): verify sort order in a single streaming pass

nh13:
## Summary

The sort-verify engine backing `fgumi compare bams --command sort` read each input BAM **twice**: two per-file monotonic sort-order passes (`fgumi_sort::verify_sort_order`) plus a reopened run-grouped multiset-comparison pass — **four full BAM traversals where two suffice**. `--command sort` runs on WGS-scale BAMs where BAM I/O + BGZF decode is the documented bottleneck, so this was roughly 2x the necessary decode work.

This refactor reads each file **exactly once** (4 → 2 traversals total), preserving byte-for-byte identical outputs.

## Design: an order-checking stream adapter

A sorted-BAM comparison is a synchronized merge of two ordered keyed streams. Both the per-file properties (monotonic order, count, first violation) and the cross-file properties (run-grouped multiset, presence) are folds over one pass.

An `OrderChecked` stream adapter layers over the single keyed-record read the run comparison already consumes:

- `OrderTracker` is the inline fold of `fgumi_sort::verify_sort_order` — it watches each key flow past and accumulates `(violations, first_violation)` (with the same 1-based record numbering and `from_utf8_lossy` read-name extraction).
- `OrderChecked::next_keyed` reads a record, extracts its key, feeds it through the tracker as a **side effect**, and yields `(key, record)` downstream unchanged.
- `pull_run` / `run_compare` are unchanged in comparison logic; they just pull records through the adapter instead of a bare reader.

The two concerns (order-checking vs. run comparison) stay decoupled and independently testable while sharing one read.

## The invariant preserved: complete per-file counts past a desync

Previously, `verify_one` walked each file end to end, so `bam1_violations` / `bam2_violations` and the `*_first_violation` fields were **complete** per-file counts regardless of whether the two streams desynced. The refactor keeps this: `run_compare`'s desync **drain** path pulls the tail *through* the same `OrderChecked` adapter (`OrderChecked::drain`), so a mis-sort in the drained tail — past the point where the comparison stopped — is still counted. This is not a partial-count shortcut; the per-file violation output is identical to the old two-pass computation for every input.

## Tests

- All existing sort-verify unit + integration tests remain green (`sort_verify` / `--command sort` / `test_sort_verify`).
[...]

PR #594, opened 2026-07-19
perf(compare): bound sort-verify run comparison to O(order divergence)

nh13:
Stacked on #568 — review that first; this PR is the single commit on top.

## The problem

The sort-verify engine materialized each equal-core-sort-key run in full (`pull_run` → `Vec<RawRecord>`, both files at once) before comparing the two runs' record multisets. Run length is data-controlled, and one run in particular is the whole file: `extract_coordinate_key_inline` packs every `tid = -1` record to a single constant, so a coordinate-sorted BAM's unmapped tail is **one run**.

That is not a degenerate input — unmapped reads sort last in every coordinate-sorted BAM, and 2% unmapped on a 30× genome is ~24M reads.

Measured before the change (peak RSS, `/usr/bin/time -l`):

| Unmapped tail | Peak RSS | | Control: same counts, distinct coordinates | Peak RSS |
|---|---|---|---|---|
| 100k (10bp) | 43 MB | | 100k | 18 MB |
| 400k | 125 MB | | 400k | 18 MB |
| 1.6M | 450 MB | | 1.6M | 18 MB |
| **1M × 150bp** | **1,049 MB** | | | |

The control is what makes it conclusive: at run size 1 the engine is genuinely streaming — flat across a 16× record range. Same code path, same record counts; only run size differs.

## The fix

Compare each run by **streaming cancellation** instead. Both sides advance in lockstep, and every arriving record is reduced to its canonical content key (`content_key_exact`, whose byte-equality is exactly `Exact` content-equality) and cancelled against a waiting counterpart on the opposite side, or held pending if none waits.

Files listing a run's records in the same order never hold more than one record pending, however long the run. Pending grows only with the **displacement** between the two orderings, capped by `MAX_PENDING_RUN_RECORDS`, which fails with an actionable message (re-sort both inputs) rather than exhausting memory.

After:

| Unmapped tail | Before | After |
|---|---|---|
| 100k | 43 MB | 17 MB |
| 400k | 125 MB | 18 MB |
| 1.6M | 450 MB | 18 MB |
| **1M × 150bp** | **1,049 MB** | **17 MB** (~61×) |

Flat, and matching the distinct-coordinate baseline — the unmapped tail now costs what any other run costs.

## Better diagnostics

Whatever remains pending when a run ends is the exact symmetric difference, so a mismatch names the reads involved. A 100k-record run missing one read previously reported:

```
run 50: record multiset differs (100000 record(s) in bam1 vs 99999 in bam2)
```

…which points at every unmapped read in the file at once. Now:

```
run 50: record multiset differs (100000 record(s) in bam1 vs 99999 in bam2); only in bam1: u00049946
```

Verified that name is genuinely the missing read, not merely plausible. Mismatch detection stays bounded on the differing path too (17 MB, exit 1).

## On run scoping

An earlier framing held that scoping cancellation to a run was the *soundness* constraint — that a record which migrated between sort keys could otherwise cancel across runs and produce a false MATCH. **That is wrong, and mutation testing caught it.** Disabling run scoping did not fail the test asserting it.

The real reason: a record's run membership is a function of its content. `content_key_exact` covers the core-field bytes carrying `ref_id`/`pos`/`flags`, which are exactly the fields the coordinate key `(tid, pos, reverse)` is derived from. Two records with equal content keys always fall in the same run, so a moved record can never find its old self to cancel against — scoping or no scoping.

Run scoping preserves the engine's *per-run structure*: boundary-misalignment detection, per-run counts, `run {n}` diff labels, and the deliberate no-resync behavior. That is now documented on `compare_run` so the next reader is not misled the same way.

## Testing

9 new tests (5482 → 5491), all verified load-bearing by mutation testing rather than assumed:

- `identical_run_cancels_on_arrival_and_keeps_pending_bounded` — 10k records, peak pending ≤ 1
- `unmapped_tail_mismatch_names_the_offending_read` — the named diagnostic
- `duplicate_records_cancel_by_multiplicity` (3 cases) — multiset, not set: read names are not unique in a BAM, so a set-based canceller would let a second copy match the first and report a dropped duplicate as equal
- `reversed_run_cancels_completely` — opposite order is still the same multiset
- `exceeding_the_pending_cap_reports_an_actionable_error`
- `a_record_that_moved_to_another_coordinate_is_reported`
- `residual_is_empty_iff_content_multisets_are_equal` (proptest)

Two mutations were applied and confirmed caught: removing cancellation (caught by 6 tests) and switching multiplicity to set semantics (caught by all 3 multiplicity cases).

`multiset_equal` is removed, superseded by the streaming canceller — which also deletes a doc comment that cited a design note not present anywhere in the repo.

Full suite green: 5491 passed, 26 skipped; lint and fmt clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved comparison accuracy for records with matching sort keys, including reordered and repeated records.
  * Added clearer diagnostics identifying unmatched records and trailing-run mismatches.
  * Comparisons now handle large order differences more efficiently with bounded pending state and actionable errors.

* **Documentation**
  * Clarified how exact content matching supports sort verification and bounded order comparisons.

### DIFF
--- a/src/lib/commands/compare/engines/sort_verify.rs
+++ b/src/lib/commands/compare/engines/sort_verify.rs
@@ -348,63 +348,288 @@ where
 /// using `next` as one-record lookahead state (carried across calls). Returns `None`
 /// once the stream (and lookahead) are exhausted. Every record read here flows through
 /// `stream`'s order tracker exactly once.
-fn pull_run<K, ExtractKey, IsViolation>(
+fn next_in_run<K, ExtractKey, IsViolation>(
     stream: &mut OrderChecked<'_, K, ExtractKey, IsViolation>,
     next: &mut Option<Keyed<K>>,
+    key: &K,
     same_core: &impl Fn(&K, &K) -> bool,
-) -> Result<Option<(K, Vec<RawRecord>)>>
+) -> Result<Option<RawRecord>>
 where
     K: Clone,
     ExtractKey: Fn(&[u8]) -> K,
     IsViolation: Fn(&K, &K) -> bool,
 {
-    let Some((key, rec)) = next.take() else { return Ok(None) };
-    let mut run = vec![rec];
-    while let Some((k, r)) = stream.next_keyed()? {
-        if same_core(&k, &key) {
-            run.push(r);
-        } else {
-            *next = Some((k, r));
-            break;
-        }
+    let Some((k, rec)) = next.take() else { return Ok(None) };
+    if !same_core(&k, key) {
+        // Belongs to the next run: put it back untouched so the run boundary is preserved.
+        *next = Some((k, rec));
+        return Ok(None);
     }
-    Ok(Some((key, run)))
+    *next = stream.next_keyed()?;
+    Ok(Some(rec))
 }
 
-/// Returns `true` iff `a` and `b` contain the same multiset of records under
-/// [`ContentPredicate::Exact`](super::content::ContentPredicate::Exact) (order-independent
-/// tag comparison, all core fields exact).
-///
-/// Linear-time counting: each record is reduced to a canonical content key
-/// ([`content_key_exact`]) whose byte-equality is *exactly* `Exact` content-equality (a
-/// true equivalence relation — reflexive/symmetric/transitive, no epsilon tolerance), then
-/// the two runs' key multisets are compared with a hash-count map. This replaces the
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #601, opened 2026-07-20
fix: reject degenerate CLI values instead of silently misbehaving

nh13:
Five independent validation gaps where an out-of-range or non-finite argument is accepted and then silently changes behaviour rather than erroring.

## `downsample --fraction NaN` → empty BAM, exit 0

The range test `fraction <= 0.0 || fraction > 1.0` is false for `NaN`, so `NaN` passed validation; every `rng < fraction` sampling decision is then false and the command writes an **empty output BAM with exit code 0**. Adds a leading `is_finite` guard, extracted into a testable `validate_fraction`.

## `downsample -i -` spuriously rejected

Input validation called `validate_file_exists` directly instead of `BamIoOptions::validate`, which exempts stdin. Piping into downsample failed with "file does not exist" even though the BAM reader handles stdin fine.

## `codec --max-duplex-disagreement-rate NaN` disables rejection

Same NaN hole. A `NaN` rate makes every downstream `rate > threshold` comparison false, **silently disabling duplex-disagreement rejection** so molecules that should be rejected are emitted as consensus reads.

## `codec --single-strand-qual` / `--outer-bases-qual` unbounded

Both are bare `Option<u8>` written straight into the consensus record's QUAL array, so a value above the maximum Phred score emits BAM quality bytes outside the legal SAM `!`..`~` range — corrupt output, no error. Now bounded by the existing `ConsensusCallingOptions::MAX_PHRED` (93), which already guards the four sibling quality options; widened to `pub(crate)` so there is one source of truth rather than a second literal.

## `simplex` and `duplex --min-reads 0` are silent no-ops

0 was accepted, making the minimum-family-size filter dead since `raw_records.len() < 0` can never hold. `codec` already rejects this. simplex's bounds are extracted into `validate_read_bounds`, and the same lower bound is added to `duplex` — whose caller validates emptiness (`duplex_caller.rs:376`) and high-to-low ordering (`:394`, `:397`) but never a lower bound. All three consensus commands now agree on the same flag.

## Note on removed tests

This also drops three pre-existing downsample tests that were **vacuous**: they built a command with `fraction: 0.0` and then asserted `cmd.fraction <= 0.0`, never invoking validation, so they passed regardless of whether it existed. The new rstest tables call the validators directly and cover the NaN/infinity cases.

## Testing

- `cargo ci-test`: 5564 passed, 22 skipped

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation for disagreement rates, quality overrides, downsampling fractions, and read-count settings.
  - Rejects non-finite values such as `NaN` and infinity where applicable.
  - Prevents invalid quality scores and empty read groups from being accepted.
  - Added clearer error messages for invalid command-line options.

### DIFF
--- a/src/lib/commands/codec.rs
+++ b/src/lib/commands/codec.rs
@@ -544,9 +544,34 @@ impl Codec {
             bail!("min-duplex-length must be >= 1");
         }
 
-        // Validate disagreement rate
+        // Validate disagreement rate. The non-finite check must come first: `NaN`
+        // compares `false` against both bounds, so a bare range test admits it, and
+        // every downstream `rate > threshold` comparison is then false — silently
+        // disabling duplex-disagreement rejection so molecules that should be
+        // rejected are emitted as consensus reads.
+        if !self.max_duplex_disagreement_rate.is_finite() {
+            bail!(
+                "max-duplex-disagreement-rate must be a finite number, got {}",
+                self.max_duplex_disagreement_rate
+            );
+        }
         if self.max_duplex_disagreement_rate < 0.0 || self.max_duplex_disagreement_rate > 1.0 {
             bail!("max-duplex-disagreement-rate must be between 0.0 and 1.0");
+        }
+
+        // Validate the fixed quality overrides. These are written straight into the
+        // consensus record's QUAL array, so a value above the maximum Phred score
+        // would emit BAM quality bytes outside the legal SAM `!`..`~` range.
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #603, opened 2026-07-20
fix(simulate): emit faithful duplex consensus tags

nh13:
Two defects made `fgumi simulate consensus-reads --duplex` produce records that no real consensus caller emits, so simulated duplex BAMs were unsound fixtures for validating `fgumi filter` and for fgbio-parity harnesses.

## Combined per-base arrays on duplex records

`CD_BASES`/`CE_BASES` were written unconditionally, but `duplex_caller` emits only the per-strand `AD_BASES`/`AE_BASES`/`BD_BASES`/`BE_BASES` arrays alongside the scalar `cD`/`cM`/`cE` (`duplex_caller.rs:1132-1177` vs `:1207-1209`). Now emitted only in simplex mode.

This is safe for the documented "suitable for input to `fgumi filter`" contract: filter branches on `is_duplex_consensus` and masks duplex reads through `mask_duplex_bases`, which reads the four per-strand arrays (`crates/fgumi-consensus/src/filter.rs:819-825`) and never the combined ones — while simplex reads still get the `CD_BASES`/`CE_BASES` that `mask_bases` requires.

The duplex `cD`/`cM`/`cE` scalars are now derived from the per-strand sums rather than the independently sampled simplex values, matching `duplex_caller`, which computes them over the combined per-base depth (`ab_i + ba_i`).

## Strand minimum depths could exceed the combined minimum

`aM` and `bM` were each computed as `strand_depth.min(cM)`, so **both could equal `cM`** and `aM + bM` routinely exceeded it. A duplex position's combined depth is the sum of its two strand depths, so the truth TSV and the BAM tags disagreed with each other — anyone validating a real run against `--truth` got impossible expectations.

`cM` is now split by the same strand fraction already used for `cD`, clamping the A share to `[cM - bD, min(aD, cM)]`. That range is feasible because `cM <= cD == aD + bD`, and it guarantees `aM + bM == cM`, `aM <= aD`, and `bM <= bD`.

## Testing

Tests sweep 64 seeds — the strand fraction is sampled per read, so a single seed proves little — asserting the summation and ordering invariants, and that the truth tuple's `cD`/`cM` agree with the emitted tags (the truth file carries sampled values while the tags are now derived; only the `aM + bM == cM` invariant keeps them equal, so it is pinned explicitly). A second test asserts duplex records carry no `CD_BASES`/`CE_BASES` while retaining all four per-strand arrays, and that simplex records still carry the combined arrays.

- `cargo ci-test`: 5536 passed, 22 skipped

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved duplex consensus depth calculations to ensure strand-level and combined minimums remain consistent.
  * Corrected duplex combined consensus summary tags for more accurate BAM metadata.
  * Updated duplex record tag output to omit combined per-base depth/error tags where not applicable (simplex output remains unchanged).
* **Tests**
  * Added regression and property-based coverage validating duplex invariants and correct combined tag generation.

### DIFF
--- a/src/lib/commands/simulate/consensus_reads.rs
+++ b/src/lib/commands/simulate/consensus_reads.rs
@@ -378,9 +378,16 @@ fn generate_consensus_pair(
         let ad = ((cd as f64) * a_frac).round() as i32;
         let bd = cd - ad;
 
-        // Min depths for each strand
-        let am = (ad.min(cm)).max(0);
-        let bm = (bd.min(cm)).max(0);
+        // Min depths for each strand. These must satisfy aM + bM == cM, because a
+        // duplex consensus position's combined depth is the sum of its two strand
+        // depths — the real duplex caller derives cM as the per-base minimum of
+        // (ab_i + ba_i). Splitting cM by the same strand fraction as cD keeps the
+        // truth file and the emitted tags mutually consistent.
+        //
+        // The split is feasible because cM <= cD == aD + bD; clamping the A share to
+        // [cM - bD, min(aD, cM)] guarantees both aM <= aD and bM = cM - aM <= bD.
+        let am = (((cm as f64) * a_frac).round() as i32).clamp((cm - bd).max(0), ad.min(cm));
+        let bm = cm - am;
 
         // Errors distributed proportionally
         let ae = ((ce as f64) * a_frac).round() as i32;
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #605, opened 2026-07-20
feat(fastq): UMI-in-read-name output and real BGZF for .gz paths

nh13:
Two gaps in `fgumi fastq`, both on the extract → align handoff.

## `-o out.fq.gz` wrote uncompressed text

The output path was opened with a bare `std::fs::File::create` and the file contained no gzip handling at all, so a `.gz` or `.bgz` output was plain FASTQ under a compressed name. Reproduced on a 20k-record extracted BAM:

```
$ fgumi fastq -i umi_subset.bam -o out.fq.gz     # before
$ file out.fq.gz
out.fq.gz: ASCII text
$ gzip -t out.fq.gz
gzip: out.fq.gz: not in gzip format
```

After, the same invocation yields `Blocked GNU Zip Format (BGZF; gzip compatible)`, `gzip -t` passes, and the decompressed bytes are identical to the previous plain-text output.

`.gz`/`.bgz` paths now route through a `BgzfFastqWriter` adapter over `InlineBgzfCompressor`, appending the BGZF EOF marker on finish so readers do not see a truncated stream. BGZF is gzip-compatible, so `zcat`/`gzip -d` read it while the output stays block-indexable. stdout deliberately stays uncompressed — that stream is meant to be piped into an aligner.

## UMI in the read name

`-a`/`--annotate-read-names` (`-U` alias for samtools users) appends the UMI from the first present tag in `--umi-tag` (default `RX,OX`) before any `/1` or `/2` suffix, with `--umi-name-delim` (default `:`) and `--umi-sep` (default `+`, replacing fgumi's stored `-` between duplex halves).

With the defaults this reproduces `samtools fastq -U` (`readname:AAA+CCC`) — the layout DRAGEN expects — replacing an external samtools or umi-tools step. Spot-checked against samtools on a real extracted BAM: `@SRR30485713.1:CTG+CCG/1`, with byte-identical sequence and quality lines.

A record carrying none of the configured tags is emitted unannotated rather than failing, since a BAM can legitimately mix UMI-bearing and UMI-free reads.

## Incidental refactors

`run_with_writer` now takes `&mut W` so `execute` can finalize the BGZF stream after the conversion loop returns, and the three per-record scratch buffers are grouped into a `FastqRecordBuffers` struct to keep the writer signature within the argument limit. Neither changes behaviour.

## Testing

Unit tests cover `is_gzip_output_path`, the `UmiNameAnnotation` delimiter/priority/fallback matrix, tag validation, and a `BgzfFastqWriter` round-trip asserting gzip magic, the EOF marker, and byte-exact decompression.

Two end-to-end tests drive `Fastq::execute` itself — one asserting the annotated read names (`readA:ACGT`, `readB:ACGT+TTTT`, and untouched names without `-a`), one asserting the `.gz` output is real BGZF that decompresses to the expected FASTQ. These exist because the unit tests alone left the `execute` dispatch arms and the UMI branch uncovered.

- `cargo ci-test`: 5553 passed, 22 skipped
- Local `cargo llvm-cov`: 98.4% of added lines covered (the two uncovered lines are the stdout arm, which cannot be driven in-process)

## Summary by CodeRabbit

* **New Features**
  * Added gzip/BGZF-compressed FASTQ output when using `.gz` or `.bgz` filenames.
  * Added optional UMI annotation for FASTQ read names with configurable SAM tags and name/separator formatting, including duplex handling.
* **Bug Fixes**
  * Improved compressed output correctness by ensuring proper BGZF/gzip-compatible stream formatting.
* **Tests**
  * Added integration coverage for UMI read-name annotation behavior and BGZF-compressed output compatibility/round-trip content matching.

### DIFF
--- a/src/lib/commands/fastq.rs
+++ b/src/lib/commands/fastq.rs
@@ -5,15 +5,17 @@
 
 use crate::commands::common::parse_bool;
 use crate::logging::OperationTimer;
+use crate::sam::SamTag;
 use crate::validation::validate_file_exists;
 use anyhow::Result;
 use clap::Parser;
 use fgumi_bam_io::create_raw_bam_reader;
 use fgumi_raw_bam::{
-    RawRecord, extract_sequence_into, quality_scores_slice, read_name as raw_read_name,
+    RawRecord, aux_data_slice, extract_sequence_into, find_string_tag, quality_scores_slice,
+    read_name as raw_read_name,
 };
 use log::{info, warn};
-use std::io::{BufWriter, Write, stdout};
+use std::io::{self, BufWriter, Write, stdout};
 use std::path::PathBuf;
 
 use crate::commands::command::Command;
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #606, opened 2026-07-21
perf(sort): narrow radix passes, reuse the stored queryname NUL, and borrow record bytes on ingest

nh13:
Three independent throughput improvements to the sort engine, ported from `feat-runall` and re-verified against `main`. Output is byte-identical in all three cases; the only behavior changes are bug fixes described below.

Suggested reading order is commit by commit — they touch different files and are independently reviewable.

## `perf(sort): build the natural queryname key from the NUL BAM already stores`

`extract_queryname_key` stripped the trailing NUL that BAM stores and re-appended one into a fresh `Vec` per record. BAM already stores the name NUL-terminated and `l_read_name` counts the terminator, so the stored bytes are directly usable.

The fast path applies only when the declared name is in bounds **and** the byte it ends on is genuinely a NUL. That second condition is a soundness requirement rather than a nicety: `RawQuerynameKey::cmp` passes `name.as_ptr()` to `natural_compare_nul`, which scans until it finds a terminator, so a key built from a record whose `l_read_name` does not land on one would read past the end of its own allocation. Anything failing either condition falls back to the previous strip-and-re-append path, which terminates unconditionally.

Tested with an rstest table checking the new extractor against the pre-optimization implementation as an independent oracle, across well-formed, empty-name, truncated, missing-terminator, non-NUL-terminated, and zero-length records.

## `perf(sort): size radix passes from the max mapped key, not the unmapped sentinel`

Mapped coordinate keys occupy ~5-6 bytes, but unmapped reads carry `u64::MAX`, which dragged `bytes_needed` to the full 8 and cost three wasted LSD passes over every record. Coordinate BAMs essentially always carry an unmapped tail, so this was paid on real input almost without exception.

Deriving the bound from the largest non-sentinel key needs no contract from the caller: `u64::MAX` is the largest `u64` and truncates to the all-`0xFF` maximum at every radix width, so unmapped records still sort to the tail and stay stable among themselves regardless of pass count.

Measured on an otherwise idle `c6a.4xlarge` over 25 contigs with a 5% unmapped tail, criterion 100 samples, confidence intervals within ±0.5%:

| records | before | after | speedup |
|---|---|---|---|
| 1M | 23.6 M/s | 29.0 M/s | 1.23x |
| 8M | 21.0 M/s | 25.0 M/s | 1.19x |

Radix is roughly a seventh of sort CPU, so this is a low-single-digit percentage of total sort time — worth having, but not a headline number.

An earlier revision of this change also tracked the bound incrementally across pushes so the sort could skip the scan entirely. That measured at a further 1.04x — about half a percent of total sort time — and is deliberately not kept: it cost a running field on `RecordBuffer`, three reset sites, an ordering hazard where the reset had to precede a macro that returns from inside itself, and a public entry point whose unchecked precondition silently mis-sorts when violated. The scan is three lines and carries no invariant.

**Bug fix included:** a derived bound of 0 does not imply the keys are all equal, so the sort is never skipped outright. A zero key can coexist with sentinels and the two do not compare equal — `PackedCoordinateKey::new` packs `tid = 0, pos = -1, reverse = false` to exactly 0, so a malformed record makes this reachable, and skipping would leave the sentinels ordered ahead of it. A single pass separates them and is a stable no-op when the keys genuinely are identical. Covered by a regression test with zero keys interleaved with sentinels above the radix threshold.

## `perf(sort): borrow record bytes from the decompressed block on ingest`

Pooled ingest copied each record twice after decompression: block into a `RawRecord`, then `RawRecord` into the sort arena. `PooledInputStream::next_record_borrowed` lends a slice straight out of the current decompressed block, falling back to a reusable scratch buffer only when the body or its 4-byte length prefix straddles a block boundary. Copy amplification goes from two to one on the coordinate and template-coordinate paths. The keyed/queryname path keeps owned records, which it requires.

`RecordSource` grows a matching `next_record_borrowed`. Both non-pooled variants already yield owned records and have nothing shared to borrow from, so each stores the record it just took and lends into it. That includes the `Stream` variant, which does not exist upstream — its producer-error handling is shared with the `Iterator` impl so a failure reaches `take_error()` identically on both paths. This matters because the ingest loops treat `Ok(None)` as end-of-input, so an error swallowed there would silently truncate a sort rather than fail it.

Tested with records and length prefixes straddling block boundaries down to 1-byte blocks, parity against the owned `read_raw_record` path, a proptest over randomized bodies and block sizes, and per-variant agreement plus mid-stream producer-error propagation.

## Verification

`cargo ci-test` (5573 tests), `cargo ci-fmt`, `cargo ci-lint`, and `RUSTDOCFLAGS="-D warnings" cargo ci-doc` all pass. Patch coverage is 100% of changed lines. Each new test was checked for non-vacuity by breaking the corresponding fix and confirming the test fails.

## Summary by CodeRabbit

- **Performance**
  - Improved coordinate and template sorting efficiency by sorting directly from borrowed BAM bytes.
  - Optimized radix sorting by ignoring unmapped sentinel keys for radix pass sizing; consistent behavior across chunked/parallel sorting.
  - Added coordinate radix sort benchmarks for multiple radix strategies and dataset sizes.
- **Bug Fixes**
  - Hardened query name key extraction to properly handle malformed or truncated BAM records while keeping names NUL-terminated.
  - Fixed sorting ingest when decompressed records span decompression block boundaries.
- **Documentation**
  - Updated high-performance “unsafe” documentation to clarify the coordinate radix-sorting hot path.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -165,7 +165,8 @@ because the hot path runs once per record (BAM workloads are millions to billion
 of records) and a safe rewrite measurably regresses sort throughput.
 
 - **`crates/fgumi-sort/src/inline.rs`** — three `#[allow(unsafe_code)]` regions:
-  the `radix_sort_record_refs` (coordinate) and `radix_sort_template_refs` /
+  the `radix_sort_record_refs_with_max` (coordinate; `radix_sort_record_refs`
+  delegates to it after scanning for the max) and `radix_sort_template_refs` /
   `radix_sort_template_field` (template) LSD radix sorts use `Vec::set_len` to
   skip per-element initialization on the auxiliary scratch buffer, plus
   raw-pointer slice swaps to avoid double-borrow restrictions across the
--- a/benches/core_functions.rs
+++ b/benches/core_functions.rs
@@ -1305,6 +1305,106 @@ fn bench_queryname_sort_strategies(c: &mut Criterion) {
     group.finish();
 }
 
+/// Coordinate-sort radix throughput across the three ways `bytes_needed` can be
+/// sized, isolating the two independent wins:
+///
+/// - `full_width` reproduces the pre-optimization behavior by passing the
+///   unmapped sentinel as the bound, forcing all 8 radix passes — which is what
+///   a sentinel-inclusive max scan yields the moment a single unmapped read is
+///   present, i.e. on essentially every real coordinate-sorted BAM.
+/// - `scan_mapped_max` scans for the largest *mapped* key, sizing the passes to
+///   the ~4–5 bytes such a key occupies. The delta against `full_width` is the
+///   pass-count win.
+/// - `tracked_max` supplies that same bound without scanning, using a value the
+///   benchmark precomputed while building the input. The delta against
+///   `scan_mapped_max` is the scan-elimination win, and it is small enough that
+///   `RecordBuffer` deliberately does not chase it: it scans once per sort
+///   rather than maintaining a running bound across pushes.
+///
+/// `sort_unstable` is a comparison-sort baseline. Note it is *unstable*, so it
+/// is not a drop-in substitute — coordinate sort order must be stable to match
+/// `samtools sort`. The input mixes 25 tids with a 5% unmapped tail to exercise
+/// the sentinel handling.
+fn bench_coordinate_radix_sort(c: &mut Criterion) {
+    use fgumi_sort::{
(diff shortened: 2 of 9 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #607, opened 2026-07-21
fix(sort): write the BAI sidecar to the samtools path, not an extension-replaced one

nh13:
`Path::with_extension("bam.bai")` replaces everything after the final `.` rather than appending, so it produced the correct sidecar name only for outputs already ending in `.bam`. For anything else it rewrote the basename itself:

```
fgumi sort --write-index -o out          wrote out.bam.bai     (indexed readers look for out.bai)
fgumi sort --write-index -o out.sorted   wrote out.bam.bai     (indexed readers look for out.sorted.bai)
```

So the index was invisible to every indexed reader. Worse, distinct outputs collapse onto a single index path: sorting `out.sorted` and then `out.other` in the same directory silently overwrites the first run's index with the second's, leaving a stale index that still reads as structurally valid.

## The fix

Adds `bai_sidecar_path` to `fgumi-bam-io`, which appends `.bai` to the full path per the samtools convention, plus `write_bai_sidecar` for the index-then-write pattern the review command open-coded in several places. Callers that already hold an in-memory index — the indexing sort writer — pair `bai_sidecar_path` with the existing `write_bai_index` instead, so nothing re-reads a BAM it just finished writing.

The read side is updated in lockstep. `Review::read_bam_index` derived its primary lookup with the same broken expression, so a BAM not named `*.bam` would have been indexed at one path and searched for at another. Its extension-replaced fallback is retained, since other tools do emit indexes under that name.

## Why the existing test did not catch it

Behavior is unchanged for outputs ending in `.bam`, and the existing sort test used `output.bam`. It also asserted:

```rust
assert!(bai.exists() || output.with_extension("bai").exists(), "index file should exist");
```

That either-naming check accommodates the bug rather than pinning the contract — it passes whichever path the index lands at.

It is now an rstest table over `.bam`, non-`.bam`, and extensionless outputs that asserts the *exact* sidecar path, and additionally asserts no index appears at the extension-replaced path. A sibling test covers the non-spilling in-memory sort path, which derives the sidecar path at a second call site that the spilling test never reaches.

The `review` integration fixture now builds its index through the same helper the command reads with, so the fixture cannot drift into agreeing with a regression.

## Verification

`cargo ci-test` (5545 tests), `cargo ci-fmt`, `cargo ci-lint`, and `RUSTDOCFLAGS="-D warnings" cargo ci-doc` all pass. Patch coverage is 100% of changed lines. Each new assertion was checked for non-vacuity by reverting the corresponding call site and confirming the matching case fails while the `.bam` case keeps passing.

## Summary by CodeRabbit

- **New Features**
  - Added support for samtools-style BAM BAI sidecar naming using the `.bai` suffix.
  - Added a helper to create and write BAM BAI sidecars and return their index paths.

- **Bug Fixes**
  - Corrected BAI sidecar path handling for BAM filenames that don’t end with `.bam`.
  - Ensured outputs use the preferred sidecar location, while still falling back to legacy locations only when needed.

- **Tests**
  - Expanded unit and integration coverage to verify the exact BAI sidecar paths written and to confirm legacy locations are not used.

### DIFF
--- a/crates/fgumi-bam-io/src/lib.rs
+++ b/crates/fgumi-bam-io/src/lib.rs
@@ -33,7 +33,7 @@ pub use reader::{
 };
 pub use reorder::{DrainReady, ReorderBuffer};
 pub use writer::{
-    BamWriter, BgzfWriterEnum, IndexingBamWriter, RawBamWriter, create_bam_writer,
-    create_indexing_bam_writer, create_optional_bam_writer, create_raw_bam_writer, write_bai_index,
-    write_bam_header,
+    BamWriter, BgzfWriterEnum, IndexingBamWriter, RawBamWriter, bai_sidecar_path,
+    create_bam_writer, create_indexing_bam_writer, create_optional_bam_writer,
+    create_raw_bam_writer, write_bai_index, write_bai_sidecar, write_bam_header,
 };
--- a/crates/fgumi-bam-io/src/writer.rs
+++ b/crates/fgumi-bam-io/src/writer.rs
@@ -1,7 +1,10 @@
 //! BAM writer factories and related types.
 //!
 //! This module provides writer factories for BAM output, including standard BAM writers,
-//! raw-bytes BAM writers, and an incremental-indexing BAM writer.
+//! raw-bytes BAM writers, an incremental-indexing BAM writer, and BAI sidecar
+//! helpers (`bai_sidecar_path` for the naming convention, `write_bai_index` for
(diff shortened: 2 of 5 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai full review

nh13:
@coderabbitai full review

nh13:
@coderabbitai full review

nh13:
@coderabbitai full review

PR #608, opened 2026-07-21
feat(sort): add per-phase --sort-threads / --merge-threads

nh13:
`fgumi sort` had a single `-@/--threads` sizing both Phase 1 (accumulate/sort/spill) and Phase 2 (merge/write). Those phases contend with different things: Phase 1 competes with whatever feeds the sort, while Phase 2 runs after the producer is finished. A user running `bwa mem -t 32 ... | fgumi sort -@ 8` had no way to cede cores to the aligner during ingest while keeping the merge wide.

## Engine

`SortWorkerPool` gains an `active_worker_limit` and `set_active_workers`. A single pool still spans both phases — sized to the wider of the two counts — and the driver flips the active cap at the ingest/merge boundary. Workers above the cap idle rather than taking new work and re-check the limit within `MAX_BACKOFF_US`, so raising it reactivates them without an explicit wake.

The invariant that makes this safe: a capped worker still drains items it *already holds* and only declines to acquire new work. Held items are per-worker, so a worker that froze while holding work would strand that output with no other worker able to advance it.

`RawExternalSorter` gains `sort_threads`/`merge_threads` builders plus `phase1_threads()`/`phase2_threads()`, each falling back to `threads` independently. The CLI exposes `--sort-threads` / `--merge-threads` with the same defaulting.

This is purely a scheduling knob: output is byte-identical for any split.

## On testing a knob that by definition changes nothing observable

An end-to-end run *cannot* tell a correctly wired flag from one that was parsed and then silently dropped — both produce identical bytes. So the sorter construction is extracted from `execute` into `Sort::build_sorter`, and the wiring is asserted directly on the resolved per-phase counts. I verified that test is non-vacuous by unwiring the flag and confirming it fails; the output-identity test, as expected, does not.

Both properties are covered separately:

- **Wiring** — `build_sorter` resolves each override, and each falls back to `--threads` independently.
- **Output identity** — four asymmetric splits produce byte-identical output to a plain `--threads` run over a spilling multi-chunk merge, exercising both the wider-Phase-1 and wider-Phase-2 pool sizings.
- **Pool behavior** — clamping of the active count, and (on a single pool, the only way to show it) that workers idled by a cap come back online when the cap is raised.

Runall-only halves of the upstream change — `SortSpillDecompress::with_max_concurrency` and the chains builder — are deliberately not ported; they depend on crates that do not exist on `main`.

## Verification

`cargo ci-test` (5557 tests), `cargo ci-fmt`, `cargo ci-lint`, and `RUSTDOCFLAGS="-D warnings" cargo ci-doc` pass. Patch coverage 96%.

## Summary by CodeRabbit

- **New Features**
  - Added independent thread controls for the sorter’s phase 1 (sort/accumulate/spill) and phase 2 (merge/output).
  - Introduced `--sort-threads` and `--merge-threads` for `fgumi sort`, with safe clamping.
- **Bug Fixes**
  - Ensured varying phase thread settings do not change output bytes.
- **Tests**
  - Added coverage for CLI parsing, phase override fallback/clamping, worker cap enforcement/reactivation, and output consistency.

### DIFF
--- a/crates/fgumi-sort/src/external.rs
+++ b/crates/fgumi-sort/src/external.rs
@@ -1494,8 +1494,15 @@ pub struct RawExternalSorter {
     /// distributed across them in free-space-aware round-robin order via
     /// [`TmpDirAllocator`].
     temp_dirs: Vec<PathBuf>,
-    /// Number of threads for parallel operations.
+    /// Number of threads for parallel operations. Used by both phases unless
+    /// overridden by `sort_threads` / `merge_threads`.
     threads: usize,
+    /// Phase-1 (accumulate/sort/spill) worker count. `None` means "use
+    /// `threads`". Lets a caller cede cores to an upstream producer during
+    /// ingest while keeping the merge wide.
+    sort_threads: Option<usize>,
+    /// Phase-2 (merge/write) worker count. `None` means "use `threads`".
+    merge_threads: Option<usize>,
     /// Compression level for output.
     output_compression: u32,
     /// Compression level for temporary chunk files (0 = uncompressed).
@@ -1568,6 +1575,8 @@ impl RawExternalSorter {
             memory_limit: 512 * 1024 * 1024, // 512 MB default
             temp_dirs: Vec::new(),
             threads: 1,
+            sort_threads: None,
+            merge_threads: None,
             output_compression: 6,
             temp_compression: 1, // Default: fast compression
             spill_codec: crate::codec::SpillCodec::default(),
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed the outside-diff finding: the `--write-index` path now honors `--merge-threads`. Both indexed writers use `phase2_threads()` instead of the base `threads` — `create_indexing_bam_writer` in the in-memory branch of `sort_coordinate_with_index`, and `writer_threads` in `merge_chunks_with_index` (which feeds all three `create_indexing_bam_writer` calls there). This matches the non-indexed `merge_chunks_generic`, whose `PooledBamWriter` already inherits the cap via the shared pool.

Also closed the coverage gap: `test_per_phase_threads_produce_identical_output` now crosses in `write_index` (false/true) and `memory_limit` (spill/in-memory) via `#[values]`, so both indexed writer sites are exercised across every thread split. Output stays byte-identical (compression thread count does not change block content), so the test guards the path against regression rather than asserting the thread count.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #609, opened 2026-07-21
fix(sort): surface worker panics mid-run and fail closed on a dead permit pool

nh13:
Two independent robustness gaps in the sort worker pool, both on failure paths. Neither affects output on a successful run.

## A mid-run worker panic could hang instead of erroring

`worker_panicked` was only set by `do_shutdown` when a join reported a panic — that is, *after* Phase 2 had already finished. A worker that panicked while holding a reader mutex left `reader.eof` unset, and because every `try_lock` site collapses `TryLockError::Poisoned` into "contended, retry", the merge loop kept spinning while the main thread waited for work that would never arrive. A diagnosable panic became a silent 100%-CPU hang with no error message.

A `WorkerPanicGuard` now publishes the flag and unparks the main thread as the worker unwinds, so the failure surfaces immediately; the normal exit path disarms it. This is cheaper than reimplementing poison handling at all three `try_lock` sites, and it catches any worker panic rather than only the held-mutex case.

## `PermitPool::acquire` drained buffered permits after close

`close` dropped the `Sender`, but `recv` keeps succeeding until the channel's buffered permits are exhausted. So after the I/O writer errored — ENOSPC on the spill directory, say — producers each took one more permit and compressed another block, up to `num_workers * 4` blocks of wasted work, before the disconnect finally surfaced.

This was never a hang and never produced wrong output; it was work thrown away on the way to failing. `acquire` now checks a closed flag before taking a permit, and `close` sets that flag before dropping the sender so any producer that sees the disconnect also sees the closed state.

The `arc_swap` lock-free `release` from the same upstream commit is deliberately not ported — adding a dependency to micro-optimize an error path needs its own justification.

## Tests

Acquiring past close with permits still buffered, release-after-close being inert, normal permit reuse, the panic flag being set by the unwinding worker rather than deferred to join, and a disarmed guard staying quiet on clean exit.

The panic test waits on a bounded deadline rather than blocking, so a regression fails in seconds instead of hanging CI — I confirmed that by reverting the fix and watching it fail at the bound. Both fixes were verified non-vacuous the same way.

`cargo ci-test` (5539 tests), `cargo ci-fmt`, `cargo ci-lint`, and `RUSTDOCFLAGS="-D warnings" cargo ci-doc` pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shutdown behavior so tasks no longer continue or receive stale permits after shutdown begins.
  * Enhanced worker failure handling by immediately detecting mid-run panics and stopping the processing loop instead of waiting.
* **Tests**
  * Added coverage for the updated shutdown/permit behavior and for worker panic detection, including verifying normal exits don’t trigger the panic flag.

### DIFF
--- a/crates/fgumi-sort/src/worker_pool.rs
+++ b/crates/fgumi-sort/src/worker_pool.rs
@@ -437,6 +437,16 @@ impl Clone for BufferPool {
 pub(crate) struct PermitPool {
     tx: std::sync::Mutex<Option<Sender<()>>>,
     rx: Receiver<()>,
+    /// Set by [`PermitPool::close`] so `acquire` fails immediately instead of
+    /// draining permits that are still buffered in the channel.
+    ///
+    /// Dropping the `Sender` alone is not enough: `recv` keeps succeeding until
+    /// the buffered permits run out, so after the I/O writer errors, producers
+    /// would each take one more permit and compress another block -- up to
+    /// `num_workers * 4` blocks of work thrown away -- before the disconnect
+    /// finally surfaced. This never produced wrong output or a hang; it just
+    /// wasted work on the failure path.
+    closed: AtomicBool,
 }
 
 impl PermitPool {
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #610, opened 2026-07-21
fix(group): do not count a failed UMI-assignment group as accepted

nh13:
When `assign_umi_groups_impl` errors for a position group, `group` warns, emits zero records for that group, and continues. But both execution paths had already tallied that group's templates as accepted, so the grouping metrics file and the run summary reported more accepted templates than the output BAM actually contained.

Both paths are affected, and they fail differently:

- **Single-threaded** merges the group's filter metrics *before* assignment runs, and the error arm returned without rolling that back.
- **Threaded** returns an empty `ProcessedPositionGroup` with `filter_metrics` intact, so the caller merges the same over-count.

The single-threaded path now subtracts the contribution it recorded; the threaded early return zeroes the accepted count to match the zero records it emits.

## Reachability

The trigger is data-driven rather than hypothetical: `--strategy paired` requires the RX tag to carry two `-`-delimited segments, and a single-segment UMI passes every filter — non-PF, mapping quality, Ns, minimum length — before failing in the assigner.

## Testing

End-to-end on both paths, asserting the reported accepted count against **the number of records actually written**, not against a literal. `--allow-unmapped` forces the threaded path via `should_use_parallel`, whose empty-group early return is a different code path from the single-threaded one.

Verified non-vacuous by reverting each fix in turn: both then report 8 accepted against 0 records written.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected grouping metrics so accepted-record counts remain accurate when UMI assignment fails.
  * Ensured failed assignments do not report records that were not written to the output BAM.
  * Applied the fix consistently to both threaded and single-threaded processing.

### DIFF
--- a/src/lib/commands/group.rs
+++ b/src/lib/commands/group.rs
@@ -1257,6 +1257,12 @@ impl Command for GroupReadsByUmi {
                     no_umi,
                 ) {
                     log::warn!("UMI assignment failed, returning empty group: {e}");
+                    // The caller merges these metrics, so zero the accepted count
+                    // to match the zero records this group contributes. Otherwise
+                    // the grouping metrics report more accepted templates than the
+                    // output BAM holds.
+                    let mut filter_metrics = filter_metrics;
+                    filter_metrics.accepted_templates = 0;
                     #[cfg(feature = "memory-debug")]
                     if debug_memory_flag
                         && let Some(stats) = stats_for_tracking.as_ref() {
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #611, opened 2026-07-21
fix(consensus): error on an over-long consensus read name instead of panicking

nh13:
Consensus read names are `<prefix>:<UMI>`, and the UMI is read from the input BAM's `MI`/`RX` tag — so its length is data-controlled. `UnmappedSamBuilder` asserted the BAM name limit, which meant a long or malformed UMI aborted `fgumi duplex`, `fgumi codec`, or `fgumi simplex` with a Rust panic **partway through writing the output BAM**, leaving a truncated file instead of failing cleanly.

Adds `UnmappedSamBuilder::try_build_record` and a `ReadNameTooLong` error, and routes all three consensus callers through it. `build_record` keeps the assertion for callers that supply their own names — tests, fixtures — and now documents that it panics and points at the fallible variant.

## Propagation

`build_consensus_record_into` (vanilla) becomes fallible; its caller already returns `Result`, so this propagates without restructuring.

The duplex and codec builders already returned `Result` — but only decoratively, as their own attributes recorded:

```rust
#[expect(clippy::unnecessary_wraps,
         reason = "Result return type kept for API consistency with other callers")]
```

Those expectations are now *unfulfilled*, because the functions can genuinely fail. Clippy flagged them, which is a nice independent confirmation that the `Result` became real rather than ornamental. Both are removed.

## Boundary

BAM stores `l_read_name` in a single byte including the NUL terminator, so the limit is 254. With the 10-byte `consensus:` prefix that puts the exact boundary at a 244-byte UMI accepted and 245 rejected — pinned as test cases rather than described.

Covered at the builder level (at-limit, one-over, far-over, empty, plus byte-identical output against `build_record` for valid names) and end-to-end through the caller. Reverting the fix makes the end-to-end cases panic inside the builder, which is precisely the reported failure mode.

`cargo ci-test` (5542 tests), `cargo ci-fmt`, `cargo ci-lint`, `RUSTDOCFLAGS="-D warnings" cargo ci-doc` pass.

## Summary by CodeRabbit

* **New Features**
  * Added a fallible unmapped BAM record builder API that rejects overlong read names with a dedicated `ReadNameTooLong` error.

* **Bug Fixes**
  * Record construction now fails gracefully when read names are invalid/too long, avoiding panics and preventing malformed output.
  * Consensus and duplex/codec writing paths now propagate record-construction errors instead of assuming success.

* **Tests**
  * Added tests covering rejection and error messaging for overlong UMI-derived read names, plus byte-identical output for valid boundary cases.

### DIFF
--- a/crates/fgumi-consensus/src/codec_caller.rs
+++ b/crates/fgumi-consensus/src/codec_caller.rs
@@ -1341,10 +1341,11 @@ impl CodecConsensusCaller {
     }
 
     /// Builds the output record from the consensus and writes raw bytes into `output`.
-    #[expect(
-        clippy::unnecessary_wraps,
-        reason = "Result return type kept for API consistency with other callers"
-    )]
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the `<prefix>:<UMI>` read name exceeds BAM's
+    /// 254-byte limit, which an over-long input-derived `umi` can cause.
     #[expect(
         clippy::too_many_arguments,
         reason = "all_records needed separately from source_raws for UMI consensus"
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #612, opened 2026-07-21
fix(sam): strip stale NM/UQ/MD when a mapped record has no reference

nh13:
`regenerate_alignment_tags_raw` returned `Ok(false)` on a negative reference id without touching the record's tags, so a record flagged mapped but carrying `ref_id < 0` kept whatever NM/UQ/MD it arrived with.

Those values describe an alignment against a reference the record does not name, and nothing downstream can distinguish a stale tag from a correct one — a wrong NM is worse than an absent one.

Such a record is malformed and there is nothing to recompute against, so this now fails closed and strips the tags, matching both the unmapped branch immediately above it and fgbio's `regenerateNmUqMdTags`.

Only reachable on malformed input.

## Testing

Plants NM/UQ/MD on a valid mapped record, clears the reference id, and asserts all three are gone. The test also asserts the record is *still flagged mapped* — otherwise the unmapped branch above would handle it and the negative-`ref_id` path would never be exercised, leaving the test green for the wrong reason.

Verified non-vacuous by reverting the fix, which leaves NM in place.

`cargo ci-test` (5535 tests), `cargo ci-fmt`, `cargo ci-lint`, `RUSTDOCFLAGS="-D warnings" cargo ci-doc` pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected handling of alignment records marked as mapped but containing an invalid (negative) reference ID.
  * Stale `NM`, `UQ`, and `MD` tags are now stripped from these malformed records to prevent outdated alignment metadata from being retained.
* **Tests**
  * Added a unit test covering the stale-tag removal behavior for records with negative reference IDs.

### DIFF
--- a/crates/fgumi-sam/src/alignment_tags.rs
+++ b/crates/fgumi-sam/src/alignment_tags.rs
@@ -236,9 +236,12 @@ use fgumi_raw_bam::{self, RawRecordView, RawTagsEditor};
 /// Regenerates NM, UQ, and MD tags for a raw BAM record after base masking.
 ///
 /// For unmapped reads, the tags are removed. For mapped reads, the tags are
-/// recalculated based on the alignment and reference.
+/// recalculated based on the alignment and reference. A record flagged mapped
+/// but carrying a negative reference id is malformed and has nothing to
+/// recompute against, so its tags are removed as well rather than left stale.
 ///
-/// Returns `Ok(true)` if tags were regenerated, `Ok(false)` for unmapped reads.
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #614, opened 2026-07-21
perf: remove three per-block and per-read allocations

nh13:
Three independent allocation removals on hot paths, grouped because none justifies its own PR. **No measured headline numbers** — these are cheap and safe rather than significant, and I have not benchmarked them.

## BGZF multithreaded writer

Each deflater returned `buffer.clone()` — a full memcpy of up to ~64 KiB on every output block, kept only so the scratch buffer stayed warm for the next call. `mem::replace` with a fresh buffer pre-sized to the frame just produced keeps it warm without the copy.

A plain `mem::take` would *regress* this: it leaves a zero-capacity `Vec` behind, so the next frame reallocates from scratch.

## BGZF reader

`read_raw_block` allocated with `vec![0u8; block_size]` — a per-block memset of up to 64 KiB that is immediately overwritten. It now reserves exact capacity and fills via a bounded `read_to_end`.

This one has a sharp edge worth calling out. Unlike `read_exact`, **`read_to_end` returns `Ok` on early EOF**, so the explicit length check is load-bearing: without it a truncated BAM short-reads silently instead of erroring. The added test demonstrates that directly — with the check removed it returns `Ok(Some(block))` carrying a short block rather than `UnexpectedEof`.

## Duplex consensus

`process_group` deep-cloned every alignment-surviving `SourceRead` — bases, quals, and two cigar `Vec`s each — through four `iter().filter().cloned().collect()` passes. The two halves of each split are exact complements on `FIRST_SEGMENT`, so `into_iter().partition(..)` yields the same groups in the same order without copying.

Binding order is the hazard: `partition` returns (predicate true, predicate false), so swapping either pair silently transposes the strand groups. I checked that existing duplex tests catch that by deliberately transposing them — 4 tests fail, including `test_duplex_ba_only_methylation_tags_use_bottom_strand`.

## Summary by CodeRabbit

- **Performance**
  - Reduced allocations and eliminated per-block copying during compressed block handling.
  - Optimized raw BGZF block reading and improved efficiency when splitting duplex alignment results.
- **Bug Fixes**
  - More reliably distinguishes clean EOF from truncated BGZF blocks and reports truncation errors.
- **Tests**
  - Expanded oracle coverage for duplex strand splitting and added cases for truncated blocks and partial header scenarios.
- **Chores**
  - Added a dev-only testing dependency and refreshed related test documentation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -726,6 +726,7 @@ dependencies = [
  "crc32fast",
  "criterion",
  "libdeflater",
+ "rstest",
 ]
 
 [[package]]
--- a/crates/fgumi-bam-io/src/vendored/bgzf_multithreaded.rs
+++ b/crates/fgumi-bam-io/src/vendored/bgzf_multithreaded.rs
@@ -482,7 +482,15 @@ fn compress_block(
         buffer[buffer.len() - 5],
     ]);
 
-    Ok((buffer.clone(), crc32, input.len()))
+    // Hand the compressed frame off instead of copying it: `buffer.clone()` was a
+    // full memcpy of up to ~64 KiB on every output block, kept only so the
+    // scratch buffer stayed warm for the next call. Swapping in a fresh buffer
+    // pre-sized to what this frame needed keeps it warm without the copy.
+    //
+    // `mem::take` would be wrong here: it leaves a zero-capacity Vec behind, so
+    // the next frame reallocates from scratch and the optimization regresses.
+    let frame = std::mem::replace(buffer, Vec::with_capacity(buffer.len()));
+    Ok((frame, crc32, input.len()))
 }
 
 fn spawn_writer<W>(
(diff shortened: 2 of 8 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #615, opened 2026-07-21
fix(cli): accept EB/EiB memory sizes instead of rejecting them as scientific notation

nh13:
Memory-size parsing rejected any input containing `e` or `E` as scientific notation before handing off to `bytesize`. But `EB` and `EiB` are the only supported units that contain an `E`, so:

```
fgumi sort --max-memory 1EB
  → Scientific notation not supported: '1EB'. Use integer values or human-readable formats like '2GB'.
```

The check now fires only on a genuine mantissa-and-exponent — a digit or `.`, then `e`/`E`, then an optional sign and at least one digit. `1EB` and `1EiB` parse; `1e5`, `1.5e10`, `2E+3`, and `2e-3` are still rejected with the same message.

Admittedly an exabyte memory limit is not a realistic setting, but the failure is a confusing message for a documented unit rather than anything subtle.

Tested through `parse_memory_size` for both accepted and rejected shapes, plus a table over the detector itself covering the boundaries it has to get right: a leading `e`, a trailing `e`, a sign with no digit after it, and a non-numeric character before the `e`.

`cargo ci-test` (5555 tests), `cargo ci-fmt`, `cargo ci-lint`, `RUSTDOCFLAGS="-D warnings" cargo ci-doc` pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Exabyte units such as `EB` and `EiB` are now accepted in supported casing variations.
  * Memory-size validation more accurately distinguishes scientific notation from valid unit names.
  * Inputs using true scientific notation, such as `1e5` or `2E+3`, continue to be rejected.

### DIFF
--- a/src/lib/validation.rs
+++ b/src/lib/validation.rs
@@ -189,6 +189,34 @@ pub fn validate_positive<T: Ord + Display + Default>(value: T, name: &str) -> Re
     Ok(())
 }
 
+/// Does `s` contain a genuine scientific-notation exponent?
+///
+/// True only for a digit or `.` followed by `e`/`E` and then an optional sign
+/// and at least one digit -- `1e5`, `1.5e10`, `2E+3`. A bare `e`/`E` test would
+/// also match the exabyte units `EB` and `EiB`, which are the only units
+/// `bytesize` accepts that contain an `E`, so `1EB` would be rejected as
+/// scientific notation.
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #617, opened 2026-07-21
fix(zipper): name the tag-skipping flag after the tag it actually writes

nh13:
The flag was `--skip-pa-tags` and its help said it skipped "`pa` (primary alignment) tags" — but the struct field is `skip_tc_tags` and the command writes `tc` (template coordinate). No `pa` tag has ever been written, so a user who grepped the output for the thing the flag named found nothing.

Renamed to `--skip-tc-tags`, with `--skip-pa-tags` **kept as an alias** rather than renamed outright, so existing scripts keep working. Repo precedent for that is the `--queue-memory*` aliases. Help text and the internal step comment now say `tc`.

One detail worth flagging: the tag bytes are lowercase **`tc`**, not `TC` — `SamTag::TC` is the Rust constant's name, not the tag's spelling (`assert_eq!(*SamTag::TC, [b't', b'c'])`). The help refers to it as `tc` accordingly, since that is what a user would actually grep for.

The existing parsing test gains cases for the new spelling alongside the retained alias, covering the bare flag, explicit `true`/`false`, and the `=` form for both — 13 cases total.

`cargo ci-test` (5539 tests), `cargo ci-fmt`, `cargo ci-lint`, `RUSTDOCFLAGS="-D warnings" cargo ci-doc` pass.

## Summary by CodeRabbit

* **New Features**
  * Added the `--skip-tc-tags` command-line option for controlling template-coordinate tags on supplementary reads.
  * Retained `--skip-pa-tags` as a backward-compatible alias.

* **Documentation**
  * Updated command help text to accurately describe template-coordinate tag behavior.

### DIFF
--- a/src/lib/commands/zipper.rs
+++ b/src/lib/commands/zipper.rs
@@ -182,11 +182,15 @@ pub struct Zipper {
     #[arg(long = "exclude-missing-reads", default_value = "false", num_args = 0..=1, default_missing_value = "true", action = clap::ArgAction::Set, value_parser = parse_bool)]
     pub exclude_missing_reads: bool,
 
-    /// Skip adding `pa` (primary alignment) tags to secondary/supplementary reads.
-    /// By default, zipper adds a `pa` tag containing the primary alignment's template
+    /// Skip adding `tc` (template coordinate) tags to secondary/supplementary reads.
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #618, opened 2026-07-21
test(raw-bam): pin the aux-tag walker against every BAM type code

nh13:
`find_i16_array_tag_in_aux` backs the test-only per-base tag accessors, so its correctness is load-bearing in an indirect and easy-to-miss way. If it advanced by the wrong width for any type code it would desynchronize, start reading payload bytes as tag identifiers, and return the wrong array — meaning **real per-base tag assertions could silently pass against the wrong bytes**. A broken oracle is worse than a missing one.

This was carried on the port list as a suspected mis-walk past the `A`, `C`, `S`, `I`, and `H` type codes.

**I investigated and it is not a mis-walk** — every type code advances by the correct width today, which I confirmed by testing each one rather than by reading the match arms. No production change here.

What was true is that *nothing pinned it*, which is presumably what made the claim plausible in the first place. This adds:

- a case placing a tag of each scalar type code (`A`, `c`, `C`, `s`, `S`, `i`, `I`, `f`, `Z`, `H`) in front of a target array
- a case for `B` arrays of each element type (`c`, `C`, `s`, `S`, `i`, `I`, `f`)
- a same-subtype array whose tag does *not* match, since the `B:s` branch returns early only on a tag match
- an absent tag

Verified non-vacuous by deliberately mis-sizing two widths (`A`/`c`/`C` to 2 bytes, `B:s` elements to 1), which fails six cases.

`cargo ci-test`, `cargo ci-fmt`, `cargo ci-lint`, `RUSTDOCFLAGS="-D warnings" cargo ci-doc` pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parsing of auxiliary BAM `B` array tags to safely handle malformed, truncated, or unknown subtype encodings, preventing unsafe reads and ensuring only correctly matching `B:s` arrays are returned.

* **Tests**
  * Added targeted test coverage for finding `B:s` int16 arrays within auxiliary data, including correct skipping across intervening tags and non-matching tag scenarios.
  * Added fail-closed and fuzz/property-based coverage for truncations, malformed scalar encodings, partial trailing tags, and invalid/overlong array counts.

### DIFF
--- a/crates/fgumi-raw-bam/src/testutil.rs
+++ b/crates/fgumi-raw-bam/src/testutil.rs
@@ -131,25 +131,31 @@ fn find_i16_array_tag_in_aux(aux: &[u8], tag: [u8; 2]) -> Option<Vec<i16>> {
         i += 3;
         match typ {
             b'B' => {
-                let sub = aux[i];
-                i += 1;
+                // Fail closed on a truncated or unrecognized array rather than
+                // indexing past the end, mirroring `tags::parse_array_tag_at`.
+                let header = aux.get(i..i + 5)?;
+                let sub = header[0];
                 let count =
-                    u32::from_le_bytes([aux[i], aux[i + 1], aux[i + 2], aux[i + 3]]) as usize;
-                i += 4;
-                if t == tag && sub == b's' {
-                    let mut vals = Vec::with_capacity(count);
-                    for _ in 0..count {
-                        vals.push(i16::from_le_bytes([aux[i], aux[i + 1]]));
-                        i += 2;
-                    }
-                    return Some(vals);
-                }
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #620, opened 2026-07-21
feat(group): accept query-grouped unaligned input with --allow-unmapped

nh13:
`group` required template-coordinate sort unconditionally. So a user with an unaligned BAM who passed `--allow-unmapped` — whose entire purpose, per the command's own log line, is "All unmapped reads will form a single position group per library/cell" — was still told:

```
Input BAM must be template-coordinate sorted ...
  fgumi sort -i input.bam -o sorted.bam --order template-coordinate
```

on data that has no coordinates. That blocked the unaligned-only workflow (ribosome display, for instance) with advice the user cannot act on.

## What changed

Template-coordinate sort exists to make position adjacency match the grouping key. Under `--allow-unmapped` every unmapped record lands in one group whatever the order, so that guarantee buys nothing — and the requirement is dropped.

**Query grouping is still required, for a different reason.** Templates are assembled by comparing each record's name to the previous one (`group_by_name_and_build`), so without adjacency an R1/R2 pair splits into two partial templates. That is corruption rather than a policy choice, so it stays enforced — and its error points at `--order queryname`, since the template-coordinate remediation is useless for a file with no coordinates.

**Whether the header declares `@SQ` is deliberately not part of the decision.** An earlier revision gated on their absence; that would reject an all-unmapped BAM that kept its header through something like `samtools view -f 4`, for no benefit. Mapped reads in query-grouped input are simply not position-adjacent, so each mapped template forms its own family — that *under*-groups rather than merging unrelated molecules, and `--allow-unmapped` already warns loudly that it groups by UMI alone. A further warning names the case explicitly when the header declares references.

Note there is no fgbio precedent to follow here: fgbio has no `--allow-unmapped` at all, so this is fgumi-specific semantics.

## What did not change

Without `--allow-unmapped`, template-coordinate sort is still required — including for extract-style `SO:unsorted GO:query` headers with no `SS`. That case is the regression test for a real earlier bug where such headers were silently mistaken for template-coordinate, and it keeps failing without the flag. Accepting it *under* the flag routes it to the unmapped-only grouping path deliberately, which is a different thing from mistaking it for a sorted file.

The pre-existing rejection test previously asserted rejection under both flag settings; it is now scoped to the no-flag contract, with the flag contract covered by the new test. Both are explicit rather than one being weakened.

## Verification

End-to-end with the real binary, not just unit tests:

- `fgumi extract` output (`SO:unsorted GO:query`, no `@SQ`) → 12/12 records out, 2 MI groups matching the 2 distinct UMIs
- all-unmapped BAM that **does** declare `@SQ` → same result, plus the warning
- same input without `--allow-unmapped` → still rejected

## Summary by CodeRabbit

* **New Features**
  * `--allow-unmapped` can now handle query-grouped inputs even when reads are not adjacent by template-coordinate sorting.
  * Updated error/remediation guidance to point users to the appropriate `--order` setting for each input type.
* **Bug Fixes**
  * Maintains strict template-coordinate sorting behavior when `--allow-unmapped` is not enabled.
  * Adds clearer warnings when reference sequences are present but inputs aren’t template-coordinate sorted.
* **Tests**
  * Expanded integration coverage for accepted/rejected query-grouped unaligned cases, including with/without reference sequences.
* **Style / Chores**
  * Minor formatting of public re-exports.

### DIFF
--- a/src/lib/commands/group.rs
+++ b/src/lib/commands/group.rs
@@ -388,6 +388,63 @@ fn parallel_threshold(strategy: Strategy, opt: &ParallelMinTemplates) -> usize {
     }
 }
 
+/// How an input header satisfies `group`'s record-ordering requirement.
+///
+/// The three cases are mutually exclusive and are what the diagnostics and the
+/// rejection message key off. They must be decided in priority order: a
+/// template-coordinate header also advertises `GO:query`, so it satisfies
+/// [`crate::sam::is_query_grouped`] too and would otherwise be misreported as
+/// merely query-grouped whenever `--allow-unmapped` is set.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum InputOrdering {
+    /// Template-coordinate sorted: records sharing a position key are adjacent,
+    /// so mapped templates group by position as intended.
+    TemplateCoordinate,
+    /// Query-grouped but not template-coordinate sorted. Accepted only under
+    /// `--allow-unmapped`, which groups unmapped reads by UMI alone.
+    QueryGroupedUnmapped,
+    /// Neither ordering holds, so a template's records may not be adjacent.
+    Unusable,
+}
+
+/// Classify how an input header orders its records.
+///
+/// Template-coordinate is the ordering `group` is built around. The streaming
+/// `RecordPositionGrouper` emits a group whenever the position key changes
+/// between consecutive records, so any input where records sharing a position
+/// key aren't already adjacent (queryname-sorted, coordinate-sorted,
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #621, opened 2026-07-21
perf(sort): restore the unstable queryname sort via a totally ordered key

nh13:
## Why

#540 made `sort --order queryname[::natural]` deterministic for exact-key ties by switching the four in-chunk sorts from `*sort_unstable_by` to their stable equivalents. That fixed a real bug: several secondary alignments of one read share a name and flag class, and an unstable sort could emit them in any order, disagreeing with `samtools sort -n`, fgbio, and fgumi's own arena/runall path.

It is also expensive. Rust's stable sort allocates an N/2-element scratch buffer and does extra moves, and the `(K, RawRecord)` element is heavy. On 219M records the in-memory sort phase went **13.4s → 31.1s (2.3x)** and end-to-end wall went **150s → 175s (1.17x)**.

## The idea

Determinism does not require a *stable sort*, only a *total order*. Read name plus flags is not one. Adding the record's ingest position within its chunk makes it one — after which an unstable sort produces exactly the output a stable sort would.

The two halves compose into a global order without a global counter:

- a `u32` position orders records **within** a chunk;
- the loser-tree merge already breaks **cross-chunk** ties by source index (`loser_tree.rs`), and sources are built in chunk (ingest) order.

The position is carried in the key behind a new `RawSortKey::set_position` whose default is a no-op, so coordinate and template-coordinate — radix sorted, where stability is free and a wider key would only add passes — are untouched.

## Cost: zero

`Vec<u8>` plus a `u16` leaves 6 bytes of tail padding, so the `u32` lands in it and both key types stay **32 bytes**. There is a test asserting this, because the entire argument is that a total order is cheaper than stable sorting — if the key ever grows, the trade needs revisiting.

## Numbers

c7g.4xlarge, `1kg-wes-HG00100.bam` (14.3 GB / 219M records), 11 spills, `--threads 8`, 3 reps, median:

| build | sort phase | wall |
|---|---|---|
| pre-#540 (unstable) | 13.4s | 150s |
| main (#540, stable) | 31.1s | 175s |
| **this change** | **14.0s** | **152s** |

The sort phase returns to within 4% of pre-#540 (the residual is the extra position comparison); wall drops 13%. Rep spread was ≤0.5s.

**Coordinate is the control** and does not move — sort phase 2.6s vs 2.3s, wall 166s vs 168s — confirming the effect is confined to the path #540 changed rather than being a general sort or I/O shift.

## Correctness

- **Output is byte-identical to the stable sort.** `compare bams --command sort` reports `IDENTICAL (sort order verified; run multisets match)` across all 219M records, through the spill path (11 spills), not just in memory.
- **#540's own regression test passes unchanged** — repeated names still emerge in ingest order across the single-threaded, `par_chunks_mut`, and spill paths. That is the point: the invariant is now preserved by the *key* rather than by the *algorithm*, so it can no longer be broken by swapping the sort primitive.
- New tests cover the invariant directly: position totally orders otherwise-equal keys (both comparators), position only applies after name and flags, and the size assertion above.
- `cargo fmt`, `clippy --all-features --all-targets -D warnings`, and the full `fgumi-sort` suite (474 tests) are clean.

## One sharp edge, documented

Keys rebuilt during merge via `extract_from_record` carry position 0, because the position is not stored in the record. So the same pair compares unequal in memory but equal at merge. That is sound — the merge only ever compares heads of *different* sources, and records within one source are already emitted in file order — but it is an asymmetry in `Ord`, so it is called out at both `set_position` and `extract_from_record`. Position must not be relied on after spill.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved queryname sorting determinism by adding an ingest-position tie-breaker for exact matches.
  - Ensures consistent, correct tie handling for queryname+flag records across both in-memory and spill workflows.

- **Performance**
  - Updated chunk spilling logic to trigger on both memory pressure and a per-chunk record cap.
  - Adjusted internal sorting behavior to maintain correct tie ordering without relying on stable sort.

- **Tests**
  - Added unit and end-to-end integration coverage for the record-cap invariant, spill predicate behavior, and output identity across comparator types and settings.

### DIFF
--- a/crates/fgumi-sort/src/external.rs
+++ b/crates/fgumi-sort/src/external.rs
@@ -58,6 +58,31 @@ use std::thread::{self, JoinHandle};
 use std::time::{Duration, Instant};
 use tempfile::TempDir;
 
+/// Maximum number of records held in one in-memory chunk.
+///
+/// Each key carries its ingest position within the chunk (see
+/// [`RawSortKey::set_position`](crate::keys::RawSortKey::set_position)) so that
+/// the key is a *total* order and the chunk sort can be unstable. That position
+/// is a `u32`, so a chunk that grew past `u32::MAX` records would have to stamp
+/// two records with the same position — exact name/flag ties would stop being
+/// totally ordered and the unstable sort could reorder them, breaking output
+/// identity with `samtools sort -n`.
+///
+/// A chunk spills on `memory_limit` long before this in any real run (holding
+/// this many records needs hundreds of GB). This is the backstop that makes the
+/// `usize` -> `u32` narrowing at the stamp site provably lossless rather than
+/// merely unlikely to overflow.
+const MAX_CHUNK_RECORDS: usize = u32::MAX as usize;
+
+/// Whether the current chunk must be spilled before accepting another record.
+///
+/// Spilling is normally driven by `memory_limit`; `MAX_CHUNK_RECORDS` is the
+/// second, independent trigger that keeps ingest positions unique within a
+/// chunk (see [`MAX_CHUNK_RECORDS`]).
+fn should_spill_chunk(memory_used: usize, memory_limit: usize, chunk_records: usize) -> bool {
+    memory_used >= memory_limit || chunk_records >= MAX_CHUNK_RECORDS
+}
+
 // ============================================================================
 // Per-Phase Timing for Sort Pipeline
 // ============================================================================
@@ -2665,8 +2690,14 @@ impl RawExternalSorter {
             stats.total_records += 1;
             progress.log_if_needed(1);
 
-            // Extract key from raw bytes
-            let key = K::extract(record.as_ref(), &ctx);
+            // Extract key from raw bytes. Stamp the ingest position within this
+            // chunk so the key is totally ordered: read name + flags alone is not
+            // (secondary alignments of one read collide), and without a tiebreak
+            // the chunk sort would have to be stable to preserve ingest order.
+            let mut key = K::extract(record.as_ref(), &ctx);
+            key.set_position(
+                u32::try_from(entries.len()).expect("chunk length is capped at MAX_CHUNK_RECORDS"),
+            );
 
             // Estimate memory: record bytes + key overhead
             let record_size = record.as_ref().len() + 50; // approximate key size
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed the outside-diff comment on `keys.rs` (sorter-level output-identity coverage): added `test_sort_queryname_output_identity_matches_stable_baseline`, which pins the entire emitted record sequence for **both** the lexicographic and natural comparators against a stable-sort baseline. Names are chosen so the two comparators disagree (`read2` < `read10` naturally, `read10` < `read2` lexicographically), each name repeats at strictly increasing positions so exact-key ties are present in every run, and the `threads` × `memory_limit` matrix exercises the single-threaded and `par_chunks_mut` in-memory paths plus the spill path.

Also resolved the inline `MAX_CHUNK_RECORDS + 1` finding: the `over_record_cap` case now uses `MAX_CHUNK_RECORDS.saturating_add(1)` so it stays representable on 32-bit targets (where the constant is already `usize::MAX`) while remaining unchanged on 64-bit.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed the remaining `keys.rs` finding (target-aware key layout): `position_fits_in_existing_key_padding` is now `#[cfg(target_pointer_width = "64")]`, and the doc explains why — on 32-bit `Vec<u8>` is 12 bytes, leaving only 2 after the `u16`, so the key grows from 16 to 20 there. That is a cost question, not a correctness one: the total order the unstable sort depends on holds on every target. The two `pos` field docs and the commit body now scope the zero-growth claim to 64-bit rather than stating it unconditionally.

Two things from a pre-push self-review, both in this branch as well:

- `RawSortKey::set_position`'s doc justified the design with an unsourced "costs 28-37% on the sort step at scale", which contradicts the only measurement in this PR. It now cites the measured numbers: 2.3x on the in-memory sort phase (13.4s -> 31.1s over 219M records) and 17% end-to-end.
- `test_sort_queryname_output_identity_matches_stable_baseline` claimed spill-path coverage but nothing pinned it — `4096` reached the spill path only because 80 fixture records happened to exceed it. The two limits are now named constants and the body asserts `stats.chunks_written > 0` exactly for the spilling one, so a change to the per-record memory estimate or the fixture size cannot silently collapse that case into a third in-memory run. Verified the assertion bites: raising the spilling limit to 32 MiB fails all four cases.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #622, opened 2026-07-21
perf(sort): derive the radix bound inside the first counting pass

nh13:
Stacked on #606 — review that first; this PR's diff is only the last commit.

#606 sizes radix passes from the largest non-sentinel key, found via a standalone scan of `refs`. It also *deliberately dropped* upstream's incremental `max_sort_key` tracking, which avoided that scan and measured 4% faster, on the grounds that the tracking needs a field kept in sync across three reset sites, has a reset-before-macro ordering hazard, and guards its bound with a `debug_assert` — meaning a drift in a release build silently mis-sorts.

**This recovers that 4% without reintroducing any of it.**

## How

The byte-0 counting pass already loads every `sort_key` to build its histogram, and it runs *before* any element is scattered. So the bound can be derived right there — for a compare per record rather than another traversal of memory — and is still known in time to size the run. The histogram that pass built is handed to its own scatter, so nothing is counted twice.

There is no invariant to check, in release or otherwise: the bound is recomputed from the data on every sort and cannot disagree with the keys it is sorting.

## Measured

Idle `c6a.4xlarge`, criterion 100 samples, confidence intervals within ±0.2%:

| records | full width | scan (#606) | **fused (this PR)** | tracked (upstream) |
|---|---|---|---|---|
| 1M | 23.76 M/s | 28.48 | **29.40** | 29.26 |
| 8M | 20.88 M/s | 24.65 | **25.40** | 25.61 |

Fusing is worth **+3.0–3.2%** over the scan and lands **within 1% of the tracked version** — i.e. it recovers the win that motivated the tracking, with none of the machinery. The benchmark carries all four arms so the claim is reproducible. A second independent run on a fresh c6a reproduced this (fused 30.15 vs tracked 30.32 at 1M; 26.33 vs 26.13 at 8M).

### Cross-architecture caveat

The "within 1%" result is x86. On an M2 Max the same benchmark puts **tracking ~7% ahead of fusing** (102.8 vs 96.2 M/s at 1M, intervals ±0.7%) — plausibly because folding the max into the counting loop inhibits some unrolling there. Both are far ahead of the scan on that machine.

So the honest claim is: fusing equals tracking on x86 and trails it by ~7% on Apple Silicon, while requiring no field, no reset sites, and no release-invisible bound. Given the in-memory sort is a small share of coordinate wall time, I think that trade still favours fusing — but the parity claim should not be read as architecture-independent.

## Also

`par_sort_into_chunks` now lets each chunk derive its own bound instead of sharing a whole-buffer one. That removes the **last production caller** of the unchecked `radix_sort_record_refs_with_max`, which is retained for benchmarks and tests that need to pin a width (passing `u64::MAX` to force full 8 passes) and documented as not-for-production. A per-chunk bound is also never wider than a shared one, so a chunk with narrow keys runs fewer passes.

## Testing

Output identity against an externally supplied bound across mixed, sentinel-free, sentinel-heavy, threshold-boundary and wide-key inputs — asserting the stable tie-break order, not just the ordering. Verified non-vacuous by corrupting the reused histogram, which fails all five cases.

`cargo ci-test` (5578 tests), `cargo ci-fmt`, `cargo ci-lint`, `RUSTDOCFLAGS="-D warnings" cargo ci-doc` pass.

## Summary by CodeRabbit

- **Performance**
  - Improved coordinate sorting by computing bounds per chunk instead of deriving a single global bound up front.
  - Optimized radix sorting to determine the effective key width during the initial pass, while safely ignoring unmapped sentinel values to avoid extra radix work.

- **Tests**
  - Added coverage ensuring the new fused-bound sorting matches the externally bounded workflow across multiple sentinel and key-width scenarios.

- **Reliability**
  - Verified stable ordering is preserved, including correct handling of ties via key+offset behavior.

### DIFF
--- a/benches/core_functions.rs
+++ b/benches/core_functions.rs
@@ -1370,11 +1370,29 @@ fn bench_coordinate_radix_sort(c: &mut Criterion) {
                 criterion::BatchSize::LargeInput,
             );
         });
-        group.bench_with_input(BenchmarkId::new("scan_mapped_max", n), &refs, |b, refs| {
+        group.bench_with_input(BenchmarkId::new("fused_bound", n), &refs, |b, refs| {
             b.iter_batched(
                 || refs.clone(),
                 |mut v| {
                     radix_sort_record_refs(&mut v);
+                    black_box(v)
+                },
+                criterion::BatchSize::LargeInput,
+            );
+        });
+        group.bench_with_input(BenchmarkId::new("scan_mapped_max", n), &refs, |b, refs| {
+            b.iter_batched(
+                || refs.clone(),
+                |mut v| {
+                    // The pre-fusion shape: a standalone traversal to derive the
+                    // bound, then the sort. Kept to measure what fusing saves.
+                    let max = v
+                        .iter()
+                        .map(|r: &RecordRef| r.sort_key)
+                        .filter(|&k| k != u64::MAX)
+                        .max()
+                        .unwrap_or(0);
+                    radix_sort_record_refs_with_max(&mut v, max);
                     black_box(v)
                 },
                 criterion::BatchSize::LargeInput,
--- a/crates/fgumi-sort/src/inline.rs
+++ b/crates/fgumi-sort/src/inline.rs
@@ -462,10 +462,11 @@ impl RecordBuffer {
         &mut self,
         threads: usize,
     ) -> Vec<InMemoryChunk<RawCoordinateKey>> {
-        // Scan once for the whole-buffer bound and hand it to every chunk's sort,
-        // so no chunk re-scans for its own maximum.
-        let max_key = max_non_sentinel_key(&self.refs);
-        let sort_chunk = |refs: &mut [RecordRef]| radix_sort_record_refs_with_max(refs, max_key);
+        // Each chunk derives its own bound inside its byte-0 counting pass, so
+        // there is no whole-buffer scan to share and no bound to pass around. A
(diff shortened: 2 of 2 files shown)


nh13:
Addressed both CodeRabbit findings:

- **`wide_keys` test case** (inline thread, `crates/fgumi-sort/src/inline.rs`) — added a `wide: bool` case parameter; the `wide_keys` case now spreads non-sentinel keys across the low 7 bytes (`& 0x00FF_FFFF_FFFF_FFFF`), so the derived bound needs 7 radix passes and genuinely exercises `bytes_needed_for`'s wide-width path instead of maxing out under 2 bytes as before.
- **Outside diff range comment** (`radix_sort_record_refs` doc, ~1625-1637; no inline thread) — rewrote the stale "scans then defers" doc block to describe the fused byte-0 counting-pass bound derivation that dispatches to `radix_sort_sized` directly, and dropped the now-misleading "skip the scan" advice.

The failing `coverage` check was a transient Codecov uploader error — the codecov CLI could not verify the GPG signature of its own downloaded binary ("Could not verify signature ... Exiting"), unrelated to the diff; codecov itself reported all modified lines covered. The re-run triggered by this push should clear it.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #623, opened 2026-07-21
perf(group): encode each UMI once in the sequential assigners

nh13:
## Summary

Each sequential UMI assigner (`edit`, `adjacency`, `paired`) re-ran `BitEnc::from_umi_str` several times per UMI and threw the result away, keeping only `.len()` or `.is_some()`. The differing-length guard added in #510 compounded this on the paired path: `underlying_umi_len` (which splits the prefixed key and encodes *both* halves) was called up to three times per UMI — in the encodability filter, the length guard, and the per-record strand resolution in the single-molecule fast path.

This encodes each UMI **once** and reuses the result everywhere:

| assigner | `from_umi_str` calls per UMI, before → after |
|---|---|
| edit | 2 → 1 |
| adjacency | 2 → 1 |
| paired | 6 → 2 (one `underlying_umi_len`) |

`BitEnc` is a 16-byte `Copy` struct, so the cached `Vec<Option<BitEnc>>` (paired: a `Vec<Option<usize>>` of underlying base lengths) is far cheaper than repeating the per-byte encode.

## Correctness

Behaviour-preserving by construction — every cached value is a pure function of a string that does not change between the cache write and its reads.

- `count_paired` now carries the underlying base length. This is sound because canonicalization only ever swaps the two halves (`A-B` ↔ `B-A`), which leaves the base multiset — and therefore both `BitEnc`-encodability and the summed base count — unchanged, so every raw UMI folding into a canonical form contributes the same value.
- `assign_with_invalid_fallback`'s `resolve` closure now receives the UMI's position so callers can index the precomputed encoding.
- The GRP-01 cross-assigner parity guarantees from #510 are untouched: `test_sequential_and_parallel_assigners_induce_same_partition` passes unchanged.
- Added two tests pinning the carried-length invariant (valid folding, and an `N`-containing UMI staying `None` through canonicalization).

Full workspace suite: **5589 passed, 23 skipped**. `cargo ci-fmt` / `ci-lint` clean.

## Performance

Measured on a `c7g.4xlarge` (Graviton3, matching the benchmark fleet), `--threads 0`, GNU `time -v` user CPU, 3 reps, with the untouched `identity` strategy as a control:

| sample | strategy | base | this PR | Δ |
|---|---|---|---|---|
| agilent-hs2 | paired | 93.35s | 92.39s | **−1.0%** |
| agilent-hs2 | identity (control) | 85.30s | 85.66s | +0.4% |

Complete rep separation on paired (every base rep slower than every branch rep). Family-size histograms and molecule counts are byte-identical before/after (paired 10,220,662; identity 10,292,162).

## Scope

This removes the redundant *computation* on the sequential paths. Profiling attributes the larger share of the paired path's recent cost to per-UMI *allocations* (`reverse()` building a fresh `String` on every adjacency comparison; `canonicalize_paired`; the already-uppercase input still being re-allocated). Those are a separate, self-contained follow-up and are intentionally not bundled here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency of UMI assignment between sequential and parallel workflows.
  * Invalid or non-encodable UMIs are now handled consistently across simple, adjacency-based, and paired assignment paths, including correct length validation and molecule-ID mapping.
  * Paired-UMI processing now preserves invalid/unknown underlying states (`None`) through canonicalization and grouping, ensuring invalid paired UMIs remain isolated.
* **Tests**
  * Extended test coverage for paired-Umi `None` propagation and isolation of invalid inputs.

### DIFF
--- a/crates/fgumi-umi/src/assigner.rs
+++ b/crates/fgumi-umi/src/assigner.rs
@@ -680,8 +680,10 @@ pub fn assert_uniform_umi_length(lengths: impl IntoIterator<Item = usize>) {
 /// Build the per-input `MoleculeId` vector, resolving each raw UMI with `resolve` and falling
 /// back to a per-string molecule for the ones `resolve` cannot place.
 ///
-/// `resolve` returns `Some(id)` for a UMI that the strategy grouped (an encodable/valid UMI) and
-/// `None` for one it could not (a non-`BitEnc`-encodable, invalid UMI). Every distinct invalid
+/// `resolve` is called with each UMI's position in `raw_umis` and the UMI itself. It returns
+/// `Some(id)` for a UMI that the strategy grouped (an encodable/valid UMI) and `None` for one it
+/// could not (a non-`BitEnc`-encodable, invalid UMI). The position lets callers read a
+/// precomputed per-UMI encoding instead of re-deriving it inside the closure. Every distinct invalid
 /// UMI string gets its own molecule from `mint`, with identical (case-folded) invalid strings
 /// sharing -- mirroring fgbio's `Map[Umi, MoleculeId]`, where every UMI it sees is assigned a
 /// molecule and identical strings collapse. This keeps the sequential and parallel assigners in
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #239, opened 2026-04-06
Remove customizable tag options in favor of SAM spec standard tags


nh13:
## Summary

fgumi currently exposes CLI options to override the SAM tags used for UMIs, molecule IDs, and cell barcodes across several commands. These should be removed in favor of always using the SAM specification standard tags.

## Affected options

| Option | Default | Commands |
|--------|---------|----------|
| `--raw-tag` / `-t` | `RX` | `group`, `dedup` |
| `--assign-tag` / `-T` | `MI` | `group`, `dedup` |
| `--umi-tag` | `RX` | `correct` |
| `--tag` / `-t` | `MI` | `simplex` |
| `--cell-tag` / `-c` | `CB` | `sort`, `group`, `dedup`, `simplex`, `duplex`, `codec` |

## Rationale

`RX`, `MI`, and `CB` are all standardized SAM tags. Allowing users to override them per-command creates a class of subtle pipeline bugs: a user who threads a non-standard tag through most commands but misconfigures a single step gets silently wrong results. See [fgbio#1153](https://github.com/fulcrumgenomics/fgbio/issues/1153) for an example of this failure mode.

Users with non-standard tags from upstream tools should normalize them once before entering the fgumi pipeline (e.g. via `samtools reheader` or a similar preprocessing step), rather than parameterizing every fgumi command.

## Proposed change

Remove all of the above CLI options. Hardcode the tag names to their SAM spec values. Update documentation and error messages accordingly.

PR #624, opened 2026-07-21
feat(correct): add -t/--target {umi,barcode} to correct RX or BC

nh13:
## Summary

Adds a bounded `-t/--target {umi,barcode}` option to `fgumi correct` so it can error-correct the `BC` sample/library-barcode tag as well as the `RX` UMI tag. The default (`umi`) reproduces today's behavior exactly.

- `--target umi` (default) — reads/writes `RX`, stashes the pre-correction value in `OX`. Unchanged behavior.
- `--target barcode` — reads/writes `BC`, stashes the pre-correction value in the fgumi-local `ob` tag.

## Motivation

`correct`'s model — error-correct observed sequences against a fixed set of known sequences — fits sample/library barcodes at least as well as UMIs, since barcodes come from a known whitelist (sample sheet / index kit). fgbio's `CorrectUmis` already supports this through its free-form `-t/--umi-tag`, so `CorrectUmis -t BC` works today.

fgumi deliberately removed that free-form override in #239 to prevent the misconfiguration bug class it describes (a non-standard tag threaded through a pipeline with one step misconfigured, yielding silently-wrong results). This PR restores the *capability* without reopening that hole: `--target` is a bounded enum over two SAM-spec standard tags (`RX` / `BC`), not an arbitrary tag string. There is no way to point `correct` at a non-standard tag.

## The `ob` tag

There is no SAM-spec "original barcode" tag (the spec has `QT` for barcode *quality*, but no analogue of `OX`). fgbio sidesteps this by writing the original barcode into `OX` — literally the "original UMI bases" tag — which this PR treats as a semantic mismatch to avoid. Instead the pre-correction barcode goes to a new fgumi-local tag `ob`. The lowercase spelling is intentional: the SAM spec reserves tags containing lowercase letters for local/end-user use and promises never to standardize them, so `ob` honestly signals a local convention rather than masquerading as a standard tag.

## Changes

- New `SamTag::OB` (`ob`) constant in `fgumi-tag`.
- New `Target { Umi, Barcode }` enum with `sequence_tag()` / `original_tag()` returning the `(RX, OX)` / `(BC, ob)` pairs; threaded through `correct`'s existing tag-agnostic plumbing (both the threaded and single-thread pipeline paths).
- `correct` gains `-t/--target <umi|barcode>` (default `umi`).
- The store-original flag is renamed to the mode-neutral `--dont-store-original`, with `--dont-store-original-umis` kept as a hidden alias for backward compatibility.
- `--help` text and module docs updated to describe both modes.

Correcting both `RX` and `BC` is two invocations, which keeps per-pass metrics and rejects unambiguous.

## Testing

- Unit: `Target` tag-pair mapping and the `ob`/`OB` constant.
- Integration (`#[rstest]` case tables over both modes): corrected value lands in the right tag, original in the right tag (`OX` / `ob`), and `--dont-store-original` (and its alias) suppresses the original write.
- A threaded-path barcode case (`--threads 2`) and a cross-contamination guard confirming a `barcode` run leaves `RX`/`OX` untouched.
- All test data is built programmatically with `SamBuilder`; no committed fixtures.

Backward compatibility: default `--target umi` is byte-for-byte identical to prior behavior, and the existing `correct` test suite passes unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added barcode correction support alongside UMI correction.
  * Added a target option to choose whether to correct UMI or barcode (default: UMI).
  * Corrected values are now written to the selected tag, with optional preservation of the original value.
  * Added a new local barcode “original barcode” tag used for storing the pre-correction value when enabled.

* **Bug Fixes**
  * Improved correctness of tag updates for both targets across single-threaded and multi-threaded processing.
  * Ensured correct behavior when records contain both UMI and barcode tags, including proper suppression of original-tag writes when no correction occurs.

### DIFF
--- a/crates/fgumi-tag/src/tag.rs
+++ b/crates/fgumi-tag/src/tag.rs
@@ -94,6 +94,13 @@ impl SamTag {
     pub const QT: SamTag = SamTag::new(b'Q', b'T');
     /// Original UMI bases, before error correction (SAM spec).
     pub const OX: SamTag = SamTag::new(b'O', b'X');
+    /// Original barcode bases, before error correction.
+    ///
+    /// fgumi-local tag (not SAM-spec): the SAM spec has no "original barcode"
+    /// tag and reserves tags containing lowercase letters for local use, so
+    /// this deliberately uses lowercase `ob` to signal a local convention
+    /// rather than masquerade as a standard tag.
+    pub const OB: SamTag = SamTag::new(b'o', b'b');
     /// Original UMI base quality scores, before error correction (SAM spec).
     pub const BZ: SamTag = SamTag::new(b'B', b'Z');
     /// Read group (SAM spec).
@@ -590,5 +597,12 @@ mod tests {
                 "REVERSE tag {t} missing from PER_BASE_CONSENSUS_TAGS",
             );
         }
+    }
+
+    #[test]
+    fn ob_tag_is_lowercase_and_valid() {
+        // fgumi-local "original barcode" tag: SAM reserves lowercase tags for local use.
+        assert_eq!(SamTag::OB.to_string(), "ob");
+        assert!(SamTag::is_valid_tag_bytes(b'o', b'b'));
     }
 }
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #625, opened 2026-07-21
perf(group): cut the paired assigner's per-UMI encoding, allocation, and iteration

nh13:
## Summary

Stacked on #623. Together with that PR, this removes the per-UMI overhead the sequential `PairedUmiAssigner` gained — the source of the `group` paired-strategy slowdown, which a full staircase bisect across v0.3.1→main attributes entirely to a single commit (the differing-length/invalid-UMI GRP-01 correctness fix), with every other commit in the range flat and the `identity` control flat throughout.

#623 removed the *redundant* encodes. This removes the rest of the per-UMI work, all in `PairedUmiAssigner`:

- **`underlying_umi_len`** — replace the `split('-')` / `rsplit(':')` iterators plus a double `BitEnc::from_umi_str` encode with a single byte scan (`position`/`rposition` + an ACGT count). The packed value was always discarded — only validity and base length were ever used. Kept bug-for-bug equivalent to `from_umi_str` (both cases accepted, >32-bases-per-half rejected); prefix bytes before `:` are never validated, so a `bb:` orientation prefix is ignored rather than rejected.
- **Single-molecule fast path** — decide the strand with a new allocation-free `orientation_is_ab` helper instead of building a reversed `String` per read. `umi < reverse(umi)` reduces to comparing the two dash-halves, because `-` (`0x2D`) is smaller than every byte that can appear in a half.
- **`matches_paired`** — compare `reverse(lhs)` to `rhs` via `reversed_matches_within` (a positional byte compare) instead of allocating the reversed string on every adjacency-graph comparison.
- **`upper_umis`** — borrow `raw_umis` directly when no UMI contains a lowercase byte (the production case), allocating an uppercased copy only for mixed-case input.
- **Paired-format validation** — locate the first `-` and confirm no second, instead of `split('-').count()`.

## Correctness

Behaviour-preserving. Grouping output is unchanged: family-size histograms and molecule counts are **byte-identical** before and after (agilent-hs2 paired: 10,220,662 molecules; identity: 10,292,162). The GRP-01 cross-assigner parity guarantees are untouched (`test_sequential_and_parallel_assigners_induce_same_partition` passes). Each rewritten helper is pinned to the definition it replaces by a new `rstest` table:

- `test_underlying_umi_len_matches_bitenc` (10 cases: prefixes, both cases, non-ACGT, the 32-base limit, empty/uneven halves)
- `test_orientation_is_ab_matches_reverse_compare` (7 cases vs `umi < reverse(umi)`)
- `test_reversed_matches_within_matches_alloc` (8 cases vs reverse-then-compare, incl. threshold boundaries)

Full workspace suite: **5614 passed, 23 skipped**. `ci-fmt` / `ci-lint` clean.

## Performance

Measured on a `c7g.4xlarge` (Graviton3, matching the benchmark fleet), `--threads 0`, GNU `time -v`, agilent-hs2 paired, with `identity` as a flat control. Against the pre-regression baseline (the offending commit's parent):

| | paired user CPU | vs pre-regression |
|---|---|---|
| offending commit | 92.96s | +4.1% |
| #623 alone | ~92.5s | +3.9% |
| **#623 + this PR** | **~90.1s** | **~+1%** |

This recovers roughly **three-quarters** of the paired regression on this workload; the residual ~1% is the irreducible per-UMI validity/length check the GRP-01 fix requires.

**Scope note (measured honestly):** agilent-hs2 is high-diversity, so its groups take the single-molecule fast path and the adjacency graph is cold. On that workload the `matches_paired` and `upper_umis` changes are **neutral** — the recovery above comes from the `underlying_umi_len` and fast-path-strand changes. `matches_paired`'s per-comparison allocation and the `upper_umis` allocation are nonetheless real; they are expected to help adjacency-heavy / low-diversity inputs (deep panels), which this measurement does not exercise. They are included as correctness-preserving allocation hygiene, not because they move this benchmark.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved paired-UMI assignment by reducing temporary allocations and optimizing paired format/validation checks.
  * Faster reverse-orientation comparison with early mismatch detection.

* **Bug Fixes**
  * More accurate UMI length handling according to supported encoding rules (including prefixed keys and invalid bases).
  * Corrected strand/orientation determination and paired-format handling, including lowercase and edge cases.

* **Tests**
  * Added targeted regression coverage for UMI length semantics, orientation logic, and reverse matching behavior.

### DIFF
--- a/crates/fgumi-umi/src/assigner.rs
+++ b/crates/fgumi-umi/src/assigner.rs
@@ -1812,11 +1812,62 @@ impl PairedUmiAssigner {
     /// every molecule (as a filter) or never fire (as the length guard).
     fn underlying_umi_len(key: &str) -> Option<usize> {
         let mut total = 0;
-        for half in key.split('-') {
-            let seq = half.rsplit(':').next().unwrap_or(half);
-            total += BitEnc::from_umi_str(seq)?.len();
+        let mut rest = key.as_bytes();
+        loop {
+            // Split off the next dash-delimited half via a byte search, avoiding the
+            // `split('-')` iterator's per-call state on this hot path.
+            let (half, more) = match rest.iter().position(|&b| b == b'-') {
+                Some(pos) => (&rest[..pos], Some(&rest[pos + 1..])),
+                None => (rest, None),
+            };
+            // The sequence is the bytes after the last ':' (orientation prefix), or the whole
+            // half when unprefixed. `rposition` is a plain byte search (memrchr); crucially the
+            // prefix bytes before it are never validated, so a prefix like `bb:` (containing the
+            // non-base `b`) is correctly ignored rather than rejected.
+            let seq = match half.iter().rposition(|&b| b == b':') {
+                Some(pos) => &half[pos + 1..],
+                None => half,
+            };
+            // Count ACGT bases and reject anything `BitEnc` cannot represent, matching
+            // `BitEnc::from_umi_str` exactly (both cases accepted; >32 bases per half rejected)
+            // without allocating the discarded 2-bit packing. This is the sole hot-path use of
+            // the result, so the packed value was pure waste.
+            let mut half_len = 0;
+            for &b in seq {
+                if !matches!(b, b'A' | b'a' | b'C' | b'c' | b'G' | b'g' | b'T' | b't') {
+                    return None;
+                }
+                if half_len >= 32 {
+                    return None;
+                }
+                half_len += 1;
+            }
+            total += half_len;
+            match more {
+                Some(r) => rest = r,
+                None => break,
+            }
         }
         Some(total)
+    }
+
+    /// Whether a paired UMI's own orientation takes the `PairedA` strand, i.e. whether it sorts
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #626, opened 2026-07-22
perf(consensus): memoize post-UMI adjusted probability tables

nh13:
## Summary

`ConsensusBaseBuilder::new` precomputed two 94-entry log-probability tables on every construction, via `ln_error_prob_two_trials` (transcendental math) plus two heap allocations. That would be negligible once per process, but `simple_umi::consensus_umis()` constructs a fresh `SimpleConsensusCaller` — and therefore a fresh `ConsensusBaseBuilder` — for **every UMI family**, so the table build ran once per family.

Both tables are a pure function of the post-UMI error rate, which is a `u8`. A 256-slot `OnceLock` cache therefore covers the input domain exactly:

```rust
static ADJUSTED_TABLE_CACHE: [OnceLock<AdjustedProbabilityTables>; 256] =
    [const { OnceLock::new() }; 256];
```

No clamping is involved, so there is no behaviour change for out-of-range Phred values. The builder's two table fields become `&'static [LogProbability]`, making `new()` a pair of atomic loads instead of 94 `exp`/`log1p` evaluations plus two allocations.

## Correctness

The memoized computation performs exactly the arithmetic the per-builder loop performed inline, in the same order, so the resulting `f64` values are bit-identical by construction — this is a pure memoization, not a reformulation.

Verified empirically on 1,000,000 consensus records: record bodies are byte-identical, checked both with `fgumi compare bams` and with a `samtools view` md5 of the record bodies.

One honest caveat on the verification method: a whole-file BAM md5 is **not** a valid oracle here, because the output filename is embedded in the `@PG CL:` header line, so two runs writing to different paths differ in the header even when every record is identical. Record-body comparison is the correct check, and that is what was used.

All 588 `fgumi-consensus` tests pass, including the `fgbio_oracle_saturation` parity tests.

## Measured impact

Consensus wall-clock, median of repeated runs:

| Workload | Before | After | Change |
| --- | --- | --- | --- |
| simplex | 16.48s | 13.87s | **-15.8%** |
| duplex | 22.79s | 20.33s | **-10.8%** |
| CODEC | 6.60s | 6.28s | **-4.9%** |

In the profile, `ConsensusBaseBuilder::new` goes from 13.1% inclusive to absent.

---

**Note on ordering:** #627 (the tie-detection fix) is stacked on top of this branch, since both touch `base_builder.rs` in the same two regions. Merge this one first and #627 will retarget to `main` automatically with no conflict.

## Summary by CodeRabbit

* **Performance**
  * Improved consensus generation efficiency by reusing shared, lazily computed post-UMI probability tables, reducing redundant per-instance work.
  * Lowered per-instance memory usage by referencing cached probability tables.
* **Bug Fixes**
  * Improved no-call/tie behavior to be order-independent and ULP-relative, with more consistent results across floating-point edge cases.
* **Documentation**
  * Updated explanations of no-call quality and tie handling to match the new behavior.
* **Tests**
  * Expanded unit tests for memoization correctness and tie/no-call resolution, including near-ties and non-finite lanes.

### DIFF
--- a/crates/fgumi-consensus/src/base_builder.rs
+++ b/crates/fgumi-consensus/src/base_builder.rs
@@ -317,6 +317,66 @@ const BASE_TO_INDEX: [u8; 256] = {
     table
 };
 
+/// The post-UMI-adjusted per-quality probability tables used by `ConsensusBaseBuilder`.
+///
+/// Both tables are indexed by the observed base quality (`0..=MAX_PHRED`) and depend *only*
+/// on the post-UMI error rate they were computed for. Because that error rate and the base
+/// quality are both integer Phred scores drawn from a tiny finite domain, the whole table is
+/// a pure function of a single `u8` — so it can be computed once and shared, rather than
+/// recomputed per `ConsensusBaseBuilder`.
+struct AdjustedProbabilityTables {
+    /// Log probability that an observation of quality `q` is correct, adjusted for post-UMI error.
+    correct: Box<[LogProbability]>,
+
+    /// Log probability of any one *specific* alternate base (total error / 3) for quality `q`.
+    error_per_alt: Box<[LogProbability]>,
+}
+
+impl AdjustedProbabilityTables {
+    /// Computes both tables for a given post-UMI error rate.
+    ///
+    /// This performs exactly the arithmetic the per-builder loop used to perform inline, in the
+    /// same order, so the resulting `f64` values are bit-identical to the un-memoized version.
+    fn compute(error_rate_post_umi: PhredScore) -> Self {
+        let ln_error_post = phred_to_ln_error_prob(error_rate_post_umi);
+        let ln_three = 3.0_f64.ln();
+
+        let mut correct = Vec::with_capacity(MAX_PHRED as usize + 1);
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #627, opened 2026-07-22
fix(consensus): make consensus tie detection scale- and order-independent

nh13:
## The bug

The inline tie check in `ConsensusBaseBuilder::call()` had two independent defects, and both made the no-call decision depend on something other than the evidence.

**1. Scale dependence.** Ties were tested with an *absolute* epsilon, `f64::EPSILON` (2.2e-16), which is the ULP of `1.0`. But consensus log-likelihoods are running sums of log-probabilities: for a realistic family they sit near `-500`, where a single ULP is already ~5.7e-14 — roughly 250x larger than the tolerance. Two adjacent representable doubles at that magnitude are further apart than the epsilon, so the tie branch could never fire.

**2. Order dependence.** Ties were measured against a *running* maximum rather than the final one. A near-tie was therefore only ever detected when the tied lane happened to appear **after** the maximum; when the likelihoods arrived in ascending order every comparison took the `Greater` branch and no tie was recorded at all. Concretely, the same two values return `None` in descending order and `Some(1)` in ascending order — the answer depended on the fixed `A,C,G,T` lane order rather than on the data.

## The fix

The check moves out of `call()` into a small `unique_max_index` free function that finds the maximum first, then counts how many lanes are within `TIE_TOLERANCE_ULPS` (4) of it. A unique maximum yields exactly one match (the maximum always matches itself); anything else is a tie and therefore a no-call.

ULP comparison is inherently relative, so a single constant behaves consistently for a 2-read family and a 2000-read family alike — no magnitude-dependent tuning.

Non-finite lanes are now handled explicitly rather than falling out of `partial_cmp` returning `None`: `NaN` lanes are ignored, and an array with no finite entry at all (no evidence for any base) is a no-call. That also removes the `expect()` the old code needed after its `is_none()` check, so the now-stale `# Panics` section on `call()` is deleted.

## Measured impact

On 1,000,000 consensus records:

- **212 records differ**, each by exactly one base — 212 bases out of 150,000,000, or **0.00014%**.
- Every single change is **base -> N**: C->N 58, A->N 54, G->N 53, T->N 47. There are **zero** N->base changes and **zero** base->other-base changes. The fix only ever converts a previously-arbitrary concrete call into an explicit no-call; it never changes one base into a different base.
- **186 of the 212** already carried the minimum quality **Q2** in the old output. In other words, at the overwhelming majority of affected positions the old code already knew the position was maximally uncertain — it just emitted an arbitrary concrete base anyway, picked by lane order.
- The near-uniform A/C/G/T split is exactly what genuine ties should look like, and is further evidence these are real ambiguities rather than a systematic bias introduced by the new tolerance.

## Relationship to fgbio

A corresponding fix is being prepared for fgbio's `MathUtil.maxWithIndex`, which contains the identical defect — the same `1.0/2^52` absolute epsilon and the same running-maximum comparison. This change therefore moves the two tools **toward** agreement, not away from it.

## Tests

Five new `#[rstest]` case tables cover the behaviour:

- `exact_ties_are_no_calls` — regression guards; these pass pre-fix too, since an exact tie always hit the `Equal` branch.
- `near_ties_are_order_independent` — the ascending/descending pair that demonstrates defect 2.
- `one_ulp_separation_is_a_tie_at_any_scale` — a one-ULP separation is summation noise, not evidence, and must read as a tie at -1, -10, -500 and -5000.
- `clear_separation_is_not_a_tie_at_any_scale` — the converse, so the fix cannot degenerate into no-calling everything.
- `tie_detection_handles_non_finite_lanes` — `-inf`/`NaN` handling.

All 608 `fgumi-consensus` tests pass, including the `fgbio_oracle_saturation` parity tests. `cargo fmt --check` and `cargo clippy --all-targets` (with `-D warnings -W clippy::pedantic`) are clean.

---

**Note on ordering:** this PR is stacked on #626 and targets `nh/perf-memoize-consensus-phred-tables`, not `main` — both changes touch `base_builder.rs` in the same two regions, so stacking avoids a merge conflict. The diff shown here is only the tie-detection change. Merge #626 first; GitHub will then retarget this PR to `main`.

The matching fgbio fix is https://github.com/fulcrumgenomics/fgbio/pull/1162 — the same defect in the algorithm this code was ported from. The two are best reviewed as a pair, since merging one without the other leaves the tools disagreeing at near-tied positions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consensus calling so tied or near-tied likelihoods are handled deterministically, independent of lane ordering and likelihood scale.
  * Non-finite likelihoods (NaN/−inf) are ignored when selecting best evidence.
  * A no-call is returned when a strictly unique best likelihood can’t be determined.
* **Documentation**
  * Updated the consensus tie-rule description to clarify the tolerance behavior.
* **Tests**
  * Added regression and parity/oracle coverage for ties, ULP-scale tolerance, and NaN/−inf edge cases.

### DIFF
--- a/crates/fgumi-consensus/src/base_builder.rs
+++ b/crates/fgumi-consensus/src/base_builder.rs
@@ -136,15 +136,18 @@
 //! ## Handling Edge Cases
 //!
 //! ### No Observations
-//! Returns `(N, 0)` - no-call with minimum quality.
+//! Returns `(N, MIN_PHRED)` - a no-call at the minimum quality, Q2.
 //!
 //! ### N Bases
 //! N (no-call) bases in input reads are ignored. They don't contribute evidence for any
 //! particular base.
 //!
 //! ### Ties
-//! If multiple bases have exactly equal posterior probability, returns `(N, 0)` to indicate
-//! ambiguity. In practice, this is rare due to quality score differences.
+//! If the greatest likelihood is not unique, returns `(N, MIN_PHRED)` to indicate ambiguity.
+//! "Not unique" means within a few ULPs, not bit-for-bit equal: the likelihoods are running
+//! sums, so a separation narrower than summation noise is not evidence. The tolerance is
+//! relative to the magnitude, so it holds at every family size — see `TIE_TOLERANCE_ULPS`.
+//! In practice ties are rare, since quality scores differ between observations.
 //!
 //! ### Low Depth
 //! With only 1-2 observations, the consensus quality will be similar to the input qualities.
@@ -191,8 +194,7 @@ use crate::phred::{
     ln_error_prob_two_trials, ln_normalize, ln_not, ln_prob_to_phred, ln_sum_exp_array,
     phred_to_ln_error_prob,
 };
-use approx::abs_diff_eq;
-use std::cmp::Ordering;
+use approx::ulps_eq;
 use std::sync::OnceLock;
 use wide::f64x4;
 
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #628, opened 2026-07-22
refactor(consensus): share one consensus read-name implementation

nh13:
## Summary

This is a **hygiene / deduplication** change. It is explicitly **not** a performance PR — see "On performance" below.

The vanilla, duplex and CODEC consensus callers each constructed their output read name with their own copy of `format!("{}:{}", read_name_prefix, umi)`. The naming convention — the `:` delimiter that matches fgbio — therefore lived in three separate places and had to be kept in sync by hand.

This PR adds a single shared implementation, `caller::write_consensus_read_name(buf, read_name_prefix, umi)`, and has all three callers use it. The name is written into a caller-owned `Vec<u8>` that is reused across records rather than allocating a fresh `String` per emitted record.

CODEC has a second naming path: records without a UMI fall back to `"<prefix>:<counter>"`. That branch now writes into the same buffer via `write!`, and both branches moved into a small private `CodecConsensusCaller::write_read_name` helper so `build_output_record_into` stays under the line limit and the two naming rules sit next to each other.

Duplex's `duplex_read_into` is an associated function that already receives a reusable `UnmappedSamBuilder`; it now receives a companion `&mut Vec<u8>` for the read name, declared alongside the builder in `call_group` so both have the same per-group lifetime.

## On performance

Removing a per-record `String` allocation is a real (small) win, but the honest ceiling here is **0.44% (simplex) / 0.68% (duplex)** `fmt` inclusive time — comfortably below run-to-run variance. **No speedup is claimed and none should be inferred.** The value of this change is that the three callers now share one implementation instead of three copies of the same string format.

## Correctness

Output is byte-identical across simplex, duplex and CODEC — this change alters how the name bytes are produced, not what they are.

All 588 `fgumi-consensus` tests pass. `cargo fmt --check` and `cargo clippy --all-targets` (with `-D warnings -W clippy::pedantic`) are clean.

## Summary by CodeRabbit

* **Performance**
  * Improved consensus record generation by reusing memory when constructing read names.
  * Reduced temporary allocations during consensus processing, which may improve throughput and resource efficiency for large datasets.

* **Consistency**
  * Standardized consensus read names using the `<read-name-prefix>:<UMI>` format across supported consensus workflows.

* **Bug Fixes**
  * Preserved validation for read names that exceed supported length limits.

### DIFF
--- a/crates/fgumi-consensus/src/caller.rs
+++ b/crates/fgumi-consensus/src/caller.rs
@@ -523,6 +523,26 @@ impl RejectionReason {
     }
 }
 
+/// Writes the consensus read name `"<read_name_prefix>:<umi>"` into `buf`, replacing whatever
+/// it held before.
+///
+/// Every consensus caller (vanilla, duplex, CODEC) names its output reads this way, once per
+/// emitted record. Sharing one implementation keeps the three callers from drifting, and
+/// writing into a caller-owned buffer instead of building a fresh `String` with `format!`
+/// keeps the name off the allocator on a path that runs once per output record.
+///
+/// # Arguments
+/// * `buf` - Reusable buffer that receives the read name
+/// * `read_name_prefix` - Prefix shared by every consensus read (see [`make_prefix_from_header`])
+/// * `umi` - Molecule identifier that distinguishes this consensus read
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #629, opened 2026-07-22
fix(extract): error on an over-long read name instead of panicking

nh13:
Stacked on #611 — it adds `UnmappedSamBuilder::try_build_record`, which this PR uses. Base will need retargeting to `main` once #611 merges. Review the second commit only.

## The problem

BAM stores the read name length in a single byte that includes the trailing NUL, so a name of 255 bytes or more cannot be represented. `extract` builds its record name from the input FASTQ header — plus `+<UMI>` when `--annotate-read-names` is set — so the length is entirely data-controlled. Both record-building paths called the asserting `build_record`, so a long FASTQ read name aborted the command with a Rust panic partway through writing the output BAM, leaving a truncated file rather than failing cleanly.

This is the same failure mode #611 fixed for the consensus callers. `extract` was the one remaining caller feeding the builder input-derived names; it was found during the sibling audit on that PR and deliberately left out of it to keep that change scoped to `fix(consensus)`.

Before, on a 300-byte read name:

```
thread ... panicked at crates/fgumi-raw-bam/src/builder.rs:172:9:
read name too long (300 bytes, max 254)
```

After:

```
Error: could not write the record for read 'NNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNN...' (name shown truncated)

Caused by:
    read name too long (300 bytes, max 254); BAM stores the name length in one byte including its NUL terminator
```

## The change

Routes `make_raw_records` (single-threaded) and `make_raw_records_static` (threaded) through `try_build_record`. Both enclosing functions already returned `Result`, so the error propagates without restructuring.

The builder’s error reports the length and the limit but not *which* read hit it, which is what you need to find the offending FASTQ record. `read_name_too_long_context` adds the name, truncated to 64 bytes so the message stays readable. It reports the final name as written, so it already includes the `+<UMI>` suffix when annotation is on.

One extra fix was required to make that message actually reach the user. The threaded path converted the `anyhow` error into the pipeline’s `io::Error` with `e.to_string()`, which keeps only the outermost context and drops everything below it — you would have seen the read name with no reason attached. It now renders the full chain with `{:#}`. That repairs the report for every error out of `make_raw_records_static`, not just this one.

The neighbouring `.map_err(std::io::Error::other)` a few lines up is left alone: it preserves the error object rather than stringifying it, so it is not lossy in the same way.

## Testing

`test_extract_rejects_over_long_read_name` is an end-to-end test through `execute` — an `#[rstest]` case table over the exact boundary, run against **both** record-building paths via `#[values]`:

| name length | expected |
|---|---|
| 254 | accepted, round-trips unchanged |
| 255 | rejected |
| 4096 | rejected |

Reverting the production change makes all four rejection cases fail with a panic in the builder, which is the reported failure mode — verified, not assumed.

Full suite: `cargo ci-test` 5549 passed / 22 skipped; `cargo ci-fmt`, `cargo ci-lint`, `cargo ci-doctest`, and `RUSTDOCFLAGS="-D warnings" cargo ci-doc` all clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of FASTQ and consensus records with read names exceeding BAM’s length limit.
  * These inputs now return clear, actionable errors instead of producing unhelpful failures or panics.
  * Error messages identify the affected read and include relevant UMI information when applicable.
  * Very long or invalid read names are safely truncated in diagnostics while preserving useful identifying details.
  * Error details are now retained consistently across single-threaded and threaded extraction modes.

### DIFF
--- a/crates/fgumi-consensus/src/caller.rs
+++ b/crates/fgumi-consensus/src/caller.rs
@@ -543,6 +543,43 @@ pub fn write_consensus_read_name(buf: &mut Vec<u8>, read_name_prefix: &str, umi:
     buf.extend_from_slice(umi.as_bytes());
 }
 
+/// Context for a consensus read name that BAM cannot represent.
+///
+/// The builder's own `ReadNameTooLong` reports the length and the limit but not *which*
+/// consensus group hit it. The UMI is input-derived, so on a large batch job a single
+/// malformed molecule identifier otherwise aborts the run with no way to find it. `name` is
+/// the fully-built consensus read name (see [`write_consensus_read_name`]), so it identifies
+/// the group by both prefix and UMI. A long name is truncated to a bounded head AND tail so
+/// it stays readable, and so the identifying bytes survive at *either* end: the read-group
+/// prefix leads and the UMI trails (see [`write_consensus_read_name`]). Showing only the head
+/// would drop the UMI — the very identifier this context exists to surface — for two groups
+/// that share a prefix. The tail is bounded too, so a name that is over-long *because* of an
+/// input-derived UMI cannot produce a multi-kilobyte line.
+///
+/// Shared by all three callers (vanilla, duplex, CODEC) for the same reason the name-writing
+/// itself is: three copies of this drift.
+#[must_use]
+pub fn consensus_read_name_too_long_context(name: &[u8]) -> String {
+    format!("could not write the consensus record for read {}", render_bounded_read_name(name))
+}
+
+/// Render a read name for a diagnostic, bounded at both ends.
+///
+/// A name within the budget is shown whole; a longer one becomes `head...tail` with the
+/// `(name shown truncated)` marker, retaining the leading `HEAD` and trailing `TAIL` bytes so
+/// distinguishing content at either end survives. `HEAD + TAIL` equals the whole-name budget,
+/// so a name at the budget is shown in full and the boundary is exact.
+fn render_bounded_read_name(name: &[u8]) -> String {
+    const HEAD: usize = 44;
+    const TAIL: usize = 20;
+    if name.len() <= HEAD + TAIL {
+        return format!("'{}'", String::from_utf8_lossy(name));
+    }
+    let head = String::from_utf8_lossy(&name[..HEAD]);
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Both outside-diff findings are addressed in the amended commit.

**Duplicate `try_build_record`+context block in `extract.rs`** — extracted into a shared `Extract::build_template_record(builder, name, flag, seq, quals, encoding)`; both `make_raw_records` and `make_raw_records_static` now call it, so the two threading paths cannot drift. I kept the quality conversion inside the helper rather than hoisting it to the caller, so the empty-sequence branch still skips the `to_standard_numeric` allocation. `test_extract_rejects_over_long_read_name` already exercises both paths and pins the refactor as behavior-preserving.

**Consensus `try_build_record` failures do not identify the offending group** — all three callers now attach context. Rather than three inline `format!` strings, the message lives in `caller::consensus_read_name_too_long_context`, next to `write_consensus_read_name` and for the same reason: three copies drift. It reports the fully-built consensus read name (prefix *and* UMI, so it identifies the group more precisely than the UMI alone) truncated to 64 bytes, mirroring `extract`'s helper. Covered by `test_consensus_read_name_too_long_context`, including the truncation boundary, invalid UTF-8, and a check that the message length does not grow with the name.

Also rebased onto the latest `main`. The consensus read-name commit that was previously part of this branch landed upstream in #628, so it dropped out of the rebase and this is now a single commit.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #630, opened 2026-07-22
fix(group): give parallel UMI assigners a persistent molecule-id counter

nh13:
## The bug

The four parallel assigners in `src/lib/umi/parallel_assigner.rs` number each `assign()` call's molecules locally as `0..k` and hold no counter across calls, so every call restarts at 0. `ParallelIdentityAssigner`'s entire struct was `{ pool }` — no counter field at all. The sequential assigners each hold an `AtomicU64` and never reuse an id.

That matters because `group` reuses a single assigner across the pair-orientation subgroups of a position group. `assign_umi_groups_impl` calls `assign()` once per FR/RF/… subgroup for every strategy whose `split_templates_by_pair_orientation()` is true — identity, edit and adjacency (paired opts out) — and then derives the group's MI offset from `max(id) + 1` over all of its templates:

```rust
let distinct_mi_count: u64 =
    templates.iter().filter_map(|t| t.mi.id()).max().map(|max_id| max_id + 1).unwrap_or(0);
```

With both subgroups numbered from 0, an FR template and an RF template carrying unrelated UMIs get the same local id; the single shared offset preserves the clash; and two distinct molecules are emitted under one MI tag. `distinct_mi_count` undercounts for the same reason, so the clash also compresses the global id space for every later position group.

The divergence is `--threads`-dependent: the sequential path is correct, so whether a given molecule is split correctly depends on whether the group cleared `--parallel-group-min-templates`.

## The fix

Add a `MoleculeIdCounter` that reserves a block of `max(local id) + 1` ids per call and rebases the call's assignments onto that block, matching the sequential assigners' contract.

Each assigner's algorithm is unchanged and still numbers locally: the existing body moves to a private `assign_local`, and the trait `assign()` rebases its result. Routing through one wrapper keeps every early return — empty input, the all-invalid-UMI fast path — covered by a single rebase site rather than needing an offset threaded through each one.

The block size is `max(local id) + 1` rather than `assignments.len()` because assigners hand out ids contiguously and several reads can share a molecule (and a paired molecule's `/A` and `/B` share one base id).

## Tests

`test_successive_assign_calls_use_disjoint_molecule_ids` is an rstest table over all eight assigners — each parallel strategy paired with its sequential counterpart — asserting that two successive `assign()` calls draw from disjoint id sets. Before the fix the four sequential cases pass and all four parallel cases fail:

```
successive assign() calls reused molecule ids [1, 0];
  first  umis=["AAAA", "CCCC"] -> [Single(0), Single(1)]
  second umis=["GGGG", "TTTT"] -> [Single(0), Single(1)]
```

`test_parallel_adjacency_deterministic_tiebreak` asserted that two `assign()` calls on one assigner returned *identical* id vectors — which is exactly the behaviour being fixed, i.e. the test encoded the bug. It now compares the induced partition, which is the determinism property it was written to check.

Full suite green: 5702 passed, plus `ci-fmt` and `ci-lint`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed parallel UMI assignment so molecule IDs remain globally unique across repeated assignments.
  - Preserved consistent read grouping while preventing ID collisions between separate assignment calls.

### DIFF
--- a/src/lib/umi/parallel_assigner.rs
+++ b/src/lib/umi/parallel_assigner.rs
@@ -12,7 +12,7 @@
 //! These produce identical results to the sequential implementations in `assigner.rs`.
 
 use std::collections::VecDeque;
-use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
 
 use ahash::{AHashMap, AHashSet};
 use rayon::prelude::*;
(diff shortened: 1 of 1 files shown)


nh13:
Heads up: #645 (index the edit assigner's neighbour search) is stacked on this branch and should land after it — it depends on the persistent molecule-id counter introduced here.

nh13:
@coderabbitai review

PR #631, opened 2026-07-22
fix(group): fail the run when UMI assignment fails

nh13:
## The bug

Both `assign_umi_groups_impl` call sites in `group.rs` caught the error, logged a warning, and substituted an empty position group:

```rust
log::warn!("UMI assignment failed, returning empty group: {e}");
...
return Ok(ProcessedPositionGroup { templates: Vec::new(), ... });
```

`fgumi group` then exited 0 having written a header-only BAM with the entire input silently discarded. Downstream consensus calling sees only an empty file, and nothing in the exit status distinguishes that from a clean run over a small input.

The trigger is easy to hit. `--strategy paired` requires the `RX` tag to carry two `-`-delimited segments, so pointing it at single-segment UMIs fails for *every* position group — the whole BAM vanishes. A missing `RX` tag or mixed UMI lengths reach the same path. In each case the reads cannot be grouped at all, which is a usage error worth surfacing, not a group worth skipping.

## The fix

Propagate the error from both paths:

- The streaming `process_fn` returns an `InvalidData` `io::Error`.
- The single-threaded path adds context and returns.

The accepted-template rollback both sites performed goes away with them — nothing continues past the failure, so there is no tally left to over-report and no `accepted_before_assignment` to remember.

## Tests

`test_umi_assignment_failure_does_not_overcount_accepted` asserted the warn-and-continue contract this PR replaces — it checked that the metrics agreed with an empty output BAM. It is rewritten as `test_umi_assignment_failure_fails_the_run`, keeping both `#[case]`s so the threaded (`--allow-unmapped`, via `should_use_parallel`) and single-threaded call sites each stay covered, since they swallowed the error independently.

Verified RED against the unfixed source — both cases fail with `group must fail when UMI assignment fails, not emit an empty BAM: ()` — then green after the fix.

Full suite green: 5694 passed, plus `ci-fmt` and `ci-lint`.

## Note on scope

This is a behaviour change: inputs that previously produced an empty BAM and exit 0 now fail loudly. That is the point, but it is worth a look if anyone is relying on the old behaviour to skip pathological groups. I could not find a case where a *partial* input failure is expected — the assigner errors are all "this input cannot be grouped with these settings" rather than "this particular group is odd".

## Summary by CodeRabbit

* **Bug Fixes**
  * `fgumi group` now stops with an error when UMI assignment fails instead of silently dropping affected reads.
  * Error messages provide details about the assignment failure, helping identify invalid UMI tags, input data, or strategy settings.
  * Paired-strategy RX UMI errors now clearly indicate when the expected two hyphen-delimited segments are missing.

### DIFF
--- a/src/lib/commands/group.rs
+++ b/src/lib/commands/group.rs
@@ -1329,26 +1329,21 @@ impl Command for GroupReadsByUmi {
                     filter_config.min_umi_length,
                     no_umi,
                 ) {
-                    log::warn!("UMI assignment failed, returning empty group: {e}");
-                    // The caller merges these metrics, so zero the accepted count
-                    // to match the zero records this group contributes. Otherwise
-                    // the grouping metrics report more accepted templates than the
-                    // output BAM holds.
-                    let mut filter_metrics = filter_metrics;
-                    filter_metrics.accepted_templates = 0;
+                    // A failed assignment means this position group's reads cannot be
+                    // grouped at all — an unparseable UMI for the chosen strategy, a
+                    // missing `RX` tag, mixed UMI lengths. Substituting an empty group
+                    // would silently drop every read in it and still exit 0, so fail
+                    // the run and let the user fix the input or the strategy.
                     #[cfg(feature = "memory-debug")]
                     if debug_memory_flag
                         && let Some(stats) = stats_for_tracking.as_ref() {
                             stats.track_position_group_memory(initial_group_size, false);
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #632, opened 2026-07-22
feat(group): accept --index-threshold always|never, reject unsatisfiable requests

nh13:
## What this does

`--index-threshold` was a bare `usize` meaning "index once a position group holds at least this many distinct UMIs". That left no way to turn the index off: suppressing it needed a value larger than any group, i.e. literally `--index-threshold 18446744073709551615`.

This makes the option a keyword-or-number, matching `--parallel-group-min-templates` one screen away in the same file:

```
--index-threshold always     index every position group
--index-threshold never      always scan all UMI pairs
--index-threshold <N>        index groups of N or more (default 100)
```

`never` is the new capability. `always` is sugar for `0`, which already meant this but read as its opposite — the trap that produced the wrong help text in the first place. Every existing integer invocation keeps working unchanged, `0` included, so no command line silently changes meaning.

Keeping this in one option rather than adding a `--no-index` flag makes the contradictory state unrepresentable: two knobs controlling one behaviour would need a precedence rule for `--no-index --index-threshold 50`.

## `always` is an assertion, so it can fail

A configuration that can never index is now a command-line error rather than a flag that is quietly ignored:

```
$ fgumi group --strategy identity --index-threshold always
Error: --index-threshold always cannot be honoured with --strategy identity:
the identity strategy never uses the UMI index. Drop --index-threshold to leave
indexing to the default threshold, or pass --index-threshold never to state that
a linear scan is intended.

$ fgumi group --strategy adjacency --edits 2 --index-threshold always
Error: --index-threshold always cannot be honoured with --strategy adjacency
--edits 2: the adjacency strategy only indexes at --edits 1. ...
```

That second case surfaces an asymmetry the help had left implicit: `build_adjacency_graph_bitenc` gates on `max_mismatches == 1` and `SimpleErrorUmiAssigner::assign` gates its defer/index branch the same way, so the adjacency and edit strategies both silently ignore the index at any other `--edits`, while paired indexes at all of them. `Strategy::can_use_index` now states it, and the check runs against the **effective** strategy and edits so `--no-umi` (which forces identity) is caught too.

A bare integer is deliberately not checked. It is a tuning knob allowed to end up inert — otherwise the default `100` could not coexist with `--strategy identity` — and that includes `0`, even though `0` admits every group exactly as `always` does.

`dedup` carries its own copy of the option, so the validation lives in `commands::common` and both call it. Assigner constructors take `T: Into<IndexThreshold>`, leaving every existing integer call site untouched.

## Integrating with edit's own index (#645)

`Edit` grew an index of its own in #645, gated on its own measured crossover (`EDIT_INDEX_THRESHOLD`, 200) rather than the shared default. `IndexThreshold::floored_at` expresses that floor: a numeric threshold is raised to the crossover, while the keywords pass through untouched, so `always` still means every group — the escape hatch that isolates the index in a profile — and `never` still means none. Every existing numeric invocation keeps the `max(flag, 200)` behaviour it had.

What `Strategy::can_use_index` reports for `Edit` is `edits == 1`, the same as adjacency. It is tempting to say otherwise: `components_via_index` hands `max_mismatches` straight to `NgramIndex::new`, which partitions each UMI into `max_mismatches + 1` pieces and pigeonholes over them, so the index is *capable* at any distance — a new brute-force test pins that across five UMI lengths and four distances, including the lengths `max_mismatches + 1` does not divide.

But `assign` never *reaches* it elsewhere. #645 gated the defer/index branch on one mismatch deliberately, because `EDIT_INDEX_THRESHOLD`'s crossover was measured there and nowhere else, and past it the index loses: the partitions get too short to be selective (4 bases at k=1, 2 at k=2 for an 8-base UMI) while single linkage collapses the group into one component, which makes the set-merge's early-exiting scan *cheaper* rather than dearer. Widening the gate is a benchmarking question — `benches/umi_assigner_threshold.rs` sweeps one mismatch only.

So both `edits == 1` gates now read from the assigner that enforces them, `SimpleErrorUmiAssigner::indexes_at_edit_distance` and its adjacency twin, rather than being restated in `can_use_index`, in the `Index threshold:` startup line, and in the parity test's rationale. Restating them is how they came to disagree in the first place: the index's capability at k>1 was read as evidence that `assign` used it there. The parity test stays at one mismatch — at any other distance both sides run the identical set-merge, so sweeping wider would compare the scan against itself.

## The gate has a name now

Rather than only rewording the help, the first commit extracts the condition so it can be tested rather than restated:

- `AdjacencyUmiAssigner::uses_index(num_umis)` — the threshold **and** the `max_mismatches == 1` restriction; `build_adjacency_graph_bitenc` calls it.
- `PairedUmiAssigner::uses_index(num_umis)` — the threshold only, reading the same `index_admits` gate `build_adjacency_graph` runs rather than a parallel copy of it.
- `SimpleErrorUmiAssigner::uses_index(num_umis)` — the threshold only, added here alongside edit's integration.

## Tests

- `index_threshold.rs` — parsing (both keywords, mixed case, integers, rejected input naming the accepted forms), `admits`, `demands_indexing`, `floored_at`, and `Display` round-trips.
- `test_strategy_can_use_index` — the strategy/edits matrix that decides whether `always` is honourable.
- `test_edit_index_is_built_at_every_edit_distance` — pinned against a real `NgramIndex` build at 0–3 mismatches, not against the flag.
- `test_edit_index_matches_scan` — `always` vs `never` produce identical molecule ids, swept across 1–3 mismatches. Naming the two sides as keywords also removes the old failure mode where a pair of numbers could silently compare the scan against itself.
- `test_index_threshold_always_rejected_when_index_unreachable` / `..._accepted_when_satisfiable` on `group`, and the equivalent end-to-end pair on `dedup`.

Full suite green: 6709 passed, plus `ci-fmt`, `ci-lint`, and docs with `RUSTDOCFLAGS=-D warnings`.

## Not addressed here

**`--edits > 1` has no index on the adjacency path.** The BK-tree branch that would serve k>1 exists but is only reachable from the paired strategy's generic builder, so `--edits 2` inherits the O(u²) behaviour on large position groups. `--index-threshold always` now reports this as an error rather than ignoring it, which is the point — but wiring the BK-tree into the BitEnc path is a separate change and wants a benchmark first.

## Summary by CodeRabbit

* **New Features**
  * Introduced an `IndexThreshold` configuration for `--index-threshold` (`always`, `never`, or minimum UMI-group size) with consistent indexing eligibility across strategies and edit distances.
* **Bug Fixes**
  * Invalid `--index-threshold always` combinations are now rejected with clear, strategy-specific error messages instead of being ignored.
  * Startup messaging now accurately reflects when indexing can be used (e.g., edit-indexing only at `--edits 1` where applicable).
* **Tests**
  * Expanded parsing, strategy eligibility, and command-level validation for `--index-threshold` semantics, including adjacency constraints and `Never` handling.

### DIFF
--- a/crates/fgumi-umi/src/assigner.rs
+++ b/crates/fgumi-umi/src/assigner.rs
@@ -205,11 +205,14 @@
 //!
 //! - **Use Edit** if: You have moderate error rates and transitive clustering is
 //!   acceptable for your application. At one mismatch (`--edits 1`), once a
-//!   position reaches `--index-threshold` distinct UMIs -- floored at
-//!   [`EDIT_INDEX_THRESHOLD`], edit's own measured crossover -- candidate
-//!   neighbours come from an N-gram index rather than from the incremental
-//!   set-merge, so large position groups no longer cost quadratic time. Other
-//!   edit distances always use the set-merge.
+//!   position reaches `--index-threshold` distinct UMIs -- a numeric threshold
+//!   being floored at [`EDIT_INDEX_THRESHOLD`], edit's own measured crossover --
+//!   candidate neighbours come from an N-gram index rather than from the
+//!   incremental set-merge, so large position groups no longer cost quadratic
+//!   time. Every other `--edits` value always uses the set-merge, whatever the
+//!   threshold: the crossover was measured at one mismatch, and past it the
+//!   index's partitions get too short to be selective while single linkage
+//!   collapses the group, which makes the scan cheaper rather than dearer.
 //!
 //! - **Use Adjacency** if: You have high sequencing depth (10-100x+ per molecule),
 //!   need conservative error correction, or are doing variant calling where
@@ -229,10 +232,7 @@ use std::sync::atomic::{AtomicU64, Ordering};
 use crate::MoleculeId;
 use fgumi_dna::BitEnc;
 
-/// Default minimum number of unique UMIs to use N-gram/BK-tree index for candidate search.
-/// Below this threshold, linear scan is used. Based on benchmarks: index provides ~10%
-/// speedup even at 100 UMIs/position with negligible construction overhead.
-pub const DEFAULT_INDEX_THRESHOLD: usize = 100;
+use crate::index_threshold::{DEFAULT_INDEX_THRESHOLD, IndexThreshold};
 
 /// Minimum distinct UMIs before the `Edit` strategy builds its N-gram index.
 ///
@@ -468,10 +468,15 @@ impl NgramIndex {
         // Need at least 1 base per partition for meaningful filtering, and at
         // most 16: `extract_bits` packs 2 bits per base into a `u32`, so a wider
         // partition would silently truncate the key (and trips its own
-        // `debug_assert!` in debug builds). Only reachable at
-        // `max_mismatches == 0`, where `partition_len == umi_len`; every caller
-        // in this crate gates on `max_mismatches == 1`, so this guards the
-        // `pub` constructor rather than closing a live hole.
+        // `debug_assert!` in debug builds). Every caller in this crate gates on
+        // `max_mismatches == 1`, so `partition_len == umi_len / 2`, and the two
+        // arms differ in reachability:
+        //
+        // - `== 0` is live: a 1-base UMI splits into two 0-base pieces. The build
+        //   declines and the caller stays on its linear scan.
+        // - `> 16` is not: `BitEnc::from_bytes` rejects anything over 32 bases, so
+        //   half of an encodable UMI is always at most 16. It guards the `pub`
+        //   constructor, whose `max_mismatches == 0` gives `partition_len == umi_len`.
         if partition_len == 0 || partition_len > 16 {
             return None;
         }
@@ -596,6 +601,32 @@ pub enum Strategy {
 }
 
 impl Strategy {
+    /// Whether this strategy consults the UMI index at all, at `edits`.
+    ///
+    /// `Identity` has no index code path. `Edit` and `Adjacency` index only at
+    /// `edits == 1`; `Paired` indexes at every edit distance, N-gram for k=1 and BK-tree
+    /// for k>1.
+    ///
+    /// Both `edits == 1` gates are read from the assigner that enforces them
+    /// (`SimpleErrorUmiAssigner::indexes_at_edit_distance` and
+    /// `AdjacencyUmiAssigner::indexes_at_edit_distance`) rather than restated here, so
+    /// what this reports and what the assigner does cannot drift apart. Restating them
+    /// is how the two came to disagree once already: the N-gram index *can* serve k>1,
+    /// but neither assigner *reaches* it there.
+    ///
+    /// Callers use this to reject an [`IndexThreshold`] that
+    /// [demands indexing](IndexThreshold::demands_indexing) under a configuration that
+    /// cannot deliver it.
+    #[must_use]
+    pub fn can_use_index(&self, edits: u32) -> bool {
+        match self {
+            Strategy::Identity => false,
+            Strategy::Edit => SimpleErrorUmiAssigner::indexes_at_edit_distance(edits),
+            Strategy::Adjacency => AdjacencyUmiAssigner::indexes_at_edit_distance(edits),
+            Strategy::Paired => true,
+        }
+    }
+
     /// Create a new UMI assigner for this strategy
     ///
     /// # Arguments
@@ -611,7 +642,7 @@ impl Strategy {
     /// Panics if `edits` is non-zero when using the Identity strategy.
     #[must_use]
     pub fn new_assigner(&self, edits: u32) -> Box<dyn UmiAssigner> {
-        self.new_assigner_full(edits, 1, DEFAULT_INDEX_THRESHOLD)
+        self.new_assigner_full(edits, 1, IndexThreshold::default())
     }
 
     /// Create a new UMI assigner with thread count specified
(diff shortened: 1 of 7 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
All four findings from the last review are addressed in `e6405f64`. The three outside-diff ones have no thread to resolve, so recording them here.

**`crates/fgumi-umi/src/assigner.rs:1285-1296` — 🟠 the `can_index` gate contradicts everything this PR added about Edit. Confirmed, and fixed by narrowing the contract, not by dropping the gate.**

You were right that five sites disagreed with one gate. Picking which side to keep took some digging, and the gate wins:

- **The gate is deliberate.** #645 was explicitly scoped to one mismatch: `EDIT_INDEX_THRESHOLD`'s crossover table is labelled "index vs. scan **at one mismatch**", and the plan was to widen it *only after adding a `-e 2`/`-e 3` benchmark arm*. That arm still does not exist — `benches/umi_assigner_threshold.rs` names its group `umi_assigner/edit_1` and hardcodes `1` in every arm.
- **`main` already said so; this PR un-said it.** Before the last force-push, `index_threshold_log_message` had a `Strategy::Edit if effective_edits == 1` branch and an explicit `"not used (edit indexes only at --edits 1)"`; `SimpleErrorUmiAssigner::uses_index` ended in `&& self.max_mismatches == 1`; the module doc said "At one mismatch (`--edits 1`)…". This PR replaced all three with the inverse. The reasoning error was reading `NgramIndex`'s *capability* at k>1 as evidence that `assign` *reaches* it there.
- **Dropping the gate would be a performance regression, not just an unmeasured risk.** At k≥2 on 8-base UMIs two effects compound against the index: `partition_len` collapses 4 → 2, so a bucket holds `n/16` instead of `n/256`; and single linkage collapses the group to one component, so `merge_umi_into_sets`' `set.iter().any(...)` early-exits sooner and the *scan gets cheaper* (≈3,870 → 962 → 197 comparisons/UMI at k=1/2/3 for 12,967 distinct). Modelled index-to-scan work ratio: **0.03× at k=1, 2.4× at k=2, 15× at k=3**. It stays a win at `L ≥ 12`, so the right shape is length-dependent — which is the argument against making it unconditional. `PairedUmiAssigner` already reflects this, switching to `BkTree` above k=1.

You also noted this is *not* a correctness question, and that is right: `find_within` re-verifies each candidate's Hamming distance, and pigeonhole gives no false negatives. That fact is now pinned rather than asserted — `test_ngram_index_matches_brute_force` compares the index's pair set against a brute-force one across five UMI lengths and four distances, including the lengths `max_mismatches + 1` does not divide (`partition_len` is a floor, so an 8-base UMI at k=2 leaves the last two bases uncovered). Verified load-bearing: restricting `find_within` to one partition kills it at every k≥1. The lengths deliberately include ones where random UMIs have *no* neighbours, so neighbours at exactly distance k are planted and their absence is asserted — otherwise the recall half would pass while checking nothing.

So both `edits == 1` gates now read from the assigner that enforces them — `SimpleErrorUmiAssigner::indexes_at_edit_distance` and its adjacency twin — instead of being restated in `can_use_index`, in the startup log line, and in a test rationale. Restating them is how they drifted; `assign`'s `can_index` is now derived from the same predicate the CLI validates against.

Beyond the five sites you listed, four more needed the same correction. One is a bug in its own right: **`validate_index_threshold`'s `reason` match sent `Strategy::Edit` into `_ => "the edit strategy never uses the UMI index"`, which is false** — edit does index, just at `--edits 1`. The Adjacency arm is widened, so `--strategy edit --edits 2 --index-threshold always` now says:

> `--index-threshold always cannot be honoured with --strategy edit --edits 2: the edit strategy only indexes at --edits 1. Drop --index-threshold to leave indexing to the default threshold, or pass --index-threshold never to state that a linear scan is intended.`

The others: the `NgramIndex::new` guard comment (under the gate, `partition_len > 16` is unreachable in-crate — `BitEnc` caps at 32 bases so half is always ≤ 16 — while `== 0` is live for a 1-base UMI); `test_strategy_can_use_index`'s `edit_zero_edits`/`edit_two_edits` expectations; and a `group.rs` case that moved from "accepted" to "rejected". Both commit message and PR body are corrected too — they carried the same inverted claim.

**`src/lib/commands/common.rs:135-175` — 🟡 the Edit log line lies at `--edits != 1`. Fixed.** Downstream of the above, as you said; since the gate stays, the Edit arm gets the "not used" branch Adjacency has, and the pinned `edit_two_edits`/`edit_zero_mismatches` cases now expect it. Added a case for `never` at two mismatches, where the edit-distance condition outranks the keyword.

**`src/lib/commands/group.rs:482-501` — 🟡 parallel assigners ignore `index_threshold`. Fixed in the help text.** Confirmed: none of `ParallelEditAssigner`/`ParallelAdjacencyAssigner`/`ParallelPairedAssigner` consults an index at all — all three go through `discover_edges_parallel_k`, which enumerates neighbours and hash-looks-them-up. Two details worth naming, since both would have made the disclosure wrong:

- It is conditional on `--threads > 1`. `create_umi_assigner` falls back to the sequential assigner at one thread regardless of `use_parallel`, so `--threads 1 --allow-unmapped` *does* honour the threshold.
- Only `group` has a parallel path. `dedup` always builds via `new_assigner_full`, so its help gets the Edit-gate correction but not this caveat.

`validate_index_threshold` still cannot see the per-group `use_parallel` decision, so this stays documentation rather than validation — as you suggested.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #633, opened 2026-07-22
fix(sort): honour --compression-level when the sort fits in memory

nh13:
## The bug

Output BGZF blocks are compressed by the worker pool, and which of a worker's two compressors gets used is decided by the pool's phase — `SortStep::CompressOutput` (Phase 2) selects `output_compressor`, `SortStep::CompressSpill` (Phase 1) selects the `temp_compression` one:

```rust
phase::PHASE1 => &[SortStep::DecompressInput, SortStep::ReadInputBlocks, SortStep::CompressSpill],
phase::PHASE2 => &[SortStep::CompressOutput, SortStep::Phase2FileWork],
```

Nothing moved the pool into Phase 2 when there was nothing to merge. The `chunk_files.is_empty()` fast paths in `sort_coordinate`, `sort_queryname` and `sort_template_coordinate` write straight through `PooledBamWriter` with the pool still in Phase 1, so every output block was compressed at `temp_compression` (default 1) and `--compression-level` was silently discarded.

Measured on a real 830 MB BAM, sorted entirely in memory:

```
fgumi sort -i schmitt-abl1.aligned.bam --order coordinate --max-memory 8G --threads 8
  before:  -c 1 -> 648,124,250 bytes    -c 9 -> 648,124,250 bytes     (byte-identical)
  after:   -c 1 -> 648,124,251 bytes    -c 9 -> 551,596,268 bytes     (-15%)
```

Record content is unchanged — 13,846,696 records both ways, `samtools view` output compares identical. Only the BGZF level differs.

Spilling sorts were always correct: they enter Phase 2 to read their chunks and pick up the right compressor on the way. That is why this went unnoticed — it only bites when the input fits in `--max-memory`, which for a default 768 MB/thread × 8 threads is a good fraction of routine use.

The indexing fast path (`create_indexing_bam_writer`) was never affected; it takes `self.output_compression` directly and does not go through the pool.

## The fix

All four sort functions already marked the transition:

```rust
// Ingest is done: everything from here is Phase 2 (merge/write), so
// lift the active-worker cap to `merge_threads`. ...
pool.set_active_workers(self.phase2_threads());
```

The comment was right; the code just never told the pool its phase had changed. Since widening the pool *without* the phase change is precisely the broken state, both halves are folded into `SortWorkerPool::begin_phase2(active_workers)` and called at those four sites.

`merge_chunks_generic` also stops making Phase 2 conditional on having disk chunks. The `MainThreadChunkConsumer` is still built only for `PoolDisk` sources — there is nothing to consume when everything is in memory — but the `Phase2Guard` is always armed, so teardown back to `LEGACY` is uniform across both paths instead of only running when spill files existed.

Entering Phase 2 with no spill files is cheap: `try_phase2_file_work` sees an empty file vector and returns `InputEmpty` on the first line, so output compression is the only Phase 2 work available.

## Tests

`crates/fgumi-sort/tests/integration/test_output_compression.rs` sorts the same 5,000-pair input at `output_compression` 0 and 9 with `temp_compression` pinned to 1, and asserts the level-9 output is smaller. Two `#[case]`s cover both merge paths, asserting `chunks_written` to prove which one ran.

Before the fix the in-memory case fails and the spilled case passes:

```
output_compression was ignored: level 9 wrote 94600 bytes and level 0 wrote 94600 bytes
(spill=false). Equal sizes mean the output was compressed at temp_compression instead.
```

Full suite green: 5696 passed, plus `ci-fmt` and `ci-lint`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured the configured BGZF `output_compression` level is applied consistently to the final BAM, including both in-memory and spill-based merge/write paths.
  * Improved the Phase 1 → Phase 2 transition so output blocks are routed through the correct output compression flow, preventing cases where temp/spill compression could be used instead.

* **Tests**
  * Added a new integration test covering `output_compression` levels (0 vs 9), validating record preservation, sort order, spill behavior expectations, and that level 9 produces a smaller BAM.

### DIFF
--- a/crates/fgumi-sort/src/external.rs
+++ b/crates/fgumi-sort/src/external.rs
@@ -1773,6 +1773,23 @@ impl RawExternalSorter {
         self.merge_threads.unwrap_or(self.threads).max(1)
     }
 
+    /// Transition the worker pool from ingest (Phase 1) to output (Phase 2).
+    ///
+    /// Every sort path calls this exactly once, immediately after ingest
+    /// completes, to hand the pool over to Phase 2 and lift the active-worker cap
+    /// to `merge_threads`. Centralized here so the invariant cannot drift between
+    /// sort orders: it is precisely the transition whose omission left the
+    /// in-memory output writing at the wrong compression level.
+    ///
+    /// Workers idled by the Phase-1 cap re-check within `MAX_BACKOFF_US`, so no
+    /// explicit wake is needed; capped workers always drained their held items, so
+    /// raising the cap can never be required for correctness, only for width. The
+    /// phase half is what routes output blocks to `output_compressor` — see
+    /// [`SortWorkerPool::begin_phase2`].
+    fn enter_output_phase(&self, pool: &SortWorkerPool) {
+        pool.begin_phase2(self.phase2_threads());
+    }
+
     /// Configured maximum number of temporary spill files kept before the
     /// oldest runs are consolidated into one (`0` means unlimited). Exposed so
     /// callers can assert how their `max_temp_files` setting resolved.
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #635, opened 2026-07-22
perf(consensus): memoize the unanimous fast-path bound per pre-UMI rate

nh13:
Stacked on #634 — please read the two together, and **land them together**. #634 alone is a large regression on every `simplex`/`duplex`/`codec` run.

## What #634 introduces

#634 correctly replaces the hardcoded `FAST_PATH_THRESHOLD = 23.0` with a bound derived from the pre-UMI error rate. That derivation bisects the full quality chain: ~200 iterations, each evaluating `ln_sum_exp_array` + `ln_normalize` + `ln_not` + `ln_error_prob_two_trials` + `ln_prob_to_phred` — on the order of a couple of thousand transcendental evaluations.

It runs in `ConsensusBaseBuilder::new`, and `consensus_umis` constructs a fresh `SimpleConsensusCaller`, and hence a fresh builder, for **every UMI family** (`simple_umi.rs:236-244` → `:46`; called from `vanilla_caller.rs:1517`, `duplex_caller.rs:1241`, `duplex_caller.rs:1732`, `codec_caller.rs:1511`). So the bisection was being paid per family.

## The change

The bound is a pure function of the pre-UMI error rate, and `PhredScore` is a `u8`, so 256 `OnceLock` slots cover the input domain exactly — no clamping, and no behaviour change for the out-of-range values the CLI already rejects. This is the same shape as the post-UMI table memoization in #626.

The cache is keyed on the Phred score rather than on an already-converted `LogProbability` so the cache key and the value provenance cannot drift apart.

## Output is unchanged

The memoized value is the same `f64` the direct computation produces. Verified two ways:

- bitwise agreement across the **whole `u8` domain** (`test_cached_gap_matches_uncached_across_the_whole_u8_domain`), and
- a `-1`/`-2` parameter sweep on real data — 15 combinations of pre ∈ {40, 50, 70, 90, 93} × post ∈ {30, 40, 90} on a grouped idt-cfdna BAM, **record bodies byte-identical** in every one. That sweep is the direct test of the cache key, since the key is exactly `-1`.

## Measured

`fgumi simplex --min-reads 1 --threads 1` over `idt-cfdna.group.adjacency.fgbio.bam`, three repetitions:

| | user CPU |
|---|---|
| #634 | 3.23 s, 3.23 s, 3.36 s |
| #634 + this | 2.10 s, 2.13 s, 2.11 s |

~35% less user CPU, with `samtools view | md5` identical on both sides.

## Note on scope

The constructor still builds `adjusted_correct_table`/`adjusted_error_per_alt` with 94 `ln_error_prob_two_trials` calls per builder — the identical "expensive pure setup keyed on a `u8`" shape. That is #626's scope, not this PR's; the two are complementary and should be read together.

## Summary by CodeRabbit

* **Performance**
  * Improved consensus-building efficiency by caching the computed unanimous-gap threshold per pre-UMI Phred/error-rate input.
* **Reliability**
  * Ensured the stored unanimous-gap threshold is derived solely from the configured pre-UMI error rate.
* **Tests**
  * Added validation that cached results match direct calculations across the full Phred `u8` range and that threshold values remain correct across different pre/post combinations.

### DIFF
--- a/crates/fgumi-consensus/src/base_builder.rs
+++ b/crates/fgumi-consensus/src/base_builder.rs
@@ -193,6 +193,7 @@ use crate::phred::{
 };
 use approx::abs_diff_eq;
 use std::cmp::Ordering;
+use std::sync::OnceLock;
 use wide::f64x4;
 
 /// The four DNA bases in the order used throughout consensus calling
@@ -249,8 +250,9 @@ pub struct ConsensusBaseBuilder {
     /// Smallest winner-minus-loser log-likelihood gap at which the unanimous fast
     /// path is provably exact for this `ln_error_pre_umi`.
     ///
-    /// Derived once in [`Self::new`] by [`min_exact_unanimous_gap`]; see there for
-    /// why a single constant cannot serve every pre-UMI error rate.
+    /// Derived once per pre-UMI error rate by [`cached_min_exact_unanimous_gap`];
+    /// see [`min_exact_unanimous_gap`] for why a single constant cannot serve every
+    /// pre-UMI error rate.
     min_unanimous_gap: f64,
 }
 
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #636, opened 2026-07-22
feat(group): accept --index-threshold always|never, reject unsatisfiable requests

nh13:
**Stacked on #632** — review that first; this PR's diff is against it.

## Why

#632 corrected the `--index-threshold` help, which had claimed "Set to 0 to always use linear scan" when `0` in fact indexes *every* group. But correcting the docs **removed a documented capability rather than delivering it**: the gate is `num_umis >= index_threshold`, so there is no value that disables the index short of one larger than any position group — literally `--index-threshold 18446744073709551615`.

## What

Make the option a keyword-or-number, matching `--parallel-group-min-templates` one screen away in the same file (`Auto | Fixed(usize)`, with a hand-written `FromStr`):

```
--index-threshold always     index every position group
--index-threshold never      always scan all UMI pairs
--index-threshold <N>        index groups of N or more   (default 100)
```

- **`never` is the new capability.**
- **`always` is sugar** for `0`, which already meant this but read as its opposite — precisely the trap that produced the wrong docs. Once `always` exists, nobody needs to write or interpret a bare `0` again.

**Backward compatible:** every existing integer invocation keeps working unchanged, `0` included. No one's command line silently flips meaning — which is the decisive argument against the alternative of redefining `0` to mean "off".

**Why not `--no-index`?** Two knobs controlling one behaviour needs a precedence rule for `--no-index --index-threshold 50`. Keeping it in one option makes the contradictory state unrepresentable. (`--index-threshold -1` was also considered: it needs an `isize` widening plus `allow_negative_numbers`, and puts a third magic value into an integer whose existing magic value already caused this bug.)

## Unsatisfiable requests are now errors

`always` asserts that indexing will happen, so a configuration that can never index is rejected rather than accepted and ignored:

```
$ fgumi group --strategy identity --index-threshold always
Error: --index-threshold always cannot be honoured with --strategy identity: the identity
strategy never uses the UMI index. Drop --index-threshold to leave indexing to the default
threshold, or pass --index-threshold never to state that a linear scan is intended.

$ fgumi group --strategy adjacency --edits 2 --index-threshold always
Error: --index-threshold always cannot be honoured with --strategy adjacency --edits 2: the
adjacency strategy only indexes at --edits 1. ...
```

That second case surfaces an asymmetry the help had left implicit and #632 could only describe: **adjacency silently ignores the index at any `--edits` other than 1**, while paired indexes at all of them. `Strategy::can_use_index(edits)` now states it in code.

The check runs against the **effective** strategy and edits, so `--no-umi` (which forces identity) is caught too.

**A bare integer is deliberately not checked.** It is a tuning knob allowed to end up inert — otherwise the default `100` could not coexist with `--strategy identity`. That includes `0`, even though `0` admits every group exactly as `always` does. This asymmetry is the one judgement call in the PR; the alternative (erroring on `0` too) is defensible and I'll change it if you prefer.

## Structure

- `IndexThreshold` lives in `fgumi-umi` next to the assigners that consult it, with `FromStr`, `Display` (round-trips), `Default`, `From<usize>`, `admits()` and `demands_indexing()`.
- `dedup` carries its own copy of the option, so `validate_index_threshold` lives in `commands::common` and both commands call it.
- Assigner constructors take `T: Into<IndexThreshold>`, so all 41 existing integer call sites are untouched.

## Tests

- 33 cases on the type: parsing (both keywords, mixed case, integers), rejection messages naming the accepted forms, `admits()` across the boundary, `demands_indexing()`, and `Display` round-trip.
- The #632 assigner tables extend into it — `#[case::usize_max_never_indexes(usize::MAX, …)]` becomes `#[case::never_skips_huge_group(IndexThreshold::Never, …)]`, which is what that case was always trying to say — plus a new case proving `never` wins at every edit distance for both strategies.
- CLI: `group` rejects `always` under identity/edit/adjacency-with-edits≠1 and accepts every satisfiable combination; `dedup` gets the same two tests through `try_parse_from`, since it has its own copy of the flag.
- All three error paths verified against the built binary, including the clap parse error for a bad value.

Full suite green: 5746 passed, plus `ci-fmt`, `ci-lint`, `ci-doc` (`RUSTDOCFLAGS=-D warnings`) and `ci-doctest`.

## Summary by CodeRabbit

- **New Features**
  - Added flexible `--index-threshold` settings: minimum UMI count, `always`, or `never`.
  - Applied threshold controls to `dedup` and `group` commands.
  - Added clear command-line validation for incompatible strategy and threshold combinations.
  - Exposed the threshold configuration for broader application use.

- **Bug Fixes**
  - Prevented unsupported indexing requests from proceeding and provided actionable error messages.
  - Improved handling of indexing decisions for different UMI assignment strategies.

- **Tests**
  - Added coverage for valid, invalid, and incompatible threshold configurations.

### DIFF
--- a/crates/fgumi-umi/src/assigner.rs
+++ b/crates/fgumi-umi/src/assigner.rs
@@ -224,10 +224,7 @@ use std::sync::atomic::{AtomicU64, Ordering};
 use crate::MoleculeId;
 use fgumi_dna::BitEnc;
 
-/// Default minimum number of unique UMIs to use N-gram/BK-tree index for candidate search.
-/// Below this threshold, linear scan is used. Based on benchmarks: index provides ~10%
-/// speedup even at 100 UMIs/position with negligible construction overhead.
-pub const DEFAULT_INDEX_THRESHOLD: usize = 100;
+use crate::index_threshold::{DEFAULT_INDEX_THRESHOLD, IndexThreshold};
 
 /// A node in the BK-tree.
 ///
@@ -545,6 +542,26 @@ pub enum Strategy {
 }
 
 impl Strategy {
+    /// Whether this strategy consults the UMI index at all, at `edits`.
+    ///
+    /// `Identity` and `Edit` have no index code path. `Adjacency` indexes only at
+    /// `edits == 1` — `build_adjacency_graph_bitenc` gates on it, and the BK-tree branch
+    /// that would serve larger distances is reachable only from the generic
+    /// `build_adjacency_graph` the paired strategy uses. `Paired` therefore indexes at
+    /// every edit distance.
+    ///
+    /// Callers use this to reject an [`IndexThreshold`] that
+    /// [demands indexing](IndexThreshold::demands_indexing) under a configuration that
+    /// cannot deliver it.
+    #[must_use]
+    pub fn can_use_index(&self, edits: u32) -> bool {
+        match self {
+            Strategy::Identity | Strategy::Edit => false,
+            Strategy::Adjacency => edits == 1,
+            Strategy::Paired => true,
+        }
+    }
+
     /// Create a new UMI assigner for this strategy
     ///
     /// # Arguments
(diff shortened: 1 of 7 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #637, opened 2026-07-22
perf(consensus): collapse ln_sum_exp to one addition when the operands are equal

nh13:
Equal operands reach `ln_sum_exp` constantly. Every lane of a never-observed base carries the identical value by construction — same SIMD lane inputs, same Kahan operations — so `ln_sum_exp_array` folds equal pairs on a large share of calls. For that case `log1pexp(0)` is `ln_1p(exp(0))` = `ln_1p(1)` = `ln 2`, so the transcendental pair collapses to a single addition.

## Why this is safe

The short-circuit is bit-identical to the general path for every finite operand and for negative infinity (the existing guards absorb `-inf` before the new check runs). `LOG1PEXP_ZERO` has to be spelled as a constant because `f64::ln_1p` is not a `const fn`, so `test_log1pexp_zero_constant_is_bit_exact` pins it to `log1pexp(0.0)` **bitwise** — a platform whose `ln_1p(1.0)` is not correctly rounded fails the build instead of silently shifting consensus qualities.

`clippy::float_cmp` is `#[expect]`-ed rather than worked around: exact equality is precisely the precondition being exploited, since only bitwise-equal operands make `ln_b - ln_a` exactly `±0.0`. An approximate comparison would change results.

## One input does change

`ln_sum_exp(+inf, +inf)` previously computed `inf - inf = NaN` and returned `NaN`; it now returns `+inf`. That is the correct sum, and no fgumi caller can reach it — likelihoods are `<= 0`, `-inf`, or `NaN`. But `ln_sum_exp` is public, so the new answer is pinned by a test and recorded in the rustdoc under the `# Divergence from fgbio` convention the file already uses for `ln_a_minus_b`.

## Measured

`fgumi simplex --min-reads 1 --threads 1` over `twist-umi.group.adjacency.fgbio.bam` (550 MB), three repetitions:

| | user CPU |
|---|---|
| `main` | 35.86 s, 35.88 s, 36.07 s |
| this branch | 35.04 s, 35.07 s, 34.91 s |

~2.6% less user CPU, with non-overlapping ranges. Record bodies (`samtools view \| md5`) are byte-identical, on that run and across a `-1`/`-2` sweep (pre ∈ {45, 70, 93} × post ∈ {30, 90}) on a grouped idt-cfdna BAM.

On the smaller idt-cfdna input the effect is inside the noise floor (2.22-2.26 s vs 2.22-2.24 s) — reporting that too, since it is the honest picture: the win needs enough consensus work to resolve.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved numerical stability when summing equal-valued logarithmic inputs by using an optimized, bit-exact path.
  * Clarified special floating-point behavior (including signed zero) and standardized results for edge cases like infinities and NaN.
  * Updated documented handling for the `+inf/+inf` scenario to avoid NaN.
* **Tests**
  * Added coverage to verify bitwise equivalence against `log1p(exp(0))`, correctness across signed-zero/equality cases, and special-value parity checks.

### DIFF
--- a/crates/fgumi-consensus/src/phred.rs
+++ b/crates/fgumi-consensus/src/phred.rs
@@ -15,6 +15,14 @@ use std::f64::consts::LN_10;
 /// Natural log of 2, used as threshold in log1mexp (Equation 7)
 const LN_TWO: f64 = std::f64::consts::LN_2; // ln(2) ≈ 0.693
 
+/// `log1pexp(0)`, the value [`ln_sum_exp`] adds when both operands are equal.
+///
+/// `log1pexp(0) = ln_1p(exp(0)) = ln_1p(1) = ln 2`. It is spelled as a constant
+/// because `f64::ln_1p` is not a `const fn`, and pinned to `log1pexp(0.0)` bit for
+/// bit by `test_log1pexp_zero_constant_is_bit_exact` — that test is what makes the
+/// short-circuit safe on a platform whose `ln_1p` is not correctly rounded.
+const LOG1PEXP_ZERO: f64 = LN_TWO;
+
 /// Natural log of 4/3, used in probabilityOfErrorTwoTrials
 const LN_FOUR_THIRDS: f64 = 0.287_682_072_451_780_9; // ln(4/3)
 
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #638, opened 2026-07-22
test(umi): pin id assignment in assign_with_invalid_fallback

nh13:
`assign_with_invalid_fallback` backs four assigner paths — the sequential identity, edit and adjacency assigners all route non-encodable UMIs through it (`assigner.rs:1096`, `:1581`, `:1619`, `:1641`) — but had no direct tests. It was covered only incidentally, through whole assigner runs.

This adds a case table over the interleavings that actually discriminate: a rejected UMI first, last, and between resolved ones; a rejected UMI repeated after a run of resolved ones; and case-folded repeats.

## The one case that carries the weight

`repeat_rejected_then_new_rejected`. A repeated rejected UMI must not consume an id, and that is **only** observable when a second *distinct* rejected UMI follows — a minted-then-discarded id never appears in the output itself, so every other interleaving passes even when the mint is eager.

I found this by mutation-testing the table rather than assuming it: rewriting `or_insert_with(&mut mint)` as `or_insert(mint())` initially left the whole table green. With the added case, that mutation fails exactly one case and nothing else.

## Why tests and not the refactor

This started as a perf change — making the `AHashMap` lazy, since the fallback fires on a tiny fraction of records. That is not worth doing: `AHashMap::new()` does not allocate and costs ~5.6 ns, so on a per-position-group call the saving is unmeasurable, and threading an `Option` through four sites would cost readability for nothing. The tests are the part that has value, so they land on their own.

## Summary by CodeRabbit

- **Tests**
  - Added regression coverage for identifier assignment when some UMIs are rejected during resolution.
  - Verified resolved UMIs receive the exact identifiers produced by resolution (based on first-seen order), unaffected by interleaving with rejected entries.
  - Confirmed rejected UMIs mint a single shared identifier per distinct case-folded spelling, reused across repeats, with minting performed only once per distinct rejected form.

### DIFF
--- a/crates/fgumi-umi/src/assigner.rs
+++ b/crates/fgumi-umi/src/assigner.rs
@@ -3808,4 +3808,101 @@ mod tests {
             );
         }
     }
+
+    /// `assign_with_invalid_fallback` must mint an id only for UMIs `resolve`
+    /// rejects, share one id between repeats of the same rejected UMI (case-folded),
+    /// and leave resolved UMIs untouched — whatever the interleaving. The cases put
+    /// rejected UMIs before, between and after resolved ones, and repeat a rejected
+    /// UMI across a run of resolved ones, because that is where an eager mint (or a
+    /// map built at the wrong time) shifts ids.
+    ///
+    /// **This asserts fgumi's intentional divergence from fgbio for invalid UMIs,
+    /// not fgbio identity.** fgbio's `GroupReadsByUmi` discards every SAM record
+    /// whose UMI contains an `N` *before* assignment (fgbio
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #639, opened 2026-07-22
perf(group): validate mate CIGARs from the decoded key, not a second aux walk

nh13:
## The redundancy

`RecordPositionGrouper::validate_mc_tag` ran `find_mc_tag_in_record` on every eligible record — a full walk of the aux TLV block plus a `str::from_utf8` — to confirm an MC tag was present. The parallel Decode step had already answered that question: `compute_group_key_from_raw` resolves the mate position from MC (`bam.rs:534-540`) and falls back to a single-ended `GroupKey` when it cannot (`:554-557`).

So the pipeline's hottest serial step was re-deriving, with a second aux walk, a fact a parallel step already had. This reads `GroupKey::has_mate_position()` instead.

It also drops `mc_validated`, which was written on every validated record and read only by tests.

## Measured

Quiet `c8g.4xlarge` (16 vCPU, gp3 3000 IOPS / 500 MB/s, `TMPDIR` on EBS), template-coordinate sorted twist-umi BAM, **7,868,052 accepted records** on every run. `fgumi group -s adjacency --threads 1`, 7 reps each, interleaved:

| | user CPU (s) | min |
|---|---|---|
| `main` | 20.18 20.26 20.35 20.26 20.32 20.32 20.24 | 20.18 |
| this PR | 19.87 19.85 19.84 19.72 19.84 19.90 19.89 | 19.72 |

Ranges do not overlap — about **2%**. A separate arm with the validation *deleted outright* (the strict ceiling) bottomed out at 19.72 s, so this captures essentially all of the available saving rather than trading one cost for another.

**Wall clock does not move**, and I want to be explicit about that because a profile of the serial step alone suggests otherwise. At 8, 12 and 16 threads the wall distributions of the two arms overlap completely. At `--threads 12` this removes ~1.0 s of user CPU — all of it from the serial Group thread — while wall moves ~0.2 s. The thread pool, not this step, is the wall-clock limiter on that hardware. This is a CPU saving; it is not a throughput win.

Output is **byte-identical**: same whole-file md5 from both arms when run with an identical output path (the path is embedded in `@PG CL:`, so it has to be held constant for the comparison to mean anything).

## One deliberate behavior relaxation

Validation now trusts the decoded key rather than re-reading the tag bytes, and the two lookups do not walk aux data the same way:

- `find_tag_position` returns the **first entry whose tag name matches, whatever its type byte**, and `find_string_tag` then rejects it for not being `Z`.
- `extract_aux_string_tags` only inspects `Z` entries and **skips past** everything else.

They therefore disagree on one malformed shape — MC present **twice** with a non-`Z` copy first:

| aux | before | after |
|---|---|---|
| `MC:A:x`, `MC:Z:100M` | error | accepted |
| `MC:i:5`, `MC:Z:50M` | error | accepted |

Duplicate aux tags violate SAM §1.5, and decode already uses the `extract_aux_string_tags` answer to build the key — so this makes validation agree with the value grouping actually uses. `test_mc_presence_matches_key_shape` pins it as a recorded decision rather than leaving it to be discovered.

## Why a key-shape check rather than a flag on `DecodedRecord`

Carrying an explicit bit was the obvious alternative. It would change the return type of `pub compute_group_key_from_raw` (a semver break), touch ~25 test call sites to be genuinely compiler-enforced, and — in the cheap variant that defaults the bit to `false` — reintroduce exactly the silent-failure mode the enforcement was meant to prevent. It also would not remove the coupling, only relocate it: the bit still has to be computed from `aux_tags.mc` at the same point the key shape is decided, and unlike the key shape it can drift out of sync with the key.

Validating inside `decode_records` was rejected outright: that path is shared with the MI-tag grouper (`mi_group.rs`), which has no MC requirement, so an unconditional check there would reject input those commands accept today.

`GroupKey::paired` gains a `debug_assert!` that both strands are boolean. That is what makes the `strand2` sentinel safe to key on, and it checks the invariant at the one site capable of violating it, in every debug and test build.

## Tests

- `test_group_key_has_mate_position` — includes two mixed-sentinel struct literals. They exist because `GroupKey::single` sets all three sentinels together, so on any byte-derived key a `ref_id2`-based implementation is indistinguishable from a `strand2`-based one; without those cases that mutation survives the whole suite.
- `test_mc_presence_matches_key_shape` — differential table over real bytes through the real decode path, including the duplicate-MC divergence.
- `test_record_position_grouper_mc_validation` — five-case eligibility table. The previous tests hand-built a `GroupKey` unrelated to the record's bytes, which cannot exercise the decode-to-group contract; these derive the key from the bytes.
- `..._dedup_mode_allows_paired_secondary_without_mc` — `fgumi dedup` builds this grouper with `with_secondary_supplementary()`, which disables the `UNKNOWN_REF` early return, so paired secondary reads reach validation carrying a default-shaped key.

Nine mutations were applied and each fails at least one test: inverting `has_mate_position`; keying it on `ref_id2`; keying it on `strand2 != 0`; dropping each of the five eligibility terms; and removing the MC check entirely.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved paired-end record grouping by consistently resolving mate-position information during decoding.
  - Corrected handling of missing or duplicated mate-coordinate metadata, including fallback to single-ended grouping when appropriate.
  - Strengthened validation of paired-end strand information to prevent inconsistent grouping results.
  - Added regression coverage for paired and single-ended record shapes and edge cases.

### DIFF
--- a/src/lib/grouper.rs
+++ b/src/lib/grouper.rs
@@ -394,6 +394,14 @@ impl MemoryEstimate for RawPositionGroup {
 /// Without MC tags, R1 and R2 would get different `position_key()` values and
 /// be incorrectly split.
 ///
+/// That requirement is enforced by reading [`GroupKey::has_mate_position`] on
+/// each eligible record — **not** by re-reading the `MC` tag bytes. Anything
+/// that changes how `compute_group_key_from_raw` falls back when `MC` is absent
+/// must keep that method meaning "the mate position was resolved", or this
+/// grouper silently stops rejecting MC-less input. A caller that hand-builds
+/// [`DecodedRecord`]s with keys from some other source is likewise not
+/// validated; both in-tree producers go through `compute_group_key_from_raw`.
+///
 /// By default, secondary/supplementary reads are skipped (they have UNKNOWN
 /// position keys). Use [`with_secondary_supplementary`](Self::with_secondary_supplementary)
 /// to include them — they are coalesced by `name_hash` into the group of their
@@ -405,8 +413,6 @@ pub struct RecordPositionGrouper {
     current_group_key: Option<GroupKey>,
     /// Records at the current position.
     current_records: Vec<DecodedRecord>,
-    /// Whether MC tags have been validated on paired records.
-    mc_validated: bool,
     /// Whether to include secondary and supplementary reads in groups.
     /// When true, secondary/supplementary reads (which have UNKNOWN position keys)
     /// are kept and coalesced by `name_hash` into the group of their primary read.
@@ -424,7 +430,6 @@ impl RecordPositionGrouper {
             current_position_key: None,
             current_group_key: None,
             current_records: Vec::new(),
-            mc_validated: false,
             include_secondary_supplementary: false,
         }
     }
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #641, opened 2026-07-22
fix: use configured error rates for the consensus UMI

nh13:
> [!IMPORTANT]
> **Do not merge until [fulcrumgenomics/fgbio#1163](https://github.com/fulcrumgenomics/fgbio/pull/1163) merges.** This is a port of that PR, so it should track whatever that review lands on — if the upstream approach changes (option names, defaults, whether the metrics tools get the options at all), this PR changes with it. Held as a draft until then.

## Summary

Port of [fulcrumgenomics/fgbio#1163](https://github.com/fulcrumgenomics/fgbio/pull/1163) — fgumi had the same bug.

The consensus UMI written to the `RX` tag was called with `SimpleConsensusCaller`'s hardcoded 90/90 error rates rather than the configured `--error-rate-pre-umi` / `--error-rate-post-umi`, so neither option had any effect on the `RX` tag emitted by `simplex`, `duplex`, or `codec`. The per-base consensus itself was always correct — `ConsensusBaseBuilder::new(options.error_rate_pre_umi, options.error_rate_post_umi)` threads them properly; only the UMI consensus was disconnected.

Following the same PR, `duplex-metrics` and `simplex-metrics` gain `-1/--error-rate-pre-umi` and `-2/--error-rate-post-umi` (matching the flags, defaults, and validation of the consensus commands) so the consensus UMI they key their counts on is calculated the same way as the one written to a consensus BAM. `simplex-metrics` has no fgbio counterpart but had the identical hardcoded default, so it gets the same treatment for consistency.

## Behavior change

The metrics tools' UMI consensus moves from 90/90 to the 45/40 defaults. The bases called are unchanged in practice — at `qError=20` the per-observation quality dominates the likelihood at both settings — but it is a real change to the effective prior and worth noting.

## Reading order

1. `crates/fgumi-consensus/src/simple_umi.rs` — `consensus_umis()` takes a `UmiErrorRates` pair instead of using the defaults. The rates travel bundled rather than as two loose `u8` arguments, which would be trivially swappable at the four call sites.
2. `vanilla_caller.rs`, `duplex_caller.rs`, `codec_caller.rs` — the call sites. The duplex caller reuses the rates its single-strand caller was built with; the CODEC `RX` block moves into a `consensus_umi()` helper to keep `build_output_record_into()` under the line limit.
3. `src/lib/commands/{duplex,simplex}_metrics.rs` — the new options.

Both commits compile and test standalone, so the branch is bisectable.

## Testing

Six new `#[rstest]` case tables, each written and watched failing before the corresponding fix:

| Path | Normal rate | Post-UMI Q1 |
| --- | --- | --- |
| vanilla / duplex / CODEC `RX` | `A` | `N` |
| `duplex-metrics`, `simplex-metrics` `umi_counts` | `AAA`, `TTT` | `NNN` |

The mechanism, mirroring fgbio's regression test: at Q1 (~79% error) an observed base becomes *less* likely than each of the three unobserved ones, so the unobserved bases tie for the maximum likelihood and the position no-calls. With a multi-base UMI that collapses every position, not just the disagreeing one — hence the single `NNN` key in the metrics cases.

Plus a 4-case validation table per metrics tool covering `0` and `94` for each rate.

`cargo ci-test` (5723 passed / 0 failed), `cargo ci-fmt`, and `cargo ci-lint` all clean.

One unrelated adjustment: the `check-tag-literals` xtask flagged the new `b"q3"` read-name fixture as a bare SAM-tag byte literal, so it is allowlisted next to the existing `q1`/`q2`.

### DIFF
--- a/crates/fgumi-consensus/src/codec_caller.rs
+++ b/crates/fgumi-consensus/src/codec_caller.rs
@@ -80,7 +80,7 @@ use crate::caller::{
     write_consensus_read_name,
 };
 use crate::phred::{MIN_PHRED, NO_CALL_BASE, NO_CALL_BASE_LOWER, PhredScore};
-use crate::simple_umi::consensus_umis;
+use crate::simple_umi::{UmiErrorRates, consensus_umis};
 use crate::vanilla_caller::{
     VanillaConsensusRead, VanillaUmiConsensusCaller, VanillaUmiConsensusOptions,
 };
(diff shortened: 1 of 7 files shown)


nh13:
@coderabbitai pause

PR #643, opened 2026-07-22
feat(sort): add --max-temp-files to tune the spill-file consolidation limit

tfenne:
## What

Adds `--max-temp-files <N>` to `fgumi sort`, exposing the previously-hardcoded temp-file consolidation limit (`DEFAULT_MAX_TEMP_FILES = 64`).

- Optional `usize`, minimum 2 (values `< 2` are rejected — a merge needs at least two inputs). For effectively unlimited, pass a large value.
- Wired through `build_sorter` to the existing `RawExternalSorter::max_temp_files` builder, following the same `Option`-override pattern as `--sort-threads` / `--merge-threads`.
- **Unset → the engine default (64, samtools-compatible) is used, so behavior is byte-identical** unless you opt in.

## Why

The external sort spills sorted runs to disk; once the number of runs reaches the limit, the oldest ~half are consolidated into a single run so the final k-way merge opens fewer files. That consolidation merge is single-threaded, so on large inputs it adds real wall-time. Sorting a 1.29 B-read WGS BAM with defaults (≈6 GiB budget → ~93 spilled runs) triggered one or two consolidation folds costing roughly 15–38% of total wall-clock — overhead a higher limit eliminates. Concurrency during the final merge is bounded by `--threads`, not the file count, so raising the limit mainly trades open file descriptors for fewer consolidation passes.

## Notes for embedders

The flag declares **no clap `default_value`** and its help text states **no number**, so a tool that embeds `Sort` via `#[command(flatten)]` can set its own default (resolving `None` in code) without the flattened help advertising a wrong value. fgumi's own default (64) is documented only in fgumi's command `long_about`, which flatten does not propagate.

## Testing

`cargo ci-fmt`, `cargo ci-lint` (clippy pedantic), and `cargo ci-test` all pass (5738 tests). New tests cover parsing (unset / explicit), rejection of values `< 2`, and the `build_sorter` wiring (asserted against a freshly-constructed sorter's default rather than a hardcoded 64).

## Summary by CodeRabbit

* **New Features**
  * Added a `--max-temp-files <usize>` option to control when temporary spilled runs are consolidated.
  * Added parse-time validation requiring values ≥ 2; invalid inputs are rejected.
  * Updated command help and examples, including the default of 64 and the effect of raising/lowering the limit.
* **Documentation**
  * Updated the performance-tuning guide to describe `--max-temp-files`, its trade-offs, and its constraints.
* **Tests**
  * Added unit tests covering unset/minimum/explicit values, rejection of invalid inputs, and that the effective limit is applied as expected without changing output correctness.

### DIFF
--- a/src/lib/commands/sort.rs
+++ b/src/lib/commands/sort.rs
@@ -154,7 +154,7 @@ PERFORMANCE:
   - Configurable temp file compression (--temp-compression)
   - Default 768M per-thread memory limit (samtools-compatible); pass
     `--max-memory auto` to detect system memory (opt-in)
-  - Spilled runs are consolidated once they exceed `--max-temp-files`
+  - Spilled runs are consolidated once they reach `--max-temp-files`
     (default 64, matching samtools); raise it to avoid repeated
     consolidation passes on very large inputs
 
@@ -342,12 +342,12 @@ pub struct Sort {
     /// the final k-way merge opens fewer files at once. Raising it avoids
     /// repeated consolidation passes on very large inputs (at the cost of more
     /// open file descriptors during the final merge); lowering it keeps fewer
-    /// files open. Must be at least 2; for effectively unlimited, pass a large
-    /// value.
+    /// files open. Must be at least 2; to effectively disable consolidation,
+    /// pass a value larger than the number of runs you expect to spill.
     ///
     /// When unset, a built-in default is used (see this command's help
     /// overview for the default value).
-    #[arg(long = "max-temp-files", value_parser = parse_max_temp_files)]
+    #[arg(long = "max-temp-files", value_parser = clap::builder::RangedU64ValueParser::<usize>::new().range(2..))]
     pub max_temp_files: Option<usize>,
 
     /// Write BAM index (.bai) alongside output.
@@ -428,19 +428,6 @@ pub(crate) fn parse_key_types(s: &str) -> Result<KeyTypesSpec, String> {
         }
     }
     Ok(KeyTypesSpec::Explicit { cb, tertiary })
-}
-
-/// Parse `--max-temp-files`: a spill-file consolidation limit of at least 2.
-///
-/// Values below 2 are meaningless (a merge needs at least two inputs) and are
-/// rejected rather than silently clamped. For an effectively unlimited limit,
-/// pass a large value.
-pub(crate) fn parse_max_temp_files(s: &str) -> Result<usize, String> {
-    let n: usize = s.parse().map_err(|_| format!("`{s}` is not a non-negative integer"))?;
-    if n < 2 {
-        return Err(format!("must be at least 2 (got {n})"));
-    }
-    Ok(n)
 }
 
 impl Command for Sort {
@@ -876,6 +863,7 @@ mod tests {
     /// builder then applies the engine default).
     #[rstest]
     #[case::unset(&[], None)]
+    #[case::minimum(&["--max-temp-files", "2"], Some(2))]
     #[case::explicit(&["--max-temp-files", "256"], Some(256))]
     #[case::large(&["--max-temp-files", "100000"], Some(100_000))]
     fn test_parse_max_temp_files(#[case] extra: &[&str], #[case] expected: Option<usize>) {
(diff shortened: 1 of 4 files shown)


src/lib/commands/sort.rs
```diff
@@ -408,6 +430,19 @@ pub(crate) fn parse_key_types(s: &str) -> Result<KeyTypesSpec, String> {
     Ok(KeyTypesSpec::Explicit { cb, tertiary })
 }
 
+/// Parse `--max-temp-files`: a spill-file consolidation limit of at least 2.
+///
+/// Values below 2 are meaningless (a merge needs at least two inputs) and are
+/// rejected rather than silently clamped. For an effectively unlimited limit,
+/// pass a large value.
+pub(crate) fn parse_max_temp_files(s: &str) -> Result<usize, String> {
```
nh13: **suggestion:** should we bound the maximum number of temp files?  Or clamp it to the system maximum?

src/lib/commands/sort.rs
```diff
@@ -328,6 +334,22 @@ pub struct Sort {
...
+    /// reaches this limit, the oldest are merged together in a single pass so
+    /// the final k-way merge opens fewer files at once. Raising it avoids
+    /// repeated consolidation passes on very large inputs (at the cost of more
+    /// open file descriptors during the final merge); lowering it keeps fewer
+    /// files open. Must be at least 2; for effectively unlimited, pass a large
+    /// value.
+    ///
+    /// When unset, a built-in default is used (see this command's help
+    /// overview for the default value).
+    #[arg(long = "max-temp-files", value_parser = parse_max_temp_files)]
```
nh13: **issue:** add a `default_value_t = fgumi_sort::DEFAULT_MAX_TEMP_FILES` here so it shows up in `--help`

src/lib/commands/sort.rs
```diff
@@ -328,6 +334,22 @@ pub struct Sort {
     #[arg(long = "temp-codec", default_value = "zstd")]
     pub temp_codec: fgumi_sort::SpillCodec,
 
+    /// Maximum number of temporary spill files kept before the oldest are
```
nh13, line 337: **issue:** Document `--max-temp-files` in the `docs/src/guide/performance-tuning.md` into the "Sort" section (make sure to update the commit hash in docs/LAST_SYNCED to this PR's head commit)

src/lib/commands/sort.rs
```diff
@@ -408,6 +430,19 @@ pub(crate) fn parse_key_types(s: &str) -> Result<KeyTypesSpec, String> {
...
 
+/// Parse `--max-temp-files`: a spill-file consolidation limit of at least 2.
+///
+/// Values below 2 are meaningless (a merge needs at least two inputs) and are
+/// rejected rather than silently clamped. For an effectively unlimited limit,
+/// pass a large value.
+pub(crate) fn parse_max_temp_files(s: &str) -> Result<usize, String> {
+    let n: usize = s.parse().map_err(|_| format!("`{s}` is not a non-negative integer"))?;
+    if n < 2 {
+        return Err(format!("must be at least 2 (got {n})"));
```
nh13: **question:** I think the library allows a value of `0` to mean unlimited, I assume for the CLI (at least 2) is on purpose?

src/lib/commands/sort.rs
```diff
@@ -408,6 +430,19 @@ pub(crate) fn parse_key_types(s: &str) -> Result<KeyTypesSpec, String> {
     Ok(KeyTypesSpec::Explicit { cb, tertiary })
 }
 
+/// Parse `--max-temp-files`: a spill-file consolidation limit of at least 2.
+///
+/// Values below 2 are meaningless (a merge needs at least two inputs) and are
+/// rejected rather than silently clamped. For an effectively unlimited limit,
+/// pass a large value.
+pub(crate) fn parse_max_temp_files(s: &str) -> Result<usize, String> {
+    let n: usize = s.parse().map_err(|_| format!("`{s}` is not a non-negative integer"))?;
```
nh13: **nitpick:** you say to put in a big number, and I put in `99999999999999999999999999`, so not a negative number just overflow.  So the message is not strictly true.  Can you add overflow to the tests below?

nh13 (approved):
Minor suggestions, when you've addressed them, feel free to squash-merge

src/lib/commands/sort.rs
tfenne: So ... I didn't do this because this option struct is flattened into mako, and I'd like to raise the 64 to a higher value in mako (say somewhere in the 256-1024 range).  The cost of finding the next record to write is log(num_tmp_files), so going up considerably from 64 doesn't cost a lot.  But mako is a general purpose tool, `fgumi sort` is targeted at UMI preps that might have substantially fewer reads, and I figured you might care about retaining the 64 default.  If I put the default on here, I either can't change it in mako _or_ mako will show an incorrect default value.

How'd you feel about upping the default to 256 or 512 and then adding the default_value_t here?

nh13: Lets leave it without a default in the help, I see your point.

nh13:
Sorry for the churn on this one — I just pushed a fix for the outstanding CodeRabbit comment (added an end-to-end test that forces spill-file consolidation and asserts the output is byte-identical between the default limit and `--max-temp-files 2`, rather than only checking the accessor). Should be good to merge once CI passes. Thanks for your patience!

PR #644, opened 2026-07-22
feat(io): accept uncompressed SAM and stdin anywhere BAM is accepted

### LINKED ISSUE #642: Every command that accepts BAM must accept SAM
## The invariant

**Any command that accepts BAM input must accept uncompressed SAM input.** Input format is a property of the data, not of the command or of which flags the user happened to pass. A command that reads `-i in.bam` must read `-i in.sam`, and it must do so identically whether or not `--threads` is set.

This is a project invariant, not a feature request for a particular command. New commands and new reader paths are expected to satisfy it.

## Why it needs stating

fgumi opens input through several distinct readers:

- the single-threaded raw-byte reader (`create_raw_bam_reader_with_opts`)
- the typed reader (`create_bam_reader_with_opts`)
- the multi-threaded pipeline reader (`create_bam_reader_for_pipeline_with_opts`)
- the sort engine's pool-integrated reader (`create_raw_bam_reader_pool_integrated`)

Commands select between these based on flags — chiefly `--threads` — that have nothing to do with the input format. That structure makes it easy for format support to end up decided per command, and even per flag combination within one command, which is what happened: on `main`, no command accepted SAM at all, and on the `feat-runall` line `--threads N` accepted SAM while `--threads` unset rejected it with a hint pointing at `--threads`.

The failure mode is not just a missing feature. It is that the answer to "does this command accept SAM?" was different depending on an unrelated performance flag, and error messages were written asserting behavior that did not hold.

## Current state

Every command that takes a BAM now accepts SAM, via content-based detection and transparent transcoding at the reader boundary (`fgumi-bam-io`'s `sam_input` module). Detection reads the input's magic bytes rather than its file extension, so a `.bam` holding SAM text and a `.sam` holding BGZF both read correctly, and stdin and FIFOs work without reopening or seeking the input.

## The one exception

`fgumi review` requires BAI indexes on `--consensus-bam` and `--grouped-bam` and performs random access (`query`) against them. Indexed random access is inherently BGZF-only — a plain SAM file cannot carry a BAI — so `review` is BAM-only by construction rather than by omission.

Partial support would be worse than none here: making `review`'s sequential opens accept SAM while its indexed opens still require BGZF would reintroduce exactly the per-path inconsistency this invariant exists to prevent. If `review` should accept SAM, it needs a separate design decision (for example, a non-indexed full-scan mode), not a reader swap.

## Keeping it true

`tests/integration/test_sam_input.rs` pins the invariant: for each command, the same data as `.bam` and as `.sam` must both succeed and produce the same records, with cases run both with and without `--threads` so that no future change can make format support depend on thread count.

The table currently covers `sort`, `group`, and `fastq`, chosen to exercise all four reader surfaces above rather than to enumerate commands. Extending it to the remaining commands needs per-command fixtures (grouped input with `MI` tags for `dedup`/`downsample`/consensus, consensus tags for `filter`, a reference for `clip`), and is worth doing so the guard is by command rather than by reader.

## Related

- Output is BAM-only in every command (`-o out.sam` writes a BAM with a `.sam` name). Whether the invariant should have an output counterpart is a separate question and is not settled here.
- The `feat-runall` line carries its own SAM support through the typed-step chain (`ReadSamChunks`/`ParseSamChunk`, and `InputSource::open_with_opts`'s suffix-then-magic detection). #547 and #549 close the last two gaps there. Those branches and `main` will need reconciling; the invariant should hold on whichever survives.

nh13:
Closes #642.

Every fgumi reader was a BGZF reader, so **no command accepted uncompressed SAM**. `-i in.sam` failed with `invalid BGZF header`, which tells the user nothing about what to do, and it failed that way regardless of `--threads` — verified against the built binary on `main` before writing any code:

```
$ fgumi simplex -i in.sam -o out.bam --min-reads 1
Error: Failed to read header from: in.sam
Caused by:  invalid BGZF header

$ fgumi simplex -i in.sam -o out.bam --min-reads 1 --threads 4      # same
```

Input format is a property of the data, not of the command, and it must not depend on which reader a flag happens to select. The same argument applies to *where* the data comes from, so this PR holds stdin to the same standard: every command that streams records accepts `-`.

## Approach

Normalize at the boundary rather than teaching every consumer a second record source.

`fgumi-bam-io`'s new `sam_input` module sniffs the input's leading bytes and, for SAM text, parses and re-encodes it on the fly into the BGZF stream the consumers already expect — so `-i in.sam` and `-i in.bam` are the same stream by the time any consumer sees them. Nothing downstream of the reader changed.

The alternative — a second, text record source threaded through the single-threaded fast paths, the pipelines and the sort engine — is what produces per-command and per-flag divergence in the first place. One conversion point cannot drift.

The transcode uses `CompressionLevel::NONE`: the bytes are decoded again microseconds later in the same process, so there is nothing to gain from deflating them.

Detection is by **content, never by extension**, so a `.bam` holding SAM text and a `.sam` holding BGZF both read correctly. Only a regular file is sniffed by path and re-opened; stdin, FIFOs and process substitution cannot be re-opened, so they are opened once and the peeked bytes replayed.

### One classifier, not three

BGZF, plain gzip and text were previously distinguished by 2-, 4- and 18-byte checks in three different places, which disagreed with each other. A plain-gzipped SAM was called BGZF by one and SAM text by another: `sort` failed with `Failed to read header`, `zipper` with `invalid flags` — the exact confusion the content sniff exists to prevent. `fgumi-bam-io::format::classify_input` (`Bgzf` / `Gzip` / `Text` / `Empty`) now answers the question everywhere. Gzip-but-not-BGZF is named as such, with the `bgzip` fix in the message, instead of failing as a corrupt BAM. `fgumi extract` keeps gzip as a *supported* input, since FASTQ legitimately arrives that way.

Input that is neither BGZF nor SAM is rejected rather than accepted as headerless SAM — a SAM header parse consumes only `@` lines, so arbitrary text would otherwise yield an empty header and transcode into a valid-looking BAM carrying no reference sequences at all. Empty input is reported as empty rather than as a malformed header.

## Reader surfaces

| Reader | Used by |
|---|---|
| `create_bam_reader_with_opts` | typed whole-record consumers |
| `create_raw_bam_reader_with_opts` | single-threaded fast paths |
| `create_bam_reader_for_pipeline_with_opts` | `--threads N` pipelines |
| `create_raw_bam_reader_pool_integrated` | the sort engine |
| `open_raw_bam_record_reader_with_header` | `sort --verify`, `compare` |
| `create_bam_reader_for_pipeline` | both `unified_pipeline` entry points |
| the shared normalizing reader | `zipper`'s mapped input |

The pipeline and pool readers parsed their header by **rewinding**, which cannot work for a transcoded stream. Both now parse through a `TeeReader` and replay the consumed bytes — the same trick the stdin branch already used. Besides enabling SAM, that means the input is opened exactly once: the file branches previously opened, seeked to 0 and re-read, which never worked for FIFOs (`fgumi -i <(samtools view …)`).

`sort --verify` and `compare` read their header through a format-aware reader and then **re-opened the path with a bare `File`**, which put them back on the BGZF-only path and made them reject the very files `sort` accepted. `zipper` chose its mapped reader by **file extension** and so was the one command a misnamed file could still fool. Both are fixed; without them the PR's central claim was simply false.

## stdin

Every command that reads records streams stdin. Three deliberately do not:

| Command | Why not |
|---|---|
| `review` | requires BAI indexes and does `query()` random access |
| `merge` | takes N pre-sorted inputs via a `-b` list; one stream cannot supply N |
| `compare` | compares two named inputs against each other |

`extract` streams when there is exactly one `--input`; `-` with two or more is rejected outright.

Getting there removed three separate double-reads:

- `fastq`, `duplex-metrics` and `simplex-metrics` rejected `-i -` at validation with `File does not exist`, because existence was checked before anything knew `-` names a stream. `validate_input_exists` now exempts stdin for streamable inputs while leaving references and include lists strictly checked.
- `sort --verify` opened its input twice, once for the header and once for the records; it now takes both from a single open.
- The metrics commands pre-scanned for a consensus BAM by opening the input a second time. **They still reject consensus BAMs** — the guard now runs against the first qualifying record of the pass they already make. Its predicate is deliberately looser than the main loop's and must stay that way: it must *not* skip unmapped reads, since consensus BAMs are unaligned and skipping them would hole the very check.
- `fgumi fastq -i -` failed outright when `--output` already existed. The guard that refuses to truncate the input BAM canonicalises both paths whenever the output exists, and `canonicalize("-")` is a `NotFound` error, so accepting stdin turned the common case of overwriting a prior run's output into an unexplained IO failure. `reject_output_clobbering_input` now exempts stdin — it names no file, so it can clobber nothing.
- `extract` opened each input **three** times — a quality-encoding sample, a BGZF probe, and the real pass — so stdin was drained before the records were read and `-i -` silently produced an empty BAM while exiting zero. The sample now reads through a tee and replays what it consumed.

## `zipper` no longer has a text mode

`zipper` kept a second, text-mode orchestration because "text-mode SAM has no raw-byte path". With SAM transcoded at the boundary it does have one, so the mapped input goes through the shared normalizing reader and the command never learns which format arrived — there is no format branch left at all.

That turned out to be **faster**, which refuted the reason it had been deferred. On a 2.4M-record zip:

| input | before | after |
|---|---|---|
| SAM file | 2.9s | **2.1s** |
| SAM pipe | 2.9s | **2.28s** |
| BAM file (control) | 1.6s | 1.6s |

Output is byte-identical across all five input shapes. The transcode's encode replaces *both* the `RecordBuf` decode and the re-encode the merge used to pay, and framing at `CompressionLevel::NONE` is nearly free.

`MappedReader` collapsed from an enum to a raw reader and then disappeared; `record_bufs_to_templates`, `file_is_bgzf` and `peek_bgzf` became dead. **zipper sheds ~255 lines.**

`SharedBuffer::is_empty` no longer takes a mutex per record either. The transcode asks the sink whether the encoder has emitted a block once for every record it writes; that is now an atomic load against a length `write` and `swap_into` maintain inside the critical sections they already hold.

Separately, `TemplateIterator` no longer copies both read names into owned `Vec`s to decide whether two records belong to the same template — two heap allocations per record, on every zipper run, to answer a `memcmp`. It compares the name bytes in place (~4.8M allocations avoided on the 2.4M-record run). The *remaining* per-record allocation there — one `RawRecord` buffer per record — needs a pool with a cross-thread return path and is tracked separately in #654.

## Deletions and API surface

- `src/lib/batched_sam_reader.rs` (662 lines) — deleted. It fed an adaptive buffer whose batch tracking was driven only by `record_parsed()`, which has **zero production callers**, so `zipper --bwa-chunk-size` was already inert before this PR. The buffer also cost a full-stream copy of the SAM text and 128 MiB of reservation, for buffering the normalizing reader already does. The flag is still parsed so existing command lines keep working, and hidden so new ones do not adopt it; removing it is a CLI break and belongs in its own change. (`fgumi fastq` has its own `--bwa-chunk-size`, which is live and untouched.)
- `run_bam_pipeline_with_header` (99 lines) — deleted; zero callers anywhere, including tests.
- `run_bam_pipeline_with_grouper` — moved behind a `test-utils` feature. Production commands all drive the pipeline through the `*_from_reader` family; this path-taking wrapper's only callers are integration tests, and leaving a generic with five type parameters in the default public API advertises it as a supported entry point that every future I/O change has to be kept in step with — a cost this PR already paid once. A self dev-dependency turns the feature on for the test targets, so `cargo test` needs no extra flag and `cargo build` never sees it. (`.cargo/config.toml` already had a `cargo t = "test --features test-utils"` alias naming a feature the crate did not define; it works now.)

## `review` stays BAM-only, deliberately

`fgumi review` requires BAI indexes on `--consensus-bam` and `--grouped-bam` and does `query()` random access against them. Indexed random access is inherently BGZF-only — a plain SAM file cannot carry a BAI — so `review` is BAM-only by construction, not by omission.

It also has two *sequential* opens that I could have converted. I did not: making half of `review`'s reads accept SAM while the indexed half still requires BGZF would reintroduce exactly the per-path inconsistency this PR removes. If `review` should accept SAM it needs a design decision (e.g. a non-indexed full-scan mode), not a reader swap.

## Errors name the stage that actually failed

The transcode is lazy, so records were being pulled while the consumer was still reading the header — a malformed record on line 4 was reported as a header failure. The header is now flushed into its own BGZF block, so the header read completes before any record is touched.

## Tests

Written test-first. The three reader-factory tests were watched failing with `invalid BGZF header` before the module existed, and `sort` failed the integration table exactly as predicted until its pool reader was converted.

- `sam_input` unit tests — transcoded stream is a readable BAM; byte-at-a-time reads match bulk reads (the staging buffer refills across `read` calls); header-only SAM yields a valid empty BAM; BGZF passes through byte-identical; empty input is rejected.
- `format` unit tests — a 7-case table over BGZF / plain gzip / SAM / empty / truncated gzip magic / short text / gzip-with-a-non-`BC`-extra.
- `reader` unit tests — the public factories accept uncompressed SAM.
- `tests/integration/test_sam_input.rs` — for each command, the same data as `.bam` and as `.sam` must both succeed and produce the same records. Cases run **with and without `--threads`**, so no future change can make format support depend on thread count. Plus misnamed extensions in both directions, piped input, a FIFO regression case, the rejection cases, and record-vs-header error attribution.
- `tests/integration/test_input_source_matrix.rs` — a `CONTRACTS` table pinning the whole contract: **every command the binary advertises must declare whether it accepts SAM and whether it accepts stdin**, and the declarations are checked against `fgumi --help`. A new command fails the suite until it states its stance rather than silently inheriting whichever reader it happened to pick. The SAM axis compares **results, not exit codes**: each run writes to its own directory (both previously wrote to one `{command}.out`, so the SAM run overwrote the BAM output it was meant to be checked against) and the BAM run is the oracle.
- `sort --verify` parity compares **violations as well as records checked** — a SAM-specific record-boundary bug could preserve the record total while corrupting order detection, and a count-only comparison would pass — with a case whose input is deliberately out of queryname order.

### One stated limit: integer aux width

Comparing the two runs as *bytes* is not achievable and the matrix does not try. SAM's `i` type carries no width, so a `cD:i:10` read from SAM text is re-encoded as `Int32` while the same tag in the original BAM may have been stored as `Int8`. Every writer picks its own rule — fgumi narrows to the smallest signed type, htslib prefers unsigned, noodles does not narrow — so requiring identical aux encoding across a SAM round-trip would require something no BAM writer guarantees. The matrix therefore renders both outputs to SAM text and compares the values, which is the property that actually has to hold. Whether fgumi should narrow on the SAM path to match is tracked in #655.

`cargo ci-test` 6000 passed / 27 skipped; `cargo ci-fmt`, `cargo ci-lint`, and `cargo check --workspace` with both `--no-default-features` and `--all-features` all clean.

## What changed since the first review

The first round of review (CodeRabbit plus an adversarial pass) found that the PR's headline claim did not hold, and the fixes cascaded well past the original scope. Summarising, since this description was first written:

- **`sort --verify`, `compare` and `zipper` did not actually accept SAM**, despite being listed as covered. Fixed; the "Reader surfaces" table above is now the real list.
- **The whole stdin half is new** — the streaming contract, the three double-reads, `extract`'s tee-and-replay.
- **The integration table is no longer a sample.** The earlier text said extending it beyond `sort`/`group`/`fastq` needed per-command fixtures and was deferred to #642. It is done: `CONTRACTS` is exhaustive and self-enforcing against `fgumi --help`.
- **`zipper` was rearchitected** rather than given a format branch, and the benchmark above refuted the deferral.
- **The three-way format sniff was unified** into one classifier, and plain gzip is now rejected with an actionable message instead of being mistaken for BGZF. The earlier claim that "BGZF begins with the gzip magic and SAM text cannot" was true but insufficient — plain gzip also begins with the gzip magic, and used to be misread as BGZF.
- **Deletions and the API-surface change** in the section above.
- **A second review round** (this one) fixed the `fastq` stdin/clobber regression above, replaced the matrix's exit-code check with an output comparison, added the `sort --verify` violation parity and the duplex cases for the consensus guard, and removed the per-record mutex from the transcode.
- Test count moved from 5746/23 to 6000/27.

## Note on `feat-runall`

That line has its own SAM support via the typed-step chain (`InputSource::open_with_opts`, suffix-then-magic detection), with #547 and #549 closing its last two gaps. This PR does not touch it — it targets `main`, which has no chain at all and is where the format gate is currently total. The two lines will need reconciling; the invariant should hold on whichever survives.

## Summary by CodeRabbit

* **New Features**
  * Commands that accept BAM input now also accept uncompressed SAM directly (including stdin/FIFOs and content-based detection even with mismatched extensions), with automatic normalization to BGZF/BAM.
  * `sort --verify` now works correctly with streamed, non-seekable stdin inputs.
  * Enhanced streaming support for `extract`, `zipper`, and related pipelines.

* **Bug Fixes**
  * Improved handling of empty/malformed/non-SAM-or-BAM inputs, with clearer fault attribution.
  * Avoids hangs with non-reopenable streaming inputs (FIFO) during compare/verification-style engines.
  * Consensus handling in metrics is now record-based (instead of rejecting the whole input up front).

* **Tests**
  * Expanded coverage for stdin/FIFO behavior, format equivalence, and `sort --verify` result consistency.


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
> **Edited to stay accurate.** This comment originally described the branch as of `64e74d4c`; three of its statements were overtaken by later work in the same PR. Those are corrected in place below and flagged with _(updated)_. The full current state is in the PR description.

Addressed both outside-diff findings from the last review, plus a set of defects found while verifying them.

**Outside-diff findings**

- `crates/fgumi-sort/src/external.rs` — `advise_sequential` is now called unconditionally after opening the input, so the synchronous sort path gets the readahead hint the two `reader.rs` sites already got. The WILLNEED hints `PrefetchReader` issues remain async-only; conflating the two is what hid this.
- `crates/fgumi-bam-io/src/reader.rs` — deduped the half that was actually duplicated: `read_header_and_replay` is now `pub` in `fgumi-bam-io` and `external.rs` calls it instead of re-implementing the tee/header/replay block. I also added `open_normalized_input` there, which gave the shared opener a real second consumer. _(updated)_ I initially declined to unify the *open* half, on the grounds that the three sites differ in buffering (2 MiB `BufReader` vs bare `File`), log level, and whether stdin honors `--async-reader`. That was wrong — those are parameters, not divergent behavior. The three near-identical copies are now one `open_normalized_with_opts`, the single place a normalized stream is minted: the async-stdin handling and the log label are arguments to it, and the buffering stays with the caller that wants it. Sort's runtime behavior is unchanged.

**Defects found while verifying**

- The PR's headline claim was false. `sort --verify` and `compare` read the header through a format-aware reader and then re-opened the path with a bare `File::open`, so both rejected the SAM files `sort` itself accepted. `zipper` dispatched on file *extension*, contradicting the "detection is by content" contract. All now go through `open_normalized_input`.
- `zipper` sniff-then-reopen broke FIFO input. Re-opening only rewinds for a regular file, so `-i <(samtools view -h in.bam)` lost the first four bytes and failed with `invalid flags`. Non-regular inputs are now opened once and the peeked bytes replayed. BAM over a FIFO now works too, which it did not before. _(updated)_ `zipper` no longer sniffs at all: its mapped input goes through the shared normalizing reader, so there is no format branch left to get wrong.
- `fgumi sort -i -` silently lost its 2 MiB input buffer in this branch; restored.
- Record-level SAM parse errors were reported as header failures, because the lazy transcode pulled records while the consumer was still reading the header. The header is now flushed into its own BGZF block.
- Non-SAM text was accepted as headerless SAM and transcoded into a reference-less BAM. Now rejected.
- `ChainedReader` dropped its replay buffer on a zero-length read, and held the buffer for the stream's lifetime after exhaustion. Guarded and freed.
- `fastq`, `duplex-metrics` and `simplex-metrics` rejected `-i -` with `File does not exist`. _(updated)_ All three now stream. The original fix only made `fastq` stream and had the two metrics commands declare their extra input scan explicitly; that scan is gone — the consensus-BAM guard now runs against the first qualifying record of the pass they already make, so they read the input exactly once. They still reject consensus BAMs.

**Coverage**

Added `tests/integration/test_input_source_matrix.rs`: every command the binary advertises must declare whether it accepts SAM and stdin, checked against `fgumi --help`, so a new command fails the suite until it states its stance. _(updated)_ the count quoted here originally is superseded: `cargo ci-test` is now 5990 passed / 27 skipped. `--workspace --no-default-features --all-targets` and `--all-features` both check clean.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed the three remaining inline threads plus the outside-diff finding, which has no thread ID and so cannot be resolved through the UI.

**Outside diff range — `src/lib/commands/fastq.rs`, output-clobber guard vs stdin.** Real, and the diagnosis was exactly right. `validate_input_exists` now lets `--input` be `-`, but the same-path check then calls `std::fs::canonicalize("-")`, which fails with `NotFound`, so `fgumi fastq -i - -o <existing-file>` died with an unexplained IO error rather than running — the common case of overwriting a prior run's output. Rather than inlining the guard clause I extracted the check into `reject_output_clobbering_input`, which exempts stdin (it names no filesystem entity, so it can clobber nothing) and is now unit-tested over the stdin, distinct-path, and same-path cases; I confirmed the stdin case fails before the fix and passes after. I also grepped for siblings: `fastq.rs` is the only command that canonicalises an input path unconditionally (`merge.rs` uses `if let Ok(..)` and does not accept stdin anyway), so there is no second instance of this.

**`test_input_source_matrix.rs` — exit status weaker than the declared contract.** Fixed, and the clobber you spotted was the more serious half: both runs wrote to one `{command}.out`, so the SAM run overwrote the BAM output it was supposed to be checked against. Each run now writes into its own directory and the BAM run is the oracle; the comparison is over every file the run produced, keyed by name so a missing file from a multi-file output (the metrics commands write several from one prefix) reads as a difference.

This immediately caught a real divergence in `filter`: `cD` came back as `Int8(10)` from BAM input and `Int32(10)` from SAM. Same value, different aux type byte — SAM's `i` carries no width, and fgumi narrows to the smallest signed type when it writes consensus tags while noodles' SAM parser does not narrow at all (htslib picks a third answer, preferring unsigned). Byte equality across a SAM round-trip is therefore not a property any BAM writer guarantees, so the matrix renders both outputs to SAM text and compares values; the reasoning is written into the test rather than left implicit. Whether the SAM path should narrow to match is now #655.

**`test_sam_input.rs` — `sort --verify` parity beyond a record count.** Fixed as suggested: a `violations` parser alongside `records_checked`, both asserted equal across formats, plus a new case whose input is deliberately out of queryname order so the violation count is non-zero and both runs must fail identically. The shared `@PG`-normalizing BAM reader moved into `tests/integration/helpers/` so the matrix and the SAM-input suites use one implementation.

**`simplex_metrics.rs` — missing duplex case.** Added, plus the `cD`+`aD`+`bD` combination, since `is_consensus` reaches the duplex branch with or without the depth tag.

`cargo ci-test` 6000 passed / 27 skipped; `ci-fmt`, `ci-lint`, `ci-doc` with `-D warnings`, and `cargo check --workspace` under both `--no-default-features` and `--all-features` are clean.

nh13:
@coderabbitai review

nh13:
Addressed the remaining review feedback.

**Inline thread — `src/lib/commands/zipper.rs`, stale `peek_bgzf` intra-doc link.** Fixed; the symbol is now `Zipper::stream_mapped_reader`. Verified it resolves by building docs with `--document-private-items -D rustdoc::broken_intra_doc_links` (the seven remaining unresolved private links in that build are pre-existing and outside this diff — `clip.rs`, `compare/bams.rs`, `review.rs:725`, `simulate/common.rs`, `unified_pipeline/fastq.rs` — and the `docs` job does not document private items, so none of them gate CI).

**Duplicate — `src/lib/commands/extract.rs`, stdin BGZF decompresses single-threaded.** Real, and the diagnosis was right: `open_input_readers` hands the *same* reader back for the main pass, so the hardcoded `1` meant `--threads` never reached the BGZF decoder on the stdin path while file inputs got the full count. Now opened with `self.threading.num_threads().max(1)`, with a comment recording that this is the main pass's thread count rather than a sampling one. `sample_detection_quals` still passes `1` deliberately and is left alone — those are short-lived, throwaway readers opened per file input purely for a bounded head-scan, and the main pass re-opens with `decomp_threads`.

**Outside diff — `src/lib/commands/duplex_metrics.rs`, `test_reject_consensus_bam` only covered `cD`.** Correct, and worth having: `is_consensus` is `is_simplex_consensus(cD) || is_duplex_consensus(aD+bD)`, so the `aD`+`bD` case exercised a genuinely distinct branch with no end-to-end coverage on the command a user is most likely to point at a duplex consensus BAM. Converted to an `#[rstest]` case table per the repo's testing conventions: simplex (`cD`), duplex (`aD`+`bD`), duplex-with-depth (`cD`+`aD`+`bD`), and a plain grouped read that must *not* be rejected as consensus.

**Duplicate — `tests/integration/test_input_source_matrix.rs`, exit-status-only comparison.** Already addressed in `74809368`, which the review predates (it ran against `34e04f33`): `run_command` takes a `run_tag` and writes into a per-run subdirectory, and `declared_sam_support_holds` treats the BAM run as the oracle and compares via `digest_run`/`describe_difference`, rendering BAM outputs to SAM text so record identity and order are compared rather than just counts.

---

**Unrelated bug found while verifying:** the full suite turned up a genuine deadlock in the BGZF FASTQ pipeline — `BlockMerge`'s early-return completion check omitted the `q2_block_parsed.is_empty()` term its sibling check has, so it could declare the pipeline done with parsed blocks stranded in the queue. It is pre-existing on `main` and `unified_pipeline/fastq.rs` is not touched by this PR, so it is fixed separately in #656 rather than folded in here.

nh13:
@coderabbitai review

nh13:
Addressed all six inline threads plus the duplicate finding, which has no thread ID and so cannot be resolved through the UI.

**Duplicate — `crates/fgumi-sort/src/external.rs`, `--async-reader` ignored for stdin.** Real, and the earlier round's fix genuinely did not reach this site: `open_normalized_with_opts` in `fgumi-bam-io` honors `async_reader` on both its stdin and file branches, but sort's *pool-integrated* reader is a separate opener, and its stdin branch handed back a plain `BufReader` whatever the flag said. `fgumi sort --async-reader -i -` therefore ran without the prefetch thread while the same command over a file got one. Now spawns `PrefetchReader::new(io::stdin())`, with the buffered reader kept for the non-async case. Tested by asserting on the log rather than on records — both readers deliver the same bytes, so output equality cannot tell whether the flag was honored — plus output equality, because honoring it must not change the result.

**`crates/fgumi-bam-io/src/format.rs`, scan FEXTRA for `BC`.** The web query is right that RFC 1952 permits subfields in any order, but I took the documented-assumption option rather than the scan, because scanning would make the classifier *disagree with the decoder it feeds*. Every decoder downstream of the classifier is keyed to the same fixed offsets. `fgumi_bgzf`'s `read_raw_block` — the sort worker pool, the BAM writers, `fastq` — checks `BC` at bytes 12-13 and takes `BSIZE` from bytes 16-17 to *frame* the block, so the layout decides where one block ends and the next begins, not just whether the header is valid. `noodles_bgzf`'s `is_valid_header` additionally requires `FLG` to be `FEXTRA` alone and `XLEN` to be exactly 6. htslib's `check_header` (`bgzf.c`) requires the same `XLEN == 6`, `BC` at 12-13 and `SLEN == 2`, differing from noodles only in tolerating other `FLG` bits alongside `FEXTRA`. Answering `Bgzf` for a member that misses this layout would hand it to a decoder that rejects it — `invalid BGZF header` from noodles, `Invalid BGZF subfield ID` from `fgumi_bgzf` — which is precisely the misleading diagnosis this classifier exists to prevent.

_Edited to correct one clause: I originally wrote that htslib's `check_header` requires the same as noodles. It matches on `XLEN`/`BC`/`SLEN` but tests `header[3] & 4`, so it tolerates `FLG` bits beyond `FEXTRA`. Because the extra field precedes `FNAME` in a gzip member, a `FLG=0x0c` file keeps `XLEN`/`BC` where they belong and htslib would read it while noodles will not — the one sub-class where fgumi is stricter than htslib rather than equally strict. The verdict is still right for fgumi (noodles rejects it, so calling it `Bgzf` would produce exactly the misleading error above), and it is the cheapest class to support if a real file ever appears, since it needs no framing change._

Going the other way was also a real (if narrow) fix: the check was *looser* than the decoder at `XLEN >= 6`, so a member with `BC` first and a second subfield trailing it classified as BGZF and then failed in the decoder. It now matches the decoder exactly, and the rationale is recorded on `classify_input` with the boundaries pinned as cases — including a test that a block written by the BGZF writer both classifies as `Bgzf` and decodes, so the two cannot drift apart silently. For FASTQ, where `Gzip` is a supported input rather than an error, the stricter verdict costs only block-parallel decode: `extract` reads it with `MultiGzDecoder`, and BGZF is valid gzip.

**`compare` engines, two opens per input.** Fixed in both, and `open_raw_bam_record_reader_with_header` was not unused — `sort --verify` already drives it (`sort.rs`); the compare engines were simply not converted. `sort_verify_compare` and `verify_records_in_order` each took a header open plus a per-arm record open, and `molecule_join_compare_capped` a header open plus `open_raw` per input; all now take one. `VerifyContext` carries the readers instead of the paths, which is what let the four per-arm opens go away.

One scope note: this makes the *engines* single-open, which is what their own entry points promise, and both now read a FIFO — pinned by new tests that hand `sort_verify_compare` and `molecule_join_compare` an input served over a FIFO. The `compare bams` CLI still opens each input once more up front for its header-compatibility precondition, deliberately, so that an incompatible pair hard-exits before any engine runs; I left that alone rather than trade a documented error-precedence guarantee for FIFO support the command does not claim (`compare` declares `stdin: NotApplicable`, since it compares two named inputs). Both helpers' docs say so, so the next reader does not have to re-derive it.

**`shared_metrics.rs`, consensus guard skipped when no paired R1 exists.** Correct, and the consequence was as described: a fragment or single-end consensus BAM has no paired R1 anywhere, so the guard never ran and the input fell through the `len() < 2` skip to an empty metrics file instead of the explicit error. The selector is now `consensus_guard_record`, which prefers a paired primary R1 and otherwise takes the first primary record — consensus tags sit on every consensus record, so any primary record answers the question. It returns `None` only for a template of nothing but secondary/supplementary records, which defers the check to the next template rather than skipping it. Covered end-to-end on `duplex-metrics` with fragment simplex and fragment duplex cases (both fail before the change), a fragment grouped read that must still not be rejected, and a record-level `#[rstest]` table over the selection itself.

**`zipper.rs`, header read loses input attribution.** Fixed as suggested. Tested by piping a BGZF stream whose contents are not BAM: it gets past format detection, fails at exactly that header read, and the error now names the mapped stream.

**`test_input_source_matrix.rs`, `compare` makes the parity assertion vacuous.** Correct, and worth catching — `compare bams` has no `{output}`, so both digests were empty and the case degenerated to exit-status-only. Both fixes are in: each run's stdout is folded into the digest under a `<stdout>` key (scrubbed for run identity like every other text output), so `compare`'s report is now compared as content; and the oracle must be non-empty, so a future command whose artifacts the digest cannot see fails loudly instead of passing silently. I verified the case really was vacuous by suppressing the stdout entry and watching `compare` trip the new guard.

**`test_sam_input.rs`, a zipper regression turns the FIFO test into a hang.** Real, and the diagnosis was right. I did not take the `O_RDWR` route, though: opening the write end read-write returns immediately instead of waiting for a reader, so the feeder can copy this small file into the pipe buffer and close both ends before zipper ever opens it — the bytes are then discarded and zipper blocks on a writer that no longer exists, trading a deterministic hang for a racy one. The blocking `O_WRONLY` open is the hand-off and is kept; the fix is to assert on the zipper result *before* joining, so a zipper that dies before opening the FIFO fails the assertion and leaves the blocked feeder to be reaped at process exit. On the success path zipper has drained the FIFO, so the join returns at once. Confirmed by pointing the run at a nonexistent reference: it now fails in under a second with the real error instead of hanging.

The same shape appears in the new compare tests, where the consumer is in-process and so would hang the test thread itself rather than a subprocess; those bound the engine call on its own thread and only join the feeder once it has returned.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Both findings from the latest review are fixed.

**`shared_metrics.rs` — "First R1 record" after the guard learned to fall back.** Correct, and it was my own regression from the previous round: `consensus_guard_record` now prefers a paired primary R1 but falls back to the first primary record, so on a fragment or single-end consensus BAM the message named a record that is not an R1. Reworded to "Record", which is accurate for every record the guard can now pick. No test pinned the old wording.

**`reader.rs` — buffer under the BGZF frame reader.** Real, and the codebase already says so twice: `zipper`'s stream path notes that "the BGZF frame reader issues small `read_exact` calls, which are expensive straight off a pipe", and the sort engine's pool reader gives stdin a 2 MiB buffer because "`io::Stdin` re-acquires a mutex and reads through an 8 KiB buffer on every call". The shared opener had neither.

I put the buffer in a slightly different place than suggested, for a reason worth stating. Buffering the stdin/file branches inside `open_normalized_with_opts` would also buffer `open_normalized_input`'s consumers — and `fgumi_sort::RawBamRecordReader` already wraps that stream in its own 256 KiB `BufReader`, so every byte on the sort path would be copied through two buffers. The buffer therefore goes in `open_bgzf_reader`, immediately under the frame reader that issues the small reads, which covers `create_bam_reader`/`create_raw_bam_reader`/the pipeline reader and leaves the already-buffered path alone. `--async-reader` is left unwrapped too: `PrefetchReader` already hands out large chunks, so it would pay the extra copy for nothing.

Measured rather than assumed, since this is a performance claim. On a 247 MB simulated BAM through `compare bams --mode content` (which reads both inputs, so ~494 MB), three runs each, release build, warm cache:

| | run 1 | run 2 | run 3 |
|---|---|---|---|
| unbuffered | 7.29s | 6.97s | 7.55s |
| buffered | 6.75s | 6.42s | 6.75s |

~7%, with non-overlapping ranges. Modest on a warm local SSD; the stdin case (mutex plus an 8 KiB buffer per call) is where it should matter more.

Buffer size is 256 KiB — several maximum-size BGZF blocks, and deliberately not the sort engine's 2 MiB, because this one is per open reader and `merge` holds one per input file.

`cargo ci-test` 6036 / `ci-fmt` / `ci-lint` clean. Note the stacked #659 has been rebased onto this branch's new head.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Both findings from the latest review are fixed.

**`test_input_source_matrix.rs` — both stdin matrices asserted exit status only.** Correct, and the sibling was worth catching: `declared_stdin_support_accepts_piped_sam` had the identical weakness and is the compound case, so both now run the named-path fixture as an oracle and compare through `digest_run`/`describe_difference`. The three matrices had drifted into three copies of the same control-plus-compare shape, so I extracted `run_named_path_oracle` and `compare_to_oracle` and pointed all three at them; `STDIN_RUN`/`STDIN_SAM_RUN` are constants now and are scrubbed longest-first, since `stdin-sam-run` contains `sam-run` and the short scrub would otherwise leave `stdin-<RUN>` behind and fail every stdin-SAM comparison on a difference that is not one.

I verified the new assertion has teeth rather than trusting that it passes: piping a header-only BAM — your exact "consumed nothing, still exits 0" scenario — turns the test red for `correct`, `fastq`, `sort`, `group`, `dedup`, `filter`, `clip` and `downsample`, with `extract` caught a step earlier by the exit status.

That same fault injection turned up something worth recording, so `compare_to_oracle` now documents it: the `oracle.is_empty()` guard proves the harness can *see* a run's artifacts, not that they are sensitive to the records that went in. Four commands are therefore still insensitive, for two reasons neither the guard nor this comparison can fix — `simplex`, `duplex` and `codec` write a header-only BAM from any input because the fixture gives them ungrouped reads (grouping it gets `simplex` to one consensus record, but `duplex` needs strand-suffixed `/A`,`/B` `MI` families and `codec` needs CODEC read structure), and `simplex-metrics`/`duplex-metrics` emit the same all-zero tables either way. Closing that needs new fixture generators, which is separate work from this PR. `zipper` is insensitive for a third reason that is *not* a gap: its output is driven by the uBAM passed as `-u`, so an empty `-i` legitimately still yields every unmapped template.

One defect of my own, found in self-review before pushing: `extract` is `stdin: Required` and its fixture is FASTQ, so calling the oracle "BAM" would have printed "failed on a named BAM path" for a `.fq` input. The helper and its messages say "a named path" instead.

**Duplicate — `sam_input.rs`, the missing `buf.is_empty()` guard.** Real, and it genuinely was absent from the final code rather than merely reported fixed: a zero-length read left the staging buffer drained, the refill loop read that as "transcode more", and `SamToBamStream::read` encoded a whole 98-byte header block to satisfy a request for nothing before returning `Ok(0)`. Nothing was lost — `offset` never advanced — so this was wasted work, matching your Minor rating. Guarded now, with a test that fails before the fix on the staged-bytes assertion and also pins that returning early costs the stream nothing, since the same reader must still deliver the whole BAM afterwards.

I swept the rest of the chain for the same gap: `ChainedReader::read` and `PrefetchReader::read` already guard (it matters more in the latter, where an unguarded empty read would block on `rx.recv()`), and `TeeReader::read` is a pure pass-through with no work to waste. With `SamToBamStream` fixed the set is uniform.

`cargo ci-test` 6022 passed / 0 failed, `ci-fmt`, `ci-lint` and `ci-doc` (with `RUSTDOCFLAGS=-D warnings`) all clean.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed the outside-diff finding from the latest review. It has no thread ID, so it cannot be resolved through the UI.

**Outside diff range — `crates/fgumi-bam-io/src/reader.rs`, `read_header_and_replay` parses the BGZF header unbuffered.** Correct, and confirmed against the current code. Two of the three callers hand it the raw `open_normalized_*` output: `create_bam_reader_for_pipeline_with_opts` here, and `fgumi_sort::open_raw_bam_record_reader_with_header` in `crates/fgumi-sort/src/reader.rs`. Both are a plain `File` or `io::stdin()`, so the frame reader's 18-byte block-header read and the block-body read each hit the source directly. The third caller, `external.rs`, already wraps its input in a 2 MiB `BufReader` and was never affected.

I did not apply the suggested fix as written. Leaving the `BufReader` resident under the returned stream fixes the header parse but puts a second buffer under the consumers that already have one — `RawBamRecordReader` wraps its input in a 256 KiB `BufReader`, and the sort worker pool reads via `read_raw_blocks`, so both pull BGZF block by block rather than in buffer-sized chunks. A resident layer would therefore copy every byte of the input a second time on the sort hot path, which is the cost `open_bgzf_reader`'s doc comment explicitly avoids paying.

Instead the buffer is scoped to the parse: it sits under the `TeeReader` while the header is read, then its readahead is drained into the replay buffer and the original stream is handed back unwrapped. Every caller gets the syscall fix, and no extra layer survives into the record path. The drain costs one memcpy of at most 256 KiB, which `ChainedReader` frees as soon as it is delivered.

Two regression tests, both verified to fail without the change:

- `test_read_header_and_replay_buffers_the_source` — a spy reader records the largest request reaching the source. Before the fix it was 80 bytes; it now sees a full buffer-sized fill.
- `test_read_header_and_replay_replays_every_byte` — asserts the replayed stream is byte-identical to the input. Without the readahead drain the entire record block and the EOF block are silently dropped, which is the failure mode that would otherwise reach a pipe consumer with no error.

Also updated the doc on `BGZF_INPUT_BUFFER_SIZE`, which named only `open_bgzf_reader` as its user and now has a second one.

PR #645, opened 2026-07-22
perf(umi): index the edit assigner's neighbour search

nh13:
Stacked on #630 — it gives the parallel assigners a persistent molecule-id counter, and this branch is cut from it. Please read #630 first.

## The gap

`Strategy::new_assigner_full` hands `index_threshold` to every strategy except one:

```rust
Strategy::Edit      => SimpleErrorUmiAssigner::new(edits),                        // no index
Strategy::Adjacency => AdjacencyUmiAssigner::new(edits, threads, index_threshold),
Strategy::Paired    => PairedUmiAssigner::new_with_threads(edits, threads, index_threshold),
```

So `edit` compared every distinct UMI at a position against every other one — quadratic in distinct UMIs — while the other two narrowed candidates with an N-gram index. That is exactly why the profiling sweep found adjacency flat across UMI diversity and edit blowing up.

This gives edit the same index.

## Output is unchanged, and provably so

Not just "tests pass". `assign` sorts components by their lexicographically smallest UMI **before** assigning molecule ids, so ids don't depend on the order neighbours were discovered in. Any method that finds the same connected components produces byte-identical output.

That the existing incremental set-merge really does compute connected components (and is order-independent) was checked two ways against a brute-force union-find oracle — 9,000 and 20,000 randomised trials, zero mismatches — and end-to-end: record bodies are byte-identical on a 7,868,052-record twist-umi run (same md5 from both arms, run from an identical path so `@PG CL:` matches).

## Two thresholds, and why they differ

**The index is chosen at 200 distinct UMIs, not the shared default of 100.** Measured index vs scan:

| distinct | indexed | scan | |
|---|---|---|---|
| 118 | 104.9 µs | 98.5 µs | index **loses** 6.5% |
| 213 | 203.7 µs | 275.8 µs | index wins 26% |

The crossover sits between them, so 100 — inherited from adjacency without being re-measured for edit — lands inside the losing band. 200 clears it and gives up none of the wins. Adjacency and paired keep 100; their crossovers weren't re-measured here.

The bench constructs `SimpleErrorUmiAssigner::new_with_index_threshold` directly rather than going through `Strategy::new_assigner_full`, which floors at 200 — routing through it would make every sub-200 arm identical to the default and leave the sweep unable to locate the crossover it exists to find.

**One shape defeats the index**, and it isn't covered by the crossover: if a UMI half has much lower entropy than the whole (a constant anchor or linker), every UMI lands in one partition bucket and each lookup degenerates to a full scan plus per-query bookkeeping — ~1.7× slower than scanning at 8 distinct first-halves, ~6× with a constant one. Raising `--index-threshold` above the largest position restores the scan for such designs.

**Merging pauses at 64 distinct, far below that.** The distinct count isn't known until the position has been read, so the choice can't be made up front — you either pre-count in a separate pass (which costs every position, including the many that never index) or start working and switch. Pausing early makes the speculative work small enough not to matter either way:

- **below** the crossover, the merge resumes from where it paused — the set-merge is incremental and order-independent, so **nothing is repeated**;
- **above** it, the only cost is having merged 64 UMIs the slow way, a couple of thousand comparisons against a job orders of magnitude larger.

Pausing *at* the crossover instead is much worse, and I measured it: **32% slower at 213 distinct**, because merging 200 UMIs is most of the work for a position holding 213, so it gets paid for twice.

## Measured

Quiet `c8g.4xlarge`, sequential edit at one mismatch, via the existing `umi_assigner_threshold` bench (extended here to reach the diversity where the index engages, and to report the actual distinct count per size — it is *not* `n/8`, because the generator's injected substitutions each mint a new sequence):

| distinct | before | after | |
|---|---|---|---|
| 213 | 274 µs | 223 µs | 1.2× |
| 446 | 1.00 ms | 480 µs | 2.1× |
| 1,778 | 13.3 ms | 2.70 ms | 4.9× |
| 6,759 | 411 ms | 20.6 ms | **20×** |
| 12,967 | 961 ms | 62.4 ms | **15.4×** |

**The cost, stated plainly: ~0.6% on low-diversity data.** `fgumi group -s edit` over twist-umi goes from 20.40 s to 20.53 s mean user CPU across 5 interleaved reps (every run validity-gated at 7,868,052 accepted records). Positions that never reach 64 distinct UMIs still record them as they go, and most real positions are singleton-dominated. Removing that means deferring the bookkeeping as well; it's a contained follow-up, not done here.

## Tests

Coverage is built around the fact that this code now has three internal regimes — merge only, merge-then-resume, and merge-then-index — and the paths must agree with each other *and* with an independent reference:

- **`test_edit_matches_brute_force_across_regimes`** — an implementation-independent brute-force union-find oracle, at 10 / 63 / 64 / 65 / 120 / 199 / 200 / 201 / 600 distinct UMIs × edits 1–2. The counts straddle **both** internal boundaries so the pause boundary, the decision boundary, and the resume path between them are each pinned. Without an outside oracle the two internal paths could agree on the same wrong answer.
- **`test_edit_index_matches_scan`** — index forced on (`index_threshold = 0`) vs off (`usize::MAX`) over pure-ACGT, all-singleton, one-component, duplicate-heavy and `N`-containing inputs, at edits 0–3. Every case is padded past `MERGE_DEFER_POINT`, because the index is unreachable below it whatever the threshold; an `assert!` on the distinct count keeps that from silently regressing. Verified by putting `panic!` at the top of `components_via_index` — all cases fail, i.e. they genuinely reach it.
- **`test_edit_dashed_umis_fall_back_within_indexed_path`** — dashed dual UMIs sized past the defer point, so `NgramIndex` declining them (its `from_bytes` rejects `-` while the caller's `from_umi_str` admits it) exercises the fallback *inside* the indexed path rather than bypassing it.
- **`test_edit_index_matches_scan_above_threshold`** — 1,000 UMIs each with a 1-mismatch neighbour, so the relation has real edges rather than being all singletons, with a sanity assert that merging actually happened.
- **`test_edit_ids_are_dense_and_ordered_by_group_minimum`** — pins the ordering property the parity argument rests on.
- **`test_ngram_index_declines_oversized_partitions`** — `NgramIndex` packs a partition into a `u32`, so at `max_mismatches == 0` a UMI longer than 16 bases would silently truncate the key. The constructor now declines instead.

Seven mutations were applied; five fail a test:

| mutation | |
|---|---|
| pause off-by-one (`>=` for `>`) | killed |
| resume slice starts too late | killed |
| never resume (drop deferred UMIs) | killed |
| index/UMI position mapping off by one | killed |
| index built over a subset, queried with the whole set | killed |
| resume slice starts one early | survives — re-merging an already-merged UMI is a no-op |
| index chosen at 64 instead of 200 | survives — a *performance* choice; both paths provably yield the same components |

The pause off-by-one was a real bug this suite caught during development: the UMI that triggered the pause was never merged, because the flag was set and the loop skipped it while the resume slice started one past it. Silent mis-grouping, not a crash.

## Also fixed

`--index-threshold`'s help said *"Set to 0 to always use linear scan"*. The gate is `distinct >= threshold`, so `0` always **indexes** — the opposite of the documentation. It also claimed the flag affects only adjacency and paired. Both corrected in `group` and `dedup`, along with the `info!` lines that logged the threshold for those two strategies only.

## Summary by CodeRabbit

* **Performance Improvements**
  * Enhanced the **Edit** strategy with a distinct-UMI–aware indexed component discovery path, plus improved switching between indexed and scan-based assignment (including scan/index fallback when indexing is not applicable).
* **Bug Fixes**
  * Tightened index construction to avoid unsafe partition sizing (including edge-case validation).
* **CLI Updates**
  * Clarified `--index-threshold` as a **minimum distinct UMIs per position** gate and updated reported “Index threshold” behavior to match **Edit**-specific flooring and when indexing is enabled.
* **Testing**
  * Expanded coverage for index partition limits, scan/index equivalence, and correctness across assignment regimes.

### DIFF
--- a/benches/umi_assigner_threshold.rs
+++ b/benches/umi_assigner_threshold.rs
@@ -24,7 +24,9 @@ use criterion::{BenchmarkId, Criterion, Throughput, criterion_group, criterion_m
 use rand::rngs::StdRng;
 use rand::{RngExt, SeedableRng};
 
-use fgumi_lib::umi::assigner::{Strategy, Umi, UmiAssigner};
+use fgumi_lib::umi::assigner::{
+    EDIT_INDEX_THRESHOLD, SimpleErrorUmiAssigner, Strategy, Umi, UmiAssigner,
+};
 use fgumi_lib::umi::parallel_assigner::{
     ParallelAdjacencyAssigner, ParallelEditAssigner, ParallelIdentityAssigner,
     ParallelPairedAssigner,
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #647, opened 2026-07-23
perf(sort): pool-integrate and optimize the coordinate --write-index merge

tfenne:
## Summary

Fixes and then optimizes the coordinate `--write-index` merge. Stacked on #643 (base `tf_expose_sort_temp_limit`), so this PR shows only the write-index changes.

Three commits:

1. **Pool-integrate the indexed merge.** `RawExternalSorter::merge_chunks_with_index` took a different, far slower path than the plain coordinate merge: it forced `reader_concurrency = 1`, `pool_decompress = false`, and a standalone `create_indexing_bam_writer`, which serialized input decompression to a single reader and bypassed the shared worker pool entirely — so the merge thread ran hot while the pool sat idle. It now mirrors `merge_chunks_generic`: `PoolDisk` sources decompressed by the worker pool, output through `PooledBamWriter::new_indexing`, with BAI virtual-offset tracking hooked into the pooled output.

2. **Borrow merge records in place (zero-copy read).** The k-way merge copied every record twice — once from the decompressed block into a per-source scratch buffer, then again into the writer's staging buffer. The first copy is now elided: when a record is contiguous in its current decompressed block (the ~100% common case, since the spill writer pre-flushes records to block boundaries) it is borrowed in place; only a record that straddles a block boundary falls back to reassembling into scratch. This is shared by the plain and indexed merges, so both get faster.

3. **Buffer the `.bai` write.** `write_bai_index` handed the output file straight to `bai::io::Writer`, which emits the index as many tiny fields — one `write()` syscall per field. It now serializes into memory and writes the sidecar in a single call.

## BAI correctness

`BaiBuilder` is extracted from `IndexingBamWriter` so the vendored-writer path and the pooled path share one virtual-offset implementation. Records are cached in write order (which is non-decreasing block order) and resolved strictly from the front of a `VecDeque`, and the `block_number -> compressed offset` map is pruned behind the front record so it stays bounded to the in-flight block window rather than growing one entry per block for the whole output.

## Validation

- Full gate green: `cargo ci-fmt`, `cargo ci-lint` (clippy pedantic), `cargo ci-test` (5746 tests).
- Output is **byte-identical** to the plain pooled writer (unit test), and the generated `.bai` **matches samtools exactly** — mapped-only region counts and per-reference `idxstats`, on a forced spill + consolidation + multithreaded input (`#[ignore]`, requires samtools).
- A new non-samtools test drives an oversized (multi-block) record through a spilled `--write-index` merge and loads the produced `.bai` back, covering `BaiBuilder`'s cross-block offset resolution.
- Real WGS (NA12813, 1.287B reads, coordinate `--write-index`, `-@ 8`): after this PR the indexed merge is ~234s and the full sort ~599s; the zero-copy read and buffered `.bai` account for ~272s → ~234s merge and ~659s → ~599s total on top of the initial pooling fix. Real-file mapped-only region counts and per-reference mapped counts match samtools exactly.

## Known limitation (pre-existing, not introduced here)

`extract_alignment_context` returns `None` for a placed-but-unmapped read — one with a valid `RNAME`/`POS` but the `FUNMAP` flag set — so such reads are not position-binned and region queries miss them. This behavior is carried verbatim from the original `IndexingBamWriter`; it is why the samtools cross-checks compare mapped-only (`-F 4`). Whether to match htslib's position-binning of these reads is worth deciding separately.

## Summary by CodeRabbit

* **New Features**
  * Added support for generating BAI indexes while sorting BAM files (including incremental index building during output).
* **Bug Fixes**
  * Improved BAI virtual-offset resolution for boundary and edge cases.
  * More reliable handling of oversized records that span compressed blocks.
  * Reduced unnecessary data copying during pooled merges for better correctness and efficiency.
* **Tests**
  * Expanded indexed-write coverage, including oversized-record spill/boundary handling.
  * Added samtools-compatibility verification for `.bai` output (guarded/disabled by default).

### DIFF
--- a/crates/fgumi-bam-io/src/lib.rs
+++ b/crates/fgumi-bam-io/src/lib.rs
@@ -33,7 +33,7 @@ pub use reader::{
 };
 pub use reorder::{DrainReady, ReorderBuffer};
 pub use writer::{
-    BamWriter, BgzfWriterEnum, IndexingBamWriter, RawBamWriter, bai_sidecar_path,
+    BaiBuilder, BamWriter, BgzfWriterEnum, IndexingBamWriter, RawBamWriter, bai_sidecar_path,
     create_bam_writer, create_indexing_bam_writer, create_optional_bam_writer,
     create_raw_bam_writer, write_bai_index, write_bai_sidecar, write_bam_header,
 };
--- a/crates/fgumi-bam-io/src/writer.rs
+++ b/crates/fgumi-bam-io/src/writer.rs
@@ -16,8 +16,9 @@ use noodles_bgzf::io::Writer as BgzfWriter;
 use noodles_csi::binning_index::Indexer;
 use noodles_csi::binning_index::index::reference_sequence::bin::Chunk;
 use noodles_csi::binning_index::index::reference_sequence::index::LinearIndex;
-use std::collections::HashMap;
+use std::collections::{HashMap, VecDeque};
 use std::fs::File;
+use std::hash::{BuildHasher, BuildHasherDefault, Hasher};
 use std::io::{self, Write};
 use std::num::NonZero;
 use std::path::{Path, PathBuf};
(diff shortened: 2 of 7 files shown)


tfenne:
@coderabbitai review

nh13:
Thank-you for putting this together!!!

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Github is having issues with pull requests, so I tried to close and reopen, now it's stuck as closed.

PR #648, opened 2026-07-23
docs(dedup,group): document that --no-umi still splits by strand of origin

nh13:
## Summary

`--no-umi` is documented as "Skip UMI-based grouping; **group by position only**". It does not group by position only. The identity assigner inherits the `split_templates_by_pair_orientation() -> true` trait default (`crates/fgumi-umi/src/assigner.rs`), so `assign_umi_groups` additionally partitions each position group by which mate defines the forward end. An F1R2 and an F2R1 template at the same unclipped 5' coordinates stay two molecules, and each keeps its own representative.

This is the dominant reason `fgumi dedup --no-umi` and Picard `MarkDuplicates` disagree on the duplicate flag, and the help text actively points users the wrong way for exactly the MarkDuplicates-swap case where `--no-umi` is the natural thing to reach for.

Documentation only — no behavior change. Whether the split *should* be optional is a separate design question and is deliberately not addressed here; the help text now states plainly that no such option exists.

## Evidence

`twist-umi` benchmark sample (SRR34589119), 15,826,646 primary reads, Picard 3.4.0, same input templates through both tools. Reconstructing duplicate families two ways and checking whether each tool marks exactly (family − 1) per family:

| Key | Families | Expected N−1 | fgumi marked | Picard marked |
|---|---|---|---|---|
| chr7 · coordinate only | 177,121 | 159,666 | off by −7,307 | **MATCH** |
| chr7 · coordinate + read-1 strand | 184,428 | 152,359 | **MATCH** | off by +7,307 |
| chr1 · coordinate only | 285,369 | 208,491 | off by −10,873 | **MATCH** |
| chr1 · coordinate + read-1 strand | 296,242 | 197,618 | **MATCH** | off by +10,873 |

Both tools build the identical set of coordinate families; fgumi sub-splits them by strand and marks N−1 of each. Both matches are exact, in both directions, on both chromosomes.

The per-template flag disagreement decomposes to the read against that split — `picard-dup-only − fgumi-dup-only` equals the strand-split count exactly (chr7 `7,398 − 91 = 7,307`; chr1 `11,041 − 168 = 10,873`). The small bidirectional residual is representative tie-breaking, not a count difference.

## What changed

- **`dedup` long help**: new `# Grouping key (strand of origin)` section — the key, why it exists (`fgbio GroupReadsByUmi` parity, duplex strand separation), the contrast with orientation-agnostic Picard, `marked duplicates = templates − molecule groups`, and that there is no orientation-agnostic option.
[...]

PR #490, opened 2026-07-08
fix(dedup): score duplicates by Picard SUM_OF_BASE_QUALITIES (DEDUP-02)

nh13:
## Summary

`fgumi dedup` selects a duplicate family's representative with `score_template`, intended to match Picard/HTSJDK's `SUM_OF_BASE_QUALITIES` `DuplicateScoringStrategy`. The implementation instead summed `min(q, 15)` over **every** base — reading Picard's `>= 15` *threshold* as a per-base *cap* (a misread of PR #27's stated Picard intent). That ranks families differently: sub-threshold bases still inflate the score, and every high-quality base is clamped to 15.

This restores the true Picard rule (htsjdk `DuplicateScoringStrategy.getSumOfBaseQualities`): **sum base qualities `>= 15` at full value; exclude qualities below 15.** No mapq term is added — fgumi `dedup` is a Picard `MarkDuplicates` drop-in, whereas fgbio's `GroupReadsByUmi` additionally adds `mapq` (intentionally omitted; **DEDUP-01** wontfix per the tracker Decisions). **DEDUP-03** (keep+mark secondary/supplementary) also remains wontfix.

## What changed

- `score_template`: `Σ min(q,15)` → `Σ { q : q >= 15 }` (full value), via a named `PICARD_MIN_BASE_QUALITY = 15` threshold constant citing htsjdk.
- **Golden test** at the q=14/15 boundary (`[14,15,16,40]` → `71`), citing htsjdk `getSumOfBaseQualities`.
- Updated the scoring and tie-break unit tests that pinned the `min(q,15)` cap. The tie-break tests assert *ordering* behavior (stable-sort / `>=` semantics), so their quality tiers were retuned so the intended orderings still hold under the threshold rule.

## Parity evidence

Source of truth is **Picard/htsjdk** (fgumi `dedup` is MarkDuplicates-style; fgbio has no standalone dedup — tracker BS3). fixture: `reports/fgbio-parity-fixtures/dedup-02-sum-base-quals.sh`.

A 2-read UMI family, **mapq held equal**, seq length 4, crafted so the two scorings pick different winners:

| read | quals | Picard `Σ(q≥15)` | old `Σ min(q,15)` |
|---|---|---|---|
| readA | `[40,40,14,14]` | **80** | 58 |
| readB | `[16,16,16,16]` | 64 | **60** |

- **before**: fgumi keeps `readB` (cap: 60 > 58) — **DIFFERs** from the Picard winner `readA`.
[...]

PR #649, opened 2026-07-23
fix(dedup)!: group --no-umi orientation-agnostically like Picard MarkDuplicates

nh13:
> Stacked on #648. Review #648 first; this PR's diff is against that branch. The base will retarget to `main` automatically once #648 merges.

## Summary

`fgumi dedup --no-umi` sub-partitioned each coordinate family by strand of origin, so an F1R2 and an F2R1 template at the same unclipped 5' coordinates became two molecules and each kept a representative. This PR gates that split on `!no_umi`, so `--no-umi` now groups orientation-agnostically like Picard `MarkDuplicates`.

The split is wrong for duplicate marking. Standard prep ligates the same asymmetric Y-adapter to both ends of a double-stranded fragment; read 1 is primed from the P5 end, so **one** input fragment yields both an F1R2 and an F2R1 read pair. Strand of origin distinguishes the two halves of one molecule, not two molecules — so splitting on it while marking duplicates keeps two reads per input fragment and double-counts it.

## Scope — dedup `--no-umi` only

- **UMI path unchanged.** `identity`/`adjacency` already separate the strands via the `A-B` vs `B-A` duplex pair; `paired` re-introduces the separation through `/A` and `/B` (which is why it reports `split == false`).
- **`group` unchanged.** Consensus calling is a positional pileup (`vanilla_caller.rs`, `for pos in 0..consensus_len`) and requires every read 1 in a molecule to be the same physical strand from the same 5' start. Merging strands there would produce a meaningless consensus. `dedup` runs no consensus, so it needs no split.

The change is one predicate, extracted into a documented `splits_by_strand_of_origin(assigner, no_umi)` helper.

## Why the split is wrong here — measured with the sample's own UMIs

The `twist-umi` library has duplex UMIs, stored in read order, so the two strands of one fragment carry swapped pairs (`L-R` vs `R-L`). For the coordinate families `--no-umi` splits:

| | chr7 | chr1 |
|---|---|---|
| Strand-split families | 7,307 | 10,873 |
| Bottom side carries the **swap** of a top-side UMI (exact) | 94.76% | 88.16% |
| ... allowing ≤1 sequencing error per half | 97.11% | 91.72% |
| Identical *without* swapping (premise check) | 0.23% | 0.12% |
| Shuffled control (swap-match by chance) | 0.00% | 0.05% |

~88–95% of split families are provably one duplex molecule against a ~0% background, so the split double-counts them.

## Verification against Picard MarkDuplicates 3.4.0

`twist-umi`, 15,826,646 primary reads, same input templates through both tools:

| | before (split) | after (this PR) |
|---|---|---|
| Mechanism test — fgumi marked vs coordinate-only key | off by −7,307 | **MATCH exactly** |
| `picard-dup-only` disagreement (chr7) | 7,398 | **109** |
| Per-template flag agreement (chr7 / chr1) | 97.78% / 97.73% | **99.94% / 99.92%** |

fgumi's marked count now equals Picard's coordinate-only expectation to the read (chr7 159,666; chr1 208,491). The residual is perfectly symmetric (109 vs 109; 186 vs 186) — pure representative tie-break (fgumi keeps the first template in template-coordinate order; Picard the earliest `read1IndexInFile`), which is intentional and unchanged by this PR. Combined with #490's `SUM_OF_BASE_QUALITIES` scoring fix, `dedup --no-umi` now agrees with Picard on 99.9% of primary templates.

## Breaking change

`fgumi dedup --no-umi` marks **more** duplicates than before (duplicate rate +~2 pp on a moderately duplicated panel; 40.99% → 43.02% on `twist-umi`). There is no flag to restore the old strand-separated no-UMI behavior; use the UMI grouping path if strand separation is needed. Labeled `fix(dedup)!` with a BREAKING CHANGE trailer; should ship in the same release as #490.

## Testing

- New `#[rstest]` cases: F1R2 + F2R1 at identical coordinates → **one** molecule under `--no-umi`, **two** with UMIs; plus a direct `splits_by_strand_of_origin` gate test. Confirmed RED without the fix (`left: 2, right: 1`), GREEN with it.
- `cargo ci-fmt`, `cargo ci-lint`, `cargo ci-test` clean — 5867 passed, 27 skipped.
- Amends #648's "Grouping key" help section, which said there was no orientation-agnostic option, so no commit ships contradictory docs. Regenerated `docs/` locally to confirm rendering.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated deduplication behavior so strand-of-origin splitting is applied only when UMI-based grouping is enabled.
  * Corrected `--no-umi` behavior to group templates by coordinate alone, aligning duplicate counts with expected results.

* **Documentation**
  * Clarified command help text and `--no-umi` option semantics.
  * Documented the distinct strand-handling behavior used during grouping for consensus calling.

### DIFF
--- a/src/lib/commands/dedup.rs
+++ b/src/lib/commands/dedup.rs
@@ -215,7 +215,7 @@ struct DedupFilterConfig {
     include_non_pf: bool,
     /// Minimum UMI length.
     min_umi_length: Option<usize>,
-    /// Skip UMI validation; group by template coordinate and strand of origin alone.
+    /// Skip UMI validation; group by template coordinate alone, orientation-agnostically.
     no_umi: bool,
 }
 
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #651, opened 2026-07-23
fix(index): position-bin placed-but-unmapped reads in the BAI

tfenne:
## What & why

A read with the UNMAPPED flag but a valid reference and position — the common "mate mapped, self unmapped" record, which carries its mate's `tid`/`pos` so it coordinate-sorts alongside its mate — was dropped from the BAI entirely: no bin, no chunk, no linear-index entry. htslib/samtools position-bin such reads, so region queries and `idxstats` over a fgumi index silently missed them (a "fetch everything in this window" would lose the unmapped mates, and the `idxstats` unmapped column disagreed with samtools).

## What it does

`extract_alignment_context` now returns a binning context for any read that has a reference and position, binning it over a 1-base `[pos, pos+1)` span when it has no CIGAR (matching htslib's `bam_endpos = pos + max(rlen, 1)`); only a read with no reference/position is treated as unplaced. The `is_mapped` flag still records the read as unmapped, so it isn't miscounted as mapped — it just becomes queryable, matching htslib. Normal mapped reads (CIGAR span ≥ 1) bin identically to before, so the blast radius is limited to placed-unmapped reads.

This changes `.bai` content for both index paths (the pooled sort writer and `IndexingBamWriter`), which is the intent: both now match samtools on content.

## Validation

- New unit test: a placed-but-unmapped record → binned over `[pos, pos+1)` with `is_mapped=false`; a mapped read → its CIGAR span; a truly-unplaced read → unplaced.
- The samtools cross-check (`test_sort_write_index_matches_samtools_index`) now compares ALL reads (dropping the previous `-F 4` mapped-only workaround) and full `idxstats` — and matches.
- Full suite green: `cargo ci-fmt`, `cargo ci-lint` (clippy pedantic), and all tests including the samtools-gated ones.

Stacked on #650 (BAI compaction); merge that first.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved BAM indexing for reads marked as unmapped but containing valid reference positions.
  * These reads are now included in the appropriate genomic regions and represented with their correct mapping status.
  * Corrected indexing for records without CIGAR-derived reference spans by treating them as covering one base.
  * Region queries and index statistics now more closely match samtools results, including placed-but-unmapped reads.

### DIFF
--- a/crates/fgumi-bam-io/src/writer.rs
+++ b/crates/fgumi-bam-io/src/writer.rs
@@ -366,8 +366,14 @@ struct CachedIndexEntry {
 
 /// Extract alignment context from raw BAM bytes.
 ///
-/// Returns `Some((ref_id, start, end, is_mapped))` for mapped reads, or `None`
-/// for unmapped reads (the indexer still counts them via a `None` context).
+/// Returns `Some((ref_id, start, end, is_mapped))` for any read that has a
+/// reference and a position — **including a placed-but-unmapped read** (e.g. the
+/// unmapped mate of a mapped read, which carries its mate's `tid`/`pos` so it
+/// sorts alongside it). Such reads are position-binned exactly as htslib/samtools
+/// do, so region queries and `idxstats` see them; the `is_mapped` flag records
+/// that they are unmapped without excluding them from the index. Only a truly
+/// unplaced read (no reference or no position) returns `None`, which the indexer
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

nh13:
Thank-you @tfenne!!!

PR #652, opened 2026-07-23
perf(consensus): resolve unanimous positions via a margin-gated gap table

nh13:
Stacked on #635 (base: `nh/perf-memoize-unanimous-gap`). Review after that one.

## What

Above pre-UMI Q69 the unanimous fast path's cap shortcut fires on fewer positions, so more unanimous positions fall through to the full log-sum-exp chain. This resolves them by binary search over a per-Phred-level table of winner-minus-loser gap thresholds, precomputed once per pre-UMI rate and memoized in a 256-slot `OnceLock`.

For a unanimous position the posterior is a pure function of the gap, `1/(1 + 3e^-gap)`, so quality is a monotone step function of the gap with one step per Phred level. The table stores those step boundaries; a `partition_point` over it yields the quality with no transcendental evaluation.

## Why it is safe

The table is built from the shifted `[0, -gap, -gap, -gap]` evaluation, which is **not** bit-identical to `call_full`'s `[w, l, l, l]` form — the latter carries the loser magnitude through the accumulation, and its final normalize cancels two `~|l|`-sized terms. **`call_full` is left untouched and remains the sole oracle**, so the fast path returns a table value only where it provably agrees, and defers otherwise:

- the observed base must be the unique maximum-likelihood lane (`gap` finite and `> f64::EPSILON`); otherwise `call_full` resolves the position, which ties or no-calls;
- **sub-cap**: the gap must clear a rounding margin from both edges of its bracket, `C * u * (|w| + |l|) / c`, where `c` is the consensus error at the bracket's far edge (its minimum, since `c` is strictly decreasing) — a mean-value-theorem mapping of the ln-posterior error bound into gap space;
- **cap region**: the ln-posterior budget `Q(gap) - Q(t_cap)` *saturates* at `~c(t_cap)`, so extra gap buys nothing. The gate is `gap - t_cap >= ln 2 && delta < c(t_cap)/2`, not a linear allowance.

Anything failing a check, or yielding a non-finite margin, defers to `call_full`. Every failure mode is "defer", never "return an unvalidated value".

`UNANIMOUS_MARGIN_HEADROOM = 16` gives ~4.16x headroom over the worst ratio observed (3.845) across 643,568 samples spanning all four winner lanes, post-UMI Q0-93, observation quality Q0-93 and depths 1-25,600. The lane axis matters: `ln_sum_exp_array` seeds with the min lane and folds in index order, so a winner at lane G/T rounds differently than at A/C (3.845 vs 1.173) — calibrating on base `A` alone understates the bound by 3.3x.

## Two pre-existing bugs this surfaced

Both also exist on #634/#635 and are worth fixing there independently:

1. **Degenerate certain-error inputs.** When the per-observation adjusted error is `>= 3/4` (`--error-rate-post-umi 0`, or an observation quality of 0), the observed base is not the maximum-likelihood lane. The old fast path still returned it; `call_full` correctly no-calls. Now deferred.
2. **Cap-boundary rounding.** #634's cap bound is derived from the shifted form, but `call_full` evaluates the unshifted form, so just above the threshold the cancellation drops `call_full` one Phred short. Reproducible at `-1 90 -2 90` with a unanimous family of ~3200 reads (Q90 vs Q89) — the shipped `SimpleConsensusCaller` defaults.

## Correctness

Consensus output is unchanged: `call() == call_full()` everywhere.

- Dense parity sweep that scans depth **contiguously** (what lands a gap near a bracket boundary) across all four bases, plus a broad rate/quality/depth grid, a targeted deep-cap-region regression, and a degenerate-input sweep.
- Direct unit tests for the internals: `unanimous_margin` (including the `+INFINITY` fail-safe), `consensus_error`, the table invariants (`cerr_min[q] == consensus_error(thresholds[q+1])`, cap indexing, sentinels), and the gate decision itself.
- The error bound is pinned by a test that asserts both the hard bound and a ceiling on the observed ratio, so headroom erosion fails the build rather than silently consuming the margin.
- Mutation-checked: zeroing the headroom, or restoring an unconditional cap return, each makes the parity tests fail with the specific reproducers above.
- End to end, `simplex` output is byte-identical (record-body md5, 200,000 records) to the base across `-1/-2` of 45/40, 70/40, 90/90, 93/93, 20/10 and 2/2.

## Performance

`simplex --min-reads 1 --threads 1` on a 796k-read grouped BAM, median of 5, single-threaded:

| `-1`/`-2` | base | this PR | delta |
|---|---|---|---|
| 45 / 40 (CLI default) | 3.03s | 2.85s | -5.9% |
| 70 / 40 | 3.42s | 2.72s | -20.5% |
| 90 / 90 | 3.65s | 2.85s | -21.9% |

Runtime is now essentially independent of the pre-UMI error rate (2.72-2.85s across all three), where the base degrades from 3.03s to 3.65s as the rate rises.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consensus base-calling accuracy and reliability in unanimous cases.
  * Added safeguards to ensure optimized decisions match the full consensus calculation.
  * Improved handling of boundary conditions and high-depth cases.

* **Tests**
  * Expanded coverage for accuracy, rounding, threshold boundaries, and cached lookup behavior.

### DIFF
--- a/crates/fgumi-consensus/src/base_builder.rs
+++ b/crates/fgumi-consensus/src/base_builder.rs
@@ -180,6 +180,86 @@
 //! The consensus quality increases logarithmically with depth, reflecting the compounding
 //! evidence that all observations support the same base.
 //!
+//! ## Unanimous Fast-Path Correctness (Margin-Gated, Not "By Construction")
+//!
+//! `ConsensusBaseBuilder::call_full` is the unmodified fgbio computation — for
+//! every position, unanimous or not, it runs the general `[w, l, l, l]` (or
+//! four-way) `ln_normalize` -> `ln_not` -> `ln_error_prob_two_trials` ->
+//! `ln_prob_to_phred` chain. It never consults `unanimous_quality_from_gap` or the
+//! gap-threshold table. `call_full` is therefore the sole oracle: fgbio parity is
+//! defined as "output equals `call_full`", by definition, not by an equivalence
+//! this module has to prove separately.
+//!
+//! `ConsensusBaseBuilder::try_unanimous_fast_path` is a *shortcut* over a
+//! precomputed per-Phred-level gap-threshold table (`unanimous_quality_from_gap`'s
+//! step function, built by `build_unanimous_gap_thresholds`). It returns the
+//! table's answer only when a stated **sufficient condition** establishes that the
+//! answer equals what `call_full` would have computed, and returns `None` (defer
+//! to `call_full`) in every other case. The sufficient condition differs between
+//! the two regions, because the quantity being bounded behaves differently in each:
+//!
+//! - **Sub-cap region** (`gap < thresholds[cap]`): `call_full`'s `[w, l, l, l]`
+//!   normalize and the table's shifted `[0, -gap, -gap, -gap]` form compute the
+//!   *same* mathematical posterior and differ only by floating-point rounding,
+//!   empirically bounded by `delta = C * u * (|w| + |l|)`, `u = f64::EPSILON/2`,
+//!   `C = 16` (an observed worst-case ratio of ~3.845x against a pinned ceiling of
+//!   5.0x; see `test_ln_posterior_full_gap_error_bound`, which asserts both the
+//!   hard bound and the ratio ceiling, and the calibration note below). Mapping
+//!   that ln-posterior perturbation to a gap perturbation via the mean value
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #653, opened 2026-07-24
fix(sort): honour --compression-level for the tail of a spilling sort

nh13:
Follow-up to #633. That PR fixed the fully-in-memory sort, which never entered Phase 2 at all and so wrote its *entire* output at `temp_compression`. This fixes the remaining leak on the **spilling** path, where only the tail of the file is affected.

## The bug

`Phase2Guard::deactivate` returned the pool to `LEGACY` *before* the output writer was finished, at all four merge sites. But `select_priorities` only offers `CompressOutput` in Phase 2 — in `LEGACY` the sole compress step is `CompressSpill`, which binds `worker.compressor` (the `temp_compression` one). Every output block still sitting in the compress queue at teardown, plus the writer's final flush, was therefore compressed at the temp level, and the tail of the BAM silently ignored `--compression-level`.

The comments at both merge sites ("workers still need to compress", "CompressOutput is eligible whenever the queue is non-empty") describe the intent correctly, and `is_available` really does gate `CompressOutput` on a non-empty queue. What they miss is that in `LEGACY` the step is never dispatched in the first place, so the eligibility check never gets consulted.

## Measured

5,000-pair coordinate sort forced to spill, writing at `--compression-level 0`:

| `temp_compression` | output size |
|---|---|
| 0 | 2,062,091 bytes |
| 9 | 2,021,843 bytes |

~40 KB (~2%) of the output came out at the temp level. The in-memory path is unaffected — it never builds a `Phase2Guard`, so there is no early teardown to get wrong, and it produces exactly the 2,062,091-byte level-0 output either way.

## The fix

Split the guard's teardown into its two halves:

- `release_sources` drops the consumer and unpublishes the spill files — which is what stops workers touching files that are about to be deleted — while leaving the pool in Phase 2, so trailing blocks still route to `output_compressor`.
- `deactivate` returns the pool to `LEGACY` after the writer is finished, and remains what `Drop` calls, so error paths are unchanged.

`try_phase2_file_work` returns immediately on an empty file vector, so output compression is the only Phase 2 work left in that window.

## Testing

New integration test `test_temp_compression_does_not_reach_the_output_bam` sorts twice at `output_compression 0`, varying only the temp level, and asserts the two outputs are the same size. It fails on all three pooled sort orders before this change:

```
temp_compression leaked into the output for Coordinate: the same sort wrote 2062091 bytes
at temp_compression=0 and 2021843 bytes at temp_compression=9, but only output_compression
(0 for both) may affect the output
```

`sort_at_level` gained a `temp_compression` parameter so both tests share it; the existing test passes `1` as before.

`cargo ci-fmt`, `cargo ci-lint` clean; `cargo ci-test` 6476 passed / 27 skipped.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed output compression so the final BAM consistently honors the selected output compression level (not temporary/spill settings), including spilling runs.
  * Improved Phase 2 merge finalization to ensure writer/index output is finalized at the correct stage, including when all sources are empty.
* **Refactor**
  * Clarified merge teardown so output finalization happens before the worker pool transitions to the earlier phase.
* **Tests**
  * Added integration coverage for BGZF compressor routing across spilling and in-memory paths, plus deterministic checks for BAI/index bytes.
  * Added unit tests validating the Phase 2 teardown contract and empty-merge behavior.

### DIFF
--- a/crates/fgumi-sort/src/external.rs
+++ b/crates/fgumi-sort/src/external.rs
@@ -1643,15 +1643,20 @@ pub struct RawExternalSorter {
 }
 
 /// RAII guard that ensures Phase 2 teardown runs on every exit path between
-/// `pool.set_phase(PHASE2)` and the explicit `deactivate()` in the merge loop.
-/// Without this, any `?` early-return would leave the pool stuck in PHASE2 with
-/// `phase2_files` still published. `deactivate()` drops the consumer (releasing
-/// the Arc snapshot of the per-file vector), resets the phase to `LEGACY`, and
-/// clears the pool's published file vector — in that order.
+/// `pool.set_phase(PHASE2)` and the end of the merge. Without this, any `?`
+/// early-return would leave the pool stuck in PHASE2 with `phase2_files` still
+/// published.
+///
+/// Merges finish by handing the guard to
+/// [`finish_output`](Self::finish_output), which owns the whole teardown order —
+/// the merge is done with its *sources* well before it is done with the *pool*,
+/// and getting that order wrong is silent. `Drop` still deactivates, so a merge
+/// that fails before it gets that far tears down completely.
 struct Phase2Guard<'a, K: RawSortKey + 'static> {
     pool: &'a Arc<SortWorkerPool>,
     consumer: Option<MainThreadChunkConsumer<K>>,
     active: bool,
+    sources_released: bool,
 }
 
 impl<K: RawSortKey + 'static> Phase2Guard<'_, K> {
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #656, opened 2026-07-25
fix(pipeline): don't declare a stage done while its items are still in flight

nh13:
Two pipeline stages could declare themselves finished while items were still in flight, stalling the run until the deadlock detector fired. Both are pre-existing on `main`.

The first surfaced as a flaky `test_extract_bgzf_async_reader_multithreaded` failure under a loaded full-suite run:

```
Queue depths: Q1=0 Q2=0 Q2b=2 Q3=0 Q4=0 Q5=0 Q6=0 Q7=0
State: read_done=true, group_done=true, draining=true
Extra: block_merge_done=true, parse_done=true, batches: read=2 block_parsed=2 merged=0
Pipeline counters: ... total_templates_pushed=0, templates_written=0
```

Two parsed blocks stranded in `q2_block_parsed` while the only stage that drains it had already finished, and nothing was ever written. Investigating it turned up the same class of bug in the BAM pipeline.

## The class

A completion predicate must account for **every** place an item can sit between produced and consumed, because a downstream stage may latch finished off that flag and orphan whatever it missed. There are three such places, and each bug here missed a different one:

| Place | Covered by | Missed in |
|---|---|---|
| the stage's input/output queue | testing the queue directly — counters don't cover it | **FASTQ `BlockMerge`** |
| a worker-local held slot | counter shape, since a held item is invisible to other workers | **BAM `FindBoundaries`** |
| the stage's own intermediate state | pending maps, surplus, remainders | — (both already covered) |

## `fix(fastq)` — `BlockMerge` declared done with blocks still queued

`BlockMerge` had two completion checks that had drifted apart:

| Term | early-return | end-of-step |
|---|---|---|
| `read_done && chunks_block_parsed == batches_read` | ✅ | ✅ |
| pending maps / surplus / suffix empty | ✅ | ✅ |
| **`q2_block_parsed.is_empty()`** | ❌ **missing** | ✅ |

The drain loop pops from a momentarily empty queue, `BlockParseFast` then pushes the final blocks and brings the counter level with `batches_read`, and the early return concludes the pipeline is done. Nothing pops them afterwards — `BlockParseFast` returns early once `block_merge_done` is set.

Both checks now go through one `block_merge_is_complete`, and both exits through one `finish_block_merge_if_complete`, so condition and action cannot drift again. The queue term is sound because `BlockParseFast` pushes *before* it counts, so observing the counter caught up through an `Acquire` load happens-after every push; reading the queue after that load distinguishes "drained" from "not yet landed".

## `fix(bam)` — `FindBoundaries` finished with a batch held for retry

`batches_boundary_processed` is incremented when a batch is popped from `q2_reorder`, *before* its boundaries are pushed. When the push fails the batch moves to `WorkerState::held_boundaries` — already counted as processed, absent from `batches_boundary_found`, and in no queue. Any worker holding the boundary lock then sees `batches_boundary_processed == next_read_serial` and sets `boundary_done` with that push outstanding. `Group` finishes on `batches_grouped == batches_boundary_found`, which excludes the held batch, and latches its one-way `finished` flag; the batch is pushed and decoded afterwards, reaching an already-finalized grouper while `q3` never drains.

This one **stalls rather than losing records** — `has_any_held_items` keeps the holding worker alive — but the grouper has already emitted its final group, so a UMI family spanning that boundary is split.

The predicate now also requires `batches_boundary_found == next_boundary_serial`. A serial is minted only immediately before a push and `found` only when one succeeds (a retried held batch reuses its serial), so the two are equal exactly when nothing is outstanding — no worker needs to see another's held slot. It cannot be built from `batches_boundary_processed`: a batch with no records is counted as processed but mints no serial and is never pushed, so `found == processed` is not an invariant even when nothing is held.

## Also

`base.rs` gains the general rule beside the existing worker-exit invariant it belongs next to — the three places, which bug came from missing which, and the three ways the other stages already avoid the held-slot half (count after the push; count at serial-assignment so a held item raises the target; or keep the un-pushed item inside the shared lock). That doc is what would have prevented both bugs; a runtime mechanism would be redundant for the six stages that are already safe for better reasons.

## Testing

Both races need a specific interleaving. Neither reproduced in 25 stress reps under full CPU saturation — **in either direction**, so the stress runs show no regression rather than validating the fixes. The regression tests therefore pin the predicates directly: reverting only the new term in each fails exactly one case (`blocks_still_queued`, `batch_held`) and no other. Empty input is pinned too, since the failure mode there is hanging on a record-less file.

`cargo ci-test` (6537 tests), `ci-fmt`, `ci-lint`, and `ci-doc` all pass.

## Follow-up (not in this PR)

`bam.rs` has two dead completion flags found during the audit: `decompress_done` is never set or read, and `decode_done` is written but never read — and its predicate already omits `held_decoded`, making it a latent third instance of this class should anyone wire it up. Both are pre-existing and untouched here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pipeline completion detection to ensure queued, held, and buffered data is fully processed before marking stages complete.
  * Prevented premature completion that could leave data stranded or cause processing to stall.
  * Added handling for empty-input completion and final pending batches.

* **Documentation**
  * Clarified worker completion requirements for queued, held, and intermediate work.

* **Tests**
  * Added regression coverage for delayed consumption, held batches, active readers, and empty inputs.

### DIFF
--- a/src/lib/unified_pipeline/bam.rs
+++ b/src/lib/unified_pipeline/bam.rs
@@ -2031,6 +2031,46 @@ fn try_step_decompress<G: Send, P: Send + MemoryEstimate>(
     }
 }
 
+/// Whether `FindBoundaries` has consumed all its input and pushed all its output,
+/// and so may set `boundary_done`.
+///
+/// Two things must hold, and the second is easy to miss.
+///
+/// **All input consumed.** `batches_boundary_processed == next_read_serial` implies
+/// `Decompress` also finished, since no more can be processed than was
+/// decompressed, so this also rules out data still sitting in `q2_decompressed`.
+///
+/// **No output still outstanding.** `batches_boundary_processed` is incremented
+/// when a batch is popped from `q2_reorder` — *before* its boundaries are pushed —
+/// so a batch whose push failed is already counted as processed while it waits in
+/// some worker's `held_boundaries`, in no queue and invisible to the other workers.
+/// Setting `boundary_done` there lets `Group` finish on
+/// `batches_grouped == batches_boundary_found`, which excludes that batch, and latch
+/// its one-way `finished` flag; the batch is then pushed and decoded, and reaches an
+/// already-finalized grouper.
+///
+/// `next_boundary_serial == batches_boundary_found` is what rules that out. A serial
+/// is minted only immediately before a push is attempted, and
+/// `batches_boundary_found` is incremented only when a push succeeds (including the
+/// retry of a held batch, which reuses its serial rather than minting a new one), so
+/// the two are equal exactly when nothing is outstanding. Note this cannot be done
+/// with `batches_boundary_processed`: a batch with no records is counted as processed
+/// but mints no serial and is never pushed, so `found == processed` is not an
+/// invariant even when nothing is held.
+///
+/// `batches_boundary_found` is loaded before `next_boundary_serial` so the comparison
+/// errs toward "not complete": a batch outstanding at the first load forces
+/// `next_boundary_serial` above the value read, and it only ever grows.
+fn boundary_finding_is_complete<G: Send, P: Send + MemoryEstimate>(
+    state: &BamPipelineState<G, P>,
+) -> bool {
+    state.read_done.load(Ordering::Acquire)
+        && state.batches_boundary_processed.load(Ordering::Acquire)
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai review

PR #657, opened 2026-07-25
chore(bam): remove two dead pipeline completion flags

nh13:
Follow-up to #656, found while auditing `BamPipelineState`'s completion graph. Independent of that PR — this branches from `main` and touches none of the same lines.

Two `*_done` flags were dead:

| Flag | State |
|---|---|
| `decompress_done` | Declared and initialized, then never touched — never set, never read. |
| `decode_done` | Set in exactly one place, read nowhere. Write-only state that looked load-bearing. |

## Neither is missed

- **`Decompress`** is tracked downstream through `batches_decompressed`, and `FindBoundaries` infers its completion from `batches_boundary_processed == next_read_serial` — nothing can be processed that was not decompressed first.
- **`Decode`** is tracked by `Group` through `batches_grouped == batches_boundary_found`.

Both are recorded as comments where the fields were, so the next reader doesn't go looking for a flag that was deliberately removed.

## Why `decode_done` is worth deleting rather than leaving

Its guard was `boundary_done && q2b_boundaries.is_empty()`, which does **not** account for a batch sitting in a worker's `held_boundaries`. Anyone who wired the flag up to a consumer would have reintroduced exactly the stranded-batch bug #656 fixes in `FindBoundaries` — the flag is inert today only by accident. Deleting it removes the trap instead of leaving a loaded gun behind a plausible-looking name.

## Behavior

Unchanged. Dropping the store left the `else if` arm that records Q2b starvation as the only live branch, so the two `state.stats()` blocks fold into one. The recording condition is the De Morgan equivalent of the original (`!(boundary_done && q2b empty)`), and the comment now explains why an empty Q2b past that point is the terminal state rather than a stall.

Both fields were `pub` on `BamPipelineState` — internal pipeline machinery with no reader outside the module.

`cargo ci-test` (6586 tests), `ci-fmt`, `ci-lint`, and `ci-doc` all pass.

## Coverage

Rewriting that stats arm dropped patch coverage to 0% on the three lines it touches — `state.stats()` is `None` unless stats collection is enabled, and nothing turned it on, so the arm had never been exercised. It is covered now by `test_decode_records_q2_starvation_only_while_boundaries_may_arrive`, which also turns the "condition unchanged" claim above into something checked rather than asserted: inverting the condition fails both cases.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved BAM processing completion and queue-handling behavior.
  * Reduced incorrect starvation reporting when no additional boundaries can arrive.
  * Improved pipeline end-of-step detection using batch and boundary progress.

### DIFF
--- a/src/lib/unified_pipeline/bam.rs
+++ b/src/lib/unified_pipeline/bam.rs
@@ -595,8 +595,11 @@ pub struct BamPipelineState<G, P: MemoryEstimate> {
 
     // ========== Step 2: Decompress (parallel) ==========
     // No state needed - each thread has its own Decompressor
-    /// Flag indicating all decompression is done.
-    pub decompress_done: AtomicBool,
+    //
+    // There is no `decompress_done` flag: downstream stages track this step through
+    // `batches_decompressed`, and `FindBoundaries` infers it from
+    // `batches_boundary_processed == next_read_serial` (nothing can be processed that
+    // was not decompressed first).
     /// Count of batches that have completed decompression (for completion tracking).
     pub batches_decompressed: AtomicU64,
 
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #658, opened 2026-07-25
refactor(sort): pick the compressor from the job, not the pipeline phase

nh13:
Follow-on to #633 and #653, which each fixed one symptom of the same root cause. This fixes the cause.

## The coupling

One `compress_queue` carries both spill and output blocks, and `CompressJob` recorded only the *codec* (BGZF vs zstd) — never which of the two kinds it was. So a worker popping a job re-derived the compression *level* from `shared.phase`, a mutable atomic that advances independently of what is already sitting in the queue. Every block that outlived a phase transition was compressed at the wrong level, silently.

That one coupling produced the same bug twice, in opposite directions:

- **Output blocks popped outside Phase 2** went through the spill compressor at `temp_compression`, so `--compression-level` was silently discarded. Fixed twice at the symptom layer: by entering Phase 2 even for an in-memory sort (#633), then by keeping the pool in Phase 2 across the writer's `finish()` (#653).
- **Phase 1 spill blocks still queued when `begin_phase2` fired** went through the output compressor at `output_compression`. This was never a live bug — it was held off only by `drain_pending_spill` happening to sit immediately before `enter_output_phase`, and was written down as a caller obligation on `begin_phase2` rather than enforced. Its cost is wasted CPU and disk, not wrong bytes, since any BGZF level decodes.

## The fix

The identity was already known for certain at submit time, by the one caller who cannot be wrong about it: there is a single production submit site, `StagingBuffer::flush`, and its two constructors are statically distinct — the output writer and the spill writer. So record it rather than infer it.

`CompressJob` gains a `target: CompressTarget { Spill, Output }` beside the existing `codec`; `StagingBuffer` carries it and stamps every job it submits; `try_compress` selects `compressor` or `output_compressor` from `job.target`. Nothing about the choice reads shared mutable state, so a block is compressed at the level its writer asked for however long it waited in the queue and whatever phase the pool is in when a worker gets to it.

This is the second form of that selection. `5df7e176`, which introduced the worker pool, read it from `shared.phase` at pop time and then narrowed it to the dispatched `SortStep` to close the window where `set_phase` fired between the pop and the choice. But a step is itself chosen from the phase, so blocks that outlived a transition were still mis-compressed — the window was narrowed, not closed.

With the level travelling on the job, `SortStep::CompressSpill` and `CompressOutput` did the same thing, so they collapse into one `Compress`. There was only ever one queue; two steps implied two, and the per-step stats buckets they fed (`CmpSpl` / `CmpOut`) could not be trusted to mean what they said. `SortStep::COUNT` drops from 5 to 4. `SortStep` is `pub` but `worker_pool` is `pub(crate)`, so this is not an API change.

Also drops `begin_phase2`'s caller obligation, which no longer exists, and pins zstd's spill-only invariant now that it is expressible: there is one zstd compressor per worker, fixed at `temp_compression`, because the output BAM is always BGZF. That one asserts unconditionally rather than in debug only — an `Output` job reaching it would be the same silent wrong-level output this mechanism exists to prevent, and the check is one compare inside the zstd arm, which the BGZF path never evaluates.

The ordering #653 introduced — release the drained sources, finalize, then leave Phase 2 — is kept, but for the reason that survives: not holding the merge's file descriptors and 2 MiB-per-chunk reorder buffers across a slow `finish`. Its compression rationale is gone, and its docs and tests say so.

## Testing

Reverting #653's ordering now leaves `test_temp_compression_does_not_reach_the_output_bam` **passing** — that is the check that this fixes the cause rather than adding a third symptom fix.

`test_compress_target_decides_level_regardless_of_phase` sweeps the phase across `LEGACY` / `PHASE1` / `PHASE2` and asserts a `Spill` job stays stored at `temp_compression = 0` while an `Output` job compresses at `output_compression = 9`. All three cases fail if the compressor is selected from the phase, including the mirror direction that had no coverage before.

`cargo ci-fmt`, `cargo ci-lint`, `cargo ci-doc` (with `-D warnings`, since this adds intra-doc links), and `cargo ci-test` (6597 tests) all pass.

## Performance

No measurable cost. The job grows by one byte in a struct that already owns a `Vec<u8>`, and the per-block work is one `match` on a `Copy` enum replacing one `match` on the step. Scheduling is untouched: the two collapsed steps shared an eligibility predicate, and neither was ever exclusive (only `ReadInputBlocks` is).

Measured on a spilling coordinate sort of a 3,689,310-record BAM (idt-cfdna, 162 MB, 4 spill chunks, 8 threads, `-m 256m`, output level 6, temp level 1), 10 interleaved reps per binary, baseline being #653's merge commit `ef0f6f8b`:

| metric | baseline | this branch | delta |
|---|---|---|---|
| CPU (user+sys) | 11.497 s ± 0.312 | 11.326 s ± 0.462 | −1.5% |
| wall | 5.663 s ± 0.941 | 5.910 s ± 1.044 | +4.4% |

Welch *t* on CPU time is −0.97, so the difference is not significant. Wall clock on this host is too noisy (σ ≈ 1 s) to resolve anything under ~15%, which is why CPU time is the reported metric. All 20 runs exited 0 and wrote exactly 3,689,310 records. The two binaries produce equivalent output: the sorted BAMs differ only in the `@PG` `CL:` field, which records the binary's path, and the record streams have identical checksums.

## Reading order

`crates/fgumi-sort/src/worker_pool.rs` is the change — `CompressTarget`, the `CompressJob` field, `SortStep`, and `try_compress`. Everything else follows from it: `bgzf_io.rs` threads the field, the two pooled writers each stamp their constant, and `external.rs` plus the integration test are documentation that had asserted the old coupling.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured temporary spill data and final BAM output are compressed using the correct, phase-independent settings.
  * Prevented compression behavior from changing incorrectly during worker-pool phase transitions.
  * Improved correctness during both disk-spill and in-memory-only processing.

* **Tests**
  * Added/updated coverage verifying compression targets stay consistent across legacy and phased execution.
  * Updated lifecycle and output-compression test messaging to reflect the refined teardown/output invariants.

* **Documentation**
  * Refreshed comments describing phase transition and compression-selection behavior.

### DIFF
--- a/crates/fgumi-sort/src/bgzf_io.rs
+++ b/crates/fgumi-sort/src/bgzf_io.rs
@@ -5,7 +5,9 @@
 //! accumulating data into ~64KB blocks before submitting compression jobs.
 
 use crate::codec::SpillCodec;
-use crate::worker_pool::{BufferPool, CompressJob, CompressResult, PermitPool, SortWorkerPool};
+use crate::worker_pool::{
+    BufferPool, CompressJob, CompressResult, CompressTarget, PermitPool, SortWorkerPool,
+};
 use anyhow::Result;
 use crossbeam_channel::{Receiver, Sender};
 use fgumi_bgzf::{BGZF_EOF, BGZF_MAX_BLOCK_SIZE};
@@ -39,6 +41,10 @@ pub(crate) struct StagingBuffer {
     result_tx: Sender<CompressResult>,
     permit_pool: Arc<PermitPool>,
     codec: SpillCodec,
+    /// Stamped onto every job this buffer submits, so the worker that pops it
+    /// compresses at the level this writer asked for regardless of the pool's
+    /// phase at that moment.
+    target: CompressTarget,
 }
 
 impl StagingBuffer {
@@ -49,6 +55,7 @@ impl StagingBuffer {
         result_tx: Sender<CompressResult>,
         permit_pool: Arc<PermitPool>,
         codec: SpillCodec,
+        target: CompressTarget,
     ) -> Self {
         Self {
             pool,
@@ -57,6 +64,7 @@ impl StagingBuffer {
             result_tx,
             permit_pool,
             codec,
+            target,
         }
     }
 
@@ -123,6 +131,7 @@ impl StagingBuffer {
             serial,
             result_tx: self.result_tx.clone(),
             codec: self.codec,
+            target: self.target,
         });
         Ok(())
     }
@@ -331,7 +340,13 @@ mod tests {
             io_writer_loop(writer, result_rx, buffer_pool, pp, codec, None)
         });
 
-        let mut staging = StagingBuffer::new(Arc::clone(&pool), result_tx, permit_pool, codec);
+        let mut staging = StagingBuffer::new(
+            Arc::clone(&pool),
+            result_tx,
+            permit_pool,
+            codec,
+            CompressTarget::Spill,
+        );
         staging.write_chunked(data).unwrap();
         staging.flush().unwrap();
         drop(staging); // closes result_tx senders → io_writer_loop exits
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #659, opened 2026-07-25
refactor(io): one BGZF header predicate, one open per compare input, and plain-gzip input

nh13:
Three follow-ups from the #644 review, each independent. Stacked on `nh/feat-accept-sam-input` because all three touch code that only exists there (`classify_input`, `normalize_to_bgzf`, `open_raw_bam_record_reader_with_header`); the base should be retargeted to `main` once #644 merges.

**Suggested reading order: one commit at a time, in order.** They share a theme — the input boundary — but nothing else, and each stands alone.

### 1. `refactor(bgzf)`: give the BGZF header layout one definition

"Is this a BGZF block?" and "where does it end?" were answered by hand at four sites, and they had drifted. `read_raw_block` checked magic, CM, `FLG & FEXTRA` and `BC` but never `XLEN` or `SLEN`; `classify_input` checked all six itself; `unified_pipeline::fastq` read `BSIZE` from bytes 16-17 twice more with no validation at all. The input sniff and the block reader could therefore disagree about the same file — the thing the one-classifier work set out to prevent, one layer further down.

`fgumi_bgzf::header` is now the single answer to both questions, and the predicate is the *intersection* of what fgumi's decoders accept rather than what RFC 1952 permits: a `true` has to be a promise every downstream path can keep. noodles is the strictest and is not ours to change, so it sets the bar.

This closes a real gap. A `BC`-first header with a trailing subfield (`XLEN` 10) framed happily — `BSIZE` is still at 16-17 in that layout — and the four extra bytes rode into the payload every consumer takes as `[18 .. len - 8]`, dying downstream on `BGZF stored block size mismatch: LEN=89, payload=16`, a message about stored-block internals that says nothing about the header.

### 2. `fix(compare)`: open each input once for the single-pass presets

The engines took one open per input after #644; the command around them did not. `CompareBams::execute` must read both headers before dispatching — an incompatible pair hard-exits ahead of any engine, deliberately — and the engine then reopened both paths for their records. Each half read its stream once, but the command as a whole still needed a re-openable input, so `compare bams --command sort` and `--command group` rejected a FIFO or a process substitution that `fgumi sort` reads happily.

`execute` now opens each input once as an `OpenedInput` and hands the streams to whichever engine runs. The path-taking entry points remain as wrappers, so the public API is unchanged and neither form can drift into opening twice. Content mode still opens by path and releases the streams first — it makes two passes over each input and cannot stream a one-shot input at all.

### 3. `feat(io)`: accept plain-gzip input at the normalization boundary

gzip that was not BGZF was rejected outright, making fgumi the one tool in the pipeline that could not read what `gzip` produced. It is decompressed once at the boundary and re-classified now, so the pipeline still sees only BGZF: BGZF passes through, SAM text takes the existing transcode, and a raw unframed BAM is re-framed by copying its bytes through a BGZF writer (no record parsing — only the framing is missing).

That last arm is what makes this more than a convenience. A BGZF file carrying a second gzip extra subfield is legal per RFC 1952 and unframeable by fgumi, noodles and htslib alike, but it is still a plain gzip member — so a spec-complete gzip reader handles the framing they reject and what falls out is exactly an unframed BAM. Such a file now reads correctly; the test builds one by rewriting a real BAM's block headers.

Two decisions worth flagging for review:

- **Gzipped BAM is accepted, not rejected.** The plan recorded in #644 was to reject it because samtools cannot read it either — but "samtools cannot" does not imply "we cannot", and what comes out of the decompressor is a perfectly good BAM.
- **The boundary decompresses exactly once.** Peeling layers until something recognisable appears would let a small crafted input cost unbounded work for one open, so two gzip layers are named instead.

The cost is block-parallel decode, warned about per input rather than left as a silent cliff. `flate2` moves to `[workspace.dependencies]` so the two crates that need it share one declaration — its `zlib-rs` pin exists for a documented throughput reason, and a second unpinned copy is precisely the feature-unification hazard that comment warns about.

### Verification

`cargo ci-test` 6036 passed / 0 failed, `ci-fmt` and `ci-lint` clean at each commit. Every behavioural change has a test that fails before it and passes after — including the FIFO cases, which **hang** rather than fail when the property regresses, so those bound the work and fail with a named message instead.

## Summary by CodeRabbit

* **New Features**
  * Added support for plain gzip-compressed SAM and BAM inputs with automatic normalization into BGZF for downstream processing.
  * Improved reliability for BAM comparisons by opening inputs once (better support for streamed/FIFO inputs).

* **Bug Fixes**
  * Strengthened BGZF header validation and block-size detection to prevent misclassification of truncated/incorrect data.
  * Added clear rejection for double-compressed gzip inputs.

* **Tests**
  * Expanded integration coverage for gzip variants, stricter BGZF validation, and non-reopenable FIFO comparison inputs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -708,6 +708,7 @@ dependencies = [
  "bytes",
  "criterion",
  "crossbeam-channel",
+ "fgumi-bgzf",
  "fgumi-raw-bam",
  "log",
  "nix",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -710,6 +710,7 @@ dependencies = [
  "crossbeam-channel",
  "fgumi-bgzf",
  "fgumi-raw-bam",
+ "flate2",
  "log",
  "nix",
  "noodles",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,15 @@ repository = "https://github.com/fulcrumgenomics/fgumi"
 license = "MIT"
 
 [workspace.dependencies]
+# `zlib-rs` selects flate2's fast Rust-zlib DEFLATE backend. It must be requested
+# explicitly: the backend is otherwise chosen by Cargo feature unification, and a
+# transitive change can silently drop it back to the ~2x slower `miniz_oxide`
+# default — which regressed extract throughput ~1.6x at high thread counts. Pinned
+# here rather than per-crate so `fgumi` (gzipped FASTQ in `extract`) and
+# `fgumi-bam-io` (gzipped SAM/BAM at the input boundary) cannot end up on
+# different backends.
+flate2 = { version = "1.1", features = ["zlib-rs"] }
+
 # Internal crates. Versions are bumped in lockstep by release-plz.
 fgumi-bam-io = { version = "0.4.0", path = "crates/fgumi-bam-io" }
 fgumi-bgzf = { version = "0.4.0", path = "crates/fgumi-bgzf" }
@@ -121,13 +130,7 @@ approx = { workspace = true }
 ahash = { workspace = true }
 bytemuck = { workspace = true }
 rand_distr = "0.6"
-# `zlib-rs` selects flate2's fast Rust-zlib DEFLATE backend for gzipped FASTQ
-# decompression (used by `fgumi extract`). It must be requested explicitly: the
-# backend is otherwise chosen by Cargo feature unification, and a transitive
-# change can silently drop it back to the ~2x slower `miniz_oxide` default —
-# which regressed extract throughput ~1.6x at high thread counts. Keep this
-# feature pinned so the producer-thread decompressor stays on zlib-rs.
-flate2 = { version = "1.1", features = ["zlib-rs"] }
+flate2 = { workspace = true }
 dhat = { version = "0.3", optional = true }
 fgumi-raw-bam = { workspace = true, features = ["noodles"] }
 fgumi-dna = { workspace = true }
--- a/crates/fgumi-bam-io/Cargo.toml
+++ b/crates/fgumi-bam-io/Cargo.toml
@@ -15,6 +15,7 @@ pedantic = { level = "deny", priority = -1 }
 
 [dependencies]
 fgumi-raw-bam = { workspace = true, features = ["noodles"] }
+fgumi-bgzf = { workspace = true }
 noodles = { workspace = true, features = ["bam", "fasta", "sam", "bgzf", "core", "vcf"] }
 noodles-bgzf = { workspace = true }
 noodles-csi = { workspace = true }
--- a/crates/fgumi-bam-io/Cargo.toml
+++ b/crates/fgumi-bam-io/Cargo.toml
@@ -20,6 +20,7 @@ noodles = { workspace = true, features = ["bam", "fasta", "sam", "bgzf", "core",
 noodles-bgzf = { workspace = true }
 noodles-csi = { workspace = true }
 bgzf = { workspace = true }
+flate2 = { workspace = true }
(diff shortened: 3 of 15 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed the outside-diff-range finding on `src/lib/unified_pipeline/fastq.rs` (unvalidated `BSIZE` underflowing the ISIZE offset).

Confirmed it as a real crash first — a structurally valid BGZF header with `BSIZE = 0` panicked with `attempt to subtract with overflow` at the `offset + block_size - 4` computation, and would index out of bounds in release. `fgumi_bgzf::block_size` documents that it reads `BSIZE` without validating it and leaves the floor to the caller, which is exactly what `read_raw_block` does, so both call sites in `fastq.rs` now enforce the same `BGZF_HEADER_SIZE + BGZF_FOOTER_SIZE` floor: `estimate_uncompressed_size` stops summing (it is only a capacity estimate and has no error channel) and `decompress_bgzf_chunk` returns `InvalidData` rather than letting `decompress_block_slice_into` silently no-op on a too-short slice. Pinned by a new regression test that fails on the old code.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed the FIFO coverage finding, plus a pre-push self-review pass.

**The finding**: `test_compare_bams_cli_reads_a_non_reopenable_input` served only `bam1` over a FIFO, so a regression that reopened just `bam2` would still have passed — `execute` opens the two independently, so one-shot coverage on one argument proves nothing about the other. Both arguments are now separately fed FIFOs with independent feeders, joined after the success assertion (joining first would hang rather than fail, since a regression leaves a feeder blocked in its `open`). The same asymmetry was present in the two engine-level tests above it — `test_sort_verify_reads_a_non_reopenable_input` and `test_molecule_join_reads_a_non_reopenable_input` — so they take two FIFOs as well rather than a FIFO plus a re-openable oracle.

Three more from a self-review of the diff:

- **The surviving private opener.** The commit message for the compare change claims `OpenedInput::open` "subsumes the engines’ two private openers, which were copies of each other", but only `molecule_join`’s was actually deleted — `sort_verify.rs::open_raw_reader_with_header` remained as a byte-identical copy, and the comment in `sort_verify_compare` still pointed readers at it even though that path now opens through `OpenedInput::open`. Removed it, moved its rationale (the tee-and-replay mechanism, the SAM normalization) onto `OpenedInput::open`, and rewrote its one remaining caller, `verify_records_in_order`. The claim in the message is now true.
- **The empty-gzip-member arm was untested.** The gzip branch of `normalize_to_bgzf` has five sub-arms and four were covered; `InputFormat::Empty` was not. It is reachable — `gzip` of a zero-byte file decompresses to nothing and the boundary’s leading-bytes emptiness check has already run by then — and it works, but nothing pinned it. Added `empty_gzip_member_is_rejected`.
- **Two parity gaps.** `BgzfFramer::produce` clears its staging buffer without the `debug_assert!` that its `SamToBamStream::stage_encoded` sibling carries for exactly that step (the invariant holds; the guard did not), and `rejections_render` exercised only the `Flags` variant of `HeaderRejection`’s `Display`, leaving six arms unrendered — including `SubfieldId`, whose `as char` formatting is the one most likely to render badly. Both fixed; the latter is now an `rstest` table with one case per variant.

Full suite green (6867 passed), `cargo ci-fmt`/`ci-lint` clean, and rustdoc builds with `-D warnings`.

nh13:
@coderabbitai review

PR #662, opened 2026-07-25
fix(simulate): join writer threads on every exit path

nh13:
Started as "why does nextest intermittently flag `simulate::fastq_writer::tests::test_multi_threaded_writer` as `LEAK`". **That flag is still unexplained** — the test calls `finish()`, and it did not reproduce in three full suite runs or five isolated ones. What the investigation did turn up is a set of real detached-thread bugs, which this fixes. The flake is left open deliberately rather than declared fixed.

### `ParallelGzipWriter` joined its threads only on the happy path

`finish()` joined behind three `?`s: a failed final dispatch returned before any join, and a panicking worker returned before the remaining workers and the I/O thread were joined. The caller got its error while detached threads were still compressing and still writing the output file. Joining is now unconditional, with the first error winning so the cause is reported.

There was also **no `Drop`**, so any `?` between construction and `finish` detached the threads entirely. Not hypothetical — `simulate fastq-reads` has:

```rust
r1_writer.finish()?;
r2_writer.finish()?;
```

If R1's finish fails, `r2_writer` is dropped unfinished and its threads keep writing `r2.fastq.gz` while the process unwinds; what lands on disk depends on how fast it exits. A failing `write_record` drops both. The new `Drop` joins, which also brings this writer in line with `noodles::bgzf::MultithreadedWriter`, which has always finished in `Drop`.

### The same shape one level up, in three commands

`consensus_reads`, `correct_reads` and `fastq_reads` each did:

```rust
drop(sender);
if let Err(e) = generation_result { return Err(...); }   // handle never joined
let count = writer_handle.join()?;
```

So a generation failure returned while the writer thread was still writing its BAM or FASTQ. And it reported the wrong error: the send only fails because the receiver hung up, i.e. the writer already died — so `"Failed to send record to writer"` was the symptom while the writer's own error, the cause, was discarded along with the unjoined handle. All three now join first and prefer the writer's error. `mapped_reads` already did this correctly via `sink.fail`, and is unchanged.

Worth noting for review: the BAM **writer** itself was never at risk — `create_raw_bam_writer` builds on noodles' `MultithreadedWriter`, which joins in `Drop`. The bug in the BAM commands is at the command level, not in the writer.

### Test

Drops a writer without finishing and reads the file immediately: it can only be complete if the drop waited. Verified to fail with the `Drop` impl removed and pass with it.

`cargo ci-test` 6595 passed / 0 failed, `ci-fmt` and `ci-lint` clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved parallel simulation reliability by always completing writer thread work before returning results or errors.
  - Enhanced error propagation so writer failures/panics are surfaced before downstream send failures.
  - Ensured consistent read counters and messaging when generation/writing encounters issues.

- **Reliability**
  - Made compressed output finalization deterministic, preserving the first failure.
  - Ensured complete gzip output even when explicit completion isn’t called.

- **Tests**
  - Added regression coverage for writer failure/error reporting across fastq-reads and BAM-writing variants.

### DIFF
--- a/src/lib/commands/simulate/common.rs
+++ b/src/lib/commands/simulate/common.rs
@@ -5,7 +5,7 @@ use crate::commands::common::{
     resolve_memory_budget,
 };
 use crate::commands::sort::{TMP_DIRS_ENV, resolve_tmp_dirs};
-use anyhow::{Context, Result, bail};
+use anyhow::{Context, Result, anyhow, bail};
 use bytesize::ByteSize;
 use clap::Args;
 use crossbeam_channel::{Receiver, Sender, bounded};
@@ -21,6 +21,7 @@ use rand::{Rng, RngExt};
 use std::fs::File;
 use std::io::BufReader;
 use std::path::PathBuf;
+use std::thread::JoinHandle;
 
 /// Build the `@HD` header map carrying the `SO`/`GO`/`SS` sort-order tags for `order`.
 ///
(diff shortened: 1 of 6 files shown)


nh13:
Pushed two fixes for the failing checks, plus a verification worth recording.

**`docs`** — my own error: `finish` is public and its doc linked `[`Self::shutdown`]`, which is private, so `-D warnings` denied it via `rustdoc::private_intra_doc_links`. Now a plain code span. My local check had only denied *broken* links; re-verified with CI's exact `RUSTDOCFLAGS="-D warnings"`.

**`codecov/patch`** (76%) — the uncovered lines were the error paths this PR is about, so they were worth covering rather than waiving. Two tests added:

- `test_finish_reports_an_io_thread_failure` drives the I/O thread's error path through a sink that fails every write. Deterministic by construction: the payload is small enough that nothing is dispatched until shutdown, so the final dispatch succeeds and the I/O error is unambiguously the first error.
- `simulate_fastq_reads_reports_the_writer_error_not_the_send_failure` covers the command-level fix end to end, by making R1's path a directory so the writer fails at its first `File::create`. Verified to fail against the pre-fix code and pass after.

That second test also let me **verify the claim in the PR description rather than assert it**. Pre-fix, at both 200 and 300,000 molecules:

```
Error: Failed to send record to writer: sending on a disconnected channel
```

Post-fix:

```
Error: Failed to create .../r1.fastq.gz
    Is a directory (os error 21)
```

So the symptom really did replace the cause, and it does so regardless of scale — my first attempt to demonstrate it was misleading because I had measured an un-rebuilt binary.

`cargo ci-test` 6597 passed / 0 failed; `ci-fmt`, `ci-lint` and `cargo doc -D warnings` all clean.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #663, opened 2026-07-26
test(input-matrix): make the SAM/stdin comparisons non-vacuous

nh13:
Stacked on #644 — please merge that one first. The base is `nh/feat-accept-sam-input`, so the diff here is only this change.

## The problem

#644 added `test_input_source_matrix.rs`, which holds every command to two contracts: uncompressed SAM must be accepted anywhere BAM is, and `-i -` must be accepted by anything that streams. Both are checked by running the command two ways and comparing what it wrote against a named-path oracle.

That comparison is only evidence when the output depends on the records that went in. For five commands it did not:

- `simplex`, `duplex` and `codec` were handed ungrouped reads and wrote a header-only BAM from *any* input, including no input at all.
- `simplex-metrics` and `duplex-metrics` emitted identical all-zero tables either way, because the fixture carried nothing for them to count.

So for those five, a reader that silently dropped the entire stream still matched the oracle and the matrices passed. This was found by fault-injecting a header-only BAM while reviewing #644 — it predates that PR's stdin work and affected the SAM axis identically.

## What this does

**Fixtures they actually emit records from.** `duplex` and `duplex-metrics` get both strands of one molecule, tagged `RX` plus a strand-suffixed `MI`. `simplex` and `simplex-metrics` get single-strand molecules instead — `simplex-metrics` rejects a base UMI seen on both strands outright ("received duplex-UMI data … run duplex-metrics"), so it cannot share the duplex shape. `codec` reuses the CODEC command's own pair builder rather than a second copy of the MI/MC/overlap invariants, and needs the same `--min-reads`/`--min-duplex-length` its own tests pass. Record counts for the five go 0/0/0 and all-zero → 4, 2, 1, 28 and 29.

**An assertion instead of a caveat.** #644 documented this limitation in a comment listing the affected commands, which is exactly the kind of thing that goes stale. Every command now declares a third contract axis, `output_depends_on_input`, and `every_command_output_depends_on_its_input` drives each fixture through its command twice — once populated, once drained of records — and fails if the two agree. A future fixture that stops producing records for a command breaks there, at the cause, rather than quietly hollowing out the other two matrices. The comment is now a pointer to that test.

`zipper` is the one declared exemption: it emits one output template per uBAM template and the uBAM arrives on `-u`, so an empty `-i` legitimately still yields every template, unmapped. Refusing an empty input outright also counts as depending on it, which is how `extract` passes.

## Verification

Fault-injected in both directions rather than trusting a green run:

- Restoring the old ungrouped shapes fails the new test for exactly the five affected commands and no others.
- Declaring `zipper` `Required` fails it for `zipper` alone, so the exemption is load-bearing rather than an escape hatch.

`cargo ci-test` 6023 passed / 0 failed; `ci-fmt`, `ci-lint`, and both `--no-default-features --all-targets` and `--all-features --all-targets` checks clean. The cross-module reference to the CODEC pair builder follows the existing precedent in `test_streaming_input.rs` and was checked against the no-default-features build, where the self dev-dependency keeps those modules compiled.

## Summary by CodeRabbit

* **Tests**
  * Expanded integration coverage for duplex and single-strand molecule workflows.
  * Added validation to ensure command results respond appropriately to their input data.
  * Added empty-input scenarios across supported commands to detect misleading or unchanged outputs.
  * Improved coverage for codec options and input formats, including FASTQ, BAM, and paired-read fixtures.

### DIFF
--- a/tests/integration/helpers/bam_generator.rs
+++ b/tests/integration/helpers/bam_generator.rs
@@ -52,6 +52,152 @@ pub fn create_grouped_family(
     )
 }
 
+/// Creates one duplex-grouped molecule: `depth` read pairs on each strand.
+///
+/// The duplex consensus caller and both metrics commands key off an `MI` tag
+/// carrying a `/A` or `/B` strand suffix, and a molecule seen on only one strand
+/// yields no duplex output at all — so the minimum shape that makes those
+/// commands emit anything is both strands of one molecule. Reads also carry
+/// `RX`, which is what the `umi_counts` table is keyed on; with `MI` alone the
+/// metrics commands write an empty UMI table.
+///
+/// Orientation mirrors a real duplex pair: `/A` is FR (R1 forward at
+/// `ref_start`, R2 reverse 100bp downstream) and `/B` is RF. The two strands see
+/// the paired UMI in opposite order, so `RX` is swapped between them.
+///
+/// `test_duplex_command` has its own private builder of the same shape that
+/// tags `MI` only. This one lives here because several test modules need it and
+/// because it also tags `RX`; folding that one into this would add `RX` to the
+/// records those duplex tests assert on, which is a change to them, not to this.
+pub fn create_duplex_grouped_family(
+    mi: &str,
+    depth: usize,
+    sequence: &str,
+    quality: u8,
+    ref_start: i32,
+) -> Vec<RawRecord> {
+    let mut records = Vec::with_capacity(depth * 4);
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #666, opened 2026-07-27
fix(consensus)!: restore fgbio-identical near-tie calls by default

nh13:
Follow-up to #626. **Breaking change.**

## The defect

#626 replaced the inline unique-maximum test with `unique_max_index`, whose `TIE_TOLERANCE_ULPS` window treats a separation of a few ULPs as a tie and no-calls it.

On the merits that rule is better than fgbio's — fgbio's tie test sits in the `else` branch of `v > max`, so it only fires for a lane arriving *after* the maximum, and its tolerance is the absolute constant `f64::EPSILON`, which at family magnitudes sits orders of magnitude below one ULP and below unit scale spans hundreds.

Better, but not compatible — and compatibility is the contract. fgumi is a drop-in replacement for fgbio, so a silent divergence in called bases is a defect regardless of which answer is more defensible.

## Why it matters more than the docs suggest

The divergence is documented as confined to near-ties, implying real pileups rarely produce them. They are routine in shallow UMI families. A 2-2 split at equal quality leaves the two lanes **mathematically identical**, so they land exactly one ULP apart on accumulation noise alone.

Measured on `library:502` of `idt-cfdna` — four R2 reads disagreeing `C,C,T,T`, all Q37, at production error rates `-1 45 -2 40`:

```
likelihoods A,C,G,T = [-36.848038684599686, -18.424618431273785,
                       -36.848038684599686, -18.42461843127378]
top two lanes: T=-1.84246184312737817379e1  C=-1.84246184312737852906e1
bit-identical? false
ULP separation = 1
```

fgbio calls `T` at Q3 — a coin flip, p(error) ~ 0.5. fgumi emitted `N` at Q2, doubling the read's `cE` (0.00403 → 0.00806; 0.5 → 1.0 expected errors over 124 bases). Across the fgumi-benchmarks corpus that moved **170 records of idt-cfdna and 7,827 of qiaseq-umi**, breaking fgbio equivalency on **42 cells across 9 samples**. Neither the changelog nor the release notes flag an output change of that size.

## This PR

Adds `TieRule`, **defaulting to `FgbioCompat`**, plumbed to simplex/duplex/codec as `--tie-rule {fgbio-compat,ulp-relative}`. `FgbioCompat` promotes the port of `MathUtil.maxWithIndex` that already lived in `fgbio_tie_oracle_tests`, so both rules now run the code those tests characterise them by rather than a paraphrase. `--tie-rule ulp-relative` keeps the order- and scale-independent rule available.

## The order-dependence objection, addressed empirically

fgbio's rule is order-dependent in principle. It does not manifest here: the four lanes are always visited in `A,C,G,T` order, and a molecule's reads are always accumulated in one thread in file order. Re-running `simplex` over the same grouped BAM and comparing against fgbio 4.0.0:

| mode | vs fgbio |
|---|---|
| fast path (no `--threads`) | **IDENTICAL** |
| `--threads 1` | **IDENTICAL** |
| `--threads 4` | **IDENTICAL** |
| `--threads 8` | **IDENTICAL** |
| previous default (`ulp-relative`) | 170 content diffs — DIFFER |

Same answer at every thread count, so the theoretical nondeterminism does not appear in practice.

## Tests

Four quadrants pinned: the new default calls `T` at Q3; `ulp-relative` no-calls the same split; **both** rules still no-call a bit-exact tie (compat is not "always call"); both agree on a clearly-separated maximum.

Full suite green — 2,620 + 467 + 9 + 34 across the workspace, 0 failures. `clippy --all-features --all-targets -D warnings` clean, `fmt --check` clean.

## BREAKING CHANGE

Near-ties between the two most likely consensus bases now call a base, matching fgbio, instead of emitting `N`. Pass `--tie-rule ulp-relative` to keep the v0.5.0 behaviour.

## Summary by CodeRabbit

* **New Features**
  * Added configurable near-tie resolution for consensus base calling.
  * Added `--tie-rule` support with ULP-relative and fgbio-compatible modes.
  * Applied tie-rule settings consistently across simplex, duplex, and CODEC workflows.
  * Added APIs for configuring tie rules through consensus callers and builders.
  * The fgbio-compatible mode is now the default for consensus calculations.

* **Tests**
  * Added coverage validating both tie-breaking modes and configuration propagation.


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #669, opened 2026-07-28
docs: mark fgumi production ready as of v0.5.0

nh13:
v0.5.0 is the fgbio-parity release, so the README's research-preview banner — and the **July 31, 2026** date we were targeting to recommend fgumi over fgbio — are both out of date.

## README

Replaces the banner with a production recommendation. It still does two honest things the old banner did: it names where fgumi and fgbio intentionally differ (`MI` tag values, and `dedup --no-umi`, which follows Picard `MarkDuplicates`), and it tells users to validate against their own data before switching a production pipeline. It also asks for an issue on any divergence that *isn't* documented, which is the useful signal to collect now that people will start trusting the output.

The migration-guide link is absolute rather than repo-relative, matching every other link in the README — relative links break when crates.io renders it.

Also adds the migration guide to the Resources list. It wasn't there, which is odd on its own and worse now that the banner sends readers to it.

## Migration guide

The banner claims the `dedup --no-umi` behavior is documented in the migration guide. It wasn't — the guide covers the `MI` divergence but had nothing on dedup. Rather than drop the clause, this adds the section, since it's a breaking change in v0.5.0 that anyone upgrading needs:

- Why coordinate-only grouping is correct for duplicate marking (one fragment yields both an F1R2 and an F2R1 pair, so strand of origin separates the two halves of one molecule, not two molecules).
- The Picard `MarkDuplicates` 3.4.0 validation, and what the residual disagreement actually is (representative tie-break, symmetric).
- Why `group --no-umi` still splits by strand while `dedup --no-umi` doesn't — consensus calling is a positional pileup and needs the split; dedup calls no consensus.
- A "Changed in v0.5.0" note with the direction of the shift and the fact that there's no flag to restore the old behavior.

Facts and figures are from #649.

## Verification

`cargo docs-build` exits 0 with mdbook-linkcheck2 clean; the three `Potential incomplete link` warnings are pre-existing, in generated tool/metric pages, and untouched by this change.

## Note for reviewers

The v0.5.0 release notes now carry a user-facing preamble above the generated changelog, which makes the same production-readiness framing. Worth a skim for consistency with the wording here: https://github.com/fulcrumgenomics/fgumi/releases/tag/v0.5.0

## Summary by CodeRabbit

* **Documentation**
  * Updated project status to production-ready as of v0.5.0.
  * Added links and guidance for migrating from fgbio.
  * Documented differences in no-UMI deduplication behavior, including coordinate-based grouping and changes introduced in v0.5.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,14 +9,21 @@
 
 # fgumi
 
-**⚠️ RESEARCH PREVIEW - USE AT YOUR OWN RISK**
+**Production ready as of v0.5.0 — the fgbio-parity release.**
 
-This software is currently a **research preview**. While we have extensively
-tested these tools across a wide variety of vendor-provided data, **no
-guarantees are made** regarding correctness or stability.
+fgumi's output is validated against
+[fgbio](https://github.com/fulcrumgenomics/fgbio) across a wide range of
+vendor-provided data, and we now recommend fgumi over fgbio for UMI
+processing. The two intentionally differ in a handful of places — most visibly
+`MI` tag values, and `dedup --no-umi`, which drops fgbio's strand-of-origin
+split to match Picard `MarkDuplicates`. The divergences we know of are
+catalogued in the
+[migration guide](https://github.com/fulcrumgenomics/fgumi/blob/main/docs/src/guide/migration-from-fgbio.md).
 
-We are targeting **July 31, 2026** to recommend fgumi over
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai pause

PR #672, opened 2026-07-30
feat(fmt,cli-common): add fgumi-fmt and fgumi-cli-common leaf crates

nh13:
First in a crate-by-crate series landing the `runall` / unified-pipeline work onto `main-runall`. Each PR targets `main-runall`, which is kept rebased on `main`; `main` itself is untouched until the whole series lands.

## What

Adds the two dependency-free leaf crates, wired into the workspace as members:

- **`fgumi-fmt`** (102 lines, 2 public items) — human-readable count and duration formatting. No workspace dependencies at all.
- **`fgumi-cli-common`** (817 lines, 21 public items) — shared CLI types and helpers. Depends only on `fgumi-fmt`.

Nothing calls into either crate yet. This is purely additive: no existing file changes behaviour, and both crates are unreachable from any command until later PRs wire them up.

## Why these two first

They are the only two of the six incoming crates that are already clean under `cargo ci-lint` (`-D warnings -W clippy::pedantic`). The remaining four need a let-chain lint sweep first — the source branch targets MSRV 1.87, this workspace is on 1.93 with let-chains adopted, so `clippy::collapsible_if` fires on code written against the older toolchain. Per the repo's guidance those get rewritten as let-chains rather than suppressed, and keeping that work out of this PR keeps both halves reviewable.

Starting small is also deliberate: this PR is really about agreeing the conventions — workspace wiring, dependency handling, lockfile discipline — that the rest of the series will repeat.

## Dependency centralization

Both crates arrived with per-crate version pins. I converted them to `{ workspace = true }` to match this workspace's existing centralization, which required promoting three dependencies into `[workspace.dependencies]`:

| dep | reason |
|---|---|
| `enum_dispatch` | now shared by two members |
| `num_cpus` | now shared by two members |
| `sysinfo` | shared by two members **and** MSRV-constrained |

`sysinfo` is the one that matters: it is held at 0.38 because >= 0.39 requires rustc 1.95, above our pinned 1.93. A per-crate pin could drift onto 0.39 and silently raise the effective MSRV — precisely the failure the centralization rule exists to prevent. Its explanatory comment moved to the workspace entry.

## Verification

| gate | result |
|---|---|
| `cargo ci-fmt` | pass |
| `cargo ci-lint` | pass |
| `cargo ci-test` | pass — 6936 run, 6936 passed, 27 skipped |
| `cargo ci-doctest` | pass |
| `cargo ci-doc` with `RUSTDOCFLAGS="-D warnings"` | pass |
| `cargo nextest run -p fgumi-fmt -p fgumi-cli-common` | 46 run, 46 passed |

`Cargo.lock` is regenerated in the same commit — three CI aliases run `--locked`, so a stale lockfile would fail on this commit specifically.

A test-inventory check (name-keyed, so module moves are not miscounted as loss) confirms no test present on `main` is absent here: 5964 before, 6005 after, 0 lost. Raw `#[case]`, `#[values]` and `#[ignore]` counts are also non-decreasing, since those are invisible to a name-keyed comparison.

## For review

1. Is promoting `enum_dispatch` / `num_cpus` / `sysinfo` into `[workspace.dependencies]` the right call to make here, or would you rather this PR stay minimal and centralization land separately?
2. Both crates are dead code until a later PR consumes them. Is that acceptable for the series, or would you prefer each crate arrive alongside its first consumer?
3. Branch naming: I used `nh/runall-NN-<topic>` so the series is orderable. Say if you want the `<issue>/<user>/<type>-<desc>` form instead.

---

**Update (force-pushed):** the first push failed `codecov/patch` at **74.9 %** against the repo's 90 % gate, with all 13 other checks green.

Cause: `fgumi-cli-common` arrived from the source branch at 73.1 % line coverage on the changed lines. Its own tests cover the parsing happy paths, but the `OperationTimer`, `validate_file_exists`, several `parse_memory_size` error arms, `parse_memory` / `parse_memory_reserve`, the binary-unit `Display` impls, and two branches of the memory-budget resolver arrived untested. `fgumi-fmt` was already at 100 %.

Rather than pad for the metric, I added tests for each of those paths — as `#[rstest]` case tables with descriptive labels, matching the repo convention. `fgumi-cli-common` goes from 33 to 63 tests; patch coverage is now **95.9 %**.

The 15 lines still uncovered are all `log::info!` / `warn!` / `debug!` macro arguments, which are never evaluated because the test binary initializes no logger, plus one defensively-unreachable `checked_mul` arm in `parse_memory_size` (its input is already bounded to ≤ 1,000,000 a few lines earlier). Both seemed like the wrong things to chase.

Amended into the original commit rather than stacked as a fixup, so the PR stays a single logical change. `tempfile` is added as a dev-dependency for the `validate_file_exists` tests.

## Summary by CodeRabbit

* **New Features**
  * Added consistent formatting for counts, durations, and processing rates.
  * Improved command-line handling for memory limits, reservations, compression settings, and boolean values.
  * Added clearer validation and error reporting for files and memory sizes.
  * Added system-aware CPU and memory detection for resource budgeting.
  * Added operation progress and completion reporting.

* **Bug Fixes**
  * Compression now defaults consistently to level 1.
  * Large memory-size suggestions now use accurate binary GiB units.
  * Improved resource-budget handling on platforms with smaller integer limits.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -135,6 +135,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "arrayvec"
+version = "0.7.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d3fb67a6e08acf24fdeccbac2cb6ac4305825bd1f117462e0e6f2f193345ad56"
+
+[[package]]
 name = "autocfg"
 version = "1.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -720,7 +726,6 @@ dependencies = [
  "nix",
  "noodles",
  "noodles-bgzf",
- "num_cpus",
  "parking_lot",
  "proptest",
  "rand 0.10.2",
@@ -770,6 +775,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "fgumi-cli-common"
+version = "0.5.0"
+dependencies = [
+ "anyhow",
+ "bytesize",
+ "clap",
+ "fgumi-fmt",
+ "log",
+ "rstest",
+ "sysinfo",
+ "tempfile",
+ "thiserror 2.0.19",
+]
+
+[[package]]
 name = "fgumi-consensus"
 version = "0.5.0"
 dependencies = [
@@ -796,6 +816,14 @@ dependencies = [
 name = "fgumi-dna"
 version = "0.5.0"
 dependencies = [
+ "rstest",
+]
+
+[[package]]
+name = "fgumi-fmt"
+version = "0.5.0"
+dependencies = [
+ "num-format",
  "rstest",
 ]
 
@@ -1087,12 +1115,6 @@ name = "heck"
 version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
-
-[[package]]
-name = "hermit-abi"
-version = "0.5.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fc0fef456e4baa96da950455cd02c081ca953b141298e41db3fc7e36b1da849c"
 
 [[package]]
 name = "iana-time-zone"
@@ -1603,6 +1625,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "num-format"
+version = "0.4.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a652d9771a63711fd3c3deb670acfbe5c30a4072e664d7a3bf5a9e1056ac72c3"
+dependencies = [
+ "arrayvec",
+ "itoa 1.0.18",
+]
+
+[[package]]
 name = "num-integer"
 version = "0.1.46"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
Both findings fixed in the amended commit.

**`enum_dispatch` cross-crate limitation** — confirmed, and it is broader than the thread describes: `#[enum_dispatch]` sits on the `Command` trait in *both* `fgumi-cli-common` and `src/lib/commands/command.rs`, but **no enum applies `#[enum_dispatch(Command)]` anywhere in the repo** — `src/main.rs:148` hand-writes the dispatch `match`. So the attribute is already inert today, not merely unable to work after migration.

Removed it from `fgumi-cli-common` along with the `enum_dispatch` dependency, and documented on the trait why it is absent. Since `fgumi-cli-common` was the second member that justified promoting `enum_dispatch` into `[workspace.dependencies]`, that promotion is reverted too — only the umbrella crate uses it now, so it belongs in the root `[dependencies]` per the "shared by two or more members" rule stated on that section.

Left the umbrella's own `#[enum_dispatch]` in place: it is pre-existing, equally inert (the trait is in the lib crate, the enum would be in the bin crate), and removing it touches live dispatch code outside this PR's scope. Worth its own change.

**Per-thread debug log** — fixed as suggested, and applied to the sibling too: `src/lib/commands/common.rs` carried the identical unconditional log in the umbrella's copy of `resolve_memory_budget`. Both now branch on `per_thread`, matching the framing `log_memory_config` already used a few lines below.

Verification: `ci-fmt`, `ci-lint`, `ci-test` (6,984 pass), `ci-doctest`, `ci-doc` all green; patch coverage 95.0% against the 90% gate.

PR #675, opened 2026-07-31
ci: retarget CodeRabbit config on measured review evidence

nh13:
Rewrites `.coderabbit.yaml` against evidence rather than intuition. The current file was written in #392 and touched once since (#453); in the meantime the code moved and the reviewer's behavior changed, and neither was reflected back into the config.

## How this was derived

Four independent passes, all reproducible:

1. **The actual CodeRabbit output on this repo.** ~100 recent PRs (#569–#674) pulled via the GitHub API: 780 bot comments, 152 distinct actionable findings, classified by severity, path, and whether a human accepted or declined them. This window is entirely after #392, so it measures the *current* config.
2. **Session history** across 380 transcripts, recovering 437 CodeRabbit artifacts spanning PR #47–#674, including the recorded dispositions where a finding was declined and why.
3. **A schema audit** against `schema.v2.json` (293 keys, 57 tools) plus the live docs, to find valid options the file was not using and to confirm the glob semantics.
4. **A reality check** of every glob and every symbol the file names, against `git ls-files` and `git grep` on `main`, with globs evaluated using minimatch — the matcher CodeRabbit actually uses.

## What was wrong

**The pipeline instruction had rotted into fiction.** It told the reviewer to require that "the must-accept `next_serial` exemption is preserved" and that errors "propagate via the shared signal so `is_done()` is observed". `is_done()` has **zero hits anywhere in this repo**. Neither do "must-accept", "transport queue", "overflow stash", "typed-step dispatch", "Parallel step", or "drain-counter". `src/lib/unified_pipeline/` also contains **no `unsafe` at all**, so the block's closing clause about new `unsafe` in the dispatch hot path was policing something that has never existed. Roughly half that instruction's concrete vocabulary did not resolve. It is replaced with what is actually there — `TransportQueue::limit_bytes`, `is_at_limit` / `is_below_drain_threshold`, and the shared `shutdown_signal: Arc<AtomicBool>`.

The other instructions held up: sort, raw-bam, and consensus named 12 of 12 constructs that still resolve. `src/lib/pipeline/**` matches nothing (the #330 rename has not landed) and stays only as the documented hedge it was.

**Coverage was 51%.** 129 of 251 `.rs` files were reached by a rule. `src/lib/commands/` — 41 files, 68k LOC, the largest module in the repo and 27% of the reviewer's file-read budget — had no instruction at all, and returned almost entirely Minor and Trivial findings as a result. It was explicitly deferred as out-of-scope in #392 and never revisited.

**The most valuable rule reached a minority of the tests it was written for.** `**/tests/**/*.rs` carries the programmatic-data / independent-oracle / assertion-strength guidance, and matches 46 files — but **168** files carry inline `#[cfg(test)] mod tests`, which per CLAUDE.md is the dominant style here.

**`tone_instructions` was spending its budget on a threat that never materialized.** It asks the reviewer to skip clippy and rustfmt nits. Across 100 PRs, clippy and rustfmt produced **zero** findings; the only linter that has ever fired on this repo is zizmor. That clause is replaced with the two asks that address the repo's actual highest-volume declines.

## What changed

**Coverage 129/251 → 219/251 by specific rule**, with a repo-wide rule reaching the rest. New instructions for `src/lib/commands`, `fgumi-bam-io`/`fgumi-bgzf`, `fgumi-sam`, `fgumi-simd-fastq`, the grouping modules, and the FASTQ/validation parsers.

`fgumi-simd-fastq` is worth calling out: 17 `unsafe` NEON/AVX2 intrinsic sites, the **only workspace member without `#![deny(unsafe_code)]` at its crate root**, absent from CLAUDE.md's unsafe allowlist, not covered by the Miri job (which runs only `-p fgumi-raw-bam sort`), and parsing untrusted input. That combination is worth a note in CLAUDE.md independently of this PR.

**Noise cut at the source rather than asked away.** Test-restyling — "convert this to `#[rstest]`" — is the single largest comment class, roughly 30% of bot threads in one inventory, and it originates from guideline ingestion of the `rstest`/`proptest` line in CLAUDE.md, not from any rule in this file. The repo-wide instruction now states that test *shape* is not a finding and only missing *coverage* is.

**Recurring declined findings are pre-answered with the invariant behind them**, so the same argument is not relitigated every few PRs:

- Unclipped coordinates are template-coordinate **sort-key lanes, not genomic positions**. Clamping one side desynchronizes a template's two mates and diverges from fgbio and `samtools sort --template-coordinate`. This was raised at least twice, the second time against an input that `saturated_clip_does_not_wrap` already pins.
- The **sequential and parallel paired assigners take different key formats** — prefixed strings versus clean UMIs. A fix proposed on the wrong side of that boundary was once applied and broke four passing tests.
- The **oracle policy** is now written down: fgbio is a JVM tool and deliberately not a CI dependency, fgbio values cited by `file:line` are acceptable, and `fgumi sort --verify` is an acceptable order oracle.
- The consensus parity ask is **scoped to output-changing diffs**. As written it fired on provably byte-identical changes, and in one case asked to replace a bitwise assertion with an approximate fgbio comparison — a downgrade.

**New capability the file was not using:** a risk-focused summary instruction, per-command label suggestions using the labels this repo already has, a conventional-commit title check (CLAUDE.md mandates the format and nothing verified it), and `github-checks.timeout_ms` raised from the 90-second default to the 15-minute maximum so reviews stop landing before the CI they depend on.

**Turned off:** the clippy and osvScanner integrations, which can only restate a check that already failed CI (`cargo ci-lint` runs clippy `-D warnings -W clippy::pedantic`; `cargo audit` runs on every PR); docstring and unit-test generation offers; and dependency/release PRs, which have not produced an actionable finding. zizmor stays **on** — nothing in CI lints workflows and its findings here have been applied essentially without exception.

**Filtered:** git-cliff-generated CHANGELOGs (12 of them, which trip markdownlint MD024 permanently and unfixably since there is no markdownlint config in the repo), `THIRDPARTY.toml` (2.2 MB, regenerated on every dependency bump), benches, and proptest regressions. Note that `path_filters` also drive CodeRabbit's `git sparse-checkout`, so an excluded path is absent from its working tree entirely — `Cargo.toml` is deliberately *not* excluded despite low yield, because removing the workspace manifests would leave the tree unresolvable.

## Second commit: making the failure loud

Documenting the 250-character cap in a comment is what the previous revision effectively relied on, and it is not enforcement. `scripts/check-coderabbit-config.py` plus a new `coderabbit-config` job in `check.yml` now fail the build on the four ways this file breaks quietly:

- the YAML does not parse;
- a documented cap is exceeded, measured **after folding**, which is the length that actually counts;
- a `path_instructions` glob matches no tracked file — the #392 rot above, caught mechanically instead of by hand;
- YAML folding split a backtick-quoted identifier across lines, turning `foo_bar` into `foo_ bar` in the text the reviewer is handed. This is not hypothetical: it happened while writing this PR, to `saturated_clip_does_not_wrap`, and was caught by the check.

Globs are matched with minimatch semantics (brace expansion, `**` crossing separators) to mirror CodeRabbit. `path_filters` is deliberately exempt from the dead-pattern check, since entries like `!**/target/**` are *meant* to match nothing tracked.

Each of the four failure modes was verified to fail the check, and the passing case to pass, before it was wired into CI.

## Third commit: stop auto-reviews from spending the shared budget

The earlier revisions of this PR treated the review limit as something to *economize* against — filter paths, skip mechanical PRs, cut noise. Measurement on 2026-08-01 shows the dominant term is not what each review costs but **how many reviews nobody asked for**.

CodeRabbit's fair-use limit is a per-developer allowance shared across the entire org, and it counts **requests, not delivered reviews** — so a rejected auto-review costs exactly as much as a useful one. Both `auto_review.enabled` (fires on PR open) and `auto_incremental_review` (fires on every push) drew on that allowance unprompted.

What that produced, measured across all of fulcrumgenomics that day:

| repo | requests | "Review limit reached" | auto-review |
|---|---|---|---|
| **fgumi** | 23 | **17 (74%)** | **enabled** |
| ferro-hgvs | 31 | 6 (19%) | disabled |

ferro-hgvs absorbed *more* requests and was rejected a quarter as often, because it already disables both settings. Same org, same allowance, opposite config.

The mechanism is worse than the ratio suggests, because a rejection is not local. Four cooldowns in a single evening — 21:14, 22:30, 23:03 and 00:04 UTC — were armed by PR-open auto-reviews on PRs 700, 702, 703 and 704. Each arms a **~58-minute org-wide cooldown**, and each overlapped the previous one, so the org sat in near-continuous cooldown and deliberate, explicitly queued reviews on every other repo were starved. This PR was itself a casualty: it has been mergeable since Jul 31 and went unreviewed through three separate review requests that day.

Pausing after the fact cannot close the gap — the PR-open review fires three to five minutes before any tooling can post `@coderabbitai pause`. It has to be off in config.

So `enabled` and `auto_incremental_review` are both `false`. On-demand `@coderabbitai review` still works, and the assertive profile and every path instruction above still apply to those reviews — this changes *when* reviews happen, not what they look like. `base_branches` is dropped with them: it only widens which base branches are auto-reviewed, which is moot once auto-review is off, and on-demand reviews reach stacked PRs regardless of base.

`ignore_title_keywords` (the deps/release skip described under **Turned off** above) is likewise inert now, since it gates auto-reviews only. It is kept deliberately, so the intent survives if auto-review is ever re-enabled; the tooling that drives our on-demand reviews filters the same titles today.

## Verification

- Validates against `schema.v2.json` with ajv (draft 2020-12): **valid**.
- `python3 scripts/check-coderabbit-config.py` passes: 12 path instructions, all globs match tracked files.
- `tone_instructions` is **249/250** characters. This cap is load-bearing and is now called out in the file: exceeding it does not truncate the field, it fails validation and **silently discards the entire config**, falling back to org defaults. That already happened once here — the original 293-character value, fixed mid-#392.
- Every instruction glob was evaluated with minimatch against `git ls-files`: **zero dead globs**. Every symbol and test name cited was confirmed with `git grep` to resolve on `main`.
- Crates that exist only on unmerged branches (`fgumi-pipeline-core`/`-io`, `fgumi-fmt`, `fgumi-cli-common`, `fgumi-cli-macros`) are deliberately **not** given globs, and are listed in the maintenance comment to be added in the PR that lands them. Adding a glob that matches nothing is the exact failure this PR is correcting.

## Reviewer notes

Config-only; no code changes. The effect is repo-wide and will be visible on the next PR review.

Note that after the third commit, opening a PR here no longer produces a review on its own — reviews run when requested (`@coderabbitai review`, or automatically via the review queue). That is the intended change, not a regression.

Two judgment calls worth a second opinion:

1. **`auto_review.enabled` stays `true`.** This repo hits CodeRabbit's fair-use review limit regularly — 22 of 100 sampled PRs, with waits up to 41 minutes, driven by ~450 manual `@coderabbitai review` invocations. `ferro-hgvs` and `coderabbit-queue` respond by disabling auto-review entirely and going request-driven. I did not go that far here, since #453 deliberately turned stacked-PR reviews back *on*; skipping dependency and release PRs is the conservative first step. Say the word if you'd rather go request-driven.
2. **`enable_prompt_for_ai_agents` stays on.** Those blocks are ~14% of all bot output by volume, but they are what the `fix-pr-review` flow consumes, so removing them would cost more than it saves.

Risk: command output changes none; `unsafe` changes none and the `CLAUDE.md` allowlist is unchanged; memory bounds, queue capacity, and thread/backpressure policy changes none. Fix: retain the evidence-based `.coderabbit.yaml` guidance and validation settings.

- Replaces stale repository references with existing constructs.
- Expands Rust review coverage from 129 to 219 of 251 files.
- Adds guidance for commands, I/O, SAM, SIMD FASTQ, grouping, parsing, validation, malformed input, unsafe code, output parity, concurrency, and memory bounds.
- Adds repository-wide test guidance, invariant documentation, oracle policies, and scoped consensus checks.
- Excludes generated artifacts, benches, and proptest regressions.
- Disables redundant analyzers and generation features.
- Adds risk summaries, label suggestions, conventional-commit validation, and a 15-minute CI timeout.
- Retains auto-review and AI-agent prompts.
- Validates the configuration schema, instruction globs, and cited symbols.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -69,19 +69,40 @@ reviews:
   suggested_reviewers: false
 
   auto_review:
-    # Review automatically on push; don't review draft PRs. `base_branches: .*`
-    # also auto-reviews PRs whose base is not the default branch (stacked PRs);
-    # without it CodeRabbit posts "auto reviews are disabled on base/target
-    # branches other than the default branch" and skips them.
-    enabled: true
+    # Reviews run on demand only. CodeRabbit's fair-use limit is a per-developer
+    # allowance shared across the whole org, and it counts REQUESTS, not
+    # delivered reviews — so a rejected auto-review costs exactly as much as a
+    # useful one. Auto-review on PR-open (`enabled`) and on every push
+    # (`auto_incremental_review`) both draw on that shared allowance without
+    # anyone asking for them.
+    #
+    # Measured across fulcrumgenomics on 2026-08-01: this repo produced 17 of
+    # the org's 23 "Review limit reached" rejections. Four cooldowns in one
+    # evening (21:14, 22:30, 23:03, 00:04 UTC) were armed by PR-open
+    # auto-reviews on #700/#702/#703/#704 — each arming a ~58-minute ORG-WIDE
+    # cooldown, and each overlapping the last, so the org sat in near-continuous
+    # cooldown and deliberate queued reviews on other repos were starved.
+    # ferro-hgvs, which already disables both settings, absorbed 31 requests the
+    # same day for 6 rejections (19%, vs 74% here). Pausing after the fact
+    # cannot fix this: the PR-open review fires 3-5 minutes before any tooling
+    # can post `@coderabbitai pause`.
+    #
+    # `@coderabbitai review` still works with `enabled: false`, and the
+    # assertive profile and path instructions still apply to those reviews.
+    # `base_branches` is dropped with them: it only widens which base branches
+    # are AUTO-reviewed (stacked PRs), which is moot once auto-review is off —
+    # on-demand reviews reach stacked PRs regardless of base.
+    enabled: false
+    auto_incremental_review: false
     drafts: false
-    base_branches:
-      - ".*"
     # Dependency and release PRs are mechanical: they have not produced an
     # actionable finding here, and `cargo audit` plus the `--locked`
     # feature-combination checks in CI already gate them. Skipping them
     # preserves review quota for PRs that carry logic — this repo runs into
-    # CodeRabbit's fair-use review limit regularly.
+    # CodeRabbit's fair-use review limit regularly. Inert while auto-review is
+    # disabled above (it gates auto-reviews only); kept so the intent survives
+    # if auto-review is ever re-enabled. The tooling that drives our on-demand
+    # reviews filters the same titles today.
     ignore_title_keywords:
       - "chore(deps"
       - "chore: release"
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -9,119 +9,478 @@
 # hot paths under an otherwise crate-wide `#![deny(unsafe_code)]`. The
 # `assertive` profile + the path instructions below aim the review at the failure
 # modes that actually bite here (output divergence vs fgbio, pipeline
-# deadlock/unbounded memory, FFI/`unsafe` unsoundness, doc/contract drift) rather
-# than generic style.
+# deadlock/unbounded memory, malformed-input panics, `unsafe` unsoundness,
+# doc/contract drift) rather than generic style.
+#
+# WHAT CI ALREADY ENFORCES — never spend a review comment re-deriving any of it.
+# `.github/workflows/check.yml` runs, on every PR: `cargo ci-lint` (clippy
+# `-D warnings -W clippy::pedantic`, all targets, all features), `cargo ci-fmt`,
+# `cargo ci-doc` with `RUSTDOCFLAGS=-D warnings` (broken intra-doc links fail),
+# `cargo ci-test` + doctests via nextest, llvm-cov with a 90% patch-coverage
+# gate, `cargo audit`, an MSRV-lockstep check, four feature-combination `cargo
+# check`s, a publish dry-run, and a samtools-backed sort-correctness suite.
+# What CI does NOT cover — and where review earns its keep — is fgbio output
+# parity (no CI job runs fgbio at all), `unsafe` soundness outside
+# `fgumi-raw-bam::sort` (the only Miri target), liveness and memory-bound
+# reasoning (stress tests are nightly-only, off the PR path), and whether a
+# test's oracle is genuinely independent.
 
 language: en-US
 
 # Terse, technical, no praise or filler — match the codebase's review style.
-# NB: CodeRabbit caps tone_instructions at 250 characters — keep this terse.
+#
+# !! HARD LIMIT: CodeRabbit caps tone_instructions at 250 characters. Exceeding
+# !! it does not truncate the field — it fails schema validation and silently
+# !! discards THIS ENTIRE FILE, falling back to org-UI defaults. That has already
+# !! happened once here (the original 293-char value, fixed in #392). Re-count
+# !! after any edit: the folded value below is 249 characters.
+#
+# The two clauses chosen are the repo's two highest-volume sources of declined
+# findings; both are restated at length in the path instructions below.
 tone_instructions: >-
-  Concise and technical. Skip praise and restating the code. Lead with the risk
-  and a one-line fix. Favor correctness, output identity vs fgbio/samtools,
-  concurrency soundness, and memory bounds over style. Skip clippy/rustfmt nits;
-  CI enforces them.
+  Concise, technical. Lead with the risk and a one-line fix; skip praise and
+  restating code. Never ask to reshape a passing test into rstest/proptest. Do
+  not re-demand an fgbio oracle when a test cites fgbio by file:line or uses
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #676, opened 2026-07-31
fix(simd-fastq): bring the SIMD lexer under the crate-wide unsafe policy

nh13:
CLAUDE.md states that `#![deny(unsafe_code)]` is set at the crate root of every workspace member. It was not. `fgumi-simd-fastq` was the sole exception, so its `unsafe` sites were the only ones in the workspace that could be added or changed without tripping the lint or showing up in the allowlist.

Found while auditing `.coderabbit.yaml` (#675), which needed an accurate picture of where `unsafe` lives.

## Why this crate in particular

The exposure was the sharpest in the repo rather than the mildest:

- `lexer.rs` carries **17 `unsafe` occurrences** across hand-written NEON and AVX2 intrinsics with runtime feature dispatch.
- It parses **untrusted input** — FASTQ record boundaries straight off a byte stream.
- **Miri cannot execute intrinsics**, and `miri.yml` runs only `-p fgumi-raw-bam sort`, so nothing in CI checks this crate for unsoundness.
- It appeared **nowhere in CLAUDE.md**, so anyone auditing the unsafe surface by reading the allowlist would not have found it.
- The crate had **no `// SAFETY:` comments at all** — not one, anywhere.

## What changed

Attributes, comments, and docs only. No behavior change, no logic touched.

- `#![deny(unsafe_code)]` at the crate root, matching every sibling.
- Seven targeted `#[allow(unsafe_code)]`: the two public dispatch functions (`lex_block`, `lex_block_full`), the three NEON helpers (`neon_movemask`, `lex_block_newlines_neon`, `lex_block_neon`), and the two AVX2 implementations (`lex_block_newlines_avx2`, `lex_block_avx2`). The nested `neon_two_bits` / `avx2_two_bits` helpers are covered by the lexically-enclosing allow rather than repeating it.
- A `// SAFETY:` comment per site. Calling a `#[target_feature]` function is `unsafe` because the caller must guarantee CPU support; the intrinsics are then safe within it. So each comment discharges exactly two obligations:
  - **Feature availability** — NEON is architecturally mandatory on aarch64, so that precondition is unconditional; every AVX2 call is guarded by `is_x86_feature_detected!("avx2")` with a scalar `else` branch.
  - **Load bounds** — the only memory accesses are vector loads from `block.as_ptr()` where `block: &[u8; 64]`. NEON issues four 16-byte `vld1q_u8` loads at offsets 0/16/32/48; AVX2 issues two 32-byte `_mm256_loadu_si256` loads at 0 and 32. Both cover exactly 64 bytes and no more, and both use unaligned load forms, so no alignment beyond `u8` is required.
- A CLAUDE.md allowlist section for the crate, recording the approval rationale, the absence of Miri coverage, and the four scalar-parity tests that stand in for it.

With this change CLAUDE.md's claim is true rather than aspirational — all **14** workspace crate roots were checked individually.

## Verification

- `cargo ci-fmt`, `cargo ci-lint` (clippy `-D warnings -W clippy::pedantic`): clean.
- `cargo ci-test`: **6890 passed**, 27 skipped.
- `cargo nextest run -p fgumi-simd-fastq`: 50 passed.
- **Both architectures compiled, not assumed.** Four of the seven allow sites and one SAFETY note are on AVX2 code that is `cfg`'d out on the aarch64 dev machine, so `cargo check -p fgumi-simd-fastq --target x86_64-apple-darwin` was run to compile that half. Both pass.
- **The deny was negative-tested**: removing a single `#[allow(unsafe_code)]` fails the build with `error: usage of an unsafe block`, confirming the lint is live rather than decorative.

## Note on what this does and does not buy

This closes a *policy* gap: new `unsafe` in this crate now has to be deliberate and has to be justified in CLAUDE.md, the same as everywhere else. It does not by itself prove the existing intrinsics sound — the bounds arguments above were derived by reading the loads, and Miri still cannot check them. If we want that, the options are a fuzz target over `lex_block_full` or asserting SIMD/scalar agreement under `proptest` rather than the fixed inputs used today. Happy to file that separately if it's worth doing.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for reviewing and maintaining SIMD implementations, including CPU feature checks, memory safety, and scalar-parity testing.

* **Refactor**
  * Clarified safety guarantees around architecture-specific SIMD processing.
  * Restricted unsafe code to documented SIMD paths while preserving existing behavior and public interfaces.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -210,6 +210,45 @@ regresses `samtools sort -n`–style throughput.
     SAFETY: the buffers are `to_vec()` + push, so the pointer is valid and
     null-terminated for the call's lifetime.
 
+### Approved SIMD intrinsics (fgumi-simd-fastq)
+
+`crates/fgumi-simd-fastq/src/lexer.rs` classifies 64-byte FASTQ blocks through
+hand-written NEON (aarch64) and AVX2 (`x86_64`) intrinsics, with a scalar
+fallback. All `unsafe` in the crate is confined to this one file. It is approved
+because the lexer runs once per 64 bytes of every FASTQ processed (billions of
+blocks on production inputs) and the scalar fallback — retained and exercised in
+tests as the reference implementation — is materially slower.
+
+Calling a `#[target_feature]` function is `unsafe` because the caller must
+guarantee the CPU supports that feature; the intrinsics themselves are then safe
+within it. Two obligations therefore apply at every site, and both are recorded
+in `// SAFETY:` comments:
+
+- **Feature availability.** `lex_block` / `lex_block_full` dispatch on the
+  architecture. On aarch64 NEON is architecturally mandatory, so the precondition
+  is unconditional. On `x86_64` every AVX2 call is guarded by
+  `is_x86_feature_detected!("avx2")` with a scalar `else` branch.
+- **Load bounds.** The only memory accesses are vector loads from
+  `block.as_ptr()`, where `block: &[u8; 64]`. NEON issues four 16-byte
+  `vld1q_u8` loads at offsets 0/16/32/48; AVX2 issues two 32-byte
+  `_mm256_loadu_si256` loads at offsets 0 and 32. Both cover exactly 64 bytes and
+  no more. Both use unaligned load forms, so no alignment beyond `u8` is
+  required.
+
+Seven `#[allow(unsafe_code)]` sites: the two public dispatch functions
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #677, opened 2026-07-31
test(simd-fastq): sweep SIMD/scalar lexer agreement with proptest

nh13:
> **Stacked on #676.** Base is `nh/fix-simd-fastq-unsafe-policy`, so review that one first; this PR's diff is the two test modules plus a CLAUDE.md line. Retarget to `main` once #676 lands.

#676 brought the SIMD lexer under `#![deny(unsafe_code)]` and documented what the scalar-parity tests are standing in for, since Miri cannot execute intrinsics and `miri.yml` does not reach this crate. It also said plainly that this closes a *policy* gap and does not by itself prove the intrinsics sound. This is the follow-up that strengthens the evidence.

## The gap

Agreement with the scalar reference was asserted only over fixed inputs: uniform blocks, one structured FASTQ block, and a loop setting a single byte at each of the 64 positions. Nothing exercised an **arbitrary combination of lanes** — which is where per-lane shuffle and movemask arithmetic is most likely to go wrong. AVX2's `_mm256_shuffle_epi8` operates within each 128-bit half, so a lane-crossing mistake needs both halves populated *differently* to surface, and no existing test does that.

## What's added

Four `proptest` sweeps in `lexer.rs`: newline bitmask vs scalar, full classification vs scalar on all three fields, the newline-only fast path vs the full classifier (separate implementations of the same predicate, so they must agree with each other and not merely each with scalar), and the same over uniformly random bytes as an adversarial complement.

Blocks come from a generator weighted toward newlines and bases. Uniform `any::<u8>()` almost never produces either, so it would leave the classification masks largely untouched while looking thorough.

Three properties in `parser.rs`: `find_record_offsets` over multi-block inputs — it is the code that stitches results across 64-byte boundaries, and the existing no-panic property stopped at 64 bytes — with its output checked in-bounds and strictly increasing; the streaming iterator; and a round-trip property over well-formed input, because no-panic properties prove only that the parser does not crash, not that it parses correctly.

## Two judgment calls worth reviewing

**`two_bits` is compared only on `is_acgt` lanes.** This is the contract, not a weakened assertion. `ENCODE_LUT` maps non-ACGT bytes to a don't-care value; the SIMD paths write it through unconditionally (`two_bits |= chunk_two_bits`) while the scalar path leaves those lanes zero. Comparing whole words would assert agreement the implementations never promised. `is_acgt` is asserted equal *first*, so no divergence can hide behind a disagreeing mask.

**`parse_records` is deliberately not fuzzed for panics.** I wrote that property first and it failed — then checking the contract showed the test was wrong, not the code. `parse_records` is documented as assuming well-formed input and routes through `parse_single_record`, which panics by contract. I confirmed `SimdFastqReader` uses the fallible `try_parse_single_record` at **both** call sites and that the panicking wrapper is reached only from this crate's bench and doc example, so no production path feeds it untrusted input. A note in the test module records this so the omission does not later read as an oversight.

## Verification

- `cargo ci-test`: **6896 passed** (up from 6890), 27 skipped. `ci-fmt` and `ci-lint` clean.
- `cargo check -p fgumi-simd-fastq --target x86_64-apple-darwin --all-targets` — the AVX2 half compiled rather than assumed.
- **Non-vacuity verified by mutation, not asserted.** Dropping `G` from the NEON ACGT comparison fails both full-classification properties while the newline-only property correctly still passes; truncating the newline fast path to three of its four loads fails both newline properties. proptest shrank each to a single offending byte (index 51 and 57 respectively).
- Regression seeds written during those deliberate failures were **discarded, not committed** — they encode bugs that never existed.
- The diff against #676 is **additive only**: 156 insertions, 0 deletions, no production code touched.

## Honest limit

A dev machine runs one instruction set, so locally these sweep the **NEON** path only; the AVX2 path is compiled here and exercised on CI's x86_64 runners. That is a genuine gain — the lane-crossing class described above is precisely what CI now sweeps and previously did not — but it does mean neither machine alone validates both paths. Worth knowing when reading a green check on a Mac.

This still does not make the intrinsics *proven* sound. Randomized agreement raises confidence; it is not Miri.

## Summary by CodeRabbit

* **Tests**
  * Added property-based coverage for SIMD FASTQ lexing across realistic and random input blocks.
  * Added validation against scalar behavior for newline detection, DNA bases, and 2-bit encoding.
  * Added parser tests for multi-block offsets, bounds, panic safety, and well-formed FASTQ round trips.

* **Documentation**
  * Expanded documentation of SIMD lexer correctness guarantees and cross-architecture validation.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -243,11 +243,21 @@ Seven `#[allow(unsafe_code)]` sites: the two public dispatch functions
 
 Note that Miri cannot execute these intrinsics, so `miri.yml` does not cover
 this crate. Correctness rests instead on the scalar-parity tests in `lexer.rs`,
-which assert the SIMD and scalar lexers agree:
-`test_simd_matches_scalar_newlines`, `test_simd_full_matches_scalar_all_fields`,
-`test_simd_full_matches_scalar_all_acgt`, and
-`test_simd_full_matches_scalar_every_position`. Any change to the intrinsic
-paths must keep these passing on both architectures.
+which assert the SIMD and scalar lexers agree. Fixed inputs pin specific shapes
+(`test_simd_matches_scalar_newlines`, `test_simd_full_matches_scalar_all_fields`,
+`test_simd_full_matches_scalar_all_acgt`,
+`test_simd_full_matches_scalar_every_position`), and `proptest` sweeps cover
+arbitrary lane combinations (`simd_newlines_match_scalar_on_arbitrary_blocks`,
+`simd_full_matches_scalar_on_arbitrary_blocks`,
+`newline_fast_path_matches_full_classifier`,
+`simd_matches_scalar_on_uniform_random_blocks`). Any change to the intrinsic
+paths must keep these passing on both architectures — note that a dev machine
+runs only one of the two, so the other is verified in CI.
+
+The sweeps compare `two_bits` only on lanes where `is_acgt` is set. That is the
+contract rather than a weakened assertion: `ENCODE_LUT` maps non-ACGT bytes to a
+don't-care value, which the SIMD paths write through unconditionally while the
+scalar path leaves those lanes zero.
 
 Any new `unsafe` site must extend this list and explain why the safe
 alternative is unacceptable. Do not introduce `unsafe` outside the crates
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #681, opened 2026-07-31
fix(duplex): count consensus reads, not templates, in --threads mode

### LINKED ISSUE #680: fgumi duplex --threads reports half the consensus reads it writes
### Description

`fgumi duplex --threads N` reports exactly **half** the consensus reads it actually writes. The single-threaded path is correct and matches fgbio; the multi-threaded pipeline path is not.

The duplex caller records one `record_consensus()` per emitted *template* (R1 + R2), while `ConsensusCallingStats::consensus_reads` is documented and consumed as a count of consensus *reads* (the vanilla/simplex caller records twice for a pair, once for a fragment). `execute_single_threaded` papers over this by overwriting `metrics.consensus_reads` with its locally tracked written-record count; `execute_threads_mode` uses the accumulated stats verbatim, so the under-count surfaces in both the log line and the `--stats` file.

### Reproduction

Same input BAM (20,000 simulated duplex molecules), `fgumi 0.5.0`:

```console
$ fgumi duplex -i grouped.bam -o out-t1.bam --min-reads 1 --stats stats-t1.txt
[INFO fgumi_lib::logging]   Consensus reads: 21,432
$ grep consensus_reads_emitted stats-t1.txt
consensus_reads_emitted	21432

$ fgumi duplex -i grouped.bam -o out-t4.bam --min-reads 1 --threads 4 --stats stats-t4.txt
[INFO fgumi_lib::logging]   Consensus reads: 10,716
$ grep consensus_reads_emitted stats-t4.txt
consensus_reads_emitted	10716

$ samtools view -c out-t4.bam
21432
```

Both runs write byte-identical output BAMs with 21,432 records. fgbio 4.1.0 on the same input reports `consensus_reads_emitted	21432`, agreeing with the `-t 1` path.

Note the pipeline's own tally is right — the same run logs `Total consensus reads written by pipeline: 21432` immediately before logging `Consensus reads: 10,716`.

### Impact

- `--stats` output is wrong for every threaded duplex run, so `frac_raw_reads_used`-style downstream QC comparisons against fgbio or against `-t 1` disagree.
- The value is thread-count dependent for identical input and identical output.

### Version

0.5.0 (also present on `main` @ 3d27019)

nh13:
Closes #680.

`fgumi duplex --threads N` reported exactly half the consensus reads it wrote — in the log line and in the `--stats` file — while `-t 1` reported the correct number for the same input and the same (byte-identical) output BAM.

`ConsensusCallingStats::consensus_reads` counts emitted *reads*: the vanilla caller records twice for an R1/R2 pair and once for a fragment. The duplex caller instead recorded a single consensus per emitted *template* at all three of its emit sites (both-strands, AB-only, BA-only). `execute_single_threaded` masked that by overwriting `metrics.consensus_reads` with its own locally tracked write count just before logging; `execute_threads_mode` consumes the accumulated stats as-is, so the under-count surfaced only under `--threads`.

The fix corrects the accounting at its source rather than patching the reporting: `record_consensus_pair` is added alongside `record_consensus`, the duplex emit sites use it, and the single-threaded overwrite is dropped so both paths derive the metric from the same accounting.

## Verification

On a 20,000-molecule simulated duplex BAM (81,368 raw reads), before:

```console
$ fgumi duplex ... --threads 4 --stats stats-t4.txt
[INFO] Consensus reads: 10,716
$ samtools view -c out-t4.bam
21432
```

After, both `-t 1` and `-t 4` report `21,432`, matching the records in the output BAM and fgbio 4.1.0's `consensus_reads_emitted` on the same input.

New test: an rstest case table over `--threads` unset/`2` asserting the reported `consensus_reads_emitted` equals the number of records actually written. It fails on `main` for the threaded case (`3` vs `6`) and passes for both cases here.

`cargo ci-test` (6,892 tests), `cargo ci-fmt`, and `cargo ci-lint` all pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected consensus statistics for paired duplex and single-strand outputs.
  * Fixed single-threaded duplex metrics to report the accurate number of consensus reads emitted.

* **Tests**
  * Added coverage verifying consensus records in output files match reported statistics across threaded and single-threaded execution.

### DIFF
--- a/crates/fgumi-consensus/src/caller.rs
+++ b/crates/fgumi-consensus/src/caller.rs
@@ -252,7 +252,10 @@ pub trait ConsensusCaller: Send + Sync {
 pub struct ConsensusCallingStats {
     /// Total number of input reads processed
     pub total_reads: usize,
-    /// Number of consensus reads generated
+    /// Number of consensus *reads* generated — not templates. A caller that emits both ends
+    /// of a pair contributes two (see [`ConsensusCallingStats::record_consensus_pair`]); a
+    /// fragment contributes one. The count is meant to equal the records written to the
+    /// output BAM.
     pub consensus_reads: usize,
     /// Number of reads filtered/rejected
     pub filtered_reads: usize,
@@ -281,6 +284,14 @@ impl ConsensusCallingStats {
     /// Records that a consensus read was created
     pub fn record_consensus(&mut self) {
         self.consensus_reads += 1;
+    }
+
+    /// Records that a consensus read pair (R1 and R2) was created.
+    ///
+    /// `consensus_reads` counts emitted *reads*, not templates, so a caller that
+    /// emits both ends of a pair must record two.
+    pub fn record_consensus_pair(&mut self) {
+        self.consensus_reads += 2;
     }
 
     /// Records that input reads were processed
--- a/crates/fgumi-consensus/src/duplex_caller.rs
+++ b/crates/fgumi-consensus/src/duplex_caller.rs
@@ -2129,7 +2129,7 @@ impl DuplexConsensusCaller {
                             cell_tag,
                             cell_barcode.as_deref(),
                         )?;
-                        stats.record_consensus();
+                        stats.record_consensus_pair();
                         return Ok((output, stats, Vec::new()));
                     }
                     stats.record_rejection(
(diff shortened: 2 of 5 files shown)


nh13:
@coderabbitai pause

PR #298, opened 2026-04-19
fix(consensus): preserve BA-only orientation for methylation tags

nh13:
## Summary

- Fix the BA-only branch of `DuplexConsensusCaller::consensus_reads` so `is_ba_only=true` is preserved, causing `duplex_read_into` to emit bottom-strand methylation tags (`bm`/`bu`/`bt`) for BA-only groups instead of top-strand tags (`am`/`au`/`at`).
- Root cause: the branch passed BA data in the AB (first) argument of `duplex_consensus`, which selects the `(Some(a), None)` match arm (`is_ba_only=false`). The fix passes BA in the BA (second) argument.

Flagged by CodeRabbit on #291; the bug is pre-existing on `main` (predates the RawRecord migration), so landing the fix here keeps #291 focused on the refactor.

## Test plan
- [x] Added `test_duplex_ba_only_methylation_tags_use_bottom_strand` — integration test through `consensus_reads_from_sam_records` with EM-Seq + B-only reads; asserts `bu`/`bt` present and `au`/`at` absent on every output record.
- [x] Verified the test fails on the unfixed code path (red) before applying the fix (green) — TDD.
- [x] `cargo ci-test` passes (2570 tests).
- [x] `cargo ci-fmt` + `cargo ci-lint` pass.

PR #683, opened 2026-07-31
fix(duplex): restore single-strand consensus via --min-reads 1,1,0

### LINKED ISSUE #678: fgumi duplex to also allow single strand consensus
### Description

We have been using fgbio for quite a long time now in our duplex data processing and variant calling pipeline, and we used to define the filter of reads in the CallDuplexConsensus step as 1 1 0 in such a way that all molecules would be compressed to a single one no matter if this would make it into duplex or not.

Now while moving to fgumi I realised that the min-reads parameter cannot receive a value of 0 for one of the strands, and then the process of calling becomes a lot more strict.

The current behavior is:
[2026-07-30T21:18:31Z INFO  fgumi_lib::logging]   Input reads: 148,583,820
[2026-07-30T21:18:31Z INFO  fgumi_lib::logging]   Consensus reads: 5,287,875

The desired behaviour is:

Be able to provide 1,1,0 as value for the --min-reads flag in fgumi duplex so that I could reproduce this previous behaviour obtained with fgbio v2.1.
[2026/07/28 07:13:45 | CallDuplexConsensusReads | Info] Total Raw Reads Considered: 148,583,820.
[2026/07/28 07:13:45 | CallDuplexConsensusReads | Info] Consensus reads emitted: 38,686,412.

### Related Issues

_No response_

### Version

0.5.0

### Impact

My guess is that this feature would not have any major drawback since this is just a filtering criteria, but I know little about the implementation so I might be missing something.

### Suggestions

I have identified the filtering condition here, but I am not sure if not having this has any downstream consequences...
https://github.com/fulcrumgenomics/fgumi/blob/3d2701932fcad794fa7ecdcd0a1d81dd135c8935/src/lib/commands/duplex.rs#L588-L594

### Test input

I can provide testing data if needed, but I guess any data should be enough to test this new feature (if implemented).

### Checklist

- [x] Checked there are no existing Issues for this feature request
- [ ] Checked there are no Discussions that address this issue
- [x] Ensured that the latest release does not contain this feature
- [x] Provided all information necessary to reproduce the current behavior

nh13:
Closes #678.

`fgumi duplex --min-reads 1,1,0` — fgbio's single-strand mode, where a molecule observed on only one strand still yields a consensus — worked in 0.4.0 and errors out in 0.5.0. `DuplexConsensusCaller` has always implemented the mode; what changed is the CLI. #601 added a lower bound to `Duplex::validate()` for consistency with `simplex` and `codec`, but its blanket `contains(&0)` also rejected the per-strand slots, where a 0 is meaningful.

The bound is dropped entirely rather than narrowed, matching fgbio, which validates only that the values run from least to most stringent. `simplex` and `codec` keep their check, because there a 0 really does make the minimum-family-size filter a silent no-op — that part of #601 stands. `duplex` differs in both slots: a 0 per strand is the single-strand mode, and a total of 0 is equivalent to 1 rather than degenerate, because the check it feeds (`min_total <= num_xy + num_yx`) is only reached for a group that already has at least one read.

Restoring the flag alone was not enough to reproduce fgbio, because two code paths that are unreachable while the mode is gated were wrong.

**The overlapping-bases consensus was skipped for single-strand groups.** It is gated on the group having both strands, "no duplex possible anyway" — which stops being true once the third value is 0, so single-strand molecules with overlapping mates kept uncorrected bases. It is now gated on whether the group can actually produce a consensus.

**BA-only molecules came out with R1/R2 swapped relative to fgbio.** #298 mapped output R1 from BA-R2 and R2 from BA-R1, to keep output R1 on the physical strand that AB-R1 would have sequenced. That formula holds only while both strand groups are present; fgbio does not extend it to a lone group, taking whichever group is present as its "AB" side, so output R1 comes from that group's R1s whether it is `/A` or `/B`. Verified in the data — within a molecule, AB-R1 and BA-R1 are on opposite strands at opposite ends:

```
molecule 13395: A-R1=fwd@44192844   B-R1=rev@44193088
molecule 14413: A-R1=rev@67604810   B-R1=fwd@67604719
```

so the previous mapping put the opposite end of the molecule in R1 for `/B`-only molecules than for every other molecule class. Each consensus read is now built from the input reads of the same end. **#298's other half is preserved**: BA still goes in the BA slot, so `is_ba_only` stays true and per-strand methylation tags are still emitted as `bm`/`bu`/`bt` — that flag records which strand *group* the bases came from, which the end does not change. `test_duplex_ba_only_methylation_tags_use_bottom_strand` is untouched and passes; `test_duplex_ba_only_mate_pair_mapping` is updated to the new expectation.

## Verification

On 20,000 simulated duplex molecules (81,368 raw reads), against fgbio 4.1.0, comparing with `fgumi compare bams --command duplex`:

| `--min-reads` | records | content diffs vs fgbio |
|---|---|---|
| `1` | 21,432 | 0 |
| `1,1,0` | 40,000 | 0 |
| `1,1,0 --threads 4` | 40,000 | 0 |
| `0` | 40,000 | 0 |

Before this change, 14,014 of those 40,000 records differed from fgbio, in two ways: R1/R2 swaps on `/B`-only molecules, and missing overlap corrections on single-strand molecules whose mates overlap. Disabling the overlapping consensus on both sides isolates the first defect, and it accounts for exactly 9,048 differing records — the R1/R2 pairs of all 4,524 `/B`-only molecules, with the 4,760 `/A`-only and 10,716 two-strand molecules matching fgbio exactly. `--min-reads 1` is unchanged.

New tests, each watched failing first: `--min-reads` lower-bound cases; single-strand molecules are emitted with `1,1,0` and still rejected with `1`; consensus reads follow input read ends for `/A`-only and `/B`-only molecules; the overlapping consensus is applied to single-strand groups. All are parameterized over single-threaded and `--threads 2`.

`cargo ci-test` (6,903 tests), `cargo ci-fmt`, and `cargo ci-lint` pass.

## Note on fgbio's docs

fgbio implements this mode but does not document it — neither `CallDuplexConsensusReads` nor `FilterConsensusReads` mentions a 0 value, and `DuplexConsensusCaller.duplexConsensus`'s own scaladoc ("If either of the incoming reads are undefined, the duplex read will be undefined") contradicts its `case (Some(a), None)` arm. This PR documents the mode on the fgumi side, in the duplex consensus guide and in `--min-reads`' help text.

## Summary by CodeRabbit

- **New Features**
  - Duplex consensus calling supports single-strand molecules with configurations such as `--min-reads 1,1,0`.
  - Single-strand consensus reads are marked with `bD:i:0`.
  - Consensus generation is consistent in threaded and single-threaded modes.

- **Bug Fixes**
  - Corrected R1/R2 orientation and methylation handling for BA-only consensus output.
  - Improved overlapping-base handling and threshold enforcement.
  - Invalid `--min-reads` configurations are rejected before output is created.

- **Documentation**
  - Added guidance on zero-valued thresholds, filtering, and fgbio equivalence.

### DIFF
--- a/crates/fgumi-consensus/src/duplex_caller.rs
+++ b/crates/fgumi-consensus/src/duplex_caller.rs
@@ -323,6 +323,76 @@ pub struct DuplexConsensusCaller {
     reason = "consensus calling uses numeric casts for quality/position math and has domain-specific variable names"
 )]
 impl DuplexConsensusCaller {
+    /// The `YX` (less-covered strand) minimum implied by `min_reads`.
+    ///
+    /// fgbio pads `--min-reads` to three values by repeating the last
+    /// (`padTo(3, last)`), so the `YX` threshold is the third value when one was
+    /// given and the last value otherwise. Returns `None` only for an empty
+    /// slice, which [`Self::new`] rejects.
+    #[must_use]
+    pub fn min_yx_reads_for(min_reads: &[usize]) -> Option<usize> {
+        min_reads.get(2).or_else(|| min_reads.last()).copied()
+    }
+
+    /// Whether `min_reads` lets a molecule observed on only one strand still
+    /// yield a consensus — fgbio's `-M 1 1 0` single-strand mode, i.e. a `YX`
+    /// minimum of 0.
+    ///
+    /// Callers that need this answer before a caller exists (the CLI's
+    /// overlapping-consensus gate, and the threaded path, which decides before
+    /// the per-batch caller is built) must go through here rather than
+    /// re-deriving the padding rule, so the gate cannot drift from the
+    /// `min_yx_reads` [`Self::new`] actually applies.
+    #[must_use]
+    pub fn allows_single_strand_consensus(min_reads: &[usize]) -> bool {
+        Self::min_yx_reads_for(min_reads) == Some(0)
+    }
+
+    /// Validates `min_reads` against the rule [`Self::new`] applies: 1-3 values,
+    /// ordered high to low (`total >= XY >= YX`) once fgbio's `padTo(3, last)`
+    /// has filled the missing slots. No lower bound is imposed, matching fgbio.
+    ///
+    /// [`Self::new`] delegates here, so a caller that must reject bad input
+    /// *before* a caller exists can fail fast without re-deriving the rule. The
+    /// CLI's `--threads` path is exactly that case: it builds its per-batch
+    /// caller inside the pipeline's process closure, i.e. after the output BAM
+    /// has already been created, so an ordering violation caught only there
+    /// surfaces as a wrapped I/O error from a worker and leaves a partial file.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if `min_reads` is empty, has more than 3 values, or the
+    /// values are not in decreasing order (`total >= XY >= YX`).
+    pub fn validate_min_reads(min_reads: &[usize]) -> Result<()> {
+        // `min_yx_reads_for` is `None` only for an empty slice, so binding it here rather
+        // than unwrapping below keeps the emptiness check in one place.
+        let (Some(&min_total_reads), Some(&last), Some(min_yx_reads)) =
+            (min_reads.first(), min_reads.last(), Self::min_yx_reads_for(min_reads))
+        else {
+            anyhow::bail!("min_reads parameter must have at least 1 value");
+        };
+        if min_reads.len() > 3 {
+            anyhow::bail!(
+                "min_reads parameter must have 1-3 values (total, [XY, [YX]]), got {} values",
+                min_reads.len()
+            );
+        }
+
+        // fgbio's `padTo(3, last)`: the missing slots repeat the last value given.
+        let min_xy_reads = min_reads.get(1).copied().unwrap_or(last);
+
+        // Validate min_reads ordering: yx <= xy <= total (matching fgbio)
+        // For depth thresholds it's required that yx <= xy <= total
+        if min_xy_reads > min_total_reads {
+            anyhow::bail!("min-reads values must be specified high to low (total >= XY)");
+        }
+        if min_yx_reads > min_xy_reads {
+            anyhow::bail!("min-reads values must be specified high to low (XY >= YX)");
+        }
+
+        Ok(())
+    }
+
     /// Creates a new duplex consensus caller
     ///
     /// # Arguments
@@ -370,31 +440,17 @@ impl DuplexConsensusCaller {
         error_rate_pre_umi: u8,
         error_rate_post_umi: u8,
     ) -> Result<Self> {
+        // Arity and ordering both live in `validate_min_reads` so the CLI can apply the
+        // same rule before any output exists.
+        Self::validate_min_reads(&min_reads)?;
+
         // Parse min_reads vector like fgbio: padTo(3, last)
         // [total, XY, YX] where XY >= YX (larger strand, smaller strand)
-        if min_reads.is_empty() {
-            anyhow::bail!("min_reads parameter must have at least 1 value");
-        }
-        if min_reads.len() > 3 {
-            anyhow::bail!(
-                "min_reads parameter must have 1-3 values (total, [XY, [YX]]), got {} values",
-                min_reads.len()
-            );
-        }
-
-        let last = *min_reads.last().expect("expected a last item");
+        let last = *min_reads.last().expect("min_reads was validated non-empty above");
         let min_total_reads = min_reads[0];
         let min_xy_reads = min_reads.get(1).copied().unwrap_or(last);
-        let min_yx_reads = min_reads.get(2).copied().unwrap_or(last);
-
-        // Validate min_reads ordering: yx <= xy <= total (matching fgbio)
-        // For depth thresholds it's required that yx <= xy <= total
-        if min_xy_reads > min_total_reads {
-            anyhow::bail!("min-reads values must be specified high to low (total >= XY)");
-        }
-        if min_yx_reads > min_xy_reads {
-            anyhow::bail!("min-reads values must be specified high to low (XY >= YX)");
-        }
+        let min_yx_reads =
+            Self::min_yx_reads_for(&min_reads).expect("min_reads was validated non-empty above");
 
         // Create single-strand consensus caller with min_reads=1 (matching fgbio)
         // Filtering happens at duplex level based on input read counts, not at SS level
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
Addressed the outside-diff comment on `src/lib/commands/duplex.rs:1203-1225` (no inline thread to resolve): `test_validate_min_reads_accepts_zero` is renamed to `test_validate_does_not_reject_zero_min_reads`, the always-true `expect_ok` parameter is dropped, and the doc comment now states explicitly that the test guards only against reintroducing the removed lower-bound rejection — ordering validation stays in `DuplexConsensusCaller::new`, covered by `test_min_reads_validation_error`.

Two further fixes from a self-review pass:

- The `--min-reads` doc text I added said the values must be given "from least to most stringent", which is backwards and contradicted this command`s own long help ("if values two and three differ, the more stringent value comes earlier", matching fgbio `CallDuplexConsensusReads.scala:101`). Corrected at all three sites.
- Added a unit test for the new `ConsensusCallingStats::record_consensus_pair`, which counts two emitted reads.

nh13:
@coderabbitai review

nh13:
Addressed the outside-diff finding on `crates/fgumi-consensus/src/duplex_caller.rs:2169-2276` (no inline thread to resolve).

The AB-only and BA-only arms now mirror the both-strand arm and re-check `duplex_consensus_has_minimum_reads` against the post-`filter_by_alignment` depths before emitting. This restores fgbio parity: fgbio applies `duplexHasMinimumNumberOfReads` to every emitted consensus regardless of strand count (`DuplexConsensusCaller.scala:221`), with its own comment naming exactly this case — "Even though we had enough reads going into consensus calling, we can lose some in the process (e.g. to CIGAR filtering)". Failure falls through to the existing rejection at the end of the function, so the `InsufficientReads` accounting and reject-BAM passthrough are unchanged.

Regression test `test_single_strand_rechecks_min_reads_after_alignment_filtering` covers both arms with the suggested threshold: five single-strand pairs clear `5,3,0` on the raw counts, but two carry `5M1I4M` (neither a prefix of nor prefixed by `10M`), so the most-common-alignment group keeps three and the molecule must be rejected. Verified it fails on the pre-fix code (emitted 2 records). The `3,3,0` control cases assert record identity rather than a count — R1/R2 flags, shared read name, base `MI`, 10bp sequence, `aD:i:3` (pinning that only the filtered reads contributed) and `bD:i:0`.

Also applied the two wording nits in `docs/src/guide/duplex-consensus-calling.md`: "afterwards" → "afterward", and "whether or not it makes duplex" → "regardless of whether it makes duplex" (the intended sense is the regardless-of one, which is what LanguageTool's `WHETHER` rule asks be made explicit).

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed both findings from the latest review.

**Template-count boundary** (inline thread, resolved). The literal suggestion — `--min-reads 2,2,0` on "the same one-template input" — misreads the fixture: `create_single_strand_molecule(..., depth = 3, ...)` builds three pairs, not one, so `2,2,0` clears the threshold and emits. The underlying concern is real though, so `test_duplex_single_strand_mode_emits_one_strand_molecules` now straddles the template count instead: `3,3,0` (equal to the 3 templates) must still emit, and `4,4,0` must not. That pair is what discriminates, since the molecule is 3 templates but 6 read records — an implementation counting records would emit in both cases. Verified by mutation: inflating the depth in `duplex_consensus_has_minimum_reads` makes the `4,4,0` case fail. (Inflating only `has_minimum_number_of_reads` does not, because the post-filter re-check added earlier in this PR catches it — the end-to-end contract is pinned either way.)

Both accepting cases now assert full pair identity (two records, exactly one R1, molecule id, bases) through a shared `assert_is_the_molecules_pair` closure, rather than a bare count.

**fgbio oracle for single-strand output** (outside-diff, no thread to resolve). Added `duplex_single_strand_matches_fgbio_oracle` in `duplex_caller.rs`'s `fgbio_oracle_tests`, with `/A`-only and `/B`-only fixtures built by `build_single_strand_fixture` and regenerated by the `#[ignore]`d `regen_duplex_single_strand_{ab,ba}_only_fixture` tests. Expectations are captured from a real `fgbio 4.1.0 CallDuplexConsensusReads -M 1 1 0` run on those exact fixture BAMs — no fgbio or JVM in default CI. R1 and R2 carry different bases (`ACGTACGT` / `TTGGCCAA`, neither the reverse complement of the other) so the read-end mapping is observable; one pair's R1 disagrees at base 0 so the error numerator is exercised.

Pinned per record: flags, bases, qualities, `MI`, `RX`, `aD`/`aM`/`aE`, `bD`/`bM`/`bE`, `cD`/`cM`/`cE`, `ad`, `ae`, `ac`, `aq` — plus the tag *shape*, i.e. that `bd`/`be`/`bc`/`bq` are absent, which is what fgbio does for a lone strand and which value-only assertions would miss. Methylation tags are not pinned: fgbio emits them only for a bisulfite run, which this fixture is not; the `/B`-only orientation stays covered by `test_duplex_ba_only_uses_bottom_strand_methylation_tags`. Verified the oracle catches the pre-fix BA mapping — restoring `BA-R2 → R1` fails the `ba_only` case on R1 bases and leaves `ab_only` passing.

The `/A` and `/B` cases pin the same bases in the same slots; only the error position differs (base 0 vs base 7), because the `/B` fixture's R1s are `REVERSE` and fgbio consensus-calls in sequencing orientation.

**Overlap-correction oracle.** `test_duplex_single_strand_mode_applies_overlapping_consensus` no longer asserts a "contains an N" predicate. `write_overlapping_single_strand_fixture` now serves both the test and a new `#[ignore]`d `regen_duplex_single_strand_overlap_fixture`, and both the corrected and uncorrected outputs are pinned to the captured fgbio run in both execution modes:

- `--consensus-call-overlapping-bases true` → R1 `ACGTACG` / `HHHHHHH`, R2 `NCGTACGT` / `#HHHHHHH`
- `--consensus-call-overlapping-bases false` → R1 `ACGTACGT` / `>>>>>>>>`, R2 `TCGTACGT` / `>>>>>>>>`

fgumi's output is byte-identical to fgbio's on both settings. The asymmetry is fgbio's: the mates disagree at the last reference base, which masks to N at qual 2 in both, and in sequencing orientation that is the end of R1 (trimmed) and the start of R2 (kept as `N` at depth 0). Confirmed by running the same command on a fixture that disagrees at the *first* reference base instead, which swaps the two records' shapes exactly.

Three fixes from a self-review pass over the diff:

- The `OVERLAP_R2_BASES` doc comment claimed the mates disagree "at reference offset 0" and described `TCGTACGT` as R2's reference bases. BAM `SEQ` is stored reference-forward, so R2's reference bases are `ACGTACGA` and the disagreement is at offset 7 — contradicting the comment 100 lines below it in the same test. Corrected; `TCGTACGT` is now named as the sequencing orientation.
- `validate_min_reads` had an unreachable `.unwrap_or(last)` on `min_yx_reads_for`, whose only `None` case the `let-else` above had already rejected. Folded into the same `let-else` so emptiness is checked once.
- The new integration regen test panicked when `FGUMI_ORACLE_BAM_OUT` was unset, where the sibling `regen_write` helper falls back to a logged temp path. Matched the ergonomics (that helper is `pub(crate)` to `fgumi-consensus`, so it cannot be shared).

nh13:
@coderabbitai review

nh13:
Good catch — the finding is correct and the assertion was genuinely vacuous.

`get_u8_array_tag` resolves `B:C`, but `bd`/`be` are written as `B:s` (`append_i16_array_tag`, `duplex_caller.rs:1247` and `:1251`), so neither branch of `get_u8_array_tag(tag).is_none() && get_string_tag(tag).is_none()` could ever see them. Confirmed by mutation: emitting zero-filled `bd`/`be` for the absent strand left both oracle cases passing. With the accessors split by written type — `get_i16_array_tag` for `bd`/`be`, `get_string_tag` for `bc` (`append_string_tag`) and `bq` (`append_phred33_string_tag`, which emits `Z`) — the same mutation fails on `R1 must NOT carry bd`. Reverted the mutation; both cases pass on the real code.

Swept the class rather than fixing only the cited site. Every tag-absence assertion in the changed files, checked against how the tag is actually written:

| Site | Written as | Accessor | Verdict |
|---|---|---|---|
| `duplex_caller.rs:7375` — `bd`/`be` | `append_i16_array_tag` → `B:s` | `get_i16_array_tag` | fixed |
| `duplex_caller.rs:7381` — `bc`/`bq` | `append_string_tag` / `append_phred33_string_tag` → `Z` | `get_string_tag` | already correct |
| `duplex_caller.rs:6567,6571` — `au`/`at` | `append_i16_array_tag` → `B:s` | `get_i16_array_tag` | already correct |
| `duplex_caller.rs:5127` — `CB` | `append_string_tag` → `Z` | `get_string_tag` | already correct |

No other site had the mismatch. The presence assertions in the same test (`ad`, `ae`, `ac`, `aq`) cannot go vacuous the same way — a wrong accessor returns `None` and fails immediately.

Added a comment at the site recording why the accessor has to match the written type, so the next edit does not reintroduce it.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #684, opened 2026-07-31
fix(zipper,clip): compute supplementary TLEN instead of copying the mate primary's

### LINKED ISSUE #673: zipper/clip: supplementary TLEN is copied from the mate primary instead of computed
`Template::fix_mate_info` sets a supplementary alignment's `TLEN` by negating its **mate primary's** `TLEN` instead of computing it from the supplementary's own coordinates. This is a port of htsjdk's `SamPairUtil.setMateInformationOnSupplementalAlignment` and inherits its defect.

The negation assumes the supplementary sits where its own primary sits — same reference, same side of the mate. For split alignments, the only records this path touches, that is false by construction.

## Where

- `src/lib/template.rs:533` — R1 supplementals (`zipper`)
- `src/lib/template.rs:578` — R2 supplementals (`zipper`)
- `src/lib/commands/clip.rs:1004` — `set_supplemental_mate_info_raw` (`clip`)

```rust
fgumi_raw_bam::set_template_length(rec, -r2_tlen);   // copied, not computed
```

The primary path computes correctly in `compute_insert_size_raw` (`template.rs:819`), guarding both unmapped and cross-reference. The supplementary path is not routed through it.

## Failure mode 1 — supplementary on a different reference

Chimeric templates of this shape are common in real data — the primaries abut on one chromosome and the supplementary lands on another:

```
primary R1  chr17:66,798,979  forward  33S115M
primary R2  chr17:66,798,853  reverse  22S127M
SUPPL   R1  chrX:153,688,036  reverse  113S35M
```

```
R1    : ref_id=0 pos=66798978 mate_ref_id=0 mate_pos=66798852 tlen=1
R2    : ref_id=0 pos=66798852 mate_ref_id=0 mate_pos=66798978 tlen=-1
R1supp: ref_id=1 pos=153688035 mate_ref_id=0 mate_pos=66798852 tlen=1
```

`bwa-mem` and `minibwa` both emit `0` for a record in this position; `zipper` overwrites it.

## Failure mode 2 — same reference, supplementary beyond its mate

Not caught by a cross-reference guard, and not covered by any test.

```
primary R1  chr1:1,000  forward  100M
primary R2  chr1:1,350  reverse  100M
SUPPL   R1  chr1:5,000  forward  100M   -->  tlen=450
```

Sign and magnitude both wrong: the supplementary is the **rightmost** segment (so `TLEN` must be negative), and `450` is the primary pair's insert size.

## Why this is a bug

The SAM spec is silent on supplementary `TLEN` — hts-specs [#522](https://github.com/samtools/hts-specs/pull/522) scoped the definition to primary reads, and [#842](https://github.com/samtools/hts-specs/issues/842) leaves it aligner-defined. So this rests on two other grounds:

1. **fgumi contradicts itself** — `compute_insert_size_raw` guards cross-reference; the supplementary path doesn't.
2. **Every independent implementation computes from real coordinates.** `bwa-mem` (`bwamem.c:887-892`) and `minibwa` (`format.c:243-262`) compute per record from its own 5′ position, emitting `0` when cross-reference — including on supplementary records; `samtools fixmate` leaves supplementaries untouched; htslib's CRAM decoder uses chain extents. Only the htsjdk lineage — Picard `MergeBamAlignment`, fgbio, fgumi — copies from another record.

Impact is low: no known consumer reads supplementary `TLEN` (`riker` skips supplementaries entirely and keys chimera detection off `SA`), and `ValidateSamFile` never cross-checks `TLEN` against `RNAME`/`RNEXT`, which is why this survived since 2014.

## Suggested fix

`compute_insert_size_raw` already returns `0` for unmapped and cross-reference pairs and otherwise computes the 5′-based insert size, so routing the supplementary path through it fixes both modes:

```rust
let tlen = compute_insert_size_raw(rec, &rr[r2_i]);   // supplementary first
fgumi_raw_bam::set_template_length(rec, tlen);
```

Apply at all three sites. Keep the existing 5′-based pairwise basis rather than htslib's chain-wide one — chain-wide extents would also change the *primaries'* `TLEN`, diverging from bwa on records everyone currently agrees on.

The three existing `..._tlen_on_*_supplementals` tests assert the negated value and were written as fgbio-parity tests; updating them states that parity is no longer the invariant here. Add cross-reference and beyond-mate cases, and mirror for `clip`.

## Upstream

Root cause is `htsjdk SamPairUtil.java:354`, introduced in [9e03608a](https://github.com/samtools/htsjdk/commit/9e03608a2) (2014), with behaviour unchanged since. Also affects Picard `MergeBamAlignment` and fgbio (`Bams.scala:118,122`, `ClipBam.scala:129,130`, `TrimPrimers.scala:215`).

- Upstream: samtools/htsjdk#1795
- fgbio tracking: fulcrumgenomics/fgbio#1165

Fixing fgumi first is an intentional divergence from fgbio until upstream lands.

---

*Investigation and reproduction assisted by Claude Code (Anthropic). All code references and the htsjdk 5.0.0 reproduction were verified by hand.*

nh13:
Closes #673.

`Template::fix_mate_info` (`zipper`) and `set_supplemental_mate_info_raw` (`clip`) set a supplementary alignment's TLEN by negating its **mate primary's** TLEN, a faithful port of htsjdk's `SamPairUtil.setMateInformationOnSupplementalAlignment`. That encodes an unstated assumption — that the supplementary occupies the same place in the template geometry as its own primary, same reference and same side of the mate — which is false by construction for the split alignments this path exists to handle.

## Two failure modes

**Cross-reference.** A supplementary mapped to a different reference than its mate gets a non-zero TLEN, so the record carries `RNAME != RNEXT` with `TLEN != 0`.

**Same reference, supplementary beyond its mate.** Both sign and magnitude are wrong. With R1 at chr1:1,000 forward, R2 at chr1:1,350 reverse and an R1 supplementary at chr1:5,000, the supplementary received `+450` — the *primary pair's* insert size — when it is the rightmost segment of the template and so must be negative.

Neither case had test coverage: every existing supplementary-TLEN test placed the supplementary on the same reference as its mate and asserted the copied value.

## The fix

Compute TLEN from the supplementary's own alignment against the mate primary. `compute_insert_size_raw` already returns `0` for unmapped and cross-reference pairs, so both failure modes fall out of routing the supplementary path through it — the correct logic was already sitting in the same file, just not wired to this path.

In `template.rs` it is split into an `InsertSizeEnd` snapshot plus `compute_insert_size_from_ends`, so the supplementary loops can compute against the mate primary without holding a second borrow of `self.records`. No per-record allocation is added on either path.

This also removes a latent ordering dependency: the old code required the primary pair to be fixed first so the supplementary would read an updated TLEN (fgbio carries the same constraint, noted at `Bams.scala:111`). Computing from coordinates makes the two steps independent.

## Why this basis

The SAM specification is **silent** on supplementary TLEN rather than violated by the old behaviour — hts-specs [#522](https://github.com/samtools/hts-specs/pull/522) scoped the definition to primary reads and left non-primary records undefined, and [#842](https://github.com/samtools/hts-specs/issues/842) leaves the computation aligner-defined. The case rests on two other grounds:

1. **fgumi contradicted itself.** `compute_insert_size_raw` guards cross-reference; the supplementary path did not.
2. **Every independent implementation computes from real coordinates.** `bwa-mem` (`bwamem.c:887-892`) and `minibwa` (`format.c:243-262`) compute per emitted record from that record's own 5′ position and emit `0` across references — including for supplementary records. `samtools fixmate` leaves supplementaries untouched. Only the htsjdk lineage copies the value from a different record.

The existing 5′-based pairwise basis is deliberately retained rather than htslib's chain-wide leftmost-to-rightmost basis: chain-wide extents would also change the **primaries'** TLEN whenever a supplementary falls outside the pair's span, diverging from bwa on records that all implementations currently agree on.

## Scope and divergence

Three sites across two commands: `template.rs:533` and `:578` (`zipper`), `clip.rs:1004` (`clip`). The issue as originally filed covered only `zipper`.

The existing supplementary-TLEN tests were written for fgbio parity and asserted the copied value; they now assert the computed value. **This is an intentional divergence from fgbio and Picard `MergeBamAlignment` on these records** until the upstream fix lands — the `compare` harness will report it.

Root cause is `htsjdk SamPairUtil.java:354`, introduced in [9e03608a](https://github.com/samtools/htsjdk/commit/9e03608a2) (2014) with behaviour unchanged since. Filed upstream as samtools/htsjdk#1795, tracked for fgbio as fulcrumgenomics/fgbio#1165.

## Tests

Eight existing assertions updated (seven unit, one integration), each previously asserting the copied value. Eight new `rstest` cases added across `zipper` and `clip` covering cross-reference, beyond-mate, before-mate, and coincident-5′-end geometries; the case tables seed a sentinel TLEN so a regression cannot pass by coincidence.

Worth noting that `test_clip_command_threads_mode_supplementary_mate_repair` already had exactly the geometry at issue — a supplementary 5 kb from its mate — and was asserting `+298` where the correct value is `-4604`. Nothing was checking supplementary TLEN correctly.

`cargo ci-fmt`, `cargo ci-lint`, and `cargo ci-test` all pass (6898 tests).

---

*Investigation and reproduction assisted by Claude Code (Anthropic). All code references and the htsjdk 5.0.0 reproduction cited in #673 were verified by hand.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected template-length calculations for supplementary alignments.
  * Supplementary records now calculate lengths from their own coordinates and their mate’s primary alignment.
  * Improved handling for alignments on different references, varied positions, strand orientation, and multiple supplementary records.
  * Preserved existing mate flags and mapping metadata behavior.

### DIFF
--- a/src/lib/commands/clip.rs
+++ b/src/lib/commands/clip.rs
@@ -13,7 +13,9 @@ use crate::metrics::writer::write_metrics as write_metrics_tsv;
 use crate::per_thread_accumulator::PerThreadAccumulator;
 use crate::reference::ReferenceReader;
 use crate::sam::SamTag;
-use crate::template::{TemplateBatch, TemplateIterator};
+use crate::template::{
+    InsertSizeEnd, TemplateBatch, TemplateIterator, compute_insert_size_from_ends,
+};
 use crate::unified_pipeline::{
     GroupKeyConfig, Grouper, MemoryEstimate, run_bam_pipeline_from_reader,
 };
@@ -853,6 +855,23 @@ struct MateSnap {
 }
 
 impl MateSnap {
+    /// Snapshots only the fields `compute_insert_size_raw` reads, skipping the `cigar_to_string`
+    /// allocation. `cigar` and `mapq` are left empty/zero, so the result must not be used to write
+    /// a mate's MC or MQ tag.
+    fn coords_of(rec: &RawRecord) -> Self {
+        use fgumi_raw_bam::flags as rflags;
+        Self {
+            ref_id: rec.ref_id(),
+            pos: rec.pos(),
+            neg: rec.flags() & rflags::REVERSE != 0,
+            unmapped: rec.flags() & rflags::UNMAPPED != 0,
+            mapq: 0,
+            cigar: String::new(),
+            aln_start: rec.alignment_start_1based(),
+            aln_end: rec.alignment_end_1based(),
+        }
+    }
+
     fn of(rec: &RawRecord) -> Self {
         use fgumi_raw_bam::flags as rflags;
         Self {
@@ -902,17 +921,21 @@ fn clear_mate_mq_mc_raw(rec: &mut RawRecord) {
 /// Computes the TLEN (inferred insert size) for the first read of a pair, mirroring
 /// htsjdk `SamPairUtil.computeInsertSize`. Returns 0 unless both reads are mapped to
 /// the same reference; the second read's TLEN is the negation of this value.
+///
+/// Only the 5'-position extraction lives here — `MateSnap` caches the alignment ends, and a
+/// read whose CIGAR yields none has no insert size. The arithmetic itself is
+/// [`compute_insert_size_from_ends`], shared with `template`'s supplementary-TLEN fix-up so
+/// the two commands cannot drift apart on ties, overflow, or the htsjdk adjustment.
 fn compute_insert_size_raw(s1: &MateSnap, s2: &MateSnap) -> i32 {
-    if s1.unmapped || s2.unmapped || s1.ref_id != s2.ref_id {
-        return 0;
-    }
     let five_prime = |s: &MateSnap| if s.neg { s.aln_end } else { s.aln_start };
     let (Some(p1), Some(p2)) = (five_prime(s1), five_prime(s2)) else {
         return 0;
     };
-    let (p1, p2) = (p1 as i64, p2 as i64);
-    let adjustment = if p2 >= p1 { 1 } else { -1 };
-    i32::try_from(p2 - p1 + adjustment).unwrap_or(0)
+    let to_i64 = |p: usize| i64::try_from(p).unwrap_or(i64::MAX);
+    compute_insert_size_from_ends(
+        InsertSizeEnd::new(s1.ref_id, s1.unmapped, to_i64(p1)),
+        InsertSizeEnd::new(s2.ref_id, s2.unmapped, to_i64(p2)),
+    )
 }
 
 /// Sets full mate-pair information on a read pair, mirroring htsjdk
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #685, opened 2026-07-31
test(consensus): pin reported consensus reads to records written

nh13:
Follow-up to #681, which fixed `duplex` reporting half the consensus reads it wrote: the caller recorded one consensus per emitted *template* while `ConsensusCallingStats::consensus_reads` counts emitted *reads*.

I checked the two sibling commands for the same defect and **neither has it** — codec emits one record per molecule and increments once per record, and the vanilla caller behind simplex records twice for an R1/R2 pair and once for a fragment. So there is no bug to fix here. What was missing is that nothing *pinned* the invariant #681 established, so it could regress in either command without a test noticing.

This adds the same rstest case table to both: run the command single-threaded and with `--threads 2`, then assert the `consensus_reads_emitted` row in `--stats` equals the records in the output BAM.

## These tests are load-bearing

A test that passes the moment it is written proves nothing on its own, so both were verified by mutation:

| Mutation | Result |
|---|---|
| codec `consensus_reads_generated += 1` → `+= 2` | codec test fails, 6 vs 3, both threading modes |
| vanilla caller's second `record_consensus()` for a pair dropped | simplex test fails, 3 vs 6, both threading modes |

The simplex case deliberately uses `create_paired_umi_family` rather than `create_umi_family`. That matters: `create_umi_family` builds *fragment* reads, so the R1/R2 branch that records twice is never reached, and with fragments the test **passed under mutation** — it would have been decoration. Worth knowing for anyone adding coverage to the consensus callers.

`cargo ci-test` (6,894 tests), `cargo ci-fmt`, and `cargo ci-lint` pass.

## Summary by CodeRabbit

* **Tests**
  * Added parameterized integration tests for codec and simplex processing.
  * Verified consensus record identities, pairing information, metadata tags, and output statistics.
  * Covered both single-threaded and two-thread execution scenarios.

### DIFF
--- a/tests/integration/helpers/assertions.rs
+++ b/tests/integration/helpers/assertions.rs
@@ -17,6 +17,42 @@ const BGZF_EOF: [u8; 28] = [
     0x1b, 0x00, 0x03, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
 ];
 
+/// Reads a required string tag off an emitted BAM record.
+///
+/// The `bam::Record` counterpart to [`assert_mi_tag`] and friends, for tests that
+/// compare a whole emitted record against an expected one rather than checking a
+/// single field. Panics rather than returning an `Option`: a missing tag is a
+/// defect in the command under test, not a case the caller should handle.
+///
+/// # Panics
+///
+/// Panics if the tag is absent, unreadable, or not a string.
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #125, opened 2026-02-26
`group`, `filter`, and possibly other subcommands produce BAM files missing BGZF EOF block


nh13:
## Summary

BAM files written by `fgumi group` and `fgumi filter` are missing the BGZF EOF marker, causing `samtools quickcheck` to fail with "missing EOF block when one should be present."

## Reproduction

Run `fgumi group` or `fgumi filter` writing to a file (not stdout) with `--compression-level 1`, then validate:

```bash
# After fgumi group writes test.grouped.bam
samtools quickcheck test.grouped.bam
# => test.grouped.bam was missing EOF block when one should be present.

# After fgumi filter writes test.cons.unmapped.bam (via pipe from fgumi duplex)
samtools quickcheck test.cons.unmapped.bam
# => test.cons.unmapped.bam was missing EOF block when one should be present.
```

Both files are otherwise readable — `samtools flagstat` works (with a warning) and downstream tools consume them correctly.

## Details

The two affected commands:

1. **`fgumi group`** — writes directly to a file with `--compression-level 1`. Output BAM is missing the EOF block.
2. **`fgumi duplex ... | fgumi filter`** — `fgumi duplex` writes to `/dev/stdout` with `--compression-level 0`, piped into `fgumi filter` which writes to a file with `--compression-level 1`. The output BAM from `filter` is missing the EOF block.

In the same pipeline, `fgumi sort` produces BAMs that pass `samtools quickcheck`, so the issue may be specific to `group` and `filter` (and possibly other subcommands).

## Expected behavior

All BAM files written to a file path should include the standard 28-byte BGZF EOF marker block upon close, per the SAM/BAM spec (§4.1.2).

## Environment

- fgumi 0.1.0

PR #689, opened 2026-08-01
fix(sort, fastq, extract): stream `-o -` to stdout instead of creating a file named `-`

### LINKED ISSUE #688: bug(sort, fastq, extract): `-o -` writes a file named `-` instead of streaming to stdout
## Summary

`fgumi sort -o -` does not write to stdout — it silently creates a regular file named `-` in the current directory. The same is true of `fgumi fastq -o -`, and of `fgumi extract -o -` but *only* when `--threads` selects the parallel pipeline. Every other BAM-writing command handles `-` correctly, so the CLI is inconsistent about its own output convention, and nothing in the test suite would notice.

## Reproducing

```console
$ printf '@HD\tVN:1.6\tSO:unsorted\n@SQ\tSN:chr1\tLN:1000\nr2\t0\tchr1\t200\t60\t4M\t*\t0\t0\tACGT\tIIII\nr1\t0\tchr1\t100\t60\t4M\t*\t0\t0\tACGT\tIIII\n' > in.sam
$ fgumi sort -i in.sam -o - --order coordinate > out.bam
$ ls -la
-rw-r--r--  333  -          # the BAM landed here
-rw-r--r--    0  out.bam    # stdout got nothing
```

Exit code is 0 and the summary line reports `Output: -`, so the run looks successful. In a pipeline (`fgumi sort -i in.bam -o - | fgumi group -i - ...`) the downstream command sees an empty stream, and a stray `-` file is left in the working directory.

## Current state, by command

Measured against `main` @ 3d270193 (0.5.0) with the release binary:

| Command | `-o -` | Notes |
| --- | --- | --- |
| `group`, `dedup`, `clip`, `correct`, `merge`, `simplex`, `duplex`, `codec`, `filter`, `downsample` | works | writer comes from `create_bam_writer` / `create_raw_bam_writer` |
| `extract` (default, single-threaded) | works | `create_raw_bam_writer` |
| `extract --threads N` | **writes `./-`** | parallel pipeline path only |
| `sort` | **writes `./-`** | every path: in-memory, spill+merge, and `--write-index` |
| `fastq` | **writes `./-`** | has a stdout mode, but only by *omitting* `-o` |

## Cause

`fgumi-bam-io` already has the plumbing: `is_stdout_path` (`crates/fgumi-bam-io/src/paths.rs:34`) recognizes both `-` and `/dev/stdout`, and `open_output_writer` (`crates/fgumi-bam-io/src/writer.rs:1167`) dispatches to `std::io::stdout()`. `create_bam_writer` and `create_raw_bam_writer` both go through it, which is why most commands are fine.

The three broken paths bypass it with a bare `std::fs::File::create`:

- `PooledBamWriter::new_inner` — `crates/fgumi-sort/src/pooled_bam_writer.rs:103`. This is *all* of `sort`'s output: the in-memory no-spill path (`external.rs:2611`) and the k-way merge path (`external.rs:3587`) both use it.
- `Extract::execute`, parallel branch — `src/lib/commands/extract.rs:1632`.
- `Fastq::execute`, both the plain and `.gz` branches — `src/lib/commands/fastq.rs:330` and `:338`.

Worth noting that `crates/fgumi-sort` *does* have a correct implementation — `MergeOutputTarget` (`external.rs:299`) checks `is_stdout_path` and streams properly. It is only reachable from `merge_bams`, i.e. the `fgumi merge` command, not from `sort`'s own `merge_chunks_generic`. That is why `merge -o -` works and `sort -o -` does not.

`--write-index` is a separate, correct case: `create_indexing_bam_writer` bails with "indexing requires a seekable file" (`writer.rs:977`). Whatever `sort` does for `-` should keep that rejection.

## Why no test caught it

`tests/integration/test_input_source_matrix.rs` is the contract test for exactly this class of bug. Its `CONTRACTS` table has an entry per advertised subcommand and `every_command_declares_an_input_contract` fails when a new command is added without one — so it is exhaustive and enforced. But all three of its axes are input-side: `sam` (`-i in.sam`), `stdin` (`-i -`), and `output_depends_on_input` (the anti-vacuity guard). There is no output axis.

Sweeping `tests/`, `scripts/`, and `docs/` for `-o -` / `--output -` returns nothing: no test passes `-` as an output path, and no test reads a BAM back from a child process's stdout. The stdout helpers are unit-tested only at the predicate level (`paths.rs:57`, `writer.rs:1360`), never end to end through a command. `zipper` is the sharpest illustration — it *defaults* `--output` to `-` (`src/lib/commands/zipper.rs:148`), its stdin default is explicitly covered (`test_zipper_command.rs:402`), and its stdout default never is, because every zipper test passes an explicit output path.

## Proposed fix

1. Route the three offending writers through `open_output_writer` (or the existing `create_*_bam_writer` helpers) so `-` and `/dev/stdout` stream, keeping the `--write-index` + stdout rejection.
2. Add a `stdout` axis to `CONTRACTS` in `test_input_source_matrix.rs`, mirroring the `stdin` axis: every command declares `Required` or `NotApplicable(reason)`, and the new test runs the command with `-o -`, captures the child's stdout, and compares it against the file-output oracle the harness already builds. Commands whose `-o` is a multi-file prefix (the metrics commands) declare `NotApplicable`.

(2) is the part that matters long term — it is what stops the next command from inheriting whatever its writer happens to do, which is precisely what the input-side matrix already does for `-i -`.

Open question for `fastq`: it currently means stdout by *omitting* `-o`. Accepting `-` as well is strictly additive and makes it consistent with the rest of the CLI, so the plan is to support both rather than deprecate the omission.

nh13:
Fixes #688.

## The bug

`fgumi sort -o -` created a regular file called `-` in the working directory, wrote the BAM into it, and exited zero — so the pipe was empty and a stray `-` file was left behind. `fgumi fastq -o -` did the same. `fgumi extract -o -` did it too, but only when `--threads` selected the parallel pipeline, so where the output landed depended on a flag that has nothing to do with where the output goes.

```console
$ fgumi sort -i in.sam -o - --order coordinate > out.bam
$ ls -la
-rw-r--r--  333  -          # the BAM landed here
-rw-r--r--    0  out.bam    # stdout got nothing
```

Every other BAM-writing command was already correct.

## The fix

`fgumi-bam-io` already had the dispatch — `is_stdout_path` recognizes `-` and `/dev/stdout`, and `open_output_writer` returns stdout for either. The commands that work reach it through `create_bam_writer` / `create_raw_bam_writer`; the three broken ones called `File::create` directly. They now go through `open_output_writer`, which is promoted from `pub(crate)` to `pub` and documented as the place the convention is honoured for BAM output. (`fastq` emits text, so it dispatches on `is_stdout_path` itself and takes `stdout().lock()` — no boxed writer needed.)

For `sort` that meant threading the choice through `PooledBamWriter`, which all three of its write paths funnel into (in-memory, k-way merge, and the indexing variant), so `io_writer_loop` is now generic over its sink instead of fixed to `File`. Spill chunks stay monomorphized on `File`, so the hot path is untouched.

`sort --write-index` combined with stdout is now rejected up front instead of attempted. A BAI needs a seekable file and a sidecar path, and a pipe has neither; `create_indexing_bam_writer` already refused it, and the check just moves earlier so the run fails before streaming a BAM it could never index.

## A second bug, found while reviewing this one

Making `-o -` a supported, contract-tested spelling exposed an adjacent hazard. `-` is honoured by *every* BAM writer, including secondary outputs, so a command given both `--output -` and `--rejects -` opened two writers on fd 1 and interleaved two BAMs — two headers, two EOF blocks, one stream — then exited zero:

```console
$ fgumi simplex -i g.bam -o - --rejects - --min-reads 1 > both.bam
$ # 678 bytes, 4 BGZF members, 2 EOF markers
$ echo $?
0
```

Nothing downstream can read that and nothing reported it — the same silent-success shape this PR exists to remove, one flag over. Pre-existing (`create_optional_bam_writer` has always resolved `-`), but this PR is what turns it from an undocumented accident into something users will reach for.

`reject_colliding_stdout_outputs` now bails when both paths name stdout, wired in immediately after input validation in all six commands that take a secondary BAM output: `simplex`, `duplex`, `codec`, `correct`, `filter`, `downsample`. Two identical *file* paths are deliberately left alone — that is last-writer-wins on disk, a different failure with a different fix.

## Why nothing caught the original

`tests/integration/test_input_source_matrix.rs` is the contract test for this class of bug — a table with an entry per advertised subcommand, enforced so a new command cannot be added without declaring a stance. But all three of its axes were input-side (`sam`, `stdin`, and the `output_depends_on_input` anti-vacuity guard). Nothing in `tests/`, `scripts/`, or `docs/` passed `-` as an *output* path, and no test read a BAM back from a child's stdout. `zipper` is the sharpest illustration: it *defaults* `--output` to `-`, its stdin default is explicitly covered, and its stdout default never was.

So the matrix grows a fourth axis, `stdout`. Every command declares `Required` or `NotApplicable` with a reason (the metrics commands and `review` take an output *prefix*, which no pipe can carry). The check runs the command with `-o -` in a scratch working directory, fails if a file named `-` appears there, and compares what arrived on the pipe against the same named-path oracle the other axes use — so a command that streams *something* but drops or reorders records still fails. The type and its test are renamed `IoContract` / `every_command_declares_an_io_contract` to match what they now cover.

`tests/integration/test_streaming_output.rs` is new, and covers what the matrix cannot reach with a command's default arguments:

- `sort` through each write path — in-memory, spill+merge, and spill+merge threaded — in both spellings (`-` and `/dev/stdout`), asserting the streamed bytes are the same BAM the file run produced. A companion test asserts the spilling fixture really does spill, so those cases cannot silently decay into duplicates of the in-memory one.
- `sort --write-index` rejection, asserting the error names *both* the flag and the stdout conflict — naming only the flag would also match the unrelated `--write-index cannot be used with --verify` error from the same function — and that no `-` or `-.bai` is left behind.
- The `--output`/`--rejects` collision, one case per affected command. The guard runs before any record is read, so one throwaway BAM drives all six regardless of the tags each would otherwise need.
- `extract` with `--threads` and with the flag omitted. The omission is deliberate: threading mode is `SingleThreaded` only when the flag is absent — `--threads 1` is `Threads(1)` and takes the parallel path — so passing `1` would have exercised the pipeline writer twice and never the raw one.
- `fastq` across `-o -`, `-o /dev/stdout`, and omitting `--output`.

Every streamed BAM is also checked for the trailing BGZF EOF block, via a new byte-slice `assert_bytes_have_bgzf_eof` beside the existing path-based helper. Records parsing back is not sufficient: noodles reads a stream missing the marker, `samtools quickcheck` rejects it, and issue #125 was commands emitting exactly that. `test_bgzf_eof.rs` covers group, dedup, simplex, duplex, codec, filter, clip, and correct — but not sort, extract, or fastq, the three commands this PR reroutes to a new sink. The marker is emitted today, so this pins an invariant rather than fixing a break.

## Verification

- Full suite: 6910 passed, 0 failed (`cargo ci-test`). `cargo ci-lint`, `cargo ci-fmt`, and `cargo ci-doc` clean.
- Each new test was confirmed to fail before its fix and pass after — the matrix axis named exactly `sort` and `fastq`, the targeted tests added `extract --threads`, and the collision test was written against the reproducer above.
- No throughput cost to routing the sort's output through `Stdout`: a 2M-record sort writing 116 MB measured 0.58–0.63 s to a file and 0.58–0.61 s to a pipe over three reps. The two outputs are record-identical (2,000,000 records, same `samtools view` body md5); they differ only in the `@PG` `CL` line, which correctly records the different `-o` argument.

## Also here

The two high-throughput pipelines in `docs/src/guide/best-practices.md` pipe `fgumi group` into `fgumi simplex` and `fgumi filter` into `fgumi sort` without giving either producer an `--output`. Both require one, so those examples fail with a clap usage error as written. Fixed in a separate commit by adding `--output -`, which is what the surrounding prose intends — and which now behaves.

## Reading order

1. `5254aaf3` — the original fix and the contract axis. Start at `crates/fgumi-sort/src/pooled_bam_writer.rs` and `tests/integration/test_input_source_matrix.rs`.
2. `8980d267` — the docs example fix.
3. `2b5b2e97` — the `--output`/`--rejects` collision guard.
4. `ea14bae1` — self-review fixes: BGZF EOF assertions, the tightened rejection assertions, the `IoContract` rename, and help-text corrections. Its regression test for (3) lives here rather than in that commit, since it shares a file with the EOF changes.

## Not in scope

No command's `--output` help text mentions that `-` means stdout, but neither does any `--input` help mention stdin, so documenting the convention is a consistent change across the whole CLI rather than something to bolt onto the commands touched here. `fastq`'s `--output` and `sort`'s `--write-index` are updated because this PR changed what they accept; the general sweep is worth a follow-up.

Risk: Output changes affect destination and streaming only, pinned by existing file output and integration comparisons; `unsafe` changes are none and the `CLAUDE.md` allowlist is unchanged; buffering adds 256 KiB for threaded `extract` and 64 MiB for `fastq`, with no queue, thread, or backpressure policy change.

## Summary

- Fixed `sort`, threaded `extract`, and `fastq` output for `-` and `/dev/stdout`.
- Added buffered stdout writers and BGZF EOF validation.
- Rejected `sort --write-index` for non-seekable stdout.
- Prevented primary and rejects outputs from targeting the same stdout or regular file.
- Expanded I/O contract and streaming tests.
- Updated pipeline examples to pass `--output -`.

### DIFF
--- a/crates/fgumi-bam-io/src/lib.rs
+++ b/crates/fgumi-bam-io/src/lib.rs
@@ -39,5 +39,6 @@ pub use reorder::{DrainReady, ReorderBuffer};
 pub use writer::{
     BaiBuilder, BamWriter, BgzfWriterEnum, IndexingBamWriter, RawBamWriter, bai_sidecar_path,
     create_bam_writer, create_indexing_bam_writer, create_optional_bam_writer,
-    create_raw_bam_writer, write_bai_index, write_bai_sidecar, write_bam_header,
+    create_raw_bam_writer, open_output_writer, write_bai_index, write_bai_sidecar,
+    write_bam_header,
 };
--- a/crates/fgumi-bam-io/src/writer.rs
+++ b/crates/fgumi-bam-io/src/writer.rs
@@ -1165,12 +1165,16 @@ pub fn create_optional_bam_writer<P: AsRef<Path>>(
 
 /// Open an output writer for a path, supporting stdout via `-` or `/dev/stdout`.
 ///
-/// This is the single place the `-` convention is honoured, so anything writing
-/// a command's output must come through here (directly, or via one of the
+/// This is the single place the `-` convention is honoured **for BAM output**,
+/// so any BAM writer must come through here (directly, or via one of the
 /// `create_*_bam_writer` helpers) rather than calling `File::create` itself. A
 /// writer that opens the path directly silently creates a regular file *named*
 /// `-` and exits zero with an empty pipe — see the stdout axis in
 /// `tests/integration/test_input_source_matrix.rs`.
+///
+/// Commands whose output is text rather than BAM (`fgumi fastq`) do not need the
+/// boxed writer and dispatch on [`is_stdout_path`] themselves, taking
+/// `stdout().lock()` directly.
 ///
 /// A stdout path yields a block-buffered handle — fd 1 duplicated into a
 /// [`File`] — rather than [`std::io::Stdout`], whose `LineWriter` would tear
--- a/crates/fgumi-bam-io/src/writer.rs
+++ b/crates/fgumi-bam-io/src/writer.rs
@@ -1164,15 +1164,63 @@ pub fn create_optional_bam_writer<P: AsRef<Path>>(
 }
 
 /// Open an output writer for a path, supporting stdout via `-` or `/dev/stdout`.
-pub(crate) fn open_output_writer<P: AsRef<Path>>(path: P) -> Result<Box<dyn Write + Send>> {
+///
+/// This is the single place the `-` convention is honoured, so anything writing
+/// a command's output must come through here (directly, or via one of the
+/// `create_*_bam_writer` helpers) rather than calling `File::create` itself. A
+/// writer that opens the path directly silently creates a regular file *named*
+/// `-` and exits zero with an empty pipe — see the stdout axis in
+/// `tests/integration/test_input_source_matrix.rs`.
+///
+/// A stdout path yields a block-buffered handle — fd 1 duplicated into a
+/// [`File`] — rather than [`std::io::Stdout`], whose `LineWriter` would tear
+/// every BGZF flush at each `0x0a`.
+///
+/// # Errors
+///
+/// Returns an error if `path` names a file that cannot be created, or if the
+/// stdout descriptor cannot be duplicated.
+pub fn open_output_writer<P: AsRef<Path>>(path: P) -> Result<Box<dyn Write + Send>> {
     let path_ref = path.as_ref();
     if is_stdout_path(path_ref) {
-        Ok(Box::new(std::io::stdout()))
+        block_buffered_stdout()
     } else {
         let file = File::create(path_ref)
             .with_context(|| format!("Failed to create output BAM: {}", path_ref.display()))?;
         Ok(Box::new(file))
     }
+}
+
+/// A block-buffered handle on this process's stdout.
+///
+/// [`std::io::Stdout`] wraps a `LineWriter`, which splits every write at the last
+/// `\n` in the buffer and issues the two halves separately. BGZF blocks carry
+/// `0x0a` at arbitrary offsets, so a single large flush from the caller's
+/// `BufWriter` would be torn into many small `write` calls on the hot output
+/// path. Duplicating the descriptor into a [`File`] bypasses the line buffering
+/// and leaves batching to the caller's buffer, where it belongs.
+///
+/// The descriptor is duplicated rather than taken, so dropping the returned
+/// handle closes only the duplicate; fd 1 stays open for the rest of the process.
+#[cfg(unix)]
+fn block_buffered_stdout() -> Result<Box<dyn Write + Send>> {
+    use std::os::fd::AsFd;
+
+    let stdout = std::io::stdout()
+        .as_fd()
+        .try_clone_to_owned()
+        .map(File::from)
+        .context("Failed to duplicate stdout for BAM output")?;
+    Ok(Box::new(stdout))
+}
+
+/// A block-buffered handle on this process's stdout.
+///
+/// Fallback for targets without file descriptors, where the line-buffering
+/// concern described on the Unix variant cannot be worked around this way.
+#[cfg(not(unix))]
+fn block_buffered_stdout() -> Result<Box<dyn Write + Send>> {
+    Ok(Box::new(std::io::stdout()))
 }
 
 #[cfg(test)]
--- a/crates/fgumi-sort/src/bgzf_io.rs
+++ b/crates/fgumi-sort/src/bgzf_io.rs
@@ -174,8 +174,8 @@ impl StagingBuffer {
 /// `compressed_start` is captured *before* the write, so it is the on-disk byte
 /// offset at which this block begins. The BGZF EOF marker is written separately
 /// and is intentionally never passed here (no record references it).
-fn write_block_in_order(
-    writer: &mut BufWriter<std::fs::File>,
+fn write_block_in_order<W: Write>(
+    writer: &mut BufWriter<W>,
     serial: u64,
     data: &[u8],
     compressed_offset: &mut u64,
@@ -196,10 +196,14 @@ fn write_block_in_order(
 ///
 /// Uses a `BTreeMap` as a reorder buffer. When the next expected serial arrives,
 /// writes it immediately. Out-of-order blocks are buffered until their turn.
-/// Releases one permit to `permit_pool` after each block is written to disk,
+/// Releases one permit to `permit_pool` after each block is written out,
 /// unblocking the corresponding `StagingBuffer::flush()` call and bounding the
 /// number of in-flight compressed blocks to the pool capacity.
 /// Writes BGZF EOF marker and flushes when all blocks are received.
+///
+/// Generic over the sink rather than fixed to `File`: spill chunks are always
+/// files, but the sort's *output* may be stdout, which reaches here as the
+/// boxed writer `open_output_writer` hands back.
 ///
 /// When `block_offset_tx` is `Some`, each written block's `(serial,
 /// compressed_start)` is emitted on it (in strict block order) for BAI virtual
@@ -207,11 +211,11 @@ fn write_block_in_order(
 ///
 /// # Errors
 ///
-/// Returns an error if any disk write fails or if a compressed block is missing
+/// Returns an error if any write fails or if a compressed block is missing
 /// (which would silently truncate the output).
 #[allow(clippy::needless_pass_by_value)]
-pub(crate) fn io_writer_loop(
-    mut writer: BufWriter<std::fs::File>,
+pub(crate) fn io_writer_loop<W: Write>(
+    mut writer: BufWriter<W>,
     result_rx: Receiver<CompressResult>,
     buffer_pool: BufferPool,
     permit_pool: Arc<PermitPool>,
@@ -233,8 +237,8 @@ pub(crate) fn io_writer_loop(
     result
 }
 
-fn io_writer_loop_inner(
-    writer: &mut BufWriter<std::fs::File>,
+fn io_writer_loop_inner<W: Write>(
+    writer: &mut BufWriter<W>,
     result_rx: &Receiver<CompressResult>,
     buffer_pool: &BufferPool,
     permit_pool: &Arc<PermitPool>,
--- a/crates/fgumi-sort/src/pooled_bam_writer.rs
+++ b/crates/fgumi-sort/src/pooled_bam_writer.rs
@@ -25,7 +25,7 @@ use crate::codec::SpillCodec;
 use crate::worker_pool::{CompressResult, CompressTarget, PermitPool, SortWorkerPool};
 use anyhow::Result;
 use crossbeam_channel::{Receiver, bounded, unbounded};
-use fgumi_bam_io::BaiBuilder;
+use fgumi_bam_io::{BaiBuilder, is_stdout_path, open_output_writer};
 use fgumi_bgzf::BGZF_MAX_BLOCK_SIZE;
 use noodles::bam::bai;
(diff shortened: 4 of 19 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
Addressed the four outside-diff-range findings from the last review; they have no inline thread to resolve, so noting them here.

**`crates/fgumi-bam-io/src/writer.rs` — `std::io::stdout()` is line-buffered.** Fixed. `open_output_writer` now goes through a new `block_buffered_stdout()`, which duplicates fd 1 into a `File` (`stdout().as_fd().try_clone_to_owned().map(File::from)`) so `LineWriter` no longer tears each BGZF flush at every `0x0a`. The descriptor is duplicated rather than taken, so dropping the handle leaves fd 1 open. `#[cfg(not(unix))]` keeps the old behaviour where there are no file descriptors. The `# Errors` section, which previously claimed the stdout case could not fail, was updated.

**`crates/fgumi-sort/src/pooled_bam_writer.rs` — `new_indexing` accepts a pipe.** Fixed. `new_indexing` now rejects a stdout path up front with an error naming both the index and the path, so the invariant travels with the writer that needs it rather than living only in `src/lib/commands/sort.rs`. Covered by a new `#[rstest]` over both spellings `is_stdout_path` honours — a guard that caught only `-` would still let `/dev/stdout` produce a `/dev/stdout.bai`.

**`crates/fgumi-sort/src/pooled_bam_writer.rs` — docs say "output file".** Fixed. `new` now documents that it opens an output *sink* via `open_output_writer` and that a non-seekable sink is accepted; `new_indexing` documents the opposite requirement and where the command-layer rejection lives.

**`tests/integration/test_input_source_matrix.rs` — failure messages describe the old two-axis contract.** Fixed. Both messages now say "I/O contract" and name all four axes (`sam`, `stdin`, `stdout`, `output_depends_on_input`); the test doc comment was updated to match.

While in that file I also took the inline suggestion to carry the spelling in `OutputSpec::Stdout(&'static str)` and run the matrix over both `-` and `/dev/stdout` for every command declaring `stdout: Required`. Verified load-bearing by mutation: dropping `/dev/stdout` from `is_stdout_path` fails `declared_stdout_support_holds`, where before it passed.

nh13:
@coderabbitai review

nh13:
Addressed all four findings from the last review, plus the failing `docs` job and a rebase onto `main`. One of the four is worth a note because the reasoning behind it turned out to be wrong, and one because I had argued against it and was wrong.

**`src/lib/commands/common.rs` — reject regular-file output collisions.** Fixed, and the commit message that declined this was factually wrong. It claimed identical file paths were "last-writer-wins on disk". They are not — two `File::create` calls give two file *descriptions* with independent offsets, both starting at zero, so the two streams overwrite each other byte for byte:

```
$ fgumi correct -i in.bam -o same.bam --rejects same.bam -d 1 -U umis.txt
  exit=0    "Read 400; kept 400 and rejected 0"
$ samtools view -c same.bam
  [E::bgzf_read_block] Invalid BGZF header at offset 392   → exit 1
$ samtools quickcheck -v same.bam                          → exit 8
```

Note it rejected nothing. The secondary writer is created eagerly (`create_optional_bam_writer`) and emits its header at offset 0 whether or not anything is ever written to it, so the corruption needs no rejected reads. This is the *worse* half of the pair the guard covers: on stdout the two writers share one file description, so every byte survives in an unparseable order; on disk bytes are destroyed.

`reject_colliding_stdout_outputs` is now `reject_colliding_outputs` and catches both. File identity is approximated deliberately — paths compare as `(canonicalized parent, file name)`, since `canonicalize` is `ENOENT` on a path that does not exist yet, which is the usual case for an output. Three mitigations keep the approximation from inventing failures: a bare file name canonicalizes against `.` rather than the empty parent `canonicalize` rejects; a parent that cannot be canonicalized falls back to the raw path rather than raising, so a run whose output directory is missing still gets the writer's message about that; and a destination that exists and is not a regular file is exempt, so `-o /dev/null --rejects /dev/null` — a legitimate compute-only run — keeps working. What the approximation still misses (symlinks, hard links, case-insensitive filesystems) is #715, along with `--stats`/`--metrics`, which have this same collision unguarded and worse: `simplex.rs:496` writes the TSV *after* `:481` finishes the BAM, so `-o out.bam --stats out.bam` truncates a completed BAM.

**`test_streaming_output.rs:246` — assert the rejected `--write-index` run wrote no bytes.** Fixed. The stray-`-` check cannot see a `/dev/stdout` run that opened its writer before failing, since that one reaches fd 1 without creating a file; an empty pipe is what distinguishes rejected-up-front from wrote-then-failed.

**`test_streaming_output.rs:310` — cover mixed stdout spellings.** Fixed; `output` and `rejects` now vary independently over both spellings, so all four pairs run against all six commands. Verified load-bearing by mutation: replacing the guard's condition with a literal path comparison fails all 12 mixed cases while the 12 matching ones still pass.

**`test_streaming_output.rs:373` — test `/dev/stdout` through the threaded extract writer.** Applied, but the stated rationale does not hold and I would rather say so than leave a test carrying a false comment. I mutated the threaded path to recognize only `-` and fall back to `File::create` for everything else — all four cases still passed, because on Unix `File::create("/dev/stdout")` reaches fd 1 on its own. There is no observable difference for a writer to get wrong there. The cases are kept as contract coverage of both advertised spellings, and the doc comment now says that rather than claiming they discriminate a `-`-only path. `-` is the spelling that has no meaning to `File::create` and so is the one the dispatch exists for.

Also in this push: the `docs` job was failing because `open_output_writer`'s public doc linked the private `block_buffered_stdout` (`RUSTDOCFLAGS=-D warnings`); the rationale is now stated inline instead of linked. Rebased onto `main` (one conflict in `tests/integration/helpers/assertions.rs`, both sides' new helpers kept).

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #690, opened 2026-08-01
docs(sort): name the accepted values for sort's boolean flags

tfenne:
`--memory-per-thread` rendered as `--memory-per-thread [<MEMORY_PER_THREAD>]`, because clap derives the value placeholder from the field name when no `value_name` is set. That reads as a request for a per-thread memory size and invites `--memory-per-thread 10G`. Short `-h` compounds it, showing only the first doc line — which never says the flag is a toggle.

Adds `value_name = "true|false"` to the three optional-value booleans on `Sort`:

```
--verify [<true|false>]
--memory-per-thread [<true|false>]
--write-index [<true|false>]
```

Help text only — parsing is unchanged, and the existing error already enumerates the accepted values. The test walks `Sort::command()` and asserts every optional-value flag names its values.

`cargo ci-fmt`/`ci-lint` clean; `ci-test` is 6890 passed / 3 failed, and those 3 (`test_input_source_matrix::declared_{stdin,sam}_support_*`) fail identically on a pristine `main` — they diff R-generated `simplex_qc.pdf` byte-for-byte and R stamps `/CreationDate` into it. Unrelated, but probably worth its own issue.

### Follow-up: should we prefer `BoolishValueParser`?

We just used clap's [`BoolishValueParser`](https://docs.rs/clap/latest/clap/builder/struct.BoolishValueParser.html) in dupblaster, and it may be the better default here than our own `parse_bool`. Worth a separate PR, with two differences to weigh first:

- **It accepts more:** `on|off|1|0` on top of our `true|false|yes|no|y|n|t|f`. Strictly backward compatible, but it diverges from sopt — `OptionLookup.convertFlagValue` accepts exactly our current set, which is what `parse_bool`'s "matches sopt/fgbio" comment refers to.
- **Its error is worse:** `value was not a boolean`, versus our `Invalid boolean value '10G'. Expected: true|false|yes|no|y|n|t|f`. It also needs `hide_possible_values`, or clap prints a misleading `[possible values: true, false]`.

So there's a third option: extend `parse_bool` with the boolish literals and keep the better error message. Either way `value_name` becomes `on|off`, so that decision and this one travel together.

## Summary by CodeRabbit

* **Improvements**
  * Boolean command-line options now clearly display their accepted `true` or `false` values.
  * Added validation coverage to ensure optional boolean arguments present their value choices consistently.

### DIFF
--- a/src/lib/commands/sort.rs
+++ b/src/lib/commands/sort.rs
@@ -211,7 +211,7 @@ pub struct Sort {
     /// Reads records sequentially and checks that each record's sort key
     /// is >= the previous record's key. Exits 0 if sorted correctly,
     /// non-zero if any records are out of order.
-    #[arg(long = "verify", default_value = "false", num_args = 0..=1, default_missing_value = "true", action = clap::ArgAction::Set, value_parser = parse_bool)]
+    #[arg(long = "verify", value_name = "true|false", default_value = "false", num_args = 0..=1, default_missing_value = "true", action = clap::ArgAction::Set, value_parser = parse_bool)]
     pub verify: bool,
 
     /// Sort order.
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #691, opened 2026-08-01
fix(sort): report the thread counts each phase actually uses

tfenne:
`--sort-threads` and `--merge-threads` each default to `--threads`, but the config log and the phase-timing summary both printed `--threads` verbatim. So a run given only the per-phase overrides reported one thread while running eight sort workers and sixteen merge workers:

```
$ fgumi sort -i in.bam -o out.bam --sort-threads 8 --merge-threads 16
INFO  Threads: 1
INFO    Threads: 1
```

Now:

```
INFO  Threads: sort 8, merge 16
INFO    Threads: sort 8, merge 16
```

It collapses to a single number when both phases agree, so the common case stays terse. The command layer reads the counts off the built sorter (`phase1_threads`/`phase2_threads`) rather than re-deriving the fallback, so the log can't drift from what the engine schedules — `build_sorter` moves above the config log to allow that, and is pure construction. The engine's debug config dump now labels its `--threads` value as the default the phases fall back to.

Logging only; scheduling and output are unchanged. Three integration tests cover the overrides-without-`--threads` case, an override below `--threads`, and the collapsed single-count case; I checked all three fail before the fix (the first reporting `left: "1"`).

`ci-fmt`/`ci-lint` clean. `ci-test` is 6892 passed / 3 failed, the 3 being `test_input_source_matrix::declared_{sam,stdin}_support_*`, which fail identically on a pristine `main` — they diff R-generated `simplex_qc.pdf` byte-for-byte and R stamps `/CreationDate` into it. Same unrelated failures noted in #690.

### Follow-up: the memory budget ignores the per-phase counts

Worth a separate look, since this PR makes it visible rather than introducing it. `resolve_memory_budget` scales by `--threads` only, so the same invocation now logs:

```
INFO  Max memory: 732.4 MiB (732.4 MiB/thread x 1 threads)
INFO  Threads: sort 8, merge 16
```

That is truthful about what was computed, but a run asking for 8 sort threads gets the 1-thread budget. Under the intended usage (`-@ 32 --sort-threads 4`, lowering a phase below `--threads`) it doesn't bite, since `--threads` is set. Whether the budget should follow the phase counts is a behavior change, so I left it out here.

### DIFF
--- a/crates/fgumi-sort/src/external.rs
+++ b/crates/fgumi-sort/src/external.rs
@@ -92,6 +92,21 @@ fn should_spill_chunk(memory_used: usize, memory_limit: usize, chunk_records: us
     memory_used >= memory_limit || chunk_records >= MAX_CHUNK_RECORDS
 }
 
+/// Render the effective per-phase worker counts for a log line.
+///
+/// Collapses to a single number when the phases agree (the common case, where
+/// neither `--sort-threads` nor `--merge-threads` was overridden) and spells
+/// both out otherwise. Shared by the engine's end-of-run summary and the `sort`
+/// command's config dump so the two emitters cannot drift.
+#[must_use]
+pub fn format_thread_counts(sort_threads: usize, merge_threads: usize) -> String {
+    if sort_threads == merge_threads {
+        format!("{sort_threads}")
+    } else {
+        format!("sort {sort_threads}, merge {merge_threads}")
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai pause

nh13 (approved):

PR #692, opened 2026-08-01
fix(sort): scale the per-thread memory budget by the sort phase

tfenne:
**Stacked on #691** — please review/merge that one first; this PR targets its branch, so the diff here is just the memory change.

`--memory-per-thread` multiplied `--max-memory` by `--threads` alone. Since `--sort-threads` defaults to `--threads` and not the reverse, a run that set only the per-phase override got the one-thread budget while sorting with eight workers:

```
$ fgumi sort -i in.bam -o out.bam --max-memory 100M --sort-threads 8
INFO  Max memory: 95.4 MiB (95.4 MiB/thread x 1 threads)
INFO  Threads: sort 8, merge 1
```

The budget sizes the in-memory accumulation buffer, which the sort phase fills, so it now scales by that phase:

```
INFO  Max memory: 762.9 MiB (95.4 MiB/thread x 8 threads)
```

### Why `max(--threads, --sort-threads)` and not just `--sort-threads`

Scaling strictly by the sort phase would be a silent regression for the pattern `--sort-threads` exists to serve. `-@ 32 --sort-threads 4` is the documented way to cede cores to an upstream producer while keeping the merge wide; under sort-phase-only scaling that run's buffer drops from 32× to 4× — an eightfold cut, trading a scheduling hint for a throughput cliff. Taking the larger of the two raises the budget exactly where `--threads` under-counts and never lowers it, so no existing invocation resolves less memory than it does today.

### Scope

- `--threads 0` still reaches `resolve_memory_budget`'s rejection rather than being clamped up by the new helper.
- `--memory-per-thread false` is unaffected (still a fixed total).
- The `Max memory` log line and the `--max-memory auto` initial-capacity cap both follow the same count, so neither can disagree with the resolved budget.
- Only `fgumi sort` is touched. `QueueMemoryOptions` (filter/clip/correct) has a single thread count with no sort/merge phases, so it needs no equivalent — and the README/performance-tuning memory notes describe that pipeline-queue path, not this one.

The helper re-derives the sort-phase fallback because the budget is needed before the sorter exists (so it can't read `phase1_threads` off it, as the `Threads:` line in #691 does). A test asserts the two definitions agree, so they can't drift.

Tests: 7 cases over the multiplier (including both zero cases), 4 drift-guard cases, and two end-to-end tests asserting the logged budget — one for the bug, one pinning the `-@ 32 --sort-threads 4` shape against the regression above. I confirmed the end-to-end tests fail before the fix, at `95.4 MiB (... x 1 threads)`.

`ci-fmt`/`ci-lint` clean. `ci-test` is 6905 passed / 3 failed — the same `test_input_source_matrix::declared_{sam,stdin}_support_*` trio that fails on a pristine `main` (they diff R-generated `simplex_qc.pdf` byte-for-byte and R stamps `/CreationDate` into it), as noted in #690 and #691.

### DIFF
--- a/src/lib/commands/sort.rs
+++ b/src/lib/commands/sort.rs
@@ -264,8 +264,10 @@ pub struct Sort {
 
     /// Scale memory limit by thread count (samtools behavior).
     ///
-    /// When enabled (default), --max-memory specifies memory per thread.
-    /// Total memory = `max_memory` × threads. Disable for fixed total memory.
+    /// When enabled (default), --max-memory specifies memory per thread. Total
+    /// memory = `max_memory` × the larger of --threads and --sort-threads, since
+    /// the sort phase is what fills the in-memory buffer. Disable for fixed total
+    /// memory.
     #[arg(long = "memory-per-thread", value_name = "true|false", default_value = "true", num_args = 0..=1, default_missing_value = "true", action = clap::ArgAction::Set, value_parser = parse_bool)]
     pub memory_per_thread: bool,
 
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

nh13 (approved):

PR #693, opened 2026-08-01
test(simd-fastq): compare long-record bytes, not just their length

nh13:
Two long-record tests in `fgumi-simd-fastq` assert only that the parsed fields have the expected length, so neither can fail on the class of defect it exists to catch.

`reader.rs::test_reader_long_records` feeds a 500-byte record through a reader with a 256-byte capacity — the point being that the record can only be produced by refilling the buffer mid-record — and then checks `rec.sequence.len() == 500` and `rec.quality.len() == 500`. A refill that dropped, duplicated, or reordered a chunk still yields 500 bytes.

`parser.rs::test_long_sequence_spanning_multiple_blocks` has the same shape one level down: a 200-base sequence spanning four 64-byte lexer blocks, asserting `sequence.len() == 200` and never looking at the quality line at all.

Both built their input from `"A".repeat(n)` / `"I".repeat(n)`, so even a byte-for-byte comparison would have been weak — every position holds the same value. Both now generate position-varying bases and qualities and compare the bytes the parser returns against the bytes that went in, which is what makes a misplaced boundary observable.

Test-only; no production code is touched.

## Summary by CodeRabbit

* **Tests**
  * Strengthened validation for long sequencing records by checking exact sequence and quality content.
  * Improved detection of missing, duplicated, or reordered data when processing records across buffer refills.

### DIFF
--- a/crates/fgumi-simd-fastq/src/parser.rs
+++ b/crates/fgumi-simd-fastq/src/parser.rs
@@ -444,16 +444,20 @@ mod tests {
 
     #[test]
     fn test_long_sequence_spanning_multiple_blocks() {
-        // 200-base sequence spans 4 blocks
-        let seq = "A".repeat(200);
-        let qual = "I".repeat(200);
+        // 200-base sequence spans 4 blocks. The bases and qualities vary with position so
+        // the assertions below pin the bytes the parser hands back, not just how many:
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #694, opened 2026-08-01
test(consensus): pin record_consensus_pair to two emitted reads

nh13:
`ConsensusCallingStats::consensus_reads` counts emitted *reads*, not templates. That is why callers emitting both ends of a pair go through `record_consensus_pair` rather than calling `record_consensus` twice — the distinction #681 introduced after duplex was found reporting one per template.

The invariant was pinned only end-to-end, by the per-command tests that compare the reported count against the records actually written. Nothing covered the accessor itself, so reverting it to a per-template increment would surface as a metrics discrepancy in duplex and simplex output rather than as a failure in the crate that owns the counter.

Mutation-checked: changing `+= 2` to `+= 1` fails the new test.

Test-only; no production code is touched.

## Summary by CodeRabbit

* **Tests**
  * Added coverage verifying that consensus pair and fragment recordings correctly increment the consensus read count.

### DIFF
--- a/crates/fgumi-consensus/src/caller.rs
+++ b/crates/fgumi-consensus/src/caller.rs
@@ -816,6 +816,20 @@ mod tests {
         assert_eq!(stats.consensus_reads, 1);
     }
 
+    /// `consensus_reads` counts emitted reads, not templates, so recording a pair must
+    /// add two — the whole reason callers that emit both ends use this instead of two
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #682, opened 2026-07-31
docs(zipper): stop denying that the `pa` tag was ever written

nh13:
Closes #679.

## The claim

`fgumi zipper --help` says of `--skip-tc-tags` (`src/lib/commands/zipper.rs:197-199`):

> The former name `--skip-pa-tags` is accepted as an alias. It described a `pa` tag that this command has never written, so grepping the output for `pa` found nothing; the tag is and always was `tc`.

Both halves are false, and I confirmed each against the tagged history:

```
$ f=src/lib/commands/zipper.rs
$ git show "v0.1.3:$f" | grep -E 'let (pa|tc)_tag = b"'
    let pa_tag = b"pa";
$ git show "v0.2.0:$f" | grep -E 'let (pa|tc)_tag = b"'
    let tc_tag = b"tc";
```

`40b685c` ("fix(zipper)!: rename sort-key tag from pa to tc to avoid bwa-mem pa:f collision", #270) is the commit that changed the literal. The "never written" wording came in with #617, which correctly renamed the flag but overcorrected on the history — its commit message carries the same claim.

## Why it matters

It is not only wrong, it is wrong in the direction that hurts. A BAM zippered by a pre-0.2.0 release carries `pa` and no `tc`. `dedup` counts `missing_tc_tag` on its secondary/supplementary reads and then bails (`src/lib/commands/dedup.rs:1428`) with a re-zipper instruction. Those BAMs also carry the `pa:B:i` vs bwa-mem `pa:f` type collision that motivated the rename in the first place. The help was telling exactly the users who hold such BAMs that nothing about their output ever changed.

## The 0.2.0 changelog

The v0.2.0 breaking-changes entry made three claims. Two hold, one does not:

| claim | verdict |
| --- | --- |
| tag renamed `pa` → `tc` | correct (`40b685c`) |
| metric `missing_pa_tag` → `missing_tc_tag` | correct |
| "The `--skip-pa-tags` flag ... is renamed to `--skip-tc-tags`" | wrong |

0.2.0 renamed only the internal Rust field. The CLI flag kept the old spelling until 0.5.0:

| version | CLI flag | internal field |
| --- | --- | --- |
| v0.1.3 | `--skip-pa-tags` | `skip_pa_tags` |
| v0.2.0 – v0.4.0 | `--skip-pa-tags` | `skip_tc_tags` |
| v0.5.0 | `--skip-tc-tags` (alias `--skip-pa-tags`) | `skip_tc_tags` |

[...]

PR #695, opened 2026-08-01
docs(zipper): name the `tc` tag in comments that still say `pa`

nh13:
`fgumi zipper` writes the template-coordinate sort key as `tc`, and has since 0.2.0, when the tag was renamed to stop colliding with bwa-mem's `pa:f` (primary-alignment score fraction). The code says so — `add_template_coordinate_tags_raw` appends `SamTag::TC` and `tc_info_from_raw` reads it — but about forty doc comments, inline comments, and assertion messages in `zipper.rs` still call it the `pa` tag. The section banner reads "Primary Alignment Tag (pa) Tests", naming the exact tag the rename exists to avoid, and failures print messages like `Supplementary should have pa tag` for a tag no fgumi release has written in four minor versions.

This renames those to `tc`, plus the `has_pa_tag` local. The `--skip-pa-tags` **flag alias** is deliberately untouched — that spelling is real and stays accepted.

One site in `fgumi-raw-bam` keeps its `pa` literal: `test_dedup_pa_tag_check_fails_on_b_array` reproduces a historical dedup bug and the legacy bytes are the point. Its comment claimed the tag was "produced by fgumi zipper", which stopped being true in 0.2.0; it now says what it actually is and notes that the tag *name* is incidental to the bug — what defeats dedup's check is the `B:i` array type, which `tc` still uses.

Comments and assertion messages only; no behavior change. Verified to merge cleanly with #682, which touches the help text in the same file.

## Summary by CodeRabbit

* **Documentation**
  * Clarified template-coordinate metadata terminology and legacy tag context.
  * Updated comments and test descriptions to consistently reference the `tc` tag.

* **Tests**
  * Improved regression test clarity for template-coordinate tags without changing existing behavior.

### DIFF
--- a/crates/fgumi-raw-bam/src/tags.rs
+++ b/crates/fgumi-raw-bam/src/tags.rs
@@ -2224,7 +2224,7 @@ mod tests {
         assert_eq!(find_string_tag(&aux, b"XI"), None);
     }
 
-    // --- B:i array (the pa tag type) ---
+    // --- B:i array (the type the `tc` template-coordinate tag uses) ---
 
     #[test]
     fn test_find_string_tag_cannot_find_b_int_array() {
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #696, opened 2026-08-01
perf(compare): make `compare bams` faster than the sort it validates

### LINKED ISSUE #686: perf(compare): `compare bams --command sort` ignores `--threads` and runs ~3x slower than `sort`
`fgumi compare bams --command sort` reads both inputs with a single-threaded BGZF reader regardless of `--threads`. It does strictly less work than sorting — no sort, no spill, no output write — yet runs at roughly a third of `fgumi sort`'s throughput on the same host at the same thread count.

Observed on `0.5.0-3df1022c`.

## Measured

Two coordinate-sorted BAMs, 779,820,469 records each, on an AWS `c7g.4xlarge` (16 vCPU, 32 GiB, gp3 provisioned to 500 MB/s):

| command | threads | wall | throughput |
| --- | --- | --- | --- |
| `compare bams --command sort` | 8 | 31m 43s | 409,673 rec/s |
| `sort --order coordinate` | 8 | 9m 50s | ~1,320,000 rec/s |

Sorting is **3.2x faster than comparing**, while additionally spilling 86 runs to disk and writing a ~48 GB output.

## Root cause

The `content` path opens its reader with a thread count:

```rust
// src/lib/commands/compare/bams.rs:590
let (mut reader, header) = create_raw_bam_reader(&path, threads)?;
```

The `sort` and `grouping` paths do not. Both inputs go through `OpenedInput::open`, which takes only a path:

```rust
// src/lib/commands/compare/bams.rs:636
let input1 = OpenedInput::open(&self.bam1)?;
let input2 = OpenedInput::open(&self.bam2)?;

// src/lib/commands/compare/engines/mod.rs:63
pub(crate) fn open(path: &Path) -> anyhow::Result<Self> {
    let (reader, header) = fgumi_sort::open_raw_bam_record_reader_with_header(path)
    ...
}
```

`src/lib/commands/compare/engines/sort_verify.rs` contains no threading primitives at all — no `rayon`, no `par_iter`, no spawned workers, no multithreaded reader.

The flag's help text is also incomplete, which is how this goes unnoticed:

```rust
// src/lib/commands/compare/bams.rs:346
/// Number of threads for BGZF decompression and parallel comparison. Used by `content`
/// mode only; the streaming molecule-join engine `grouping` mode (and `--command group`)
/// routes to is single-threaded and ignores this flag.
```

It names `grouping` as ignoring the flag but not `sort`, so `--command sort --threads 8` reads as supported and silently is not.

## Impact

At production scale (>1B records) the comparison takes longer than every sort it validates, combined. Any automated equivalency check built on `--command sort` pays this on its critical path.

## Proposed fix

1. Plumb a thread count into `OpenedInput::open` (or add an `open_with_threads`) so the `sort` and `grouping` presets get multithreaded BGZF decompression, matching what `content` already does. This is likely the bulk of the win, since the workload is decompression-bound.
2. Read the two inputs concurrently — they are independent streams until their keys are compared.
3. Parallelize the run-multiset comparison by chunking on sort-key-run boundaries, if decompression alone does not close the gap.
4. Correct the `--threads` help text, and consider warning when `--threads > 1` is passed to a path that ignores it, so a silently-ignored flag cannot mislead again.

## Acceptance criteria

- `compare bams --command sort --threads 8` on a ~780M-record pair completes in less wall time than `fgumi sort --threads 8` over the same data on the same host.
- Throughput scales measurably from `--threads 1` to `--threads 8`.
- External contract unchanged: the `RESULT: BAM files are IDENTICAL` / `DIFFER` line, `--max-diffs` behaviour, and exit-1-on-mismatch all behave exactly as before.
- Existing compare tests pass unmodified.

## Reproducing

Any two large coordinate-sorted BAMs of the same data will show it; the ratio grows with record count.

```sh
fgumi sort --input in.bam --output a.bam --order coordinate --threads 8 --compression-level 1
fgumi sort --input in.bam --output b.bam --order coordinate --threads 8 --compression-level 1
time fgumi compare bams --command sort --threads 8 a.bam b.bam
```

Compare that wall time against the sorts that produced the inputs.

nh13:
Closes #686.

`fgumi compare bams --command sort` ran ~3x slower than the `fgumi sort` it validates, and `--threads` did nothing. Profiling found why, and four changes fix it; a fifth removes a redundant traversal in `content` mode.

## Result

20.1M records/file, ~2 GB BAMs, M2 Max (12 core), page cache warmed identically before every run, minimum of 3 reps:

| | before | after |
| --- | --- | --- |
| `--command sort` `-t 8` | 23.44s | **4.14s** |
| `--command sort` scaling | flat | 8.92s (`-t 1`) → 5.58s (`-t 4`) → 4.14s (`-t 8`) |
| `--command group` `-t 8` | 24.36s | ~10s |
| `--command simplex` (content) | 47.1 CPU-s | **27.4 CPU-s** |

Both of #686's acceptance criteria hold: comparison is now **2.0x faster** than the sort it validates (4.14s vs 8.20s), having started 2.7x slower, and throughput scales measurably with `--threads`.

Content-mode results are quoted in CPU-seconds deliberately — see [Measurement](#measurement).

## What was wrong

`tricorder --trace` showed `sort` and `group` running the entire comparison in **exactly one OS thread** — `n_threads` never left 1, so `--threads` created nothing at all. Sampling put ~61% of that single thread in BGZF decode (~55% `libdeflate` inflate, ~6% the per-block CRC32) while eleven cores idled. `content` reached only 2.46 cores and additionally traversed each input **twice**, because `verify_records_in_order` ran over both paths before `positional_compare` reopened and streamed them again — four traversals for a job needing two.

## The changes

Suggested reading order is commit order; each commit is independently revertable.

1. **`f7fbf95` Read both inputs concurrently.** `OpenedInput::open_pair` gives each input a share of the `--threads` BGZF budget and its own read-ahead thread, so the two decodes overlap each other and the comparison. The budget is *split* rather than handed to each input in full: they are symmetric and both must be fully decoded, so giving each the whole count would request double the parallelism asked for. (`zipper` splits asymmetrically because its two inputs differ in cost; a comparison's do not.)

2. **`352cfa1` Cancel byte-identical records without building content keys.** `RunCanceller` built a canonical content key per record — a `Vec` per aux tag, a sort, a concatenation, a hash and a map insert. Comparing the raw bytes first skips all of it for pairs that agree.

3. **`13d8907` Verify sort order during the content pass.** Folds the order check into the comparison's own record pass via `OrderCheck`, removing the two dedicated traversals.

4. **`388ff00` Skip key and content checks for byte-identical pairs (content mode).** `record_keys_match` ran on every pair before `content_diffs`, whose first act is the same byte comparison. Also splits content mode's decompression budget, which was still starting 8 BGZF workers *per* reader at `-t 8`.

5. **`f0126b3` Read the MI tag in one aux scan.** `get_mi_tag_raw` probed `find_int_tag` then `find_string_tag`, each walking the aux block from the start; `fgumi group` writes `MI:Z:`, so the integer probe always missed and its scan was wasted.

### Why the byte-equality fast paths are sound

Both rest on the same argument, and neither replaces the general path. A `RecordKey` is a pure function of a record's bytes, and `content_key_exact` excludes `bin`, width-normalizes integer tags and sorts the tag multiset — so byte equality *strictly implies* key and content equality. The converse fails, which is exactly why these are fast paths: records that are content-equal but differ in tag order, integer width or `bin` fall through and still match. For the run comparison there is a second step: removing one record from each side *with the same key* leaves the multiset symmetric difference unchanged whatever else is pending, so the fast path needs no precondition on canceller state.

## Correctness

The reader change carried a real trap. `RawReadAheadReader` is `Iterator<Item = RawRecord>` — no `Result` — and signals a read failure by *ending iteration* and parking the error for a later `take_error()`. Adopted directly, a truncated or CRC-corrupt BAM would have read as a **shorter** file: a record-count `DIFFER`, or a false `IDENTICAL` when both inputs are damaged alike. Rather than requiring every engine loop to remember a trailing check, `CheckedRecords` restores the `Result` contract at the boundary, so the engines' existing `rec?` propagation is correct by construction and forgetting the check is not an available mistake.

Every change was gated against a **121-comparison differential corpus** — matched, missing, extra, edited, tag-reordered, intra-run reordered, mis-sorted, truncated, header-conflicting and byte-corrupt inputs, in both directions and at three thread counts — requiring byte-identical verdicts, exit codes and RESULT lines against a binary built from `main`. The corpus is generated (`fgumi simulate`), not committed. Two cases carry particular weight: `tag-reorder` and `swap-adjacent` are content-equal but *not* byte-equal, so they prove the fast paths did not swallow the general checks.

Error text changed only for damaged inputs, where the messages now name the failing file and diagnose it more precisely (`block data checksum mismatch` rather than a generic open failure).

New tests: sort-order violations are pinned by count, by which file is named, and by acceptance of correctly-ordered input at more than one thread count; the damaged-input regression test is parameterized over thread count as well as damage mode, covering the worker-pool error route that `open_pair` actually uses. An existing test asserting only that an error `contains("order")` was strengthened to assert the actual contract.

## Measurement

Content-mode numbers are CPU-seconds, not wall. This host's wall time proved unreliable under interactive load — identical `-t 8` runs measured 3.56s and 16.56s — while total CPU stayed within 24–27s across the same sweep. Where wall is quoted, arms were interleaved with the page cache warmed identically before every run and the minimum of 3 reps taken; the outliers are all slower, i.e. contention.

All numbers are from this workstation. #686's criteria were written against a 780M-record `c7g.4xlarge`, and that run has not been done — the effects here (5.7x) are far outside the noise, but the issue's own scale has not been reproduced.

## Notes for reviewers

- **`fgumi-sort` gains one export and one signature generalization.** `RawReadAheadReader` is re-exported, and `verify_sort_order` now takes any `Iterator<Item = io::Result<RawRecord>>` rather than a concrete reader. The `Result` item type is deliberate: it keeps a read failure distinguishable from end-of-stream.
- **`positional_compare` gained a required parameter**, so this is a breaking change for out-of-tree callers of that feature-gated API. A defaulting overload was considered and rejected: it would let a caller opt out of order verification by omission.
- **`verify_records_in_order` is removed** — `OrderCheck` subsumes it and it had no callers left. `fgumi sort --verify` continues to use `fgumi_sort::verify_sort_order`.
- **Grouping mode keeps its ~10s** beyond the reader win; its engine (`molecule_join`) was otherwise untouched, and its remaining hot spots are `find_tag_position` and `get_mi_tag_raw`.

## Not done, deliberately

Migrating the readers to the sort engine's pooled architecture was investigated and rejected for now. `PooledInputStream::next_block` blocks on `std::thread::park()` and pool workers `unpark()` the thread captured at construction — correct for the sort pipeline's single stream and single consumer. Two pools feeding one comparison thread would let stream A's workers wake the consumer while it waits on stream B, producing a cross-stream unpark storm: a CPU-burning busy-wait, the opposite of the intent. Fixing it properly needs per-stream condvars or a per-stream consumer thread, and the latter reintroduces the channel it would be removing.

Raising the read-ahead batch from 256 to 4096 records was also measured and reverted: CPU was unchanged (26.5 vs 27.0 CPU-s) while peak RSS tripled. That null result is informative — the time in `crossbeam_channel::recv` is the comparison thread *waiting on decode*, not per-handoff overhead, so the remaining lever is decode throughput rather than handoff granularity.

`CLAUDE.md` records the profiling ladder this work established (`tricorder` first, then in-tree phase timers, then `perf` on Linux) along with three macOS sampling traps that produced confidently wrong profiles here — including "38% of CPU in `clap_builder::error::Error::print`" on a *successful* run, from symbolicating another image's addresses against the fgumi binary.

Risk: comparison verdicts, exit codes, `RESULT` lines, `--max-diffs`, and diagnostics remain pinned; `unsafe` changes are none and the `CLAUDE.md` allowlist is unchanged; thread allocation and read-ahead behavior change, with no reported memory-bound or queue-capacity change.

- `fgumi compare bams` reads both BAM inputs concurrently.
- `--threads` applies to sort and grouping comparisons through split decompression budgets.
- Sort-order checks run during comparison instead of a separate traversal.
- Byte-identical records use fast paths.
- MI tags use one auxiliary-tag scan.
- `CheckedRecords` preserves deferred read errors and identifies the affected input.
- Tests cover ordering diagnostics, damaged BAMs, thread counts, and differential comparison behavior.
- Benchmarking guidance and performance data were added to `CLAUDE.md`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,7 @@ docs/src/metrics/
 
 # Heap profiling output produced when binaries are built with the dhat-heap feature
 dhat-heap.json
+
+# Scratch space for local benchmark corpora, profiles, and campaign notes.
+# Never committed: the compare-bams benchmark corpus alone is ~20 GB of BAMs.
+/ephemeral/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -265,6 +265,124 @@ listed in this section.
 
 ## Benchmarking Notes
 
+### Profiling fgumi: what works, in order
+
+Reach for these in order. The first two answer most questions and need no new
+code; only drop to sampling when you genuinely need per-function attribution.
+
+**1. `tricorder` — core utilization, memory, and I/O.** The fastest way to learn
+whether a command is CPU-bound, I/O-bound, or leaving cores idle:
+
+```sh
+tricorder --interval 0.25 --summary --out agg.tsv --trace trace.tsv -- fgumi <cmd> ...
+# tricorder: wall=23.68s cpu=23.22s mean_load=98% max_rss=20.3MiB io_in=4.5MiB
+```
+
+`mean_load` is the number to read first: **98% means one core busy**, so on a
+12-core box the command is single-threaded regardless of what `--threads` was
+passed. `--trace` gives cores-over-time, which distinguishes "parallel
+throughout" from "parallel phase followed by a serial tail". Note that `io_in`
+collapses to near zero once the input is in page cache — a low `io_in` means the
+run was CPU-bound *on that invocation*, not that the tool does little I/O.
+
+**2. In-tree phase timers — the primary attribution tool.** `fgumi sort` already
+carries `SortPhaseTimer` (`crates/fgumi-sort/src/external.rs`): per-phase `f64`
+accumulators, a `time()` helper wrapping each span, and a `log_summary()` that
+emits `=== Sort Phase Timing ===` at `info!` (read+decompress / in-memory sort /
+spill write / consolidation / k-way merge / write output, each with a percentage).
+Prior campaigns got full baseline phase attribution from this with **zero new
+code** — just `RUST_LOG=info`.
+
+Mirror that pattern when optimizing any other command: the phases you would name
+in a design doc are exactly the phases worth timing, it works identically on
+macOS and Linux, and it cannot be defeated by inlining or missing symbols the way
+a sampling profiler can. Prefer adding a phase timer over fighting a profiler.
+
+**3. `perf record` / `perf stat` on Linux (EC2)** when per-function or
+per-instruction attribution is genuinely required. This is the reliable path for
+function-level work; see the EC2 host recipe in the user-level CLAUDE.md.
+
+#### macOS sampling: pitfalls that produce confidently wrong profiles
+
+`samply` works on macOS but has three traps with this codebase. All three
+silently yield a plausible-looking, wrong profile:
+
+- **The release binary has no symbols.** `[profile.release]` builds get
+  `-C strip=debuginfo`, so everything symbolicates to raw addresses. Use the
+  existing `[profile.profiling]` (inherits release, `debug = "line-tables-only"`),
+  or `CARGO_PROFILE_PROFILING_DEBUG=2` for full DWARF.
+- **Blocked threads are counted as CPU.** samply samples every thread on a timer,
+  including ones parked on a channel or condvar. Counting raw samples credited an
+  idle BGZF worker pool the same CPU as the working thread and made an 18 s run
+  look like 285 s across 21 threads. Weight samples by `threadCPUDelta`
+  (microseconds of real CPU between samples) instead of counting them.
+- **`atos` mis-resolves addresses from other images.** Feeding every sampled
+  address to `atos -o target/profiling/fgumi` resolves `libsystem_kernel` /
+  `libsystem_pthread` / `libsystem_malloc` addresses against the fgumi image,
+  landing on whatever symbol happens to sit at that offset. This is where
+  nonsense like "38% of CPU in `clap_builder::error::Error::print`" on a
+  *successful* run comes from. Map each frame to its library first (Gecko
+  `funcTable.resource` → `resourceTable` → `libs`) and only symbolicate frames
+  belonging to the fgumi image. Under fat LTO, local symbols are gone, so even
+  in-image resolution can land on a neighbouring function — treat any
+  suspiciously hot dependency symbol as misattribution until corroborated.
+
+Corroborate a sampling result against `tricorder` or a phase timer before acting
+on it. In the `compare bams` work, the sampling profile and `tricorder` agreed
+that BGZF decode dominates (~61% self time; `mean_load=98%`), which is what made
+the conclusion trustworthy — not the profile alone.
+
+### `fgumi compare bams` baseline (recorded 2026-08-01, issue #686)
+
+M2 Max (12 core), 20.1M records/file, ~2 GB BAMs, page-cached, `--threads 8`:
+
+| mode | wall | cpu | mean_load | cores busy | peak OS threads | max_rss |
+| --- | --- | --- | --- | --- | --- | --- |
+| `--command sort` (coordinate) | 23.7s | 23.2s | 98% | 0.98 / 12 | **1** | 20.3 MiB |
+| `--command group` | 24.7s | 24.3s | 98% | 0.98 / 12 | **1** | 21.2 MiB |
+| `--command simplex` (content) | 18.3s | 45.1s | 246% | 2.46 / 12 | 21 | 82.5 MiB |
+
+`sort` and `group` ignored `--threads` entirely: the tricorder `--trace`
+`n_threads` column stayed at **1** for the whole run, so the flag created no
+threads at all. `OpenedInput::open` took no thread count, so both inputs were
+decoded on the single main thread.
+
+#### After round 1 (concurrent per-input readers)
+
+`OpenedInput::open_pair` now gives each input its own read-ahead thread and a
+share of the `--threads` BGZF budget. Same host and corpus, interleaved arms with
+the page cache warmed identically before every run (minimum of 3 reps; the
+outliers are all slower, i.e. contention):
+
+| mode | baseline | round 1 | speedup |
+| --- | --- | --- | --- |
+| `--command sort` | 23.44s | 8.93s | 2.6x |
+| `--command group` | 24.36s | 9.92s | 2.5x |
+
+`mean_load` rises 98% → 370% and peak threads 1 → 13.
+
+**The win is from overlap, not from `--threads`.** Thread scaling is flat —
+`-t 1` 8.64s, `-t 2` 8.47s, `-t 4` 8.97s, `-t 8` 8.89s — because the single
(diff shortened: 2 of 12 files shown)


nh13:
@coderabbitai pause

nh13:
Ran this branch at production scale on `c7g.4xlarge`, which the PR notes as not done. The workstation result transfers: **5.84x throughput**, against 5.7x measured on the 12-core host.

## Numbers

Both arms `--command sort --threads 8` on `c7g.4xlarge` (16 vCPU / 30 GiB), inputs on gp3 provisioned to 500 MB/s, comparing two independently-sorted BAMs:

| | fgumi | records | wall | throughput |
| --- | --- | --- | --- | --- |
| before | `3df1022c` (v0.5.0, pre-#696) | 779,820,469 | 31m 43s | 409,673/s |
| after | `7585aba3` (this branch) | 1,327,219,598 | **9m 14s** | **2,392,619/s** |

Three comparisons ran on the after side, all at 1.33B records:

- coordinate, fgumi vs samtools — 9m 14s (2,392,619/s)
- coordinate, mako vs fgumi — 9m 15s (2,390,736/s)
- template-coordinate, mako vs fgumi — 10m 1s (2,205,211/s)

## #686's acceptance criteria at this scale

> comparison completes in less wall time than `fgumi sort --threads 8` over the same data on the same host

Met, and not narrowly. On the coordinate cell the `fgumi sort` that produced one of these inputs took **2529.2s**; comparing its output took **554s** — the comparison is **4.6x faster than the sort it validates**. Against the same sort without its consolidation pass (1535.5s, the mako arm) it is still 2.8x faster.

## Caveats, since these are not an interleaved A/B

The two rows are different datasets and different record counts — the before is a 1000 Genomes HG00096 pair from an earlier run, the after is GIAB HG002 60x. I have normalized to items/s, and the instance type, thread count and storage configuration are identical, but this is not the same-inputs comparison your workstation numbers are. Treat 5.84x as "the effect is the same order at production scale on the target hardware", not as a precise replication.

I also did not sweep `--threads` at this scale, so the second criterion — measurable scaling from 1 to 8 — remains verified only on the workstation.

## Also

The reader change held up on inputs where per-record cost is high: the template-coordinate cells run over a BAM with 77.3M both-ends-unmapped records, and the only thing that stopped one of those four comparisons was the pending-window abort that #700 fixes — not throughput or memory.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #700, opened 2026-08-01
fix(compare): compare a displaced sort-key run by digest instead of aborting

### LINKED ISSUE #699: compare bams --command sort: cannot compare two template-coordinate sorts with a large unmapped tail
`fgumi compare bams --command sort` cannot compare two independently-produced template-coordinate sorts of a BAM with a large unmapped tail. It aborts on `MAX_PENDING_RUN_RECORDS`, and the remedy the error suggests is circular when the two inputs are themselves the output of two sorters being compared.

Observed on `0.5.0-7585aba3` (the head of #696).

## What happened

Comparing `fgumi sort --order template-coordinate` output against `samtools sort --template-coordinate` output, same input, 1,327,219,598 records:

```
Using --command sort preset: sort-key-run verification (no positional pairing)
Error: sort-verify: more than 5000000 records held pending while comparing one
equal-sort-key run — the two inputs disagree about the order of records within
that run by more than this window (an unmapped tail is one such run). Re-sort
both inputs into a common order and retry.
```

The same pair in **coordinate** order compares fine:

```
Record counts: 1327219598 vs 1327219598
Sort-key-run multiset mismatches: 0
RESULT: BAM files are IDENTICAL (sort order verified; run multisets match)
```

Both outputs are correctly sorted — `fgumi sort --verify` reports `PASS - file is correctly sorted by TemplateCoordinate`.

## Why

The input is GIAB HG002 60x, which contains **77,270,534 records flagged `0xC`** (both ends unmapped) out of 1.33B. Under template-coordinate those share one sort key and form a single equal-key run. fgumi and samtools order records *within* that run differently, so `RunCanceller`'s pending set grows with the displacement between the two orderings — 77.3M records against a 5M window.

`MAX_PENDING_RUN_RECORDS`' own doc comment describes this precisely:

> Pending grows only with the *displacement* between the two orderings, so this fires when the inputs disagree about intra-run order beyond this window — e.g. two independently-produced unmapped tails that were never re-sorted into a common order.

## The problem with the suggested remedy

The error says *"Re-sort both inputs into a common order and retry"*, and the doc comment names `samtools sort` / `fgbio SortBam` as the fix. That works when comparing a pipeline's output against a reference, but not here: **the two inputs are the outputs of the two sorters under comparison.** Re-sorting them into a common order would destroy the very property being verified. There is no third order to normalize to.

This is not a hypothetical setup — it is what `--command sort` appears to exist for: checking that two implementations sort the same data identically.

## Impact

Any cross-implementation template-coordinate comparison on a whole-genome BAM hits this, because unmapped tails at that scale are normal (5.8% here, and the fraction matters less than the absolute count). The coordinate arm of the same comparison passes, so the failure is specific to sort orders that collapse the unmapped tail into one key.

There is currently no CLI flag to raise the cap, so there is no workaround short of patching the constant.

## Possible directions

Roughly in order of how well they seem to fit the engine's existing design — the maintainers will know better which is right:

1. **Compare the unmapped tail as an order-insensitive multiset.** Records with no alignment have no meaningful order to disagree about, so requiring a common intra-run order over-specifies the contract. This would make the comparison correct at any tail size rather than merely larger.
2. **Expose the cap as a flag** (e.g. `--max-pending-run-records`). Cheapest, but pushes an O(N) memory decision onto the caller and only moves the wall.
3. **Spill the pending set** once it exceeds the window, so a large displacement costs disk rather than aborting.

Option 1 seems the most principled, if unmapped-tail ordering is genuinely not part of the template-coordinate contract; that judgment is yours.

## Reproducing

Any whole-genome BAM with a few million both-unmapped records, sorted two ways:

```sh
fgumi sort --input in.bam --output fgumi.tc.bam --order template-coordinate --threads 8
samtools sort --template-coordinate -@ 8 -o samtools.tc.bam in.bam
fgumi compare bams --command sort --threads 8 fgumi.tc.bam samtools.tc.bam
```

The input must have MC on every mapped record for samtools to accept it (`samtools fixmate -m` on a name-sorted file); both-unmapped pairs legitimately have no MC and samtools does not require it.

nh13:
Closes #699. Stacked on #696 — review that first; this PR's diff is against its head (`7585aba3`).

## The problem

`compare bams --command sort` compares each equal-core-sort-key run as a multiset. `RunCanceller` holds unmatched records keyed by canonical content key until their counterparts arrive, so pending memory grows with the *displacement* between the two files' orderings — and past 5M records it was a hard error.

A BAM's both-ends-unmapped tail is a single run: every such record packs to one constant coordinate key, and to one core template-coordinate key. Two independently-produced sorts order that tail arbitrarily differently, and on a whole-genome BAM the tail is tens of millions of records — 77.3M of 1.33B in #699. So the comparison aborted.

The error's advice was to re-sort both inputs into a common order. That is circular in exactly the case `--command sort` exists for: **the two inputs are the outputs of the two sorters under comparison.** There is no third order to normalize to, and re-sorting would destroy the property being verified.

## The fix

`RunCanceller` degrades instead of aborting. On the observation that pushes pending past the window, both pending sets are folded into per-side `MultisetDigest`es, the maps are dropped, and the rest of the run accumulates into those digests in constant memory.

A `MultisetDigest` is a commutative fold — count, wrapping sum, and xor over a 128-bit hash of each record's canonical content key. Commutativity is the whole point: it is what makes the accumulated value independent of arrival order, which is what the exact canceller was paying O(displacement) memory to achieve.

### Why the fallback is sound

The comparison decides the *symmetric difference* of the two sides' record multisets. Writing `A` and `B` for the two sides' records and `C` for the pairs already cancelled — each of which removed one element from `A` and an **equal** element from `B` — the digests cover exactly `A \ C` and `B \ C`, and `A == B` iff `A \ C == B \ C`.

So the fallback is a change of representation mid-run, not a restart: a partially-displaced run still cancels exactly everything it can before degrading, and that work carries across. Pinned by `cancellations_made_before_degrading_survive_it`.

The key material is `content_key_exact`, unchanged from #696. Its byte-equality already *is* exact content-equality (there's a proptest pinning that), so hashing it inherits the contract rather than inventing a weaker one. Read names never enter the comparison — they aren't unique across mates and supplementaries, so they'd be unsound as identity.

### Why all three digest fields

- `count` alone catches a missing or extra record.
- `xor` alone does **not**: it is self-inverse, so `{A,A,B}` and `{B,C,C}` share both a count and an xor while being different multisets.
- `sum` separates those, since duplicates add rather than annihilate; `xor` in turn covers sums that coincide modulo 2¹²⁸.

`digest_distinguishes_xor_cancelling_multisets` pins this, and it is genuinely load-bearing — deleting the `sum` update makes it fail.

## The guarantee this changes

**This trades exactness for constant memory, and that is worth stating plainly.** Digest equality is probabilistic: two genuinely different runs agreeing on count, sum *and* xor would be reported equal. At 128 bits that is ~2⁻¹²⁸ per run and not reachable by construction from non-adversarial BAM data — but it is not zero, and `--command sort` is a validation oracle, which is the one place a false MATCH costs most.

The alternative on the table was aborting on the input entirely. Exact-at-any-displacement would mean spilling the pending set to disk (#699's option 3), which is a lot of machinery for a feature-gated developer tool; I'd rather add it later if a false MATCH ever plausibly bites than build it speculatively now.

Runs that degrade are counted and reported, so an `IDENTICAL` verdict that rested on a digest is visible as one rather than silently indistinguishable from an exact match.

## Window: 5M → 1M

The window no longer bounds correctness, only diagnostics — within it a mismatched run still names the reads that went unmatched; past it the run is reported by count with an explicit note saying why no reads are named. So it is now sized to cover any plausible genuine *difference* rather than any plausible *displacement*, and 5M pending content keys was several GB of maps for no remaining benefit.

It is threaded through `VerifyContext` so tests can reach the fallback with a handful of records instead of a million. It is deliberately **not** a CLI flag (#699's option 2): exceeding it is no longer something the caller has to do anything about.

## Validation on the real shape

11,964,096 both-ends-unmapped records taken from 1000 Genomes HG00096, sorted independently by `fgumi sort --order template-coordinate` and `samtools sort --template-coordinate`. The two orderings are fully divergent — the first record of one file sits 557,077 records into the other.

| | #696 (`7585aba3`) | this branch |
| --- | --- | --- |
| verdict | `Error: ... more than 5000000 records held pending` | `RESULT: BAM files are IDENTICAL` |
| peak RSS | 4.6 GB | **985 MB** |
| wall | 6.8s (aborted) | 10.1s (completed) |

Both arms then re-run against a deliberately corrupted copy, and both are still caught:

- **one record dropped** → `DIFFER` (counts 11964096 vs 11964095)
- **one base flipped** → `DIFFER` with **equal record counts on both sides**, so only the digest could have found it — one altered base among 11.96M records

A note on scale, since it is easy to under-size this reproducer: a fully-scrambled run of N records peaks at only ~N/2 pending, so the 1.33M-record tail I started with cleared neither cap and reproduced nothing. #699's 77.3M tail peaks around 38M. That is why the fixture is ~12M.

## Tests

- **`digest_and_exact_verdicts_agree`** (proptest) — the load-bearing one. For arbitrary multiset pairs, the verdict with `window = 0` (everything through the digest) must equal the verdict with `window = usize::MAX` (never degrades). A digest collision, a record lost across the fold, or a mishandled multiplicity all fail this.
- `a_run_past_the_window_is_compared_by_digest_rather_than_aborting` — the #699 regression at the canceller level.
- `a_reordered_unmapped_tail_compares_identical_via_the_digest` — the same, end to end through the engine.
- `a_difference_inside_a_degraded_run_is_reported_without_names` — a degraded run still catches a real difference, and says why it names nothing.
- `a_degraded_run_does_not_cost_other_runs_their_diagnostics` — the fallback is per-run: a huge unmapped tail must not cost the mapped runs their read names.
- `pending_never_exceeds_the_window` — bounded memory asserted directly rather than inferred from RSS.
- `the_digest_decides_a_mismatched_run`, `digest_distinguishes_xor_cancelling_multisets`, `cancellations_made_before_degrading_survive_it`.

Existing exact-mode residual-naming tests are unchanged and still pass — that's the regression guard on the common path. Full suite: 6916 passed.

## Notes for reviewers

- **`RunResidual` is now an enum** (`Exact { .. }` / `Digest { matched }`), and `is_empty()` became `is_match()` — "empty" stopped describing the digest arm.
- **`RunCanceller::observe_bam1`/`observe_bam2` collapsed into `observe(Side, ..)` and became infallible.** Removing the error was the point; `compare_run` loses two `?`.
- **The digest hashers are fixed-seed, not `RandomState::new()`.** A comparison oracle that could change its mind about the same pair of files between invocations would be worse than no oracle, so the verdict has to be reproducible across runs, hosts and architectures.
- `SortVerifyOutcome` gains `runs_compared_by_digest`, which deliberately does **not** feed `is_match()` — degrading is not an error.

Risk: sort comparison output adds deterministic digest-comparison reporting; no grouping, consensus, sort-order, corrected-UMI, or metrics output changes; no `unsafe` change and no CLAUDE.md allowlist change; exact pending-record tracking is bounded with constant-memory digest fallback.

When the exact window is exceeded, `sort_verify` preserves prior cancellations and switches to a deterministic 128-bit multiset digest. The digest uses a fixed seed and tracks record count, wrapping sum, and XOR of canonical record-content hashes. Digest comparisons preserve multiplicity but cannot identify unmatched records.

The change adds degraded-run reporting through `runs_compared_by_digest` and retains exact diagnostics for non-degraded runs. Tests cover digest behavior, bounded memory, residual diagnostics, and the large unmapped-tail regression. Full validation passed 6,916 tests.

### DIFF
--- a/src/lib/commands/compare/bams.rs
+++ b/src/lib/commands/compare/bams.rs
@@ -926,6 +926,17 @@ impl CompareBams {
             }
             println!();
             println!("Sort-key-run multiset mismatches: {}", outcome.run_mismatches);
+            // Printed only when it happened: on the overwhelmingly common in-order pair it
+            // would be a line of zero every time, and its whole purpose is to qualify a
+            // verdict that was reached without naming records.
+            if outcome.runs_compared_by_digest > 0 {
+                println!(
+                    "Runs compared by order-insensitive digest: {} (the two inputs' orderings \
+                     diverged within these runs by more than the exact window; the multiset \
+                     verdict still holds, but unmatched records in them cannot be named)",
+                    outcome.runs_compared_by_digest
+                );
+            }
             println!();
 
             if is_equal {
--- a/src/lib/commands/compare/engines/sort_verify.rs
+++ b/src/lib/commands/compare/engines/sort_verify.rs
@@ -29,6 +29,12 @@
 //!    tolerates intra-run reordering while still catching a missing/extra record or any
 //!    content difference.
 //!
+//! The run comparison is exact — reporting the specific reads that went unmatched — for as
+//! long as the two files' orderings stay within [`EXACT_WINDOW_RECORDS`] of each other. Past
+//! that it degrades to an order-insensitive digest ([`MultisetDigest`]) that still decides
+//! the same question in constant memory, but can no longer name individual reads. See
+//! [`RunCanceller`] for why that fallback exists and what it costs.
+//!
 //! Reading each file once (rather than an order pass per file *plus* a reopened comparison
 //! pass — four full BAM traversals) roughly halves the BAM I/O + BGZF decode this
 //! benchmarking command spends on WGS-scale inputs. Because the comparison — including its
@@ -42,6 +48,7 @@
 
 use std::cmp::Ordering;
 use std::path::Path;
+use std::sync::LazyLock;
 
 use anyhow::{Result, anyhow, bail};
 
@@ -53,7 +60,7 @@ use fgumi_sort::{
 };
 use noodles::sam::Header;
 
-use ahash::AHashMap;
+use ahash::{AHashMap, RandomState};
 
 use super::{CheckedRecords, OpenedInput};
 use crate::sam::SamTag;
@@ -82,6 +89,17 @@ pub struct SortVerifyOutcome {
     /// Number of equal-core-sort-key runs whose record multiset differed between the two
     /// files (a content difference, or a missing/extra record within the run).
     pub run_mismatches: u64,
+    /// Number of runs whose comparison fell back from the exact canceller to the
+    /// order-insensitive digest because the two files' orderings diverged by more than
+    /// this engine's exact-cancellation window (`EXACT_WINDOW_RECORDS`, enforced by
+    /// `RunCanceller`). Both are private to this module, so they are named here rather
+    /// than linked — an intra-doc link from this public field would not resolve.
+    ///
+    /// Purely informational, and deliberately *not* part of [`Self::is_match`]: degrading is
+    /// not an error, and a run compared this way is still compared. It is reported so a
+    /// reader can tell that an `IDENTICAL` verdict rested on a digest rather than on an
+    /// exact multiset, and so a `DIFFER` verdict that names no reads is explicable.
+    pub runs_compared_by_digest: u64,
     /// `true` if the two files' run sequences desynchronized — one file's run boundaries
     /// don't line up with the other's (a whole run present on only one side), or one file
     /// ran out of records before the other. Once this happens, no further runs are
@@ -413,13 +431,14 @@ fn compare_run<K, ExtractKey, IsViolation>(
     next2: &mut Option<Keyed<K>>,
     key: &K,
     same_core: &impl Fn(&K, &K) -> bool,
+    exact_window: usize,
 ) -> Result<RunPairOutcome>
 where
     K: Clone,
     ExtractKey: Fn(&[u8]) -> K,
     IsViolation: Fn(&K, &K) -> bool,
 {
-    let mut canceller = RunCanceller::new(MAX_PENDING_RUN_RECORDS);
+    let mut canceller = RunCanceller::new(exact_window);
     let (mut count1, mut count2) = (0u64, 0u64);
 
     loop {
@@ -456,11 +475,11 @@ where
             continue;
         }
         if let Some(rec) = from1 {
-            canceller.observe_bam1(rec.as_ref())?;
+            canceller.observe(Side::Bam1, rec.as_ref());
             count1 += 1;
         }
         if let Some(rec) = from2 {
-            canceller.observe_bam2(rec.as_ref())?;
+            canceller.observe(Side::Bam2, rec.as_ref());
             count2 += 1;
         }
     }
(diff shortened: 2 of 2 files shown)


nh13:
@coderabbitai pause

nh13:
Validated this branch against the actual #699 inputs — the 77,270,534-record unmapped tail, not a fixture. It fixes it, and the constant-memory claim holds at 6.4x the scale of the PR's own reproducer.

## Setup

The two BAMs from the run that produced #699: GIAB HG002 60x (1,327,219,598 records) sorted independently by `fgumi sort --order template-coordinate` and `samtools sort --template-coordinate`, 134 GB each. `c7g.4xlarge` (16 vCPU / 30 GiB), `--threads 8`, both files on local NVMe-backed gp3.

Built from `5862032d` with `--features compare`.

## Result

```
Detected sort order: TemplateCoordinate
Record counts: 1327219598 vs 1327219598
bam1 sort-order violations: 0
bam2 sort-order violations: 0
Sort-key-run multiset mismatches: 0
Runs compared by order-insensitive digest: 1 (the two inputs' orderings diverged
within these runs by more than the exact window; the multiset verdict still holds,
but unmatched records in them cannot be named)

RESULT: BAM files are IDENTICAL (sort order verified; run multisets match)
```

| | #696 (`7585aba3`) | this branch (`5862032d`) |
| --- | --- | --- |
| verdict | `Error: ... more than 5000000 records held pending` | `IDENTICAL` |
| peak RSS | aborted | **969,700 KB (969.7 MB)** |
| wall | aborted | 10m 58s |
| CPU | — | 376% |

## What this adds over the PR's own numbers

The fixture in the PR is 11,964,096 records (~6M peak pending) and measured 985 MB. This is **77,270,534 records, ~38M peak pending** — 6.4x the records, ~6x the peak displacement — and peak RSS came in at **969.7 MB**, marginally *below* the fixture's. Flat across a 6.4x scale-up is the evidence the constant-memory argument needed; the fixture alone couldn't distinguish "constant" from "grows slowly".

Steady-state RSS during the run sat around 22 MB, so the ~970 MB is a transient at the fold rather than a sustained footprint.

**Exactly one run degraded**, and it is the right one. The unmapped tail is 5.8% of this file; every mapped run stayed on the exact path and kept its diagnostics. That is `a_degraded_run_does_not_cost_other_runs_their_diagnostics` holding on real data rather than a constructed case.

The reported degradation line matters as much as the verdict here: an `IDENTICAL` that rested on a digest is visibly distinguishable from an exact one, which is the right call for an oracle.

## On the exactness trade-off

Agree with the framing in the PR. For our use — a benchmark suite checking that two sorters agree — a probabilistic verdict on the unmapped tail specifically is the right trade: those records have no meaningful order to disagree about, so the exact path was paying O(displacement) memory to verify something that isn't part of the contract. The mapped runs, where a real regression would show up, are still exact.

Worth noting for anyone sizing a reproducer later: the scale note in the PR is correct and easy to get wrong. Our first attempt at reproducing #699 on a 1.33M-record tail cleared neither cap and showed nothing.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #702, opened 2026-08-01
ci: report scheduled-workflow failures as a tracking issue

nh13:
## Summary

A failing cron notifies nobody. GitHub emails only the person who last touched the workflow file — private, easy to filter, invisible to the rest of the team. `miri.yml` failed on **every one of its 13 runs** over two weeks and that surfaced nowhere. It was found by chance, by manually dispatching the workflow for an unrelated reason.

This adds a composite action that opens a tracking issue when a scheduled workflow fails, wired into both crons as an `if: failure()` step.

## The dedupe is the point

A naive "file an issue on failure" would have produced fourteen issues for the outage above, which is just as easy to tune out as producing none — the same silence reached by a different route.

So: one open issue per workflow, reused. The first failure files it; every subsequent failure comments on it. The comment count then reads as *how long this has been broken*. Closing the issue once the workflow is green means the next failure opens a fresh one.

Matching is on the exact issue title (`Scheduled workflow failing: <name>`) rather than `gh`'s full-text `--search`, which would also match issues that merely mention the workflow — and would let the two crons adopt each other's issue.

## Scope

Both cron workflows, not just the broken one. `stress.yml` has the identical blind spot; it has simply been green every day so far, so its blind spot is untested rather than absent.

## Permissions

Both workflows gain `issues: write`, used only by the reporting step. Neither writes to repository contents, and `persist-credentials: false` on checkout is unchanged — the step authenticates `gh` through `github.token` in the environment, not through git credentials.

The `ci-failure` label is created on demand inside the action (idempotent, failure tolerated), so this needs no out-of-band setup and no manual step before the first failure can be reported.

No third-party action is introduced — `gh` and `jq` are both preinstalled on GitHub-hosted runners, so there is nothing new to pin by commit hash.

## Verification

The failure path is the one that never runs in normal operation, which is exactly how the original bug survived, so I tested it directly rather than by inspection. I extracted the action's script and ran it against a stubbed `gh` across three cases:

| Case | Expected | Result |
|---|---|---|
| No open issue with the label | file a new issue | 1 create, 0 comments |
| Open issue with this workflow's exact title | comment, don't duplicate | 1 comment on that issue, 0 creates |
| Open issue with the *other* workflow's title | file its own issue | 1 create |

Also confirmed the issue body renders flush-left — indented heredoc content would have turned the whole body into a Markdown code block.

## What this does not solve

`miri.yml` failed on its very first run, so there was never a green baseline to regress from. Notification would have caught it on day one, which is enough here. But nothing automatable enforces "confirm a new gate actually passes before trusting it" — that stays a review-time habit.

## Related

- #701 fixes the underlying `miri.yml` failure. The two PRs touch different regions of that file and should not conflict.

Risk: command output changes: yes, issue reporting is pinned by exact workflow-title matching; `unsafe`: none, CLAUDE.md allowlist: unchanged; memory, queue, and thread/backpressure policy: none.

Fix: The composite action reports scheduled `miri.yml` and `stress.yml` failures through one open tracking issue per workflow. It creates the `ci-failure` label when needed, comments on an existing exact-title issue, or creates a new issue after closure. Both workflows now grant `issues: write`.

### DIFF
--- a/.github/actions/report-cron-failure/action.yml
+++ b/.github/actions/report-cron-failure/action.yml
@@ -1,0 +1,87 @@
+name: Report cron failure
+description: >-
+  Open (or update) a tracking issue when a scheduled workflow fails, so a red
+  cron is visible in the repository rather than only in per-user email.
+
+# A scheduled workflow that fails notifies nobody except the person who last
+# touched its file, by email, which is easy to filter and invisible to everyone
+# else. `miri.yml` failed on every run for two weeks before anyone noticed. This
+# action turns that silence into an issue in the repository.
+#
+# One issue per workflow, reused: a failure that persists for weeks comments on
+# the existing issue rather than filing a new one each night. Filing duplicates
+# would be just as easy to tune out as filing nothing.
+
+inputs:
+  token:
+    description: "Token with `issues: write` on this repository."
+    required: true
+  label:
+    description: Label applied to the tracking issue.
+    required: false
+    default: ci-failure
+
+runs:
+  using: composite
+  steps:
+    - name: Open or update the tracking issue
+      shell: bash
+      env:
+        GH_TOKEN: ${{ inputs.token }}
+        LABEL: ${{ inputs.label }}
+        REPO: ${{ github.repository }}
+        WORKFLOW: ${{ github.workflow }}
+        REF_NAME: ${{ github.ref_name }}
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #703, opened 2026-08-01
feat(cli): accept on/off/1/0 for boolean flags via BoolishValueParser

nh13:
Follow-up to the question raised in #690: *should we prefer `BoolishValueParser`?*

## What changes

`parse_bool` is replaced by clap's [`BoolishValueParser`](https://docs.rs/clap/latest/clap/builder/struct.BoolishValueParser.html) at all 44 boolean `#[arg]` sites, and the function is deleted. Each of those 44 sites also gains `value_name = "true|false"`, extending repo-wide the convention #690 established for `Sort`'s three flags.

| | accepted (case-insensitive) |
|---|---|
| before | `true` `false` `yes` `no` `y` `n` `t` `f` |
| after | those, plus `on` `off` `1` `0` |

A strict superset, verified against `clap_builder-4.6.2`: `TRUE_LITERALS` and `FALSE_LITERALS` contain all eight previous spellings. Nothing that parsed before parses differently.

Unrecognized input is still **rejected**. Worth stating explicitly because clap's own doc comment on `str_to_bool` claims "Any other value will be considered as `true`" — that comment is wrong. The function returns `None`, and `BoolishValueParser::parse_ref` maps that to a validation error. Confirmed on the binary: `--memory-per-thread 10G` errors.

## The tradeoff, stated plainly

`parse_bool` documented itself as *"Matches sopt/fgbio behavior"*, so this deliberately diverges from sopt's accepted set — that is the one real consequence, and it is the point of the change rather than a side effect. The flags now take the spellings users coming from other CLIs reach for first.

The other cost named in #690 was the error message. Ours was `Invalid boolean value '10G'. Expected: true|false|yes|no|y|n|t|f`; clap's is `invalid value '10G' for '--memory-per-thread [<true|false>]': value was not a boolean`. Less enumerative, but it names the offending flag, which ours never did.

## `hide_possible_values` and the placeholder

These two are one decision, which is why the `value_name` sweep rides here rather than in a follow-up.

`BoolishValueParser::possible_values()` yields all twelve literals but marks ten hidden, so help would print `[possible values: true, false]` — advertising two of the twelve accepted spellings, which is worse than advertising none. Each site therefore sets `hide_possible_values = true`.

That suppression makes the placeholder the only in-help signal about what the flag takes, and without `value_name` clap derives it from the field name. `--memory-per-thread [<MEMORY_PER_THREAD>]` reads as a request for a per-thread memory size and invites `--memory-per-thread 10G`. #690 fixed that for the three flags on `Sort`; the remaining 41 were left rendering the derived name, so the *same* flag rendered two different ways depending on which command you asked — `QueueMemoryOptions::memory_per_thread` is flattened into `dedup`, `group`, `filter`, `clip`, `correct`, and the consensus callers. `--allow-unmapped` was worse still: it set `value_name = "ALLOW_UNMAPPED"` explicitly.

Verified on the built binary across `sort`, `filter`, `dedup`, `group`, `clip`, `correct`, and `simplex`: every one now renders `--memory-per-thread [<true|false>]`, no `[possible values: ...]` line appears anywhere, and no upper-snake placeholder survives on a boolean flag.

## Tests

The two rstest tables that called `parse_bool` directly now drive the same cases through `TestBoolFlags::try_parse_from`, asserting the CLI contract rather than a private function, so they survive another change of parser.

`on`/`off`/`1`/`0` **moved from the rejected table to the accepted one**. That migration is the behaviour change, made explicit rather than implied — a third table, `test_extended_bool_values_in_cli_invalid`, asserted those four were rejected, and it failed on the first run of this change exactly as it should have.

`test_bool_args_name_their_accepted_values` (added by #690) still guards `Sort` only, and it identifies boolean args by the structural proxy `num_args = 0..=1` rather than by their parser. Both are addressed in a follow-up PR stacked on this one: `Arg::get_possible_values()` now surfaces `BoolishValueParser`'s literal set, so the test can identify boolean args exactly and walk every command instead of `Sort`.

`cargo ci-fmt` / `ci-lint` clean.

## Note for the runall landing series

`feat-runall` and `main-runall` are deliberately **not** changed here. `main-runall` inherits this for `src/lib/commands/` on its next rebase, but two copies of `parse_bool` survive on that side and must not become the live parser — `crates/fgumi-cli-common/src/lib.rs:526` (landed by #672) and `crates/fgumi-sort-cli/src/sort.rs` (lands at P4). Both are recorded in the landing tracker's duplication ledger so the port converts them rather than reintroducing the function.

Risk: command output changes: none; `unsafe` changes: none, and `CLAUDE.md` allowlist updates: none; memory bounds, queue capacity, and thread/backpressure policy changes: none.

Replaced the custom `parse_bool` parser at 44 CLI boolean argument sites with clap’s `BoolishValueParser`.

Boolean arguments now accept `true`, `false`, `yes`, `no`, `y`, `n`, `t`, `f`, `on`, `off`, `1`, and `0`. Invalid values remain rejected.

Standardized boolean help output with `value_name = "true|false"` and hidden possible values.

Updated CLI parsing tests for the expanded value set and invalid input. Formatting, lint, tests, and help rendering pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Features
+
+- Boolean command-line flags now additionally accept `on`, `off`, `1`, and `0`, alongside the
+  existing `true`/`false`/`yes`/`no`/`y`/`n`/`t`/`f`. All spellings remain case-insensitive, and
+  values outside that set are still rejected.
+- Boolean command-line flags now render a `true|false` placeholder in help, so `fgumi filter -h`
+  shows `--memory-per-thread [<true|false>]` rather than `--memory-per-thread
+  [<MEMORY_PER_THREAD>]`, which read as a request for a per-thread memory size.
+
 ## [0.5.0] - 2026-07-28
 
 ### Bug Fixes
--- a/src/lib/commands/clip.rs
+++ b/src/lib/commands/clip.rs
@@ -35,7 +35,7 @@ use std::sync::atomic::{AtomicU64, Ordering};
 use super::command::Command;
 use super::common::{
     BamIoOptions, CompressionOptions, QueueMemoryOptions, SchedulerOptions, ThreadingOptions,
-    build_pipeline_config, parse_bool,
+    build_pipeline_config,
 };
 
 /// Clips reads in a BAM file to remove overlaps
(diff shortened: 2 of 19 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #714, opened 2026-08-05
feat(bgzf): add slice decompression and caller-driven buffer recycling

nh13:
Phase 2 of the `feat-runall` landing series, on top of #697. Three additive entry points in `fgumi-bgzf` that later phases consume, landed ahead of their callers so the crate arrives complete rather than growing under each subsequent port.

**Do not read this as a port of `feat-runall`'s `fgumi-bgzf`.** `git diff main-runall feat-runall -- crates/fgumi-bgzf` reads +512/−578, but almost all of the deletions are `main` work that `feat-runall` predates: `header.rs` in full (#659's single BGZF header predicate), the `Cargo.toml` conversion to workspace dependencies, and every CHANGELOG entry from 0.3.1 through 0.5.0. Applying that diff would silently revert #614 and #659. The three additions were hand-ported instead, and `BGZF_HEADER_SIZE` is taken from `header.rs` rather than reintroduced as a local constant.

### What lands

`decompress_into_slice` — the fixed-slice analogue of `decompress_block_slice_into`. It decompresses a block straight into a caller-sized `&mut [u8]` rather than appending to a `Vec`, so the sort ingest path can decompress directly into an arena slot instead of into a staging buffer it then copies out of. Same integrity guarantee as its siblings: the payload must exactly fill `out` and match the footer CRC32, and level-0 input takes the same deflate stored-block fast path.

`parse_stored_frame` — extracted, not new. `copy_stored_and_verify` had its framing checks inlined in its body, so the slice variant would have had to duplicate them. Both copy paths now call the one parser, which is what keeps the `Vec` and slice entry points from drifting apart on what counts as a well-formed stored frame.

`uncompressed_size` — the accessor that makes `decompress_into_slice`'s contract safe to satisfy. A caller has to size the slot *before* calling, and ISIZE is a `u32` sitting in the file, so sizing straight off the footer means allocating from unvalidated input — up to 4 GiB from a corrupt block. This reads the footer off the same `&[u8]` (no owned `Vec`, so no copying the block to read four bytes) and bounds the claim to `MAX_UNCOMPRESSED_BLOCK_SIZE`. `decompress_into_slice` resolves the size through it too, so what a caller allocates and what the decompressor accepts cannot diverge.

`parse_stored_frame`, `is_stored_block`, `deflate_into_slice_and_verify` — extracted, not new. `copy_stored_and_verify` had its framing checks inlined in its body, so the slice variant would have had to duplicate them. The BTYPE dispatch predicate and the inflate-then-verify tail were duplicated for the same reason. All three are now shared with `decompress_and_verify`, so the `Vec` and fixed-slice entry points cannot drift on what a well-formed stored frame is or on the exact-fill invariant.

`InlineBgzfCompressor::recycle_buffer` — lets a `take_blocks` consumer hand a drained block buffer back to the pool. Only `write_blocks_to` recycled before, so a consumer driving the compressor with `write_all` + `flush` + `take_blocks` left the pool permanently empty and allocated a fresh output `Vec` for every block.

`pub use libdeflater::Decompressor` — so a consumer can name the type every `decompress_*` entry point takes without declaring its own `libdeflater` dependency, and so the version it names is necessarily the one this crate decompresses with.

### Two things here are not purely additive

Both fall out of the additions rather than being bundled with them, but they are behaviour changes and should be reviewed as such.

**`write_blocks_to` now bounds the pool and survives a write error.** Routing it through `recycle_buffer` means draining into a temporary rather than holding a borrow on `completed_blocks`, since `recycle_buffer` takes `&mut self`. Two consequences. The pool is now capped at `MAX_POOLED_BUFFERS`; previously this method pushed one buffer per block written with nothing capping the growth, so the "bounded" claim in `recycle_buffer`'s docs would have been false the moment the two paths disagreed. And the write-error path became recoverable, so it is handled: the failing block and its tail are restored to the queue instead of being dropped by the `Drain` guard. The old `for block in self.completed_blocks.drain(..) { output.write_all(&block.data)?; }` silently discarded every unwritten block on error, and the error says nothing about how far the write got.

**`decompress_into_slice` validates `out.len()` against the footer ISIZE up front.** The doc stated this precondition; nothing enforced it. Both decompress paths compare against `out.len()` on the assumption that it *is* the ISIZE, so a mis-sized slot was reported as a fault in the block — the stored path claimed `BGZF stored block ISIZE mismatch: footer=105, LEN=100` for a block whose footer says 100. That names a value the footer does not contain and sends a reader after file corruption that isn't there, when the defect is in the arena that sized the slot.

**`write_blocks_to`'s retained blocks are documented as safe to inspect, not to replay.** `io::Write::write_all` loops over `write`, advancing past each `Ok(n)`, so it can commit part of the failing block before an error surfaces — and the whole block is re-queued, not the unwritten remainder. Writing the queue again would repeat those bytes inside a gzip member and produce a stream no BGZF reader can decode.

### Both new functions land without callers

`decompress_into_slice` gets its first consumer at P3 (arena slots) and P4 (`fgumi-pipeline-io`'s `spill_decompress`). `recycle_buffer`'s eventual consumer is the pipeline BGZF compress step.

Worth naming explicitly: `crates/fgumi-sort/src/worker_pool.rs:2193` is *already* a `take_blocks` consumer of exactly the shape `recycle_buffer` was written for, and it still allocates a fresh `Vec` per block. It is left alone here because P3 rewrites that file wholesale; wiring it now would be undone in two phases. That gap is sequencing, not an oversight.

### Verification

`cargo ci-fmt`, `cargo ci-lint`, `cargo ci-doc` clean; **7554 tests pass**, 27 skipped.

Every guard here was mutation-tested rather than assumed — revert the fix, confirm a named test fails:

| Mutation | Test that fails |
|---|---|
| Delete the stored-path CRC/size verify | `case_2_bad_crc_stored` |
| Make the exact-fill check tautological | `case_3_short_fill` |
| Remove the ISIZE upper bound | `case_4_isize_above_max` |
| Off-by-one on that bound (`>` → `>=`) | `uncompressed_size_accepts_the_block_maximum` |
| Bypass the shared size accessor | `case_4_isize_above_max`, `case_7_too_short_block` |
| Drop the pool capacity bound | `test_recycle_buffer_refuses_an_oversized_buffer` |
| Pop-and-discard instead of reusing | `test_recycle_buffer_repopulates_the_pool` |
| Let `write_blocks_to` bypass the cap | `test_write_blocks_to_recycles_up_to_the_cap` |

The first two matter most: both mutations previously left the whole suite green, so neither integrity check this crate advertises was actually pinned.

Two constants were measured, not assumed. `MAX_POOLED_BUFFER_BYTES` (130560) against real block buffers, which are 65376 bytes at levels 0/6/12 on incompressible input — so no legitimate buffer is refused. And dropping the pop-site `clear()` is safe because `bgzf` 0.4's `resize_uninit` opens with `Vec::clear`, i.e. the compressor clears unconditionally.

### Still outstanding from P1

`fgumi-sort` does not yet have the `test-utils` feature. The tracker files it under P1, but it is inert until P3 gates `chunk_sorter.rs` on it, so it belongs in that phase rather than shipping here as a dead feature flag.

## Summary by CodeRabbit

- **New Features**
  - Added direct decompression into caller-provided buffers.
  - Added utilities for checking uncompressed block sizes.
  - Exposed decompression and buffer-recycling capabilities for advanced integrations.
  - Added safeguards limiting uncompressed BGZF blocks to 64 KiB.

- **Bug Fixes**
  - Improved validation of block sizes, framing, output lengths, and CRC checks.
  - Write failures now preserve queued data for retry.

- **Performance**
  - Reuses compression buffers to reduce unnecessary allocations.

### DIFF
--- a/crates/fgumi-bgzf/src/lib.rs
+++ b/crates/fgumi-bgzf/src/lib.rs
@@ -12,12 +12,19 @@ pub mod reader;
 pub mod writer;
 
 // Re-export commonly used types
+
 pub use header::{
     BGZF_HEADER_SIZE, HeaderRejection, MIN_BLOCK_SIZE, block_size, block_size_checked,
     is_bgzf_header,
 };
+/// The libdeflater decompressor every `decompress_*` entry point here takes.
+///
+/// Re-exported so a consumer can name the type it must construct and reuse
+/// without also declaring a direct `libdeflater` dependency — and so the
+/// version it names is necessarily the one this crate decompresses with.
+pub use libdeflater::Decompressor;
 pub use reader::{
     BGZF_EOF, BGZF_FOOTER_SIZE, RawBgzfBlock, decompress_block, decompress_block_into,
-    decompress_block_slice_into, read_raw_blocks,
+    decompress_block_slice_into, decompress_into_slice, read_raw_blocks, uncompressed_size,
 };
 pub use writer::{BGZF_MAX_BLOCK_SIZE, CompressedBlock, InlineBgzfCompressor};
--- a/crates/fgumi-bgzf/src/reader.rs
+++ b/crates/fgumi-bgzf/src/reader.rs
@@ -47,6 +47,21 @@ pub use crate::header::BGZF_HEADER_SIZE;
 /// Size of the BGZF block footer (CRC32 + ISIZE).
 pub const BGZF_FOOTER_SIZE: usize = 8;
 
+/// Largest uncompressed payload a single BGZF block can hold.
+///
+/// A block's `BSIZE` is a `u16` holding the total size minus one, which caps a
+/// block at 64 KiB; the SAM/BGZF spec and htslib's `BGZF_MAX_BLOCK_SIZE` both
+/// use this bound, and the `bgzf` crate refuses to *write* a larger one
+/// (`BlockSizeExceeded`). The footer's ISIZE is a `u32`, so a corrupt or
+/// hostile block can claim up to 4 GiB; anything a caller sizes from that value
+/// has to be bounded first.
+///
+/// Note this is deliberately 64 KiB rather than [`crate::writer::BGZF_MAX_BLOCK_SIZE`]
+/// (65280), which is the size *we* fill blocks to. Other writers legitimately
+/// emit up to the spec limit, so validating against our own chunk size would
+/// reject valid input.
+pub const MAX_UNCOMPRESSED_BLOCK_SIZE: usize = 64 * 1024;
+
 /// BGZF EOF marker block (empty block signaling end of file).
 pub const BGZF_EOF: [u8; 28] = [
     0x1f, 0x8b, 0x08, 0x04, 0x00, 0x00, 0x00, 0x00, 0x00, 0xff, 0x06, 0x00, 0x42, 0x43, 0x02, 0x00,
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed in `bf39a451`.

`decompress_into_slice` now returns `Ok(0)` when the footer's ISIZE is zero, and there are two regression tests: `decompress_into_slice_accepts_the_eof_marker` and `decompress_into_slice_rejects_a_sized_slot_for_an_empty_block`.

One correction on the diagnosis, since it changes what the change is for. The finding states that with an empty `out`, `deflate_decompress` returns `InsufficientSpace`, mapped to `InvalidData`. It does not — it returns `Ok(0)`, and `decompress_into_slice(&BGZF_EOF, .., &mut [])` already returned `Ok(0)` before this change. Probed against the built crate:

```
PROBE isize=Ok(0)
PROBE compressed=[03, 00] is_stored=false
PROBE decompress_into_slice(BGZF_EOF, out=[]) -> Ok(0)
PROBE raw deflate_decompress(payload, &mut []) -> Ok(0)
```

Removing the new short-circuit leaves all 87 crate tests green, which is the same thing said another way: this is not repairing a failure.

It is still worth taking, for the reason the finding is right about — `03 00` is a fixed-Huffman frame, so `is_stored_block` is false and the EOF marker does reach libdeflater with a zero-length output slice. Resting on an undocumented edge of the C library for a block every stream ends with is a bad bet across libdeflater versions, and both `Vec` siblings already short-circuit here, so the three entry points now agree.

One deviation from the suggested placement. The prompt says to return before the stored/DEFLATE dispatch while preserving the output-length validation; those two conflict if taken literally, because the size check sits between them. Placing the short-circuit ahead of the size check makes a caller passing a wrongly-sized slot for a zero-ISIZE block get a silent `Ok(0)` instead of being told the slot is wrong — the exact misreport that check exists to prevent. It goes after the check instead, and `decompress_into_slice_rejects_a_sized_slot_for_an_empty_block` fails if it is moved ahead of it (verified by mutation).

nh13:
@coderabbitai review

PR #716, opened 2026-08-05
test(streaming): reap the `cat` children the streaming tests spawn

nh13:
`tests/integration/test_streaming_input.rs` pipes a `cat` child's stdout into an fgumi command's stdin at five sites. None of them waited on the child. Two things kept that quiet: a crate-level `#![allow(clippy::zombie_processes)]`, which suppressed the lint that exists to catch exactly this, and a module doc comment asserting the opposite of what the code did —

> These tests spawn cat processes whose stdout is piped to fgumi commands. The child processes are properly cleaned up when their stdout is consumed.

Consuming a child's stdout does not reap it. An unwaited `cat` can outlive the test that spawned it while still holding the stderr pipe it inherited from the test binary, which is what nextest watches for when it marks a test leaky.

### The change

A `CatPipe` guard owns the child and waits in `Drop`, so none of the five call sites can forget:

```rust
let (_cat, cat_stdout) = CatPipe::spawn(&input_bam);
```

The `wait` does not block. By the time a guard drops, the consuming command has exited and closed the read end, so `cat` sees EPIPE and exits on its own.

### Dropping the `allow` is the part that keeps it fixed

The guard alone would be a fix that decays — nothing stops the next call site from being written the old way. With the suppression gone, that fails the build. Verified by reverting one call site to the open-coded form:

```
error: spawned process is never `wait()`ed on
   = help: ... #index.html#zombie_processes
```

And removing the `wait` from `Drop` fails too, though via `dead_code` (`field 0 is never read`) rather than `zombie_processes` — the `Child` becomes a field nothing reads.

### On the leak this came out of

This was found while chasing an intermittent `(1 leaky)` nextest result, and it is the most plausible mechanism I identified for it. It is **not** established as the cause: I could not reproduce the symptom across 11 consecutive runs, so there is no before/after evidence tying this fix to it. That investigation stays open.

It stands on its own regardless — an unreaped child process in a test, a lint suppressed rather than satisfied, and a comment claiming the opposite of the code.

### Verification

`cargo ci-fmt` and `cargo ci-lint` clean; **6967 tests pass**, 27 skipped, no leaky results. The 35 `test_streaming_input` tests pass in 1.3s, so no guard is blocking on its `wait`.

Risk: command output changes none; `unsafe` changes none; memory bounds, queue capacity, and thread/backpressure policy changes none. Fix: streaming tests now reap spawned `cat` processes with a `CatPipe` RAII guard.

- Added `CatPipe` to own each child process and wait for it on drop.
- Replaced open-coded process handling in streaming input tests.
- Removed the crate-level `clippy::zombie_processes` suppression.
- Corrected module documentation.
- Added baseline record assertions and reusable `/dev/stdin` arguments.
- Validation: 6,967 tests passed, 27 skipped, no leaky results, and all 35 streaming input tests passed.

### DIFF
--- a/tests/integration/test_streaming_input.rs
+++ b/tests/integration/test_streaming_input.rs
@@ -1,20 +1,50 @@
 //! Integration tests for streaming input support (stdin/pipes).
 //!
-//! These tests spawn cat processes whose stdout is piped to fgumi commands.
-//! The child processes are properly cleaned up when their stdout is consumed.
-#![allow(clippy::zombie_processes)]
+//! These tests pipe a `cat` child's stdout into an fgumi command's stdin. Every
+//! such child is owned by a [`CatPipe`], which waits on it, so none outlives the
+//! test that spawned it.
 
 use noodles::bam;
 use noodles::sam::alignment::io::Write as AlignmentWrite;
 use std::fs::{self, File};
 use std::io::{BufReader, Read};
-use std::path::PathBuf;
-use std::process::{Command, Stdio};
+use std::path::{Path, PathBuf};
+use std::process::{Child, ChildStdout, Command, Stdio};
 
 use rstest::rstest;
 use tempfile::TempDir;
 
 use crate::helpers::bam_generator::{create_minimal_header, create_umi_family, to_record_buf};
+
+/// A `cat <path>` child whose stdout is piped, reaped when the guard drops.
+///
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #717, opened 2026-08-05
docs(best-practices): clarify mate tags and update aligner recommendations

### LINKED ISSUE #713: "Best Practices" specifies BWA v0.7.17 or later but MQ tag wasn't added until 0.7.18
### Description

...as @nh13 knows since he added it :)

Not a bug or a feature request, just a request for doc clarification. The BP explicitly says samtools isn't required; but if the aligner used doesn't output both `MC` and `MQ` tags, including versions of BWA older than 0.7.18, then am I correct that either samtools `fixmate` or fgbio `SetMateInformation` (or another similar tool) would be required for fgumi `group` to work? In which case perhaps this should be noted in the BP? (Unless I'm mistaken and `group` has the ability to fetch `MC` and `MQ` on the fly, unlike fgbio `GroupReadsByUmi`, in which case explicitly mentioning that in the BP might also be worthwhile.)

BWA 0.7.19 fixed a header bug in 0.7.18 so probably the BP should be updated to just recommend 0.7.19.

### Related Issues

_No response_

### Version

0.5.0

### Impact

Adding this feature could have the following drawbacks...

Adding this feature will introduce a breaking change...

Adding this feature will require updating the documentation...

### Suggestions

This feature could be implemented in a few different ways...

This feature should be tested in the following ways...

### Test input

_No response_

### Checklist

- [ ] Checked there are no existing Issues for this feature request
- [ ] Checked there are no Discussions that address this issue
- [ ] Ensured that the latest release does not contain this feature
- [ ] Provided all information necessary to reproduce the current behavior

nh13:
Closes #713.

The Best Practices page said "no samtools is required" and recommended bwa 0.7.17 without saying anything about mate information, which left open the question the issue raises: does a pipeline need `samtools fixmate` / fgbio `SetMateInformation` when the aligner does not write `MC` and `MQ`?

It does not. `fgumi zipper` recomputes mate information for every template it emits — `RNEXT`, `PNEXT`, `TLEN`, and the `MC`, `MQ`, and `ms` tags — from the alignments themselves, so the aligner's tag output is irrelevant in the recommended pipeline. Verified end to end against a mapped SAM with no `MC`/`MQ` and no mate coordinates:

```
# input (mapped.sam)
q1  99  chr1  101  60  50M  *  0  0  ...  AS:i:50
q1 147  chr1  301  60  50M  *  0  0  ...  AS:i:50

# fgumi zipper output
q1  99  chr1  101  60  50M  =  301   250  MQ:i:60  MC:Z:50M  ms:i:50  RX:Z:ACGT  AS:i:50
q1 147  chr1  301  60  50M  =  101  -250  MQ:i:60  MC:Z:50M  ms:i:50  RX:Z:ACGT  AS:i:50
```

The path that does depend on pre-existing mate information is skipping `zipper` and feeding an externally aligned BAM into the rest of the pipeline. Two steps read mate tags and neither errors when they are missing, so both are now documented:

- `fgumi group` applies `--min-map-q` to each read's own MAPQ *and* to the `MQ` tag, and a read with no `MQ` tag skips the mate half of that check.
- `fgumi simplex` / `fgumi duplex` clip bases extending past the mate via `num_bases_extending_past_mate_raw`, the `MC`-tag variant, which returns 0 when the tag is absent (`crates/fgumi-raw-bam/src/overlap.rs`) — so overlapping bases get counted twice in the consensus, silently.

While here, the aligner list is updated:

- **bwa mem 0.7.19** rather than 0.7.17. `MC` has been written since 0.7.16 (lh3/bwa@3b96dce, in v0.7.16) but `MQ` only since 0.7.18 (lh3/bwa@d8dd308, in v0.7.18) — so the issue's premise is right — and 0.7.19 fixed the internal `@HD` header bug introduced in 0.7.18.
- **bwa-mem3 0.8.0+** is now listed as an alternative, with a worked `--bam` alignment example, and the interaction with `--compat=bwa-mem2` is spelled out: that target suppresses `MQ` (plus `HN:i` and the default `@HD`) for byte-identity with bwa-mem2 v2.2.1, which is harmless here because `zipper` re-derives `MQ`, and it changes no alignments.

Docs-only; no code changes.

Risk: command output changes—none; `unsafe` changes—none, and CLAUDE.md allowlist changes—none; memory, queue, and thread/backpressure changes—none. Fix: documentation only.

- Clarifies that `fgumi zipper` recomputes mate coordinates and the `MC`, `MQ`, and `ms` tags.
- Removes the separate `samtools fixmate` or `fgbio SetMateInformation` requirement from the recommended pipeline.
- Documents mate-information requirements for externally aligned BAM files.
- Updates recommendations to BWA 0.7.19+ and bwa-mem3 0.8.0+, including bwa-mem3 usage details.

### DIFF
--- a/docs/src/guide/best-practices.md
+++ b/docs/src/guide/best-practices.md
@@ -7,9 +7,85 @@ This document describes the recommended best practice pipeline for processing FA
 This pipeline uses only fgumi and a read aligner:
 
 - **fgumi** (version 0.1 or higher)
-- **bwa mem** (version 0.7.17 or higher recommended)
+- a read aligner — one of:
+  - **[bwa-mem3](https://github.com/fg-labs/bwa-mem3)** (version 0.8.0 or higher), a drop-in
+    bwa-mem2 replacement that can write uncompressed BAM (`--bam`) straight into `fgumi zipper`
+  - **bwa mem** (version 0.7.19 or higher recommended)
 
 Unlike fgbio-based pipelines, **no samtools is required** - fgumi provides native `fastq`, `sort`, and `merge` commands.
+
+### Mate Information and Aligner Version
+
+No separate mate-fixing step (`samtools fixmate`, fgbio `SetMateInformation`) is required
+either, for any aligner or aligner version. `fgumi zipper` recomputes mate information for
+every template it emits — `RNEXT`, `PNEXT`, `TLEN`, and the `MC` (mate CIGAR), `MQ` (mate
+mapping quality), and `ms` (mate score) tags — from the alignments themselves, so it does not
+matter whether the aligner wrote those tags. (`ms` is the one exception to "from the alignments
+themselves": it carries the mate's `AS` value, so it is written only when the aligner scored the
+mate with `AS`, which every supported aligner does.)
+
+The version recommendations are primarily about what each aligner emits natively:
+
+- `bwa mem` has written `MC` since 0.7.16 but only added `MQ` in 0.7.18.
+- `bwa-mem3` writes both `MC` and `MQ` by default. `--compat=bwa-mem2`, which shapes output to
+  be byte-identical to bwa-mem2 v2.2.1, suppresses `MQ` (along with the `HN:i` tag and the
+  default `@HD` line); that is safe in this pipeline for the same reason — `fgumi zipper`
+  re-derives `MQ` — and it changes no alignments.
+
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #722, opened 2026-08-07
docs: check private-item doc links in CI and fix the 30 that had accumulated

nh13:
## What

`cargo ci-doc` builds the workspace docs with `RUSTDOCFLAGS="-D warnings"`, which reads as "every rustdoc warning is an error." It was not. Rustdoc only checks intra-doc links on items it actually *renders*, and without `--document-private-items` that means public items only — so a broken link in the doc comment of a private fn, a private field, or a `#[cfg(test)]` helper never reached the linter.

Thirty such links had accumulated across the workspace, all of them invisible to a green docs job. This adds the flag to the `ci-doc` alias and fixes every link it surfaces, so the job now checks what its name implies.

## Why it matters

This is a gap in the check, not a pile of typos. It was found only because a broken link in `crates/fgumi-sort/src/inline.rs` was caught by review on #720 — CI could not have caught it, and could not catch the other twenty-nine either. Closing the gap is the point; the fixes are the cost of closing it.

None of the thirty were reachable defects. The cost was a doc build that quietly under-checked itself, plus links that render as literal brackets instead of hyperlinks.

## The four kinds

- **Not in scope at the link site**, so the path was required — `PipelineError::*`, `EdgeMetrics`, `MAX_ARITY`, `TwoInputHandles`, `SortWorkerPool::phase1_queue_depths`, `PooledBamWriter::new_indexing`.
- **Associated fns linked bare** instead of through `Self::` — `merge_chunks_generic`, `merge_chunks_with_index`.
- **A symbol in another crate** — `read_raw_record` lives in `fgumi-raw-bam`, not `fgumi-sort`.
- **Prose that rustdoc parses as a reference-style link** — byte-layout notation `[len(4)][record(len)]`, index expressions `buffer[0]` and `R1[i]`, and `[`X`](path)` pairs whose explicit target is redundant because the link text already resolves.

## Scope

Touches `fgumi-sort`, `fgumi-pipeline-core`, `fgumi-bam-io`, and the main crate. Docs and one cargo alias only — no functional change.

Based on `main-runall` rather than `main` because eight of the fixes are in `fgumi-pipeline-core`, which does not exist on `main`.

Three further instances of the same class, in files that #720 introduces, are fixed in #720 itself rather than here.

## Verification

- `RUSTDOCFLAGS="-D warnings" cargo ci-doc` — clean workspace-wide *with* `--document-private-items`, which fails on the parent commit.
- `cargo ci-fmt`, `cargo ci-lint`, `cargo ci-tag-literals` — clean.
- `cargo ci-test` — 7835 passed, 30 skipped.

Risk: command output changes — none; `unsafe` changes — none, and `CLAUDE.md` is unchanged; memory bounds, queue capacity, and thread/backpressure policy changes — none.

Fix: `cargo ci-doc` now checks private items and test helpers with `--document-private-items`.

- Corrected 30 broken or misparsed Rustdoc links and code formatting issues.
- Updated documentation across sorting, pipeline, BAM I/O, and command modules.
- No functional or public API changes.
- Verification passed: 7,835 tests passed and 30 were skipped.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -7,7 +7,14 @@ ci-tag-literals = "run --package xtask -- check-tag-literals"
 ci-doctest = "test --doc --workspace --features compare,simulate,profile-adjacency --locked"
 # Build the docs and fail on any rustdoc warning (e.g. broken intra-doc links).
 # RUSTDOCFLAGS="-D warnings" is set by the caller (CI job / pre-push).
-ci-doc = "doc --no-deps --workspace --features compare,simulate,profile-adjacency --locked"
+#
+# `--document-private-items` is load-bearing, not cosmetic. Without it rustdoc
+# only checks links on items it renders — i.e. public ones — so a broken link in
+# the doc comment of a private fn, a private field, or a `#[cfg(test)]` helper
+# passes silently. That is not hypothetical: several such links accumulated
+# undetected and had to be found by hand. The flag makes the docs job check every
+# doc comment in the workspace, which is what the job's name implies it does.
+ci-doc = "doc --no-deps --workspace --document-private-items --features compare,simulate,profile-adjacency --locked"
 # Run only the #[ignore]-d integration tests (the sort-correctness suites in
 # tests/integration/{test_sort_correctness,test_async_reader,test_sort_write_index}.rs).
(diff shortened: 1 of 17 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

Issue #724, opened 2026-08-07
Reads dropped by consensus downsampling are counted as used and never reach --rejects


nh13:
### Problem

The consensus downsampling paths discard reads without recording them: neither `downsample_source_reads` nor `downsample_reads` (`crates/fgumi-consensus/src/vanilla_caller.rs`) nor the codec equivalent (`codec_caller.rs:710`) calls `record_rejection` or appends to `rejected_reads`.

Since `raw_reads_used = total_input_reads - filtered_reads` (`crates/fgumi-metrics/src/consensus.rs:392`) and only `record_rejection` increments `filtered_reads`, a family of 10 capped to 3 reports `raw_reads_used = 10` while the consensus carries `cD:i:3`. The 7 discarded reads also never appear in the `--rejects` BAM.

### Fix

Add a `RejectionReason::Downsampled` variant (`crates/fgumi-consensus/src/caller.rs:397`) and, at each downsample site, record the discarded count and — when `track_rejects` is set — append the discarded raw records.

The metrics side is straightforward: fgumi keeps a flat `rejection_reasons` map with no per-caller "used by" tables. The plumbing is the cost — the cap is applied inside `consensus_call`, which receives only `SourceRead`s (which carry `original_idx`, not raw bytes) and never sees the original `Vec<RawRecord>`. `process_subgroup` does have both in scope and already emits `group_reads[sr.original_idx]` for other rejection reasons, so one option is to hoist the cap up to that level.

### Related

Pre-existing and shared with fgbio: called out as a known gap in fulcrumgenomics/fgbio#1166 and now tracked as fulcrumgenomics/fgbio#1167. Fixing it in fgumi first means `raw_reads_used` diverges from fgbio under a cap until fgbio does the same, so it may be worth landing the two together.

Issue #723, opened 2026-08-07
simplex --max-reads caps the whole MI group instead of each end


nh13:
### Problem

`fgumi simplex` applies `--max-reads` to the entire MI group before splitting it into fragment/R1/R2 subgroups (`crates/fgumi-consensus/src/vanilla_caller.rs:1163`), whereas fgbio's `CallMolecularConsensusReads` applies the cap independently to each subgroup inside `consensusCall`. A cap of N therefore admits N reads *total* across both ends in fgumi, but N reads *per end* in fgbio.

Because the retained reads are split arbitrarily between R1 and R2, the effective per-end depth is roughly halved, and families whose split leaves one end empty produce no consensus pair at all.

### Measured

Same grouped BAM (~77k molecules, IDT cfDNA library prep), `--max-reads 2` / `-M 1`:

| | consensus reads | `cD` distribution |
|---|---|---|
| fgbio 4.1.0 | 734,536 | `1`: 702,047 · `2`: 32,489 |
| fgumi | 726,854 | `1`: 726,854 |

fgumi uncapped emits 734,536 reads, matching fgbio's capped total — so the 7,682 missing reads are families the group-level cap collapsed to a single end, and every surviving consensus is built from one read instead of up to two.

### Fix

`process_subgroup` already builds per-end `SourceRead`s and calls `consensus_call`, which applies its own per-end cap (`vanilla_caller.rs:687`). Removing the group-level `downsample_reads` call should give per-end semantics directly. duplex and codec do not route through `process_group` and are unaffected.

This changes output for anyone setting `--max-reads` on `simplex` and needs a release note.

### Related

fulcrumgenomics/fgbio#1166 makes fgbio's consensus downsampling deterministic and documents the cap as "applied independently to each end of the family". That wording is accurate for fgbio but not for fgumi simplex until this is fixed. The fgumi port of that PR is tracked separately; this issue is fgumi-only and has no fgbio counterpart.

PR #727, opened 2026-08-08
fix(consensus)!: make consensus read downsampling deterministic

### LINKED ISSUE #725: Consensus downsampling is not reproducible across execution modes and cannot match fgbio
### Problem

Consensus downsampling selects which reads to retain by shuffling with a seeded `StdRng` held on the caller (`crates/fgumi-consensus/src/vanilla_caller.rs:375`, `codec_caller.rs:340`). The RNG is stateful, so which reads survive a given family depends on how many families that caller instance already downsampled.

fgumi does not have fgbio's thread-count bug: the unified pipeline builds a fresh caller per batch with a fixed batch size, so `--threads 1` and `--threads 8` agree. Verified byte-identical on a ~77k-molecule grouped BAM at `--max-reads-per-strand 2`.

It does have the analogous bug **across execution modes**. The no-`--threads` path holds one caller for the entire run (`src/lib/commands/simplex.rs:378`, `duplex.rs:420`), so its RNG is at a different position for every family than the per-batch callers are:

- duplex `--max-reads-per-strand 2`: 566 of 77,804 consensus records differ (0.73%) between `--threads N` and no `--threads`
- simplex `--max-reads 2`: outputs differ
- with no cap, both modes are byte-identical, so the divergence is cap-specific

Two further consequences of RNG-based selection:

- **Both ends of a template are downsampled independently.** duplex calls `consensus_call` four times (`duplex_caller.rs:1992-1995`), once per strand/end, so R1 and R2 sample different templates. codec is already immune (it downsamples R1 and R2 through one shared index vector).
- **Selection cannot match fgbio.** fgumi shuffles with `StdRng` (ChaCha) while fgbio uses `java.util.Random` (an LCG), so the same seed gives a different subset. Measured against fgbio 4.1.0 at `--max-reads-per-strand 2`: 487 of 77,804 duplex consensus reads (0.63%) differ in SEQ/QUAL. Depth distributions match exactly, so selection is the only difference. This is already documented as a known limitation in `downsample_source_reads`.

### Fix

Port fulcrumgenomics/fgbio#1166: rank reads by a Murmur3 hash of the read name and keep the lowest-ranking ones. The retained subset becomes a pure function of the family — independent of caller history, execution mode, and the order reads arrive in — and because mates share a read name they share a rank, so both ends are retained or discarded together.

fgumi already contains a byte-exact port of htsjdk's `Murmur3.hashUnencodedChars`, validated against reference vectors captured from htsjdk 3.1.2 (`src/lib/commands/shared_metrics.rs:157`). It needs to move into a crate the consensus caller can see.

Once fgbio#1166 merges, this also restores byte parity with fgbio under a cap. A parity oracle test will follow separately rather than blocking on that merge.

Also in scope, mirroring the second commit of fgbio#1166: `fgumi duplex` does not validate `--max-reads-per-strand`, so `--max-reads-per-strand 0` exits 0 and writes an empty BAM. simplex and codec already validate `max >= min_reads >= 1`.

### Not in scope

- fgumi#723 — simplex applies `--max-reads` to the whole MI group instead of each end
- fgumi#724 — reads dropped by downsampling are counted as used and never reach `--rejects`

### Behaviour change

Which reads are retained changes for anyone setting `--max-reads`, `--max-reads-per-strand` or codec `--max-reads`, in every execution mode. There is no way to fix the determinism bug while preserving the old selection, since the old selection was the bug. Needs a release note.

nh13:
## Problem

Consensus downsampling selected which reads to keep by shuffling with a seeded `StdRng` held on the consensus caller:

```rust
reads.shuffle(&mut self.rng);
reads.truncate(max_reads);
```

The RNG is stateful, so which reads survive a tag family depends on how many families that caller instance has already downsampled.

fgumi does **not** have fgbio's thread-count bug. The unified pipeline builds a fresh caller per batch at a fixed batch size, so `--threads 1` and `--threads 8` already agreed — I verified this byte-for-byte. But the no-`--threads` path holds a single caller for the entire run (`simplex.rs:378`, `duplex.rs:420`), putting its RNG at a different position for every family. The two execution *modes* therefore disagreed under a cap:

- duplex `--max-reads-per-strand 2`: **566 of 77,804 consensus records differed** (0.73%) between `--threads N` and no-`--threads`
- simplex `--max-reads 2`: differed
- **with no cap, both modes were byte-identical** — the divergence is cap-specific

Two further consequences of RNG-based selection:

- **Both ends of a template were downsampled independently.** duplex calls `consensus_call` once per strand per end (`duplex_caller.rs:1992-1995`), so R1 and R2 sampled different templates. Beyond reproducibility, this silently *lost* reads: families whose split left one end empty produced no consensus pair at all.
- **Selection could not match fgbio.** fgumi shuffles with `StdRng` (ChaCha); fgbio uses `java.util.Random` (an LCG). Against fgbio 4.1.0 at `--max-reads-per-strand 2`, 487 of 77,804 duplex records (0.63%) differed in SEQ/QUAL. Depth distributions matched exactly, so selection was the only difference. This was already documented as a known limitation in `downsample_source_reads`.

## Fix

Rank reads by a Murmur3 hash of the read name and keep the lowest-ranking ones, porting fulcrumgenomics/fgbio#1166. The retained subset becomes a pure function of the family — independent of caller history, execution mode, and arrival order — and because mates share a read name they share a rank, so a template is retained or discarded as a unit.

fgumi already contained a byte-exact port of htsjdk's `Murmur3.hashUnencodedChars`, validated against reference vectors captured from htsjdk 3.1.2. It moves from the root crate down into `fgumi-raw-bam`, where the consensus callers can reach it, alongside the samtools-parity read-name comparator already living there.

Three details are load-bearing for parity, and each has a dedicated test:

- The rank is computed **once per read** and stored on `SourceRead` / `ClippedRecordInfo`, never inside a sort key closure. Rust's `sort_by_key` may evaluate its closure more than once per element, so hashing there would re-hash on every comparison — the same defect the review of fgbio#1166 flagged in its Scala (25–30× the element count at n=100k).
- Ranks compare as **signed `i32`**, matching fgbio's Scala `Int`. About half of all read names hash negative, so an unsigned comparison silently reorders them.
- The sort is **stable**, matching Scala's `sortInPlaceBy`, so tied ranks keep input order.

Selection now lives in one place, `caller::select_lowest_ranking`, shared by the raw-record and codec paths. With no RNG left on the caller there is no per-instance state for downsampling to depend on, so the `seed` option and the `rand` dependency are removed — the determinism is structural rather than incidental.

## Behaviour changes

**1. Which reads are retained differs from previous releases** for any run that sets `--max-reads`, `--max-reads-per-strand`, or codec `--max-reads`. There is no way to fix the reproducibility bug while preserving the old selection, since the old selection *was* the bug. Runs without a cap are byte-identical, verified against a pre-change binary.

**2. Both ends of a template are now retained or discarded together**, which also recovers reads the old shuffle lost. On a 77k-molecule library:

| | reads emitted | `cD` distribution |
|---|---|---|
| before | 726,854 | `1`: 726,854 |
| after | 734,536 | `1`: 734,536 |
| fgbio 4.1.0 | 734,536 | `1`: 702,047 · `2`: 32,489 |

The count now matches fgbio. The depth does not, because simplex still applies the cap to the whole tag family rather than to each end — tracked separately as #723, and deliberately not fixed here.

**3. `duplex --max-reads-per-strand 0` is now rejected** (second commit). It previously exited 0 having written an empty BAM.

## Tests

Every selection test was verified **red under a mutant that breaks the property it claims to guard**, not merely green afterward:

- the subset is independent of how many families the caller downsampled previously
- the subset is invariant to input read order
- signed comparison — mutating to `wrapping_abs` (the abs form that exists elsewhere in this repo) fails it
- stability — mutating to `sort_unstable_by_key` fails it, using mixed tied/distinct keys at n=30, since all-equal keys can never separate stable from unstable
- both ends of a template retained together — a `hash(name + end)` mutant fails exactly this test and nothing else
- a golden test pinning *which* reads are retained, derived independently from htsjdk `Murmur3(42)` by running real htsjdk. Without it, a deterministic-but-*biased* rule passes everything above — sorting by read name, which on Illumina names sorts by tile/x/y, would systematically favour one region of the flowcell.

Plus the regression test for the bug that is actually live in fgumi: capped output must be identical across `--threads 1`, `--threads 8`, and no-`--threads`, for simplex and duplex. **Verified failing at `e2ad27bf`** — both capped cases fail, both uncapped cases pass.

Full suite: 7182 passed, 30 skipped.

## Deliberately not included

**No fgbio byte-parity oracle.** fgbio#1166 is still open, so there is nothing stable to pin against; a fixture generated from released 4.1.0 would disagree by design, since it still uses the shuffle. Once that PR merges, this change also restores byte parity under a cap, and the oracle can follow. This is a known gap, not an oversight.

Also out of scope, both filed: **#723** (simplex caps the whole MI group instead of each end) and **#724** / fulcrumgenomics/fgbio#1167 (reads dropped by downsampling are counted as used in `raw_reads_used` and never reach `--rejects` — pre-existing and shared with fgbio).

## Reading order

Two commits, mirroring fgbio#1166's structure. Both build and test standalone.

1. `fix(consensus)!: make consensus read downsampling deterministic` — start at `crates/fgumi-raw-bam/src/hash.rs` (the ranking primitive), then `caller::select_lowest_ranking`, then the three call sites in `vanilla_caller.rs` and `codec_caller.rs`.
2. `fix(duplex): reject --max-reads-per-strand 0` — reviewable independently.

Closes #725

Risk: output changes for capped consensus runs, pinned by signed fgbio-compatible read-name hashes and stable pair selection; `unsafe` changes: none, and CLAUDE.md allowlist changes: none; memory bounds, queue capacity, and thread/backpressure policy changes: none.

- Replaced seeded random downsampling with deterministic lowest signed Murmur3 read-name ranks.
- Preserved input-order stability and retained paired ends together.
- Applied shared selection logic to raw-record, vanilla, and CODEC paths.
- Removed the obsolete RNG and `seed` option.
- Added fgbio-compatible UTF-16 Murmur3 hashing and public hash helpers.
- Added unit, property-based, and integration coverage across execution modes, ordering, ties, pairing, hash vectors, and cap boundaries.
- Rejected `--max-reads-per-strand 0`.
- Updated command documentation.
- Uncapped runs remain byte-identical. Capped runs can retain different reads.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,12 @@ All notable changes to this project will be documented in this file.
   old selection was the bug. Runs that do not set a cap are unaffected and remain
   byte-identical. Ports [fgbio#1166](https://github.com/fulcrumgenomics/fgbio/pull/1166).
 
+### Bug Fixes
+
+- `duplex --max-reads-per-strand 0` is now rejected at startup. It previously exited 0 having
+  written an empty BAM, because a cap of zero empties every strand. `simplex` and `codec`
+  already validated their equivalents.
+
 ### Features
 
 - Boolean command-line flags now additionally accept `on`, `off`, `1`, and `0`, alongside the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,28 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Changed
+
+- **Breaking:** consensus downsampling now selects which reads to retain by a hash of the read
+  name rather than by shuffling a seeded random number generator. This affects `simplex
+  --max-reads`, `duplex --max-reads-per-strand`, and `codec --max-reads`.
+
+  The retained subset is now a pure function of the tag family, so it is reproducible across
+  runs, thread counts, and execution modes. Mates share a read name and therefore a rank, so
+  the two ends of a template rank together; `codec --max-reads` selects whole pairs and so
+  always keeps or drops both ends, while `simplex --max-reads` (which caps the whole family by
+  record count, fgumi#723) and `duplex --max-reads-per-strand` (which caps each end
+  independently) can retain one end of a template without the other. Previously the surviving
+  reads depended on how many families the
+  consensus caller had already processed, so `--threads N` and the no-`--threads` path could
+  produce different consensus reads from the same input — measured at 566 of 77,804 records
+  (0.73%) on one library at `--max-reads-per-strand 2`.
+
+  **Which reads are retained differs from previous releases** for any run that sets a cap.
+  There is no way to fix the reproducibility bug while preserving the old selection, since the
+  old selection was the bug. Runs that do not set a cap are unaffected and remain
+  byte-identical. Ports [fgbio#1166](https://github.com/fulcrumgenomics/fgbio/pull/1166).
+
 ### Features
 
 - Boolean command-line flags now additionally accept `on`, `off`, `1`, and `0`, alongside the
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -785,7 +785,6 @@ dependencies = [
  "log",
  "noodles",
  "proptest",
- "rand 0.10.2",
  "rstest",
  "tempfile",
  "thiserror 2.0.19",
--- a/crates/fgumi-consensus/Cargo.toml
+++ b/crates/fgumi-consensus/Cargo.toml
@@ -14,7 +14,6 @@ approx = { workspace = true }
 bstr = { workspace = true }
 log = { workspace = true }
 noodles = { workspace = true, features = ["bam", "sam", "core"] }
-rand = { workspace = true }
 thiserror = { workspace = true }
 wide = { workspace = true }
 fgumi-dna = { workspace = true }
--- a/crates/fgumi-consensus/src/caller.rs
+++ b/crates/fgumi-consensus/src/caller.rs
@@ -633,11 +633,40 @@ pub fn make_prefix_from_header(header: &Header) -> String {
     }
 }
 
+/// Selects which of `ranks` survive a cap of `max_reads`, returning their indices in
+/// ascending (input) order.
+///
+/// The lowest-ranking entries are kept, where rank is the fgbio-compatible Murmur3 hash of
+/// the read name (see [`fgumi_raw_bam::hash::fgbio_read_name_rank`]). Ranks are compared as
+/// **signed** `i32`, matching fgbio's Scala `Int` — about half of all read names rank
+/// negative, so an unsigned comparison would silently reorder them. The selection sort is
+/// **stable**, matching Scala's `sortInPlaceBy`, so tied ranks resolve to input order.
+///
+/// The returned indices are re-sorted ascending after truncation so callers emit records in
+/// input order. That second sort orders *indices*, not ranks, and is independent of the
+/// selection above.
+///
+/// Callers pass ranks already stamped on their records rather than hashing here: Rust's
+/// `sort_by_key` may evaluate its key closure more than once per element, so hashing inside
+/// it would re-hash per comparison.
+pub(crate) fn select_lowest_ranking(ranks: &[i32], max_reads: usize) -> Vec<usize> {
+    let mut indices: Vec<usize> = (0..ranks.len()).collect();
+    if ranks.len() <= max_reads {
+        return indices;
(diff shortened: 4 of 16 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #731, opened 2026-08-09
fix(codec)!: downsample after filtering to the most common alignment

### LINKED ISSUE #730: codec --max-reads downsamples before filtering to the most common alignment, diverging from fgbio
`fgumi codec` applies the `--max-reads` cap *before* filtering each strand to its most common alignment. fgbio does the opposite: it filters first, checks `--min-reads` on the filtered lists, and only then caps. The two orderings select different reads, and fgumi's can reject a family that fgbio consenses.

## Where

`crates/fgumi-consensus/src/codec_caller.rs`, in `consensus_reads_raw`, the order today is:

1. `min_reads_per_strand` check on the raw infos
2. cap to `max_reads_per_strand`
3. `filter_to_most_common_alignment_raw` per strand
4. `min_reads_per_strand` check on the filtered lists

fgbio's `CodecConsensusCaller.scala` runs them as:

1. `filterToMostCommonAlignment` per strand (lines 196-197)
2. `minReadsPerStrand` check on the filtered lists (line 198)
3. the cap, applied inside `ssCaller.consensusCall(r1s)` / `(r2s)` (lines 234-235), with `ssCaller` constructed with `maxReadsPerStrand` (line 115)

fgumi instead constructs its `ss_caller` with `max_reads: None` and does the capping itself, at the earlier stage.

## Why it matters

Capping first samples across a family that still mixes alignment patterns, so the majority pattern *among the sampled reads* can differ from the majority in the full family. `filter_to_most_common_alignment_raw` then cuts the family a second time, and can push it below `--min-reads` — rejecting a family fgbio would consense.

Concretely: a family of 10 R1 pairs, 6 carrying alignment A and 4 carrying alignment B, with `--max-reads 3 --min-reads 3`. fgbio filters to the 6 A-reads, then caps to 3 of them, and emits a consensus. fgumi caps first; if the 3 lowest-ranking names happen to be 1 A and 2 B, the alignment filter keeps only the 2 B-reads, which is below `--min-reads 3`, and the whole family is rejected as `InsufficientReads`. Nothing about a read name correlates with its alignment group, so this is a plain sampling artifact of the ordering.

The ordering itself predates the deterministic-downsampling work — the old RNG shuffle sat in the same place. What changed is that the divergence used to be one arbitrary sample among many, and is now a fixed, reproducible disagreement with fgbio for every over-cap CODEC family whose reads span more than one alignment pattern.

## Impact

Measured on a CODEC library (`CODEC.group.adjacency.fgbio.bam`, 10 templates/family typical), filtering before capping recovers consensus records that the current order drops:

| cap | current | filter-first | delta |
| --- | --- | --- | --- |
| `--max-reads 2` | 103,144 | 103,541 | +397 |
| `--max-reads 3` | 105,478 | 105,539 | +61 |

Runs that do not set `--max-reads` are unaffected.

## Note

Fixing this also changes what the cap guarantees about mates: capping each strand independently after its own alignment filter means R1 and R2 of a template can diverge when one end survives the filter and the other does not. That matches fgbio, and nothing downstream pairs the two strand lists by index — phase 4 takes the longest of each independently, and phase 5 builds a separate single-strand consensus from each. The `codec --max-reads` help text and the CHANGELOG entry need to stop claiming both ends are always retained together.

nh13:
Closes #730.

`codec --max-reads` applied its cap *before* `filter_to_most_common_alignment_raw`, while fgbio filters each strand first, checks `--min-reads` on the filtered lists, and only then caps — inside `ssCaller.consensusCall`, which is constructed with `maxReadsPerStrand` (`CodecConsensusCaller.scala:115, 196-198, 234-235`).

Capping first samples across a family that still mixes alignment patterns, so the majority pattern *among the sampled reads* can differ from the majority in the full family. The alignment filter then cuts the family a second time and can push it below `--min-reads`, rejecting a family fgbio consenses. See #730 for the worked example.

The cap now runs after the filter and the read-count check, and independently per strand.

## Dropping the shared index vector is safe

The old code computed one index vector from R1's ranks and applied it to both `r1_infos` and `r2_infos`, which kept the two lists index-aligned. Nothing downstream depends on that: `filter_to_most_common_alignment_raw` already ran independently per strand and routinely returns different lengths, phase 4 takes the longest of each list independently, and phase 5 builds a separate single-strand consensus from each. So the lists were already free to diverge before this change.

## Impact

Measured on `CODEC.group.adjacency.fgbio.bam`:

| cap | before | after | delta |
| --- | --- | --- | --- |
| `--max-reads 2` | 103,144 | 103,541 | +397 |
| `--max-reads 3` | 105,478 | 105,539 | +61 |

Runs that do not set `--max-reads` are unaffected.

## Behaviour note

Because each strand is now capped on its own ranks, a template can be retained on one end and dropped on the other when only one end survives the alignment filter. That matches fgbio. The `codec --max-reads` help text and the CHANGELOG entry added in #727 claimed both ends are always retained together, which is no longer true — both are corrected here.

## Testing

`codec_cap_runs_after_the_alignment_filter` builds the worked example from #730 — ten templates, six carrying the majority alignment and four a minority one, with the three lowest-ranking names deliberately split across both groups — and asserts a consensus is produced. Verified failing against the previous ordering (`count == 0`, the family rejected as `InsufficientReads`) and passing after.

`cargo ci-fmt`, `ci-lint` and `ci-test` are clean: 7195 passed, 30 skipped.

## Stacked on #727

Based on `725/nhomer/fix-deterministic-consensus-downsampling`, since it rewrites the same block. Merge #727 first; this PR retargets to `main` once that lands.

Risk: `codec` consensus output changes, pinned by fgbio-compatible alignment filtering and reproducible hash-based downsampling; `unsafe` changes: none, and the `CLAUDE.md` allowlist is unchanged; memory bounds, queue capacity, and thread/backpressure policy changes: none.

Fix: Filter each strand to its majority alignment, validate `--min-reads`, then downsample independently with `--max-reads`.

- Update help text and the changelog to document independent strand downsampling and possible paired-end divergence.
- Add regression coverage for post-filter capping, zero-cap rejection, independent strand selection, and consensus generation.
- Recover 397 consensus records with `--max-reads 2` and 61 with `--max-reads 3` on `CODEC.group.adjacency.fgbio.bam`.
- Runs without `--max-reads` remain unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,10 +12,10 @@ All notable changes to this project will be documented in this file.
 
   The retained subset is now a pure function of the tag family, so it is reproducible across
   runs, thread counts, and execution modes. Mates share a read name and therefore a rank, so
-  the two ends of a template rank together; `codec --max-reads` selects whole pairs and so
-  always keeps or drops both ends, while `simplex --max-reads` (which caps the whole family by
-  record count, fgumi#723) and `duplex --max-reads-per-strand` (which caps each end
-  independently) can retain one end of a template without the other. Previously the surviving
+  the two ends of a template rank together, but none of the three commands guarantees that both
+  ends survive: `simplex --max-reads` caps the whole family by record count (fgumi#723), and
+  `duplex --max-reads-per-strand` and `codec --max-reads` cap each end independently, so a
+  template can be retained on one end and dropped on the other. Previously the surviving
   reads depended on how many families the
   consensus caller had already processed, so `--threads N` and the no-`--threads` path could
   produce different consensus reads from the same input — measured at 566 of 77,804 records
@@ -25,6 +25,15 @@ All notable changes to this project will be documented in this file.
   There is no way to fix the reproducibility bug while preserving the old selection, since the
   old selection was the bug. Runs that do not set a cap are unaffected and remain
   byte-identical. Ports [fgbio#1166](https://github.com/fulcrumgenomics/fgbio/pull/1166).
+
+- **Breaking:** `codec --max-reads` now downsamples *after* filtering each strand to its most
+  common alignment, and independently per strand, matching fgbio's
+  `CodecConsensusCaller`. Previously the cap ran first, across a family that still mixed
+  alignment patterns, so the majority pattern among the sampled reads could differ from the
+  majority in the full family — the alignment filter then cut the family a second time and could
+  push it below `--min-reads`, rejecting a family fgbio consenses. On a CODEC library this
+  recovered 397 consensus records at `--max-reads 2` (103,144 → 103,541) and 61 at
+  `--max-reads 3` (105,478 → 105,539). Runs that do not set `--max-reads` are unaffected.
 
 ### Bug Fixes
 
--- a/crates/fgumi-consensus/src/codec_caller.rs
+++ b/crates/fgumi-consensus/src/codec_caller.rs
@@ -156,6 +156,11 @@ pub struct CodecConsensusOptions {
     pub min_reads_per_strand: usize,
 
     /// Maximum number of read pairs to use per strand (downsample if exceeded)
+    ///
+    /// `Some(0)` is not a meaningful cap: it retains no reads on either strand. The `codec`
+    /// command rejects it at startup (`--max-reads` must be at least `--min-reads`, itself at
+    /// least 1); a caller that sets it directly gets every family rejected as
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #733, opened 2026-08-09
ci: add cargo ci-publish-order to the local gate

nh13:
Adding a workspace crate without adding it to the `CRATES` list in `scripts/publish-crates.sh` is invisible locally: every `cargo ci-*` check passes, and only the `publish-order` CI job catches it — after the push. This closes that gap.

Found the hard way on a branch that adds a new crate: full local gate green (`ci-fmt`, `ci-lint`, `ci-tag-literals`, `ci-doc` under `-D warnings`, `ci-test`, loom), then `publish-order` failed in CI in 14 seconds on a one-line omission.

## What this adds

- `cargo xtask check-publish-order` and a matching **`cargo ci-publish-order`** alias.
- The check now runs in the **pre-commit hook**, alongside `ci-fmt`, `ci-lint`, and `ci-tag-literals`.

## Why it shells out instead of reimplementing the check in Rust

`scripts/publish-crates.sh` is already the single source of truth for the publish list — both the release workflow and the `publish-order` CI job invoke it. A second, independent implementation in xtask could drift from it, and a local gate that disagrees with CI is worse than no local gate. `run_mdbook` sets the existing precedent for `std::process::Command` in xtask.

It is invoked through `bash` rather than executed directly, so the check does not depend on the file's executable bit surviving a checkout or export.

CI is unchanged: the `publish-order` job keeps calling the script directly, so it stays a ~14s job with no xtask build.

## Verification

Checked in both directions, because a gate that cannot fail is not a gate:

```
$ cargo ci-publish-order
publish list OK: 12 crates, complete and in dependency order
$ echo $?
0

# with fgumi-sort deleted from the CRATES list:
$ cargo ci-publish-order; echo $?
ERROR: workspace crates missing from the publish list:
fgumi-sort
Add them to CRATES in .../scripts/publish-crates.sh in the correct dependency order.
Error: publish-crates.sh --check failed with exit code: exit status: 1
1
```

`ci-fmt`, `ci-lint`, `ci-tag-literals`, and `ci-publish-order` all exit 0 on this branch.

Note that the hook change only takes effect after re-running `./scripts/install-hooks.sh`, since the installed copy lives in `.git/hooks`.

Risk: command output changes: none; `unsafe` changes: none, and no CLAUDE.md allowlist update applies; memory bounds, queue capacity, and thread/backpressure policy changes: none.

Adds a local publish-order validation gate.

- Adds `cargo xtask check-publish-order`.
- Adds the `cargo ci-publish-order` alias.
- Runs `scripts/publish-crates.sh --check` through `bash`.
- Adds the check to the pre-commit hook.
- Updates hook-installation guidance and documentation.
- Reuses the publish list as the single source of truth.
- CI remains unchanged.
- Tests cover valid and invalid publish-list cases, script failures, workspace execution, argument forwarding, and missing scripts.


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #737, opened 2026-08-10
fix(consensus)!: require mapped reads for consensus unless --allow-unmapped

nh13:
Companion to fulcrumgenomics/fgbio#1168, which fixes the same defect in the Scala implementation.

## Problem

Unmapped reads contribute bases and depth to consensus reads, with no alignment supporting them.

`fgumi group --allow-unmapped` deliberately supports wholly unmapped templates (ribosome display and similar), but its gate is on templates with **both** reads unmapped, so a *half*-mapped pair is kept regardless of that flag. The consensus pre-group filter then kept any record that was mapped **or** had a mapped mate, which admitted the unmapped end too.

That end has no cigar, and `is_cigar_prefix` returns `true` for an empty left hand side — `a.len() > b.len()` is false, `last_index = 0.saturating_sub(1)`, and the loop body never runs:

```rust
pub fn is_cigar_prefix(a: &[(Kind, usize)], b: &[(Kind, usize)]) -> bool {
    if a.len() > b.len() { return false; }
    let last_index = a.len().saturating_sub(1);
    for (i, &(op_a, len_a)) in a.iter().enumerate() { ... }
    true
}
```

So in `select_most_common_alignment_group` an empty cigar matches *every* alignment group instead of forming one of its own, and the read is retained rather than rejected. Whether it is retained at all depends on where it lands in the length-descending sort, so the outcome turns on a read-length tiebreak.

Observed for a molecule of three read pairs where one pair has an unmapped R2:

| Command | Symptom |
|---|---|
| `simplex` | `cD=3` for R2 where only two reads are eligible, and the unmapped read's bases reach the consensus |
| `duplex` | `aD=3` for the AB single-strand consensus of R2 where only two AB-R2s are mapped |
| `codec` | Unaffected — its alignment grouping is reached only for templates forming a primary FR pair |

## Fix

Two layers.

**1. `consensus_pregroup_keep_flags` now requires the record to be mapped** (`src/lib/commands/common.rs`). This is fgumi's copy of fgbio's `ConsensusCallingIterator` filter and is already shared by simplex, duplex and codec, so all three are covered by construction — and it is already parameterised by `--allow-unmapped`, so the relaxation has a home. The mapped end of a half-mapped pair is unaffected; it is mapped in its own right.

**2. Under `--allow-unmapped`, mapped reads still win.** `drop_unmapped_if_any_mapped` removes unmapped reads from any set being consensus called that also holds a mapped read, and leaves a wholly unmapped set intact. Dropped reads are attributed to `RejectionReason::Unmapped` — which existed in the enum with no production call site until now — and reach the rejects output.

So: unmapped reads are not consensus input by default; with `--allow-unmapped` they are admitted but never displace mapped reads.

## Behaviour change

**This is breaking for input containing half-mapped pairs.** A family of such pairs previously produced an R2 consensus built from unaligned reads. It now produces no consensus for that end, and therefore no read pair.

That is deliberate: it restores parity with fgbio, where the same family produces nothing at all. Gating the fallback behind `--allow-unmapped` rather than applying it unconditionally is what keeps the default in step with fgbio while preserving the wholly-unmapped capability fgbio does not have.

## Tests

Written first, watched failing, then fixed — the two commits are `test:` then `fix:` in that order, so the first is red by design. Happy to squash if you would rather not carry a red commit.

At the test commit, verified: 3 failing, 2 passing guards.

| Test | At `test:` commit | Purpose |
|---|---|---|
| `test_simplex_ignores_unmapped_end_when_mapped_reads_present` | FAIL | reproduces the defect |
| `test_duplex_ignores_unmapped_end_when_mapped_reads_present` | FAIL | reproduces the defect |
| `test_simplex_prefers_mapped_reads_even_when_unmapped_are_allowed` | FAIL | pins layer 2, under `--allow-unmapped` |
| `test_codec_ignores_half_mapped_pair_when_mapped_reads_present` | PASS | guards codec's existing FR-pair protection |
| `test_simplex_allow_unmapped_still_consenses_a_wholly_unmapped_family` | PASS | guards the `--allow-unmapped` capability against this change |

The simplex tests assert R2's *bases* as well as its depth. Depth alone can look correct while the consensus is still contaminated — that is how the `--max-reads` interaction was caught in the fgbio equivalent.

`test_consensus_pregroup_keep_flags`'s `unmapped_paired_mate_mapped` case is inverted to pin the new behaviour, and a case is added for the mapped end of a half-mapped pair, which is still kept.

After: `cargo ci-test` 7247 passed, 0 failed; `cargo ci-fmt` and `cargo ci-lint` clean.

## Not addressed here

**The root cause in `is_cigar_prefix` is untouched.** An empty cigar is still a prefix of every cigar. This PR makes that unreachable from the consensus callers by default, and harmless under `--allow-unmapped`; any future path producing an empty cigar would hit it again.

Risk: output changes for consensus grouping and consensus depth/bases, pinned by mapped-read filtering; sort order, corrected UMIs, and metrics: none. `unsafe` changes: none, and `CLAUDE.md` allowlist updates: none. Memory bounds, queue capacity, and thread/backpressure policy: none.

Fix: Require mapped records by default. With `--allow-unmapped`, retain wholly unmapped families but prefer mapped records when both types exist. Attribute removed records to `RejectionReason::Unmapped`.

Tests cover simplex, duplex, and CODEC behavior, including half-mapped pairs and mapped-read depth.

### DIFF
--- a/crates/fgumi-consensus/src/vanilla_caller.rs
+++ b/crates/fgumi-consensus/src/vanilla_caller.rs
@@ -1088,16 +1088,50 @@ impl VanillaUmiConsensusCaller {
         if self.options.max_reads.is_some() { fgbio_read_name_rank(name) } else { 0 }
     }
 
+    /// Drops unmapped reads when the set also contains at least one mapped read.
+    ///
+    /// An unmapped read has no cigar, and an empty cigar is a prefix of every cigar, so
+    /// `select_most_common_alignment_group` would match it against whichever alignment group it
+    /// was tested against first rather than letting it form a group of its own. It would then
+    /// contribute bases and depth to the consensus with no alignment supporting it.
+    ///
+    /// A wholly unmapped set is returned untouched, so molecules from
+    /// `fgumi group --allow-unmapped` (e.g. ribosome display) still produce consensus reads.
+    ///
+    /// Returns the retained reads and the original indices of any dropped unmapped reads.
+    fn drop_unmapped_if_any_mapped(
+        &mut self,
+        source_reads: Vec<SourceRead>,
+    ) -> (Vec<SourceRead>, HashSet<usize>) {
+        let is_unmapped = |sr: &SourceRead| sr.flags & flags::UNMAPPED != 0;
+
+        if !source_reads.iter().any(is_unmapped) || source_reads.iter().all(is_unmapped) {
+            return (source_reads, HashSet::new());
+        }
+
+        let (retained, dropped): (Vec<SourceRead>, Vec<SourceRead>) =
+            source_reads.into_iter().partition(|sr| !is_unmapped(sr));
+
+        self.stats.record_rejection(RejectionReason::Unmapped, dropped.len());
+        (retained, dropped.into_iter().map(|sr| sr.original_idx).collect())
+    }
+
     /// Filters `SourceReads` to only include those with the most common alignment pattern.
     /// This matches fgbio's filterToMostCommonAlignment behavior.
+    ///
+    /// Unmapped reads are dropped first whenever any mapped read is present: fgumi always
+    /// prefers to build a consensus from mapped reads. A wholly unmapped set is left intact so
+    /// that `fgumi group --allow-unmapped` workflows still produce consensus reads.
     ///
     /// Returns the filtered `SourceReads` and a set of rejected original indices.
     fn filter_source_reads_by_alignment(
         &mut self,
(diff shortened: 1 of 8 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #740, opened 2026-08-12
fix(dedup)!: report per-reason filtered template counts

### LINKED ISSUE #739: `DedupMetricsOutput` omits `filter_metrics`
### Description

from `fgumi dedup --help` in fgumi 0.5.0, lines 52–53:

`The counts of templates dropped for each reason are reported in the metrics output
(--metrics). For the templates that remain, representative selection (the duplicate score),
the 0x400 flag, and --remove-duplicates follow Picard `MarkDuplicates` semantics; the
grouping key does not — see below.`

Here is what I see when I look at the the file `--metrics` actually produces:

```
total_templates       unique_templates        duplicate_templates     duplicate_rate  total_reads     unique_reads    duplicate_reads secondary_reads supplementary_reads     missing_tc_tags
```

If possible I think we need to flatten filter_metrics into DedupMetricsOutput or update the `--help` information.

### Related Issues

_No response_

### Version

0.5.0

### Reproduction steps

```bash
`fgumi --help`
```

### Input

_No response_

### Logs/Error messages

```bash
compare against:

/// Metrics collected during deduplication.
#[derive(Debug, Default, Clone)]
pub struct DedupMetrics {
    /// Total templates processed
    pub total_templates: u64,
    /// Templates marked as duplicates
    pub duplicate_templates: u64,
    /// Templates kept (not duplicates)
    pub unique_templates: u64,
    /// Total reads processed
    pub total_reads: u64,
    /// Reads marked as duplicates
    pub duplicate_reads: u64,
    /// Unique reads (not duplicates)
    pub unique_reads: u64,
    /// Secondary reads processed
    pub secondary_reads: u64,
    /// Supplementary reads processed
    pub supplementary_reads: u64,
    /// Secondary/supplementary without `tc` tag
    pub missing_tc_tag: u64,
    /// Filter metrics from position grouping
    pub filter_metrics: FilterMetrics,
}
```

### Screenshots

```bash
![DESCRIPTION](LINK.png)
```

### OS

Other (describe in "Reproduction Steps")

### Checklist

- [ ] Checked there are no existing Issues for this bug
- [ ] Checked there are no Discussions that explain this issue
- [ ] Tried reproducing this issue using the latest release
- [ ] Provided all information necessary to reproduce this issue

nh13:
Closes #739.

`fgumi dedup` is a read *filter* as well as a duplicate marker: templates that fail the pre-grouping filter are dropped and never reach the output. It counted why it dropped each one, merged those counts across workers — and then discarded them at the serialization boundary, because `DedupMetricsOutput` had no field to hold them. A run that filtered out its entire input wrote a row of zeros and logged `Deduplication complete: 0 templates`, with nothing to indicate records had been dropped or why.

Reproducer on 574 simulated templates, all failing `-q 61`, before this change:

```
total_templates unique_templates duplicate_templates duplicate_rate total_reads ...
0               0                0                   0              0
```

After:

```
filtered_templates  filtered_low_mapping_quality  total_templates
574                 574                           0
```

plus a new log line — `Filtered out 574 templates before marking: 574 low_mapping_quality` — so the drops are visible even without `--metrics`.

## Why the fix is not just "add the missing columns"

`grouper::FilterMetrics` was `group`'s fgbio output schema being used as a measurement primitive. Its four `discarded_*` buckets are named after fgbio's `UmiGroupingMetric` columns, down to preserving fgbio's `discarded_umis_to_short` typo. That is a *column* vocabulary, not a *reason* vocabulary, and the filter has nine distinct rejection sites — so five of them collapsed into `discarded_poor_alignment`, including missing-`RX` and truncated-record, which are not alignment problems. `dedup` borrowed the struct and inherited both the vocabulary and a unit mismatch: `group` increments it by primary-record count, `dedup` by one per template, with nothing in the type to tell them apart. The field named `accepted_templates` has been holding a record count in `group` all along.

So the three collapsed layers are separated:

1. `TemplateFilterReason` — one variant per rejection site in the code, not per output column.
2. `TemplateFilterCounts` — a collector keyed by that enum, recording every decision in **both** template and primary-record units, so the unit is an explicit rendering choice rather than an implicit property of a shared `u64`.
3. Per-command rendering — `group` projects the nine reasons back onto fgbio's four columns in record units; `dedup` gets eleven new `filtered_*` columns in template units.

## `group`'s output is unchanged

This is the hard constraint on the change and it is enforced, not asserted. Golden `.grouping_metrics.txt` files were captured from the pre-change binary across three filter regimes (nothing filtered, all filtered for mapping quality, all filtered for UMI length, so more than one fgbio column is exercised) and diffed after every subsequent phase. All byte-identical. A characterization test pinning the five-column header at the command level was committed *before* the refactor so it could catch a regression during it, and `UmiGroupingMetrics::from_filter_counts` routes through an exhaustive match, so adding a tenth reason is a compile error until someone assigns it a column.

## Breaking change

`dedup --metrics` gains eleven columns. Consumers keyed on column names are unaffected; consumers parsing positionally must be updated. Note that `fgumi compare metrics` treats a column-set difference as `DIFFER`, so comparing a pre- and post-change dedup metrics file will fail by design.

## Suggested reading order

1. `feat(metrics): add reason-keyed template filter accounting` — the primitive, in isolation.
2. `refactor: record filter rejections by reason in group and dedup` — the migration. Both commands convert together because deleting `FilterMetrics` breaks `dedup`, and every commit builds.
3. `fix(dedup)!: report filtered template counts in the metrics file` — the actual fix.
4. `refactor: share one template filter between group and dedup` — the largest diff, and optional to the fix. A normalized diff of the two filter implementations showed them semantically identical except for one condition (`allow_unmapped`); they are now one function.

The rest are a mechanical rename, the new metrics type, and docs.

## Notes for the reviewer

- The reason split immediately caught a mislabelled existing test: `test_filter_template_raw_truncated_record_no_panic` was recorded as a poor-alignment rejection, but the record has a truncated *aux block*, clears the length check, and is rejected for a missing UMI. Its own comment said so. The coarse bucket had been hiding the discrepancy.
- The accounting invariant `filtered_templates + total_templates == templates read` is asserted in unit tests and end-to-end. That is the test that would have caught the original bug.
- `total_templates` deliberately keeps its name and means "templates *written*". Adding columns without renaming existing ones keeps name-keyed consumers working; the doc comment says so explicitly.
- There is no `input_templates` column — it is derivable, matching how `UmiGroupingMetrics` treats its derivable fields.
- `docs/src/metrics/deduplication-metrics.md` is generated by `cargo xtask` and gitignored, so it does not appear in the diff. Moving the schema into `fgumi-metrics` is what makes it generate at all; `dedup` was previously the only metric type with no reference page.

## Verification

`cargo ci-test` (7,277 tests), `cargo ci-lint`, `cargo ci-fmt` all clean. `group` goldens re-diffed after every phase including the final cleanup pass; the #739 reproducer re-run against the final binary.

Risk: `dedup --metrics` output changes, pinned by exact schema and reconciliation tests; `unsafe`: none and the CLAUDE.md allowlist is unchanged; memory, queue, and backpressure policy: none.

Fixes `fgumi dedup --metrics` so filtered templates and rejection reasons are serialized. Adds eleven `filtered_*` columns and passthrough counts. Preserves `group`’s five-column fgbio-compatible output.

Shares template filtering and accounting between `group` and `dedup`. Adds coverage for filter reasons, unmapped pass-through, output reconciliation, and metrics column order.

Documents metric units, reconciliation rules, and the positional parsing breaking change.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -329,7 +329,7 @@ reviews:
         require it to be called out. Do NOT file wording-only nits on help text
         or doc comments unless the text states a behavior the code does not
         implement.
-    - path: "src/lib/{grouper,template,mi_group,read_info}.rs"
+    - path: "src/lib/{grouper,template,template_filter,mi_group,read_info}.rs"
       instructions: >-
         Core read-grouping and template-assembly logic that sits outside the
         crates but feeds the same fgbio-validated output as `fgumi-umi` and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
+- **Breaking:** `dedup --metrics` gains eleven columns: one `filtered_*` column per template
+  filter reason, `filtered_templates` for the total, and `passthrough_templates` for the
+  `--include-unmapped` templates that bypass the filter. Consumers that parse the file
+  positionally must be updated; consumers keyed on column names are unaffected. Note that
+  `fgumi compare metrics` treats a column-set difference as `DIFFER`, so comparing a
+  pre-change dedup metrics file against a post-change one will fail by design.
+
+  These columns count *templates*, whereas the `discarded_*` columns of `group`'s
+  fgbio-shaped `.grouping_metrics.txt` count *primary records*. `group`'s output is unchanged.
+
 - **Breaking:** consensus downsampling now selects which reads to retain by a hash of the read
   name rather than by shuffling a seeded random number generator. This affects `simplex
   --max-reads`, `duplex --max-reads-per-strand`, and `codec --max-reads`.
@@ -27,6 +37,14 @@ All notable changes to this project will be documented in this file.
   byte-identical. Ports [fgbio#1166](https://github.com/fulcrumgenomics/fgbio/pull/1166).
 
 ### Bug Fixes
+
+- `dedup --metrics` now reports the templates dropped by its template filter, broken out by
+  reason. `dedup` is a read filter as well as a duplicate marker, but the drop counts were
+  collected, merged across workers, and then discarded at the serialization boundary: a run
+  that filtered out every input template wrote a row of zeros and logged only "Deduplication
+  complete: 0 templates", with nothing to indicate records had been dropped or why. A
+  per-reason summary is now also logged, so the drops are visible without `--metrics`
+  ([#739](https://github.com/fulcrumgenomics/fgumi/issues/739)).
 
 - `duplex --max-reads-per-strand 0` is now rejected at startup. It previously exited 0 having
   written an empty BAM, because a cap of zero empties every strand. `simplex` and `codec`
--- a/crates/fgumi-metrics/src/dedup.rs
+++ b/crates/fgumi-metrics/src/dedup.rs
@@ -1,0 +1,164 @@
+//! Metrics for the `dedup` command.
+
+use serde::{Deserialize, Serialize};
+
+use crate::Metric;
+
+/// Metrics written by `fgumi dedup --metrics`.
+///
+/// Three sections, in column order:
+///
+/// 1. **Filtering** — `filtered_templates` through `passthrough_templates`.
+///    `dedup` is a read *filter* as well as a duplicate marker: templates failing
+///    the pre-grouping filter are dropped and never reach the output. The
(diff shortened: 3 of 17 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #734, opened 2026-08-10
feat(bam-io): add the shared grouping and library-lookup domain types (R0)

nh13:
Prerequisite for porting `src/lib/pipeline/steps/`. Additive: nothing calls the new modules yet.

## Why

The typed-step tree imports `fgumi_bam_io::{DecodedRecord, GroupKey, GroupKeyConfig, Grouper, compute_group_key_from_raw, name_hash_key}`. On this branch those types live in `src/lib/unified_pipeline/{base,bam}.rs`, and `name_hash_key` does not exist at all — so the new pipeline tree would have to import from the tree it is meant to replace, which inverts the dependency and breaks when `unified_pipeline` is eventually deleted.

This moves them next to the raw-record helpers they operate on, matching where `feat-runall` put them. `feat-runall`'s own `pipeline/mod.rs` states the intent: *"Shared grouping/decoded-record domain types … live in the `fgumi-bam-io` crate, next to the raw-record helpers they operate on."*

## What

- `crates/fgumi-bam-io/src/grouping.rs` (779 lines) — `DecodedRecord`, `GroupKey`, `GroupKeyConfig`, `Grouper`, `compute_group_key_from_raw`, `name_hash_key`
- `crates/fgumi-bam-io/src/library.rs` (193 lines) — `LibraryLookup`, `LibraryIndex`, `build_library_lookup`, which the group key depends on
- module declarations + re-exports in `lib.rs`
- `ahash` added to the crate's dependencies, via `workspace = true` (this branch's convention; `feat-runall` pins `"0.8"` directly). It was already a workspace dependency, used by `fgumi-consensus`, `fgumi-pipeline-core`, `fgumi-sort`, `fgumi-umi`.

Both files are self-contained: their only imports are `noodles::sam` and each other.

## Duplication

`unified_pipeline` keeps its own copies of the equivalent types. That duplicate is intentional and temporary — it retires when `unified_pipeline` is deleted, after the commands migrate. Nothing is changed in `unified_pipeline` here, so no existing behaviour moves.

## Verification

Full local gate on this branch:

- `cargo ci-fmt`, `cargo ci-lint`, `cargo ci-tag-literals` — clean
- `RUSTDOCFLAGS="-D warnings" cargo ci-doc` — clean
- `./scripts/publish-crates.sh --check` — clean
- `cargo ci-test` — **8192 passed**, 30 skipped (up 6 from 8186; the ported files bring their own tests)

[...]

PR #741, opened 2026-08-12
refactor(bam-io): collapse the duplicated grouping domain types onto one definition (R0b)

nh13:
Prerequisite for R1c. Discovered while porting `steps/group/`, which is the first ported code that cannot be written without it.

## Why

#734 moved `DecodedRecord`, `GroupKey`, `GroupKeyConfig`, and the library-lookup helpers into `fgumi-bam-io`, and deliberately left the `unified_pipeline` / `read_info` copies in place so nothing had to be rewired. That was the right call at the time: the ported step library only ever *produced* these values, so two identical definitions coexisted harmlessly.

`steps/group/` breaks the arrangement. `GroupBam` calls `TemplateGrouper::add_records`, which is `fgumi_bam_io::Grouper` on one side and the umbrella `Grouper` on the other — typed over two different `DecodedRecord`s:

```
error[E0308]: expected `fgumi_bam_io::DecodedRecord`, found `base::DecodedRecord`
```

Two textually identical definitions are still two types. The call cannot be written until they are one.

## What changed

Keep the `fgumi-bam-io` definition; re-export it from both former homes so every existing `crate::unified_pipeline::DecodedRecord` and `crate::read_info::LibraryIndex` path resolves unchanged. Collapsed: `DecodedRecord`, `GroupKey`, `GroupKeyConfig`, `LibraryIndex`, `LibraryLookup`, `build_library_lookup`, `unknown_library`, and the `Grouper` trait.

Before removing each copy I diffed the two. All were identical apart from field visibility and doc-comment formatting; where they differed in API the `fgumi-bam-io` side was already a superset (`GroupKeyConfig::name_hash_only`, `DecodedRecord::{record, raw_bytes_mut, cached_umi_position_opt}`, `GroupKey::name_hash_only`). One exception went the other way and was ported across:

- `GroupKey::has_mate_position` existed only on the umbrella copy, so it moved to `fgumi-bam-io`. The surviving definition is now a superset of both.

`grouper.rs` reached into `DecodedRecord`'s fields directly. Those are private on the `fgumi-bam-io` copy — made so during #734's review — so five sites now go through `into_raw_bytes()` / `record()` / `raw_bytes()` instead.

## Risk

No behavior change; this is deletion plus re-export. Net **-483 lines**. The suite is unchanged at **8322 passing**, which is the main evidence: every caller of these types kept compiling and kept passing against a single definition.

The duplication ledger from the landing plan shrinks accordingly — R6 has this much less to unpick when `unified_pipeline` is deleted.

## Gate

`ci-fmt`, `ci-lint`, `ci-tag-literals`, `ci-doc` (`-D warnings`), `publish-crates.sh --check`, `ci-test` — all exit 0.

Risk: command output changes none; `unsafe` changes none and the `CLAUDE.md` allowlist is unchanged; memory bounds, queue capacities, and thread/backpressure policies change none.

Fix: Consolidate grouping types in `fgumi-bam-io` and preserve existing import paths through re-exports.

- Resolve incompatible duplicate types used by `GroupBam` and `TemplateGrouper`.
- Move `GroupKey::has_mate_position` into `fgumi-bam-io`.
- Replace direct `DecodedRecord` field access with accessors.
- Share `unknown_library()` across `ReadInfo` fallbacks.
- Remove 483 lines without behavior changes.
- CI checks and 8,439 tests pass.

### DIFF
--- a/crates/fgumi-bam-io/src/grouping.rs
+++ b/crates/fgumi-bam-io/src/grouping.rs
@@ -133,6 +133,16 @@ impl GroupKey {
             self.cell_hash,
         )
     }
+
+    /// Whether this key carries a mate position, i.e. it was built by
+    /// [`Self::paired`] rather than [`Self::single`].
+    ///
+    /// `strand2` is set to [`Self::UNKNOWN_STRAND`] by the single-end
+    /// constructor, so it is the discriminator between the two shapes.
+    #[must_use]
+    pub fn has_mate_position(&self) -> bool {
+        self.strand2 != Self::UNKNOWN_STRAND
+    }
 }
 
 impl PartialOrd for GroupKey {
--- a/crates/fgumi-bam-io/src/lib.rs
+++ b/crates/fgumi-bam-io/src/lib.rs
@@ -31,7 +31,7 @@ pub use format::{FORMAT_PREFIX_LEN, InputFormat, classify_input};
 pub use grouping::{
     DecodedRecord, GroupKey, GroupKeyConfig, Grouper, compute_group_key_from_raw, name_hash_key,
 };
-pub use library::{LibraryIndex, LibraryLookup, build_library_lookup};
+pub use library::{LibraryIndex, LibraryLookup, build_library_lookup, unknown_library};
 pub use mem_estimate::MemoryEstimate;
 pub use paths::{is_stdin_path, is_stdout_path};
 pub use progress::ProgressTracker;
--- a/src/lib/grouper.rs
+++ b/src/lib/grouper.rs
@@ -222,7 +222,7 @@ impl Grouper for TemplateGrouper {
         // records into the same template.
         for decoded in records {
             let name_hash = decoded.key.name_hash;
-            let raw = decoded.data;
+            let raw = decoded.into_raw_bytes();
             let read_name = fgumi_raw_bam::read_name(&raw);
             let same_template = match (self.current_name_hash, self.current_name.as_deref()) {
                 (Some(h), Some(name)) => h == name_hash && name == read_name,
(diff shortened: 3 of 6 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
Fixed. The local `UNKNOWN_LIBRARY` static is gone and both `ReadInfo` fallbacks call the re-exported `unknown_library()`.

Worth naming why this one stings: it is the PR's own thesis violated one level down. The change collapses duplicated *types* onto a single definition, and I re-exported `unknown_library` while leaving a second `LazyLock<Arc<str>>` for the same logical value sitting right beside it — so "unknown" existed as two allocations, one of which `build_library_lookup` reserves as index 0 and the other of which `ReadInfo::from` fell back to.

No behavior changed: nothing compares library `Arc`s by pointer (checked — the only `ptr_eq` uses in the tree are on sort arenas and pipeline storage, not libraries), so the two allocations always compared equal by value. That is also why there is no new test: the fix is allocation sharing with no observable difference, which is the cosmetic case the review conventions exempt. Flagging that explicitly rather than leaving it as a silent omission.

Sibling audit for the same mistake across the rest of the change: the other statics in the touched files are `THREAD_ID_UNSET`, `NUM_STEPS`, and three buffer-size constants — unrelated tuning, not duplicates of anything this PR re-exports. Clean.

The re-export comment now explains the value-level duplication too, so the next reader sees why there is no local static.

8439 tests passing, all static gates green.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai full review

PR #744, opened 2026-08-12
feat(commands): project per-stage options out of the CLI structs (R1e-a)

nh13:
Prerequisite for porting the chain builder. Based on `main-runall` rather than stacked on #736/#741/#743 — it touches only `src/lib/commands/`, disjoint from those, so it can merge in parallel.

## Why

The chain builder needs each stage's tuning knobs without depending on how they were supplied, so a chain can be constructed with no CLI struct behind it. Today those knobs are reachable only as fields on the clap command structs.

This adds one plain `<Command>Options` struct per stage plus a `to_<command>_options()` projection, for **correct, group, simplex, duplex, codec, filter, zipper**.

## Why these are not `clap::Args`

Flattening them into the command structs — the shape the fused command will eventually want, so it can expose `--<stage>::<flag>` — would move the fields off those structs and rewrite every `self.<field>` reference and its tests. That buys the chain builder nothing: it only ever reads the values, and never touches clap. So that refactor is deferred to the change that actually needs prefixed flags.

The consequence is that this PR is **additive**: no CLI surface moves, no existing field is relocated, and the existing suite covers the seven commands unchanged. The trade is that those fields get touched twice — once here, once when they become flattened `Args`. The field *names* do not change, so nothing downstream churns.

## Resolved vs. raw values

Two `GroupOptions` fields hold resolved values, because grouping cannot be configured from the raw flags alone:

- `min_map_q` applies the default `execute` would otherwise apply at run time.
- `effective_strategy` / `effective_edits` carry the `--no-umi` and identity-implies-zero-edits rules.

Both now come from `GroupReadsByUmi::resolved_min_map_q` and `resolve_strategy_and_edits`, **extracted from `execute`** so the command and the chain builder read them from one place rather than duplicating the rules. `execute` keeps its own validation and its override log line; the new methods only compute. This is the only behavior-adjacent change in the PR, and a case table pins both sets of rules as a regression test for the extraction.

`GroupOptions::index_threshold` keeps the richer `IndexThreshold` type rather than a bare count, since the flag also accepts `always` / `never`.

## Testing

Each projection has a test that drives the command through `try_parse_from` rather than a struct literal, with non-default values throughout — so a renamed or unwired flag fails the test instead of silently compiling, and a field read from the wrong source fails rather than coincidentally matching its default. Zipper additionally has a defaults test, since an all-non-default assertion cannot catch a field hard-coded to a non-default constant.

`Strategy` gains `PartialEq`/`Eq` (additive, fieldless enum) so the resolution table can assert on it.

8322 → 8339 tests, all passing. `ci-fmt`, `ci-lint`, `ci-tag-literals`, `ci-doc` (with `RUSTDOCFLAGS="-D warnings"`), and `publish-crates.sh --check` all pass.

## Deliberately not included

`warn_unwired_pipeline_flags` is part of the same upstream surface, but it warns that `--scheduler` is a deprecated no-op. On this branch the pluggable scheduler is still live — `SchedulerOptions::scheduler` is a `SchedulerStrategy` with a default, not an `Option`, and `unified_pipeline` still consumes it. Porting the warning now would announce a removal that has not happened, so it travels with the change that actually removes it.

Risk: command output changes: none; `unsafe`: none, and `CLAUDE.md` allowlist: unchanged; memory bounds, queue capacity, and thread/backpressure policy: none. Fix: add stage option projections without changing CLI interfaces.

- Added `CorrectOptions`, `GroupOptions`, `SimplexOptions`, `DuplexOptions`, `CodecOptions`, `FilterOptions`, and `ZipperOptions`.
- Added `to_*_options()` methods for each stage.
- Shared group resolution for MAPQ defaults, UMI strategy, and edit behavior while preserving validation and logging.
- Added projection and default-value tests.
- Added `PartialEq` and `Eq` to `Strategy`.

### DIFF
--- a/crates/fgumi-umi/src/assigner.rs
+++ b/crates/fgumi-umi/src/assigner.rs
@@ -565,7 +565,7 @@ pub const BOTTOM_STRAND_DUPLEX: &str = "/B";
 ///
 /// Determines how reads are grouped based on their UMI sequences. Each strategy makes
 /// different tradeoffs between speed, error tolerance, and grouping behavior.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 #[cfg_attr(feature = "cli", derive(clap::ValueEnum))]
 pub enum Strategy {
     /// Only reads with identical UMI sequences are grouped together
--- a/src/lib/commands/codec.rs
+++ b/src/lib/commands/codec.rs
@@ -277,6 +277,93 @@ pub struct Codec {
     pub queue_memory: QueueMemoryOptions,
 }
 
+// ─────────────────────────────────────────────────────────────────────────────
+// CodecOptions — the stage's tuning knobs, projected out of the CLI struct
+// ─────────────────────────────────────────────────────────────────────────────
+
+/// Codec-stage tuning, independent of how the values were supplied.
+///
+/// See [`crate::commands::zipper::ZipperOptions`] for why this is a plain
+/// struct rather than a flattened `clap::Args`. Note that the consensus-calling
+/// knobs are held **flat** here even though [`Codec`] nests them behind
+/// `#[command(flatten)]` sub-structs: the chain builder wants one bag per stage,
+/// not a re-run of the CLI's grouping.
+#[derive(Debug, Clone)]
+pub struct CodecOptions {
+    /// Pre-UMI error rate (phred).
+    pub error_rate_pre_umi: u8,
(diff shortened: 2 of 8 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #745, opened 2026-08-14
perf(pipeline): decouple deadlock liveness from stats and stop polling at teardown

nh13:
Makes the pipeline's deadlock monitor cheap enough to arm, and adds the benchmark that proves it. Split out of the #736 review — that PR fixed one *producer* of gapped ordinals; this one is about the framework's response to a wedge from any producer.

## The problem

The scheduled runtime's deadlock monitor ships disarmed (`deadlock_timeout_secs: 0`), so a wedged pipeline hangs silently and forever — measured directly while reviewing #736: a gapped ordinal sequence ran past 120s with no diagnostic. The fused path, which has no monitor, carries an unremovable 60s stall bound precisely because it knows nothing would otherwise catch it. The two runtimes disagree about whether a wedge should be survivable, and the one every multi-threaded run uses is the unprotected one.

It was disarmed for a reason, though — arming it cost real throughput. This removes both costs.

## Liveness no longer rides on instrumentation

The monitor needs one fact: has anything progressed since the last poll? It read that from `PipelineStats`, which is deliberately not free — `dispatch_one_step` times every dispatch with `Instant::now()` (~20–50ns aarch64, ~50–100ns x86_64) and gates it on `stats.is_some()` to keep the uninstrumented path zero-cost. Arming the monitor therefore meant buying per-dispatch timing for a fact that is one integer.

`LivenessCounter` supplies it directly: one counter per worker, each padded to its own cache line so a bump is an uncontended increment on a line nobody else touches, summed by the monitor once per poll interval. A single shared atomic would have been a false-sharing hotspot that got worse with thread count — the wrong shape, since more threads is when a wedge matters most. Profiling stays opt-in.

## Teardown no longer polls

`sleep_until_stop` slept in 25ms slices and re-checked a flag, so `Pipeline::run` waited up to a full slice for the monitor to notice the stop before joining it — dead time on the critical path of **every** run. A condvar wakes it immediately. The existing comment already predicted this shape ("a fixed dead-time tail … a large *relative* regression on short ones"); the slices made it smaller, not absent.

## Measurements

No command drives the typed pipeline on this branch yet and it had no benchmark, so a hot-path change was unmeasurable. `benches/pipeline_dispatch.rs` closes that: a near-no-op chain whose wall time is dominated by scheduler loop, queue push/pop, and reorder — read it as relative-only, since real steps do BGZF decode and parsing that dwarf dispatch.

Run on a quiet `c8g.4xlarge` (aarch64), three reps per configuration against one shared baseline, with a **clean-vs-clean control first** to establish the noise floor:

| configuration | 1 thread | 4 threads | 8 threads |
| --- | --- | --- | --- |
| control (identical code) | −0.8% | −7.9% | +2.1% |
| liveness counter alone | +8.3 / +1.6 / +0.3% | +6.2 / −1.6 / −9.8% | −6.0 / −0.1 / −2.8% |
| monitor armed, polled teardown | +1.8 / +11.8 / +2.4% | **+80.7 / +79.8 / +80.7%** | **+23.41 / +23.41 / +23.42%** |
| monitor armed, condvar teardown | +13.3 / +11.6 / +2.3% | +2.2 / +6.3 / +2.4% | −2.0 / −2.2 / +1.5% |

The counter's numbers scatter across reps and straddle zero — that is drift, not cost. The polled-teardown numbers reproduce to three significant figures, which is what makes them a real regression, and they match the mechanism: ~one 25ms slice added to a 27ms run. After the condvar they fall back inside the control's own spread.

Worth stating plainly: the first attempt at this measurement was run on a loaded laptop and produced +84%/+291%/+221% **for identical code**. The control run is the only reason that was caught, and it is why the table above leads with one.

## Why the default is still 0

Neither cost blocks it any more. `MonitorBlindTransport` does: an armed monitor *rejects* any chain with a non-`ByteBounded` edge, because its stall verdict counts bytes in flight to tell "idle waiting on a slow source" from "wedged". `Process2` uses `CountBounded`, so flipping the default today would fail those chains outright rather than watch them. Teaching the verdict to handle byte-blind edges is a separate change; this PR makes it the *only* remaining blocker, and exports `DEFAULT_DEADLOCK_TIMEOUT_SECS` for callers who want to arm it now on byte-bounded chains.

## Notes for review

- `apply_stall_verdict` takes `Option<&PipelineStats>` now: an uninstrumented run still reports the stall, just without the per-step snapshot. Losing the snapshot beats losing the detection.
- `LivenessCounter::bump` masks rather than takes a modulo; slot counts round up to a power of two. A runtime integer division per productive dispatch is the wrong thing to put there regardless of what the benchmark can resolve.
- The `#[allow(clippy::too_many_arguments)]` additions are the liveness shard threading through `dispatch_one_step` / `run_worker_loop`; a parameter struct would rename the problem, not fix it.

Full gate green: fmt, clippy, tag-literals, `RUSTDOCFLAGS="-D warnings"` doc, publish-check, 8327 tests.

Risk: command output changes—none; `unsafe` changes—none, so no `CLAUDE.md` allowlist update; memory or queue bounds and thread/backpressure policy—none. Fix: decouple deadlock monitoring from `PipelineStats` with cache-padded `LivenessCounter`s and replace teardown polling with condition-variable shutdown.

- Add public `LivenessCounter` and `DEFAULT_DEADLOCK_TIMEOUT_SECS`.
- Propagate liveness state through worker, detached-driver, and single-thread paths.
- Validate monitor-blind transports only when monitoring is enabled.
- Add stats-less monitoring diagnostics and `Duration::MAX` deadline handling.
- Add the `pipeline_dispatch` benchmark.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -216,6 +216,12 @@ harness = false
 required-features = ["simplex"]
 
 [[bench]]
+name = "pipeline_dispatch"
+harness = false
+# Dispatch-overhead A/B for the typed-step runtime; no command drives the
+# pipeline yet, so this is the only way to measure a hot-path change.
+
+[[bench]]
 name = "raw_bam_accessors"
 harness = false
 
--- a/benches/pipeline_dispatch.rs
+++ b/benches/pipeline_dispatch.rs
@@ -1,0 +1,223 @@
+//! Dispatch-throughput benchmark for the typed-step pipeline runtime.
+//!
+//! This measures the runtime's *per-dispatch overhead*, not the work steps do:
+//! every step here is a near-no-op, so wall time is dominated by the scheduler
+//! loop, queue push/pop, and the reorder stage. That is exactly the path any
+//! change to the driver's inner loop lands on.
+//!
+//! It exists because the pipeline had no benchmark at all — no command on this
+//! branch drives it yet (the chain builders and command rewiring arrive later),
+//! so there was no way to tell whether a hot-path change cost anything.
+//!
+//! Read the numbers as *relative* only. Absolute throughput here is meaningless
+//! as a product metric (real steps decompress BGZF and parse records, which
+//! dwarfs dispatch); the point is A/B on the same host.
+
+use std::collections::VecDeque;
+use std::hint::black_box;
+use std::io;
+use std::sync::Arc;
+use std::sync::atomic::{AtomicU64, Ordering};
+
+use criterion::{Criterion, criterion_group, criterion_main};
+use fgumi_pipeline_core::Unpushed;
+use fgumi_pipeline_core::builder::{Pipeline, PipelineConfig};
+use fgumi_pipeline_core::held::HeldSlot;
+use fgumi_pipeline_core::item::{HeapSize, Ordered};
+use fgumi_pipeline_core::outputs::OrderedBytesSingle;
+use fgumi_pipeline_core::queues::QueueSpec;
+use fgumi_pipeline_core::reorder::BranchOrdering;
+use fgumi_pipeline_core::step::{Step, StepCtx, StepKind, StepOutcome, StepProfile};
+
+/// Per-edge byte budget. Large enough that the queues never bind — this
+/// benchmark measures dispatch cost, and backpressure stalls would swamp it
+/// with scheduling noise.
+const EDGE_LIMIT_BYTES: u64 = 64 * 1024 * 1024;
+
+/// Items pushed through the chain per iteration.
+const N_ITEMS: u64 = 50_000;
+
+/// A minimal ordered item. `heap_size` is a small constant rather than 0 so the
+/// byte-bounded accounting does real work per item, as it would in production.
+#[derive(Clone, Copy)]
+struct Item {
+    ordinal: u64,
+}
+
+impl HeapSize for Item {
+    fn heap_size(&self) -> usize {
+        64
+    }
+}
+
+impl Ordered for Item {
+    fn ordinal(&self) -> u64 {
+        self.ordinal
+    }
+}
+
+/// Emits `n` items and finishes. `Exclusive` so it owns its cursor.
+struct CountingSource {
+    remaining: VecDeque<Item>,
+    held: HeldSlot<Unpushed<Item>>,
+}
+
+impl CountingSource {
+    fn new(n: u64) -> Self {
+        Self { remaining: (0..n).map(|ordinal| Item { ordinal }).collect(), held: HeldSlot::new() }
+    }
+}
+
+impl Step for CountingSource {
+    type Input = ();
+    type Outputs = OrderedBytesSingle<Item>;
+
+    fn profile(&self) -> StepProfile {
+        StepProfile {
+            name: "BenchSource",
+            kind: StepKind::Exclusive,
+            sticky: false,
+            output_queues: vec![QueueSpec::ByteBounded { limit_bytes: EDGE_LIMIT_BYTES }],
+            branch_ordering: vec![BranchOrdering::ByItemOrdinal],
+        }
+    }
+
+    fn try_run(&mut self, ctx: &mut StepCtx<'_, Self>) -> io::Result<StepOutcome> {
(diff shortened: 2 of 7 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
Addressed the outside-diff-range finding on `builder.rs#L998-L1000` (no inline thread exists for it, so noting it here).

The finding is correct, and it is a gap this PR introduced. The monitor arms on `deadlock_timeout_secs > 0` alone, but the transport guard still required `stats_arc.is_some()` — a condition that was implied back when liveness was read out of `PipelineStats`, and stopped being implied the moment `LivenessCounter` decoupled them. An armed, stats-less run could therefore start on a `CountBounded`/`Unbounded` edge, where `in_flight_bytes` reports 0, `classify_stall` reads that as `Starving`, and `Starving` resets the stall clock on every poll — so the wedge never reaches the fatal timeout.

Changes:

- `builder.rs`: the guard is now keyed on `deadlock_timeout_secs > 0` alone, matching the monitor-spawn condition below it, with a comment recording that the two must track each other.
- `builder.rs` (`with_deadlock_timeout` doc): states that any non-zero value arms the monitor and that `stats` is optional, adding only the per-step snapshot.
- `liveness.rs` (module doc): replaced "The monitor can then be armed by default" — the default is still `0` — with what is actually true: the monitor can now run with no `PipelineStats` attached, which removes the *cost* argument for shipping disarmed without arming anything by itself. The remaining blocker (`MonitorBlindTransport` vs `CountBounded` chains such as `Process2`) is named there.
- New run-level regression test `pipeline_run_rejects_a_monitor_blind_transport_without_stats`: non-zero timeout, no stats, blind transport, asserting `PipelineError::MonitorBlindTransport`.

On the test: it goes through `Pipeline::run` rather than calling `ensure_monitor_visible_transports` directly, because the existing helper unit tests pass whatever condition guards the call site and so cannot catch a guard that skips the check. Verified by reverting the guard to `&& stats_arc.is_some()` — the new test then hangs to the 120s harness timeout instead of failing an assert, which is exactly the silent failure mode described above.

I did not add the converse test (blind transport legal when disarmed): most chains in this suite already pair a `CountBounded` edge with the default `deadlock_timeout_secs: 0`, so an over-broad guard would fail them by the hundred.

Full suite: 8328 passed. Lint, docs, and publish-order checks clean.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed the two remaining CodeRabbit findings in `crates/fgumi-pipeline-core/src/builder.rs` (pushed as `c0300c4b`):

**`sleep_until_stop` — unrepresentable deadline (inline thread, now resolved).** `Instant::now() + dur` panics when the deadline is not representable, and `Pipeline::run` discards this helper thread's join error — so a panic there would silently disarm the monitor while the run reported success. The deadline is now built with `Instant::checked_add`; on `None` (effectively-never deadline) the helper waits untimed on the condvar until `StopSignal::stop()` wakes it, preserving the timed-wait/expiry path for representable deadlines. Added the regression test `sleep_until_stop_survives_an_unrepresentable_deadline` (passes `Duration::MAX`; panicked before the guard, wakes on stop after it).

**Fatal-stall message with no `stats` (outside-diff finding on `builder.rs#L1655-1664`, no inline thread).** The `StallVerdict::Wedged` branch claimed "Snapshot follows." but emitted nothing when `stats` is `None`. It now mirrors the `StallVerdict::Stalled` sibling exactly: when `stats` is absent it logs the same `(no per-step snapshot: PipelineConfig::stats is not set; attach a stats handle to see which step is stuck)` explanation. I kept the "Snapshot follows." lead-in rather than deleting it (as the proposed diff did) so the two stall branches stay structurally identical.

`cargo ci-fmt`, workspace `ci-lint` (`-D warnings -W clippy::pedantic`), `ci-tag-literals`, and the full `fgumi-pipeline-core` test suite (368 lib tests) pass.

nh13:
@coderabbitai review

PR #750, opened 2026-08-14
fix(compare): qualify the content-mode diff count when pairing stopped early

### LINKED ISSUE #747: compare bams prints "Content diffs: 0" when the comparison stopped early
## Summary

In content mode, when the two inputs desync, `compare bams` stops pairing at the first `RecordKey` mismatch and prints `Content diffs: 0`. That count covers only the records compared *before* the stop. The value is correct for what it counted, and `PositionalOutcome` documents the semantics precisely ("among records paired before any key mismatch was found"), but the printed summary does not carry the qualifier — so the line reads as a global "no content differences found".

## Example

Comparing two consensus BAMs that differ only in record count (one emits 122 extra records out of ~103M):

```
=== BAM Comparison Results (content mode) ===
Record counts: 103045729 vs 103045851
Content diffs: 0
First RecordKey mismatch: record 2468528 (pairing stopped)

RESULT: BAM files DIFFER
```

Only ~2.4% of the file was compared for content; the remaining ~100M records were never examined. `RESULT: BAM files DIFFER` is correct, so the tool's verdict is not wrong — but `Content diffs: 0` is the line that gets copied into summaries, and in isolation it states close to the opposite of what happened.

## Suggestion

Qualify the count when `key_mismatch_at` is set, in `Bams::report` (`src/lib/commands/compare/bams.rs`, the `println!("Content diffs: {}", ...)` call):

```
Content diffs: 0 (of 2,468,528 records compared before pairing stopped)
```

That keeps the happy path unchanged — when there is no key mismatch the line means what it says today — while making the truncated case self-describing.

nh13:
Closes #747.

## What

`compare bams` content mode pairs records by index and stops at the first `RecordKey` mismatch, never resyncing — so `PositionalOutcome::content_diffs` counts only the records paired *before* that stop. The struct documents exactly that; the printed summary did not. A run that examined 2,468,528 of 103,045,729 records reported a bare `Content diffs: 0`, which in isolation reads as "these files have no content differences" — close to the opposite of what the comparison established. The `RESULT: BAM files DIFFER` verdict was always correct; the count line is the one that gets copied into summaries.

The count now names the prefix it actually covers whenever `key_mismatch_at` is set:

```
Record counts: 103045729 vs 103045851
Content diffs: 0 (of 2468528 records compared before pairing stopped)
First RecordKey mismatch: record 2468528 (pairing stopped)
```

When pairing ran to completion the line is byte-for-byte what it has always been, so nothing grepping the completed-pairing report changes, and the `RESULT:` contract is untouched either way.

## How

The formatting is extracted into a pure `content_diffs_line(content_diffs, key_mismatch_at)` rather than inlined into the `println!`, so both branches are unit-testable directly instead of only through a subprocess capturing stdout. `key_mismatch_at` is the 0-based index of the first mismatching pair, hence also the number of pairs examined before the stop — that identity is what the qualifier reports, and it is pinned by its own cases (index 0 → "0 records", index 1 → the singular "1 record").

Two notes on the rendering, both deliberate:

- The count is printed unseparated (`2468528`, not `2,468,528` as sketched in the issue) to match the adjacent `Record counts:` line and every other number in these reports; the repo has no thousands-separator helper and adding one for a single line would make this report internally inconsistent.
- The qualifier repeats "pairing stopped" from the line below it. That redundancy is the point — the failure being fixed is this line being read on its own.

## Tests

Six `#[rstest]`/`#[test]` cases on `content_diffs_line` covering the unqualified branch, the qualified branch, singular/zero record counts, and a non-zero diff count surviving the qualifier. Plus two subprocess tests in `tests/integration/test_compare_bams.rs`: one pinning that a real desync report carries the qualifier (a report that formats it correctly but passes `None` for the stop index is precisely the reported bug, and would leave the unit tests green), and one pinning that a completed pairing leaves the line unqualified and never claims pairing stopped.

`cargo ci-fmt`, `cargo ci-lint`, `cargo ci-test` (7502 passed), `cargo ci-doctest` and `RUSTDOCFLAGS=-D warnings cargo ci-doc` all pass locally.

Risk: `compare bams` content-mode output changes; tests pin qualified early-stop counts, while completed output and `RESULT:` remain unchanged; `unsafe`: none, so the `CLAUDE.md` allowlist is unchanged; memory bounds, queue capacity, and thread/backpressure policy: none.

Fix: Added `content_diffs_line` and unit and integration tests for early desynchronization, completed pairing, zero and non-zero counts, and singular/plural wording.

### DIFF
--- a/src/lib/commands/compare/bams.rs
+++ b/src/lib/commands/compare/bams.rs
@@ -623,6 +623,33 @@ pub(crate) fn start_raw_batch_reader(
     Ok((rx, header))
 }
 
+/// Renders the `Content diffs:` line of the content-mode report.
+///
+/// `content_diffs` counts only the pairs compared *before* the first `RecordKey`
+/// mismatch: pairing stops at the first desync and never resyncs, so every record from
+/// there on is left unexamined (see
+/// [`PositionalOutcome`](super::engines::positional::PositionalOutcome)). Unqualified,
+/// `Content diffs: 0` on such a run reads as "these files have no content differences",
+/// close to the opposite of what was established. So when `key_mismatch_at` is set the
+/// count carries the size of the prefix it actually covers; when pairing ran to
+/// completion it covers every index-paired record and the line is exactly what it has
+/// always been. (Pairing running to completion is not the same as both files being fully
+/// examined: an unequal record count leaves the longer file's tail unpaired. That case is
+/// already explicit one line above, in `Record counts: {bam1} vs {bam2}`, which a
+/// truncated pairing has no counterpart for.)
+fn content_diffs_line(content_diffs: u64, key_mismatch_at: Option<u64>) -> String {
+    let Some(records_compared) = key_mismatch_at else {
+        return format!("Content diffs: {content_diffs}");
+    };
+    // `key_mismatch_at` is the 0-based index of the first mismatching pair, hence also
+    // the number of pairs examined before the stop.
+    let plural = if records_compared == 1 { "" } else { "s" };
+    format!(
+        "Content diffs: {content_diffs} \
+         (of {records_compared} record{plural} compared before pairing stopped)"
+    )
+}
+
 impl Command for CompareBams {
     fn execute(&self, _command_line: &str) -> Result<()> {
         validate_file_exists(&self.bam1, "First BAM")?;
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #753, opened 2026-08-14
docs(cli): correct the TieRuleArg default and pin it with a test

nh13:
The `--tie-rule` variant docs claimed `ulp-relative` was the default. It has not been since #666 ("restore fgbio-identical near-tie calls by default"), which moved `#[default]` to `FgbioCompat` and set `default_value = "fgbio-compat"` but left the variant doc comment untouched.

This is not a typo-class fix: `--tie-rule` decides whether fgumi reproduces fgbio's base calls on near-ties, and the doc told readers the opposite of what the tool does. Anyone reading the enum to decide whether they needed the flag would have concluded, wrongly, that fgumi defaults away from fgbio parity.

## Which side was stale

Two independent corroborations that the attribute is right and the prose is wrong, both from the same commit that introduced the drift:

- The `--tie-rule` field doc in `ConsensusCallingOptions` already says `` `fgbio-compat` (default)``.
- The domain enum `fgumi_consensus::TieRule` documents the same choice **correctly** — "The default:" sits on `FgbioCompat`.

So only the CLI mirror drifted. Both variants are reworded to match the domain enum's phrasing rather than newly invented text, which keeps the two enums readable side by side.

## Why a doc fix carries tests

The default is declared **three times, independently**, with nothing tying them together:

1. `#[default]` on the `TieRuleArg` variant
2. `default_value = "fgbio-compat"` on the `--tie-rule` arg
3. the hand-written `ConsensusCallingOptions::default`

That is the channel this drift came through, and nothing pinned the value at all. `tie_rule_defaults_to_fgbio_compat_everywhere` asserts all three agree *and* that the value is `fgbio-compat` — pinning the value matters as much as pinning the agreement, because `fgbio-compat` is a compatibility contract and a silent flip would change consensus output on near-ties rather than fail anything.

A second `#[rstest]` case table pins the `From<TieRuleArg> for TieRule` mapping. The CLI enum mirrors the domain enum, so that impl is the seam where the mirrored pair could be crossed — and crossing it would select the opposite tie rule silently rather than fail to compile.

The guard was verified by mutation: moving `#[default]` to `UlpRelative` fails the test with `left: UlpRelative, right: FgbioCompat`.

## Verification

- `cargo ci-test` — 7497 passed, 0 failures
- `cargo ci-fmt`, `cargo ci-lint`, `cargo ci-tag-literals` — clean
- `RUSTDOCFLAGS="-D warnings" cargo ci-doc` — clean
- `./scripts/publish-crates.sh --check` — clean

## Scope

`main-runall` and the three open PR branches stacked on it all carry the same stale doc, since they descend from `main`. Only `main` is fixed here; the integration branch picks it up on its next sync, which avoids four conflicting copies of the same edit.

Risk: command output changes none; `unsafe` changes none and `CLAUDE.md` allowlist changes none; memory bounds, queue capacity, and thread/backpressure policy changes none.

Fix: Correct `TieRuleArg` documentation and verify `fgbio-compat` defaults and CLI-to-domain mappings.

- Added coverage for enum, Clap, and `ConsensusCallingOptions` defaults.
- Confirmed both CLI variants map to the correct domain variants.
- Tests, formatting, linting, documentation, and crate-publishing checks pass.

### DIFF
--- a/src/lib/commands/common.rs
+++ b/src/lib/commands/common.rs
@@ -54,15 +54,17 @@ impl From<MethylationModeArg> for fgumi_consensus::MethylationMode {
 pub enum TieRuleArg {
     /// No-call a maximum separated from the runner-up by only a few ULPs.
     ///
-    /// The default. A separation that small is summation-order noise rather than evidence,
-    /// and the rule is independent of both read order and family depth.
+    /// A separation that small is summation-order noise rather than evidence, and the rule is
+    /// independent of both read order and family depth — the better rule on the merits, but
+    /// not the fgbio-compatible one, which is why it is not the default.
     #[value(name = "ulp-relative")]
     UlpRelative,
     /// Reproduce fgbio's tie rule exactly, defects included.
     ///
-    /// For byte-parity with fgbio output — chiefly cross-tool equivalency testing. Inherits
-    /// fgbio's order dependence and scale dependence, so it will call a base off one ULP of
-    /// accumulation noise.
+    /// The default: fgumi is a drop-in replacement for fgbio, so matching its output is the
+    /// contract. For byte-parity with fgbio output — chiefly cross-tool equivalency testing.
+    /// Inherits fgbio's order dependence and scale dependence, so it will call a base off one
+    /// ULP of accumulation noise.
     #[default]
     #[value(name = "fgbio-compat")]
     FgbioCompat,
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #754, opened 2026-08-14
fix(codec): emit the five codec stats rows fgbio reports

### LINKED ISSUE #748: codec --stats omits five metrics that fgbio emits, though all five are already computed
## Summary

`fgumi codec --stats` writes 18 rows where fgbio's `CallCodecConsensusReads --stats` writes 23 for the same input. Five keys are absent from fgumi's output. All five values are **already computed and stored** by fgumi — they are simply never serialized, so this is a reporting gap rather than missing functionality.

## Missing keys

These appear in fgbio's stats file and not in fgumi's (fgbio's own descriptions):

| key | fgbio description |
| --- | --- |
| `consensus_reads_rejected_hdd` | Consensus Reads Rejected: High Duplex Disagreement |
| `consensus_bases_emitted` | Total consensus bases emitted in consensus reads |
| `consensus_duplex_bases_emitted` | Consensus bases emitted with support from both strands of the duplex |
| `duplex_disagreement_base_count` | Number of consensus bases at which the top and bottom strands disagreed |
| `duplex_disagreement_rate` | Rate of top/bottom strand disagreement within duplex regions of consensus reads |

## They already exist in fgumi

`CodecConsensusStats` in `crates/fgumi-consensus/src/codec_caller.rs` carries all four counters, and the rate is derived:

- `consensus_reads_rejected_hdd`
- `consensus_bases_emitted`
- `consensus_duplex_bases_emitted` (incremented at `codec_caller.rs:1279`)
- `duplex_disagreement_base_count`
- `CodecConsensusStats::duplex_disagreement_rate()` (`codec_caller.rs:252`)

The conversion that produces the emitted rows is `to_metrics()` in `crates/fgumi-metrics/src/shared.rs:112`; these five do not make it through.

## Why it matters

`#513` moved codec stats to fgbio's KV format precisely so the two tools' metrics files can be compared directly. With five rows missing, that comparison is incomplete, and a `diff` of the two files reports absent keys rather than value differences.

`duplex_disagreement_rate` is the practically important one: it is the headline quality signal for a duplex consensus run. Without it, assessing the strand concordance of a set of consensus reads from fgumi's own output requires reconstructing it from per-read `ac`/`bc` tags by hand, whereas fgbio reports it directly.

## Suggested fix

Serialize the four counters and the derived rate through `to_metrics()`, matching fgbio's key names and description strings exactly so the two files remain diffable.

nh13:
Closes #748.

`fgumi codec --stats` wrote 18 rows where fgbio's `CallCodecConsensusReads --stats` writes 23 for the same input. The five missing rows are now emitted, with fgbio's exact key names, description strings and row order, so a `diff` of the two tools' stats files reports value differences rather than absent keys.

## Where the rows were lost

`CodecConsensusStats` carried four of the five values, but `CodecConsensusStats::to_metrics` (`src/lib/commands/consensus_runner.rs`) dropped them at the boundary into `ConsensusMetrics` — which is the only thing `to_kv_metrics` can see — so nothing downstream could serialize them. `ConsensusMetrics` gains the four counters plus a derived `duplex_disagreement_rate()`, and `to_kv_metrics` appends the five rows after `consensus_reads_emitted` for `ConsensusCallerKind::Codec` only, mirroring fgbio's `super.statistics ++ IndexedSeq(...)` in `CodecConsensusCaller.statistics`. The rate uses fgbio's zero-denominator guard and renders through the same `format_float` helper as the existing `frac_raw_reads_used` row, so every float row in the file agrees.

**`simplex` and `duplex` do not share this gap.** Only `CodecConsensusCaller` overrides `statistics` in fgbio (`UmiConsensusCaller.scala:524` is the sole other definition), so there is no simplex/duplex analogue of these five rows to drop.

## Two counters behind the rows were also wrong

The issue's premise is that all five values are already computed. Two of them were not, and serializing them as-is would have published known-wrong numbers, so each is fixed in its own commit for reviewability:

- `ff0b46f3` — `consensus_bases_emitted` was declared, merged across threads, and logged, but nothing ever incremented it. It reported 0 for every run, including in the run log at `codec_caller.rs:1717`. It is now accumulated at the point the consensus read is produced, matching fgbio's `totalConsensusBases += consensus.length`.
- `c9a8a5db` — `consensus_duplex_bases_emitted` and `duplex_disagreement_base_count` were accumulated *before* the two high-duplex-disagreement checks, so a molecule rejected for disagreeing contributed its bases and its disagreements to both totals anyway. fgbio increments them only in the accepted branch. The practical effect is that `duplex_disagreement_rate` — a rate over *emitted* duplex bases, and the headline signal this PR exists to expose — was inflated by exactly the molecules the threshold excluded, and grew with the rejection rate. The new end-to-end case adds a molecule whose strands disagree at every base; before the fix it reported 32 duplex bases against 24 bases of output.

Both are counter-population fixes, not changes to consensus base or quality math: no emitted record changes, and every existing test passes unmodified.

## Tests

- `crates/fgumi-metrics/src/consensus.rs` — `test_to_kv_metrics_codec_emits_codec_only_tail` pins the five rows as the *tail* of the file (keys, order, values, and fgbio's descriptions verbatim), `test_to_kv_metrics_non_codec_omits_codec_only_tail` pins that vanilla and duplex still end at `consensus_reads_emitted`, and `test_duplex_disagreement_rate_zero_denominator` pins the zero-denominator guard.
- `src/lib/commands/consensus_runner.rs` — `test_codec_stats_to_metrics_carries_codec_counters` pins the boundary the rows were lost at, across a merge.
- `crates/fgumi-consensus/src/codec_caller.rs` — `test_consensus_bases_emitted_totals_emitted_consensus_bases` checks the counter against the emitted record's own base count; `test_rejected_molecules_contribute_no_emitted_bases` covers both rejection triggers and asserts all three counters and the rate stay at zero.
- `tests/integration/test_codec_command.rs` — `test_codec_stats_emits_fgbio_codec_only_rows` runs the command end to end and reads the written stats file, checking counts against the output BAM rather than hard-coding them. Its second case adds a rejected molecule.

The metrics guide claimed `codec` used the horizontal TSV format; it has used the vertical key-value format since the codec stats moved to fgbio's layout. Corrected, and the five new rows are documented there.

Risk: codec stats output changes, pinned by fgbio-compatible keys, descriptions, and row order; `unsafe`: none; CLAUDE.md allowlist: unchanged; memory bounds, queue capacity, and thread/backpressure policy: none.

Fix: `fgumi codec --stats` now emits five fgbio-compatible codec metrics and tracks emitted and rejected consensus counters correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,9 +53,16 @@ All notable changes to this project will be documented in this file.
   `duplex_disagreement_rate` — using fgbio's key names, descriptions and row order, so a diff
   of the two tools' stats files reports value differences rather than absent keys.
   `duplex_disagreement_rate` is the headline strand-concordance signal for a duplex run and
-  previously had to be reconstructed by hand from per-read `ac`/`bc` tags. Relatedly,
-  `consensus_bases_emitted` was never incremented and so reported 0 everywhere it was
-  surfaced, including the run log ([#748](https://github.com/fulcrumgenomics/fgumi/issues/748)).
+  previously had to be reconstructed by hand from per-read `ac`/`bc` tags
+  ([#748](https://github.com/fulcrumgenomics/fgumi/issues/748)).
+
+  Two counters behind those rows were also wrong. `consensus_bases_emitted` was never
+  incremented and so reported 0 everywhere it was surfaced, including the run log. And
+  `consensus_duplex_bases_emitted` / `duplex_disagreement_base_count` were accumulated before
+  the high-duplex-disagreement check rather than after it, so molecules that were *rejected*
+  for disagreeing were counted as emitted — inflating the logged `duplex_disagreement_rate` by
+  exactly the templates it is meant to exclude. All three now count only emitted consensus
+  reads, matching fgbio.
 
 - `dedup --metrics` now reports the templates dropped by its template filter, broken out by
   reason. `dedup` is a read filter as well as a duplicate marker, but the drop counts were
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,16 @@ All notable changes to this project will be documented in this file.
 
 ### Bug Fixes
 
+- `codec --stats` now writes the five codec-specific rows fgbio's `CallCodecConsensusReads
+  --stats` writes — `consensus_reads_rejected_hdd`, `consensus_bases_emitted`,
+  `consensus_duplex_bases_emitted`, `duplex_disagreement_base_count` and
+  `duplex_disagreement_rate` — using fgbio's key names, descriptions and row order, so a diff
+  of the two tools' stats files reports value differences rather than absent keys.
+  `duplex_disagreement_rate` is the headline strand-concordance signal for a duplex run and
+  previously had to be reconstructed by hand from per-read `ac`/`bc` tags. Relatedly,
+  `consensus_bases_emitted` was never incremented and so reported 0 everywhere it was
+  surfaced, including the run log ([#748](https://github.com/fulcrumgenomics/fgumi/issues/748)).
+
 - `dedup --metrics` now reports the templates dropped by its template filter, broken out by
   reason. `dedup` is a read filter as well as a duplicate marker, but the drop counts were
   collected, merged across workers, and then discarded at the serialization boundary: a run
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #756, opened 2026-08-14
fix(codec): route per-read rejections to the --rejects output

### LINKED ISSUE #751: codec --rejects omits per-read rejections, so the rejects BAM does not reconcile with --stats
## Summary

`fgumi codec --rejects` writes only reads from groups that produced **no** consensus (or that hit a duplex-disagreement error). Reads rejected *within* a group that still yields a consensus are counted in `--stats` but never written to the rejects BAM, so the two outputs do not reconcile.

## Reproduction

On a small CODEC input (524 records, 95 molecules) run with `--rejects` and `--stats`:

- `--stats` reports `raw_reads_rejected = 42`, all under `raw_reads_rejected_for_minority_alignment`
- the rejects BAM contains **0 records**

Every one of those 42 reads was dropped inside a group that went on to emit a consensus, so none of them reached the rejects output.

## Cause

In `crates/fgumi-consensus/src/codec_caller.rs`, records are appended to `rejected_reads` at exactly two sites:

```rust
if self.track_rejects && e.is_duplex_disagreement() && !records.is_empty() { ... }
```
```rust
// When a group fails to produce consensus, all its records are rejected
if self.track_rejects && result.count == 0 && !records.is_empty() { ... }
```

Both are whole-group conditions. Per-read rejections recorded via `reject_records_count` (minority alignment, and the other caller-level reasons) increment the counters but are never routed to the writer.

fgbio's `CallCodecConsensusReads` routes rejected records to its `rejectsWriter` at the point of rejection, so its rejects BAM does reconcile with its stats.

## Impact

`--rejects` is the mechanism for auditing *which* reads were dropped and why. For the most common rejection reason on real data it currently yields nothing, so a divergence investigation cannot use it and has to reconstruct the picture from per-read consensus tags instead.

## Suggested fix

Route per-read rejections to the rejects output at the same site the counters are incremented, so that the rejects BAM record count equals `raw_reads_rejected` from `--stats`. That equality is a good invariant for a test to pin.

nh13:
Closes #751.

## What was broken

`codec --rejects` appended to the rejects buffer at exactly two sites, both whole-group conditions in `consensus_reads_typed`: a recoverable duplex disagreement, and a molecule that produced no consensus at all. Reads dropped *inside* a molecule that still consensed were counted in `--stats` and written nowhere, so the two outputs could not be reconciled. The issue's reproduction reports `raw_reads_rejected = 42` (all `minority_alignment`) against a rejects BAM holding zero records.

## Where the issue was wrong

The issue says per-read rejections are "recorded via `reject_records_count`" and never routed onward. The routing claim is right; the attribution is not. `MinorityAlignment` — the reason the reproduction is entirely made of — did **not** go through `reject_records_count` at all. `filter_to_most_common_alignment_raw` bumped `stats.rejection_reasons` and `stats.reads_filtered` inline, bypassing the shared tallying helper. The doc comment on `test_clip_overlap_failed_counted_and_labeled` asserting `reject_records_count` "is the single site every reject flows through" was therefore inaccurate when written. It is accurate now, and has been reworded to say so precisely.

Everything else in the issue holds: both append sites are whole-group, the per-read reasons never reach the writer, and record-count-equals-`raw_reads_rejected` is the right invariant to pin.

## The fix

Every rejection site records *which* records it rejected as it counts them, via a new `reject_records_at(indices, reason)` that both tallies and marks a per-group `reject_mask`. `consensus_reads_typed` then moves exactly the marked set into the rejects buffer, in input order, taking the group by value so the bytes are moved rather than copied.

That replaces the blanket "on `result.count == 0`, dump every record in the group" pass. Worth being explicit that this is a deliberate narrowing rather than a pure addition: the old pass matched the counters only by coincidence (the per-reason counts happen to sum to the group size when no record is filtered out uncounted), it would have double-written anything the new per-read routing emits, and it also emitted records that nothing counted. The records that are now *not* written are the ones neither used nor counted: secondary/supplementary alignments (which `codec` already drops before grouping via `consensus_pregroup_keep_flags`, so this is unreachable from the CLI) and reads dropped by the `--max-reads` cap. fgbio leaves both outside its rejects BAM as well — its whole-family codec rejects pass `(r1s.view ++ r2s.view)`, i.e. the survivors of `filterToMostCommonAlignment`, not the raw group.

## Tests

TDD; all three failed before the fix with an empty rejects output and passed after, with the counters already correct in the failing runs.

- `test_minority_alignment_rejects_are_tracked_when_the_family_still_consenses` — four templates, one carrying a minority indel pattern on both strands. Asserts a consensus is still emitted, the two counters, and that the retained bytes are *exactly* the minority template's two records (identity, in input order), not merely the right count.
- `test_fragment_read_rejects_are_tracked_when_the_family_still_consenses` — the same shape for a second reason, so the fix is pinned as a class rather than one branch.
- `test_codec_rejects_bam_reconciles_with_stats` (integration, single-threaded and `--threads 2`) — pins the invariant end to end: the rejects BAM record count equals `raw_reads_rejected` from `--stats` for an input where one molecule both emits a consensus and drops reads. Each reject is compared byte-for-byte against its input record, so a rejects BAM that dropped a tag or rewrote a CIGAR fails rather than passing a name-and-count check.

The existing rejects tests keep passing unchanged, which is what covers the narrowing: `test_codec_command_with_rejects` (whole molecule below `--min-reads`) and `test_high_duplex_disagreement_tracks_rejects` (byte identity on the HDD path) both still get every record of the failed molecule.

## Not changed here

- **`simplex` and `duplex` do not share this gap** — checked site by site rather than assumed. `VanillaUmiConsensusCaller` already routes its per-read rejections (`MinorityAlignment`, `Unmapped`, `ZeroLengthAfterTrimming`, `SecondaryOrSupplementary`) into `rejected_reads` at the point of rejection, and every duplex-level rejection is whole-group and routes the same records it counts. Both reconcile today. Duplex does have the *dual* defect — reads the single-strand caller drops inside a surviving molecule are counted into `ss_caller.stats`, which `DuplexConsensusCaller::statistics()` never merges, so they appear in neither output and `raw_reads_rejected` under-reports. That is a different bug and belongs in its own issue.
- **No `rr` tag.** fgbio tags every rejected record with `rr` = the reason code (`minority_alignment`, `insufficient_support`, …) in `UmiConsensusCaller.rejectRecords`, first-writer-wins. fgumi writes no `rr` tag anywhere, on any command — so this is a pre-existing gap across the whole rejects surface, not something codec lost. Adding it means rewriting the record bytes to append a tag and touching all four commands, so it wants its own change rather than being smuggled in here.

## Memory

No new growth. `reject_mask` is a `Vec<bool>` sized to one MI group and reused across groups. The rejected records themselves are moved out of the group's own `Vec<RawRecord>`, which is already fully resident, and both drain sites (`codec.rs` single-threaded per group, threaded per batch) already existed and are unchanged — the pre-fix worst case, a batch in which every group fails, dominates the post-fix one.

## Gate

`cargo ci-fmt`, `cargo ci-lint`, `cargo ci-test` (7498 passed, 30 skipped) and `cargo ci-doctest` all clean.

Risk: rejects output changes, pinned by original record bytes, input order, and matching `raw_reads_rejected` statistics; `unsafe` changes: none, with no `CLAUDE.md` allowlist update; memory bounds, queue capacity, and thread/backpressure policy changes: none.

The fix routes per-read rejections to the rejects output, including reads from molecules that still produce a consensus. Whole-group fallback rejection avoids duplicate or uncounted records. Single-threaded and multithreaded tests verify byte-preserving rejects output and statistics reconciliation.

### DIFF
--- a/crates/fgumi-consensus/src/codec_caller.rs
+++ b/crates/fgumi-consensus/src/codec_caller.rs
@@ -294,6 +294,21 @@ struct SingleStrandConsensus {
     is_negative_strand: bool,
 }
 
+/// Per-molecule duplex base tallies produced while building a duplex consensus.
+///
+/// These are returned rather than folded straight into [`CodecStats`] so the caller can
+/// apply them only *after* the consensus record has been serialized. A molecule whose output
+/// fails to build — an over-long input-derived UMI makes `try_build_record` reject the read
+/// name — emits nothing, so its duplex bases must not be counted, exactly as
+/// `consensus_reads_generated`/`consensus_bases_emitted` are only bumped on the success path.
+#[derive(Debug, Clone, Copy, Default)]
+struct DuplexBaseTally {
+    /// Duplex bases the molecule would contribute to `consensus_duplex_bases_emitted`.
+    bases_emitted: u64,
+    /// Duplex disagreements the molecule would contribute to `duplex_disagreement_base_count`.
+    disagreement_bases: u64,
+}
+
 /// Convert a per-base depth/error array to the `i16` `Array[Short]` form fgbio
 /// emits, saturating any value above the `Short` ceiling at `i16::MAX` (32767)
 /// rather than letting a raw `as i16` cast wrap it to a negative value. Matches
@@ -358,6 +373,12 @@ pub struct CodecConsensusCaller {
 
     /// Rejected raw BAM records (only populated when `track_rejects` is true)
     rejected_reads: Vec<Vec<u8>>,
+
+    /// Per-group mark of which input records were rejected, indexed by position in the
+    /// group's record slice. Sized at the top of [`Self::consensus_reads_raw`] and
+    /// consumed by [`Self::drain_marked_rejects`]; left empty (so marking is a no-op)
+    /// when `track_rejects` is false.
+    reject_mask: Vec<bool>,
 }
 
 #[expect(
@@ -422,6 +443,7 @@ impl CodecConsensusCaller {
             read_name_buf: Vec::new(),
             track_rejects: false,
             rejected_reads: Vec::new(),
+            reject_mask: Vec::new(),
         }
     }
 
@@ -455,6 +477,7 @@ impl CodecConsensusCaller {
         self.stats = CodecConsensusStats::default();
         self.ss_caller.clear();
         self.rejected_reads.clear();
+        self.reject_mask.clear();
     }
 
     /// Returns a reference to the rejected reads (raw BAM bytes).
@@ -605,6 +628,13 @@ impl CodecConsensusCaller {
     ) -> std::result::Result<ConsensusOutput, CodecConsensusError> {
         self.stats.total_input_reads += records.len() as u64;
 
+        // Size the per-group reject mark before any rejection site can run. Left empty when
+        // tracking is off, which makes `reject_records_at`'s marking a no-op.
+        self.reject_mask.clear();
+        if self.track_rejects {
+            self.reject_mask.resize(records.len(), false);
+        }
+
         if records.is_empty() {
             return Ok(ConsensusOutput::default());
         }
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #749, opened 2026-08-14
codec rejection reasons are attributed differently from fgbio, making the stats files non-comparable


nh13:
## Summary

`fgumi codec --stats` and fgbio's `CallCodecConsensusReads --stats` attribute the same rejected reads to different reasons. On a CODEC dataset of ~502M raw reads the two rejection-reason rows are near-mirror images, while the outcome rows barely move — so the great majority of the discrepancy is *attribution*, not a different accept/reject decision.

```
key                                                    fgbio      fgumi         delta
─────────────────────────────────────────────────────────────────────────────────────
raw_reads_rejected_for_clip_overlap_failed            192140      40344      -151796
raw_reads_rejected_for_indel_error_between_strands 106447346  106639085      +191739
raw_reads_rejected_for_r1_r2_overlap_too_short       5124422    5084049       -40373
raw_reads_rejected_for_minority_alignment            1511782    1511528         -254
─────────────────────────────────────────────────────────────────────────────────────
raw_reads_used                                     284047076  284047712         +636
consensus_reads_emitted                            103045729  103045851         +122
```

The four reason deltas sum to roughly the 636-read difference in `raw_reads_used`, which is what you would expect if the bulk is reads moving between buckets rather than changing verdict.

(Two further rows, `raw_reads_considered` -5,637,497 and `raw_reads_rejected_for_not_primary_fr_pair` -5,637,449, agree to within 48 reads and are the expected consequence of #509 filtering secondary/supplementary records before counting rather than rejecting them after. Those are not part of this issue.)

## Mechanism

The two tools evaluate the same conditions in a different order, so a read that would trip more than one is attributed to whichever check runs first.

fgumi's `ClipOverlapFailed` site is `crates/fgumi-consensus/src/codec_caller.rs:847` (`consensus_length < ss.bases.len()`), mirroring `CodecConsensusCaller.scala:243`. A test comment at `codec_caller.rs:3017` states the site is only reached when the read does not "first trip the earlier `IndelErrorBetweenStrands` phase check", and notes that this reason previously leaked into `IndelErrorBetweenStrands` before a fix. That adjacency is exactly what the mirrored deltas above show at scale.

## Requested change

[...]

PR #758, opened 2026-08-14
fix(duplex): reconcile single-strand rejections with --stats and --rejects

### LINKED ISSUE #757: duplex --stats omits single-strand rejections, so --rejects and --stats do not reconcile
## Summary

`DuplexConsensusCaller` never merges its single-strand caller's statistics, so reads dropped by the single-strand layer inside a molecule that still yields a duplex consensus are written to `--rejects` but are absent from `--stats`. The two outputs do not reconcile.

This is the mirror image of #751. There, reads were counted in `--stats` but never written to `--rejects`; here they are written to `--rejects` but never counted in `--stats`. The same invariant — rejects BAM record count == `raw_reads_rejected` — fails, in the opposite direction.

## Cause

`crates/fgumi-consensus/src/duplex_caller.rs`:

```rust
// :2406
self.stats.merge(&group_stats);
```

```rust
// :2437
fn statistics(&self) -> ConsensusCallingStats {
    self.stats.clone()
}
```

`self.stats` accumulates only `group_stats` from the duplex layer. `self.ss_caller` keeps its own `ConsensusCallingStats`, and nothing in the crate ever folds them together — `grep -rn 'ss_caller.statistics\|ss_caller\.stats' crates/fgumi-consensus/src/` returns no hits.

The single-strand layer does record rejections. `duplex_caller.rs:1977-1978`:

```rust
let (filtered_xs, _) = ss_caller.filter_by_alignment(x_sources);
let (filtered_ys, _) = ss_caller.filter_by_alignment(y_sources);
```

which reaches `vanilla_caller.rs:1095`:

```rust
self.stats.record_rejection(RejectionReason::MinorityAlignment, rejected_count);
```

`self` there is the `ss_caller`, so the count lands in the sub-caller's stats and stops.

Meanwhile the *records* are routed correctly (`duplex_caller.rs:2414-2419`):

```rust
let ss_rejected_raw = self.ss_caller.take_rejected_reads();
if duplex_rejected_raw.is_empty() {
    self.rejected_reads.extend(ss_rejected_raw);
} else {
    self.rejected_reads.extend(duplex_rejected_raw);
}
```

When the duplex layer rejects the whole group, `duplex_rejected_raw` already contains every raw record and the ss-layer set would be a duplicate subset — that branch is correct. But when the molecule *survives*, the ss-layer rejects are emitted to the rejects BAM while their counts remain stranded in `ss_caller.stats`.

`log_statistics()` (:2441-2455) does call `self.ss_caller.log_statistics()`, so the numbers are visible in the run log. They are simply absent from the serialized metrics.

## Impact

`raw_reads_rejected` under-reports for the duplex command by the number of single-strand drops inside surviving molecules, and the per-reason breakdown omits them entirely. Any check that reconciles `--rejects` against `--stats` — the invariant #751 asks for — will fail on duplex even after #751 lands, and will fail in a direction that looks like unexplained extra records rather than missing ones.

## Suggested fix

Merge the single-strand caller's statistics into the duplex caller's before returning them from `statistics()`, taking care not to double-count reasons already recorded at the duplex layer. Pin the same reconciliation invariant #751 pins: rejects BAM record count == `raw_reads_rejected`, asserted for a molecule that survives with at least one single-strand drop.

Note the codec caller should be checked for the same shape while fixing this; it also composes a single-strand caller.

nh13:
Closes #757.

## Where the issue was wrong

The issue says the records "are routed correctly" and reach `--rejects` while only the counts are stranded. They do not. `DuplexConsensusCaller` reaches the single-strand caller through `filter_by_alignment`, which *counts* what it dropped on `ss_caller.stats` but returns the rejected indices rather than the records — the raw bytes belong to the duplex caller, and the duplex caller discarded the indices (`let (filtered_xs, _) = ...`). The vanilla caller populates `rejected_reads` only inside `process_subgroup`, which the duplex path never enters, so `ss_caller.take_rejected_reads()` at the drain site always returned an empty vec.

So the single-strand drops reached **neither** output, not just `--stats`. `raw_reads_rejected` under-reported *and* the rejects BAM was missing records — which is what #756's own "Not changed here" section said ("they appear in neither output"), and is the stronger claim.

The issue is also wrong that the numbers were "visible in the run log". `log_statistics()` has no callers anywhere in the repo — the commands log through `log_consensus_summary(&metrics)`, derived from `statistics()`. The counts were invisible everywhere.

Everything else holds: `statistics()` returned only `self.stats`, nothing merged the sub-caller's, the reasons involved are recorded at `vanilla_caller.rs`'s alignment filter, and record-count-equals-`raw_reads_rejected` is the right invariant to pin.

## The fix

Two halves, both required for the invariant:

- `filter_by_alignment`'s rejected indices are mapped back to their raw records and handed to the sub-caller via a new `record_rejected_raw`, in **input order** (iterating the record slice, not the `HashSet`). They then drain through the existing `take_rejected_reads` path.
- `take_statistics` drains the sub-caller's counters per molecule, and `consensus_reads` folds that delta into its own.

## Avoiding the double count

A whole-group duplex rejection counts *and* emits every raw input record, so the single-strand layer's rejections are a subset already covered; adding them on top reports more rejected reads than the rejects BAM holds. The fold therefore runs on exactly one of the two paths.

The discriminator is the load-bearing detail. The pre-existing code used `duplex_rejected_raw.is_empty()`, which works only because it ran under `if track_rejects` — the payload is *also* empty whenever tracking is off, so reusing it to gate a *statistics* decision would double-count every run made without `--rejects`, i.e. the common case for `--stats`. `process_group` now returns an explicit `DuplexGroupOutcome::{Kept, RejectedWholeGroup(records)}` instead, and the enum's doc says why the payload cannot stand in for it.

Both hazards are pinned by tests that were confirmed to fail against the naive fix, not merely written alongside the right one:

- folding unconditionally makes `test_single_strand_rejections_are_not_double_counted_on_whole_group_rejection` report 20 rejected reads against 16 input records;
- returning `Kept` when tracking is off makes `test_whole_group_rejection_counts_are_independent_of_rejects_tracking` report 16 tracked vs 20 untracked.

Which reasons can collide was worked out rather than assumed. The sub-caller records exactly two reasons on this path — `Unmapped` (from `drop_unmapped_if_any_mapped`) and `MinorityAlignment` — and they are disjoint by construction, since the unmapped reads are removed before the minority pass counts what is left. Its `record_input` / `record_consensus` sites are in `process_group` and `consensus_reads`, which the duplex path does not call, so `total_reads` and `consensus_reads` are structurally zero in the delta; both are asserted to stay at the duplex layer's own values rather than left to trust.

## Tests

TDD; all four failed before the fix.

- `test_single_strand_rejections_reach_the_duplex_statistics` — a molecule that survives with one minority-alignment template on `/A`. Pins the two counters, that `total_reads`/`consensus_reads` are not inflated by the fold, and that the retained bytes are *exactly* that template's two records — byte-for-byte, in input order — not merely the right count.
- `test_single_strand_rejections_are_not_double_counted_on_whole_group_rejection` — `min_reads = 4` is met before single-strand filtering and missed after it, the only way to reach a whole-group rejection downstream of that layer. A control run pins the fixture is not vacuous: four clean templates per strand clear the same threshold, so the rejection is caused by the single-strand drops rather than by the raw depth.
- `test_whole_group_rejection_counts_are_independent_of_rejects_tracking` — the same molecule with and without tracking must report identical counters and per-reason breakdown.
- `test_duplex_rejects_bam_reconciles_with_stats` (integration, single-threaded and `--threads 2`) — the end-to-end invariant: rejects BAM record count equals `raw_reads_rejected` for an input where one molecule both emits a consensus and drops reads, with each reject compared byte-for-byte against the record it came from.

`test_duplex_command_rejects_contain_no_duplicates`'s doc claimed to cover the single-strand source; every molecule in it rejects at the duplex layer, so it never could. Reworded to say what it actually pins and to point at the new test for the other half.

`log_statistics` no longer logs the sub-caller separately — its counters are drained per molecule now, so the block would print zeroes and imply the drops went uncounted.

## Parity with fgbio

This moves the surviving-molecule path onto fgbio's behavior exactly. `DuplexConsensusCaller.scala:321-322` calls the inherited `filterToMostCommonAlignment`, so `rejectRecords(..., MinorityAlignment)` lands on the duplex caller's own counter and its own rejects writer — one layer, both outputs.

## Not changed here

- **The per-reason split on the whole-group path.** fgbio's `rejectRecords` is first-writer-wins (`recs.filterNot(_.contains(RejectReasonTag))`), so a record already rejected as `MinorityAlignment` keeps that reason and the whole-group reason applies only to the survivors — `DuplexConsensusCaller.scala:359-362` says so in a comment. fgumi attributes the entire raw group to the duplex reason instead. The **totals** agree either way (one record, one count), which is what this issue is about; the split does not, and fixing it means touching all five whole-group sites and their reject payloads. It belongs with the rejection-attribution work in #749.
- **Reads dropped by `create_source_read` returning `None`** (zero length after trimming, absent qualities). The duplex path drops them silently via `filter_map`, where `process_subgroup` counts them as `ZeroLengthAfterTrimming` and writes them. They are missing from both outputs symmetrically, so they do not break reconciliation — a separate under-count, not this one.
- **`codec`** — audited, no equivalent defect. Its composed single-strand caller is reached only through `consensus_call`, which records no statistics and touches no rejects buffer; codec does its own alignment filtering and counts it on its own stats. Nothing is stranded there.

## Memory

Bounded by one molecule's rejected records, and only when `--rejects` is set: the bytes are copied into the sub-caller's buffer and moved into the duplex caller's on the same call. Worth noting separately that single-threaded `duplex` never drains its rejects buffer inside the group loop the way `simplex` and `codec` do (`simplex.rs:467`, `codec.rs:513`), so that buffer holds every rejected record until the loop ends; this change adds to what accumulates there. That is a pre-existing shape, not introduced here, and the `--threads` path drains per batch.

## Gate

`cargo ci-fmt`, `cargo ci-lint`, `cargo ci-test` (7499 passed, 30 skipped) and `cargo ci-doctest` all clean.

Risk: output changes affect `--stats` and `--rejects` only, pinned by raw-record order, deduplication, and reconciliation tests; grouping, consensus, sort order, and corrected UMIs: none; `unsafe`: none; memory bounds, queue capacity, and thread/backpressure policy: none.

Fixes duplex rejection accounting. Single-strand rejected records now use the rejects buffer in raw-record order. Single-strand statistics now merge into duplex statistics without double counting whole-group rejections. Post-filter read thresholds prevent invalid single-strand consensus output.

Adds regression coverage for statistics, reject records, deduplication, tracking modes, alignment filtering, and single- versus multi-threaded execution.

### DIFF
--- a/crates/fgumi-consensus/src/duplex_caller.rs
+++ b/crates/fgumi-consensus/src/duplex_caller.rs
@@ -214,6 +214,23 @@ use crate::vanilla_caller::{
 use crate::{ReadType, SourceRead};
 use fgumi_raw_bam::{self as bam_fields, RawRecord, RawRecordView, UnmappedSamBuilder, flags};
 use fgumi_sam::SamTag;
+use std::collections::HashSet;
+
+/// What the duplex layer did with one MI group's raw input records.
+///
+/// The two variants cannot be told apart by the payload: records are collected only when
+/// rejects tracking is on, so an empty payload means "nothing to emit", not "the group
+/// survived". The single-strand layer's rejections are folded in on exactly one of these
+/// two paths, so conflating them would double-count every run made without `--rejects`.
+enum DuplexGroupOutcome {
+    /// The duplex layer did not reject the group as a whole. Any rejections for this group
+    /// came from the single-strand layer, and are held by the single-strand caller.
+    Kept,
+    /// The duplex layer rejected the whole group and already counted every one of its raw
+    /// records, so the single-strand layer's subset is covered. The payload is every raw
+    /// input record when rejects tracking is on, and empty when it is off.
+    RejectedWholeGroup(Vec<Vec<u8>>),
+}
 
 /// Duplex consensus read - matches fgbio's `DuplexConsensusRead`
 ///
@@ -304,7 +321,9 @@ pub struct DuplexConsensusCaller {
     produce_per_base_tags: bool,
     /// Cell barcode tag (e.g., CB) for preserving cell barcodes
     cell_tag: Option<Tag>,
-    /// Statistics tracking
+    /// Statistics tracking, including the single-strand caller's rejections: `consensus_reads`
+    /// folds that caller's per-molecule delta in here, so `ss_caller` keeps no running totals
+    /// of its own and this is the one place duplex counters live.
     stats: ConsensusCallingStats,
     /// Single-strand consensus caller (used for both /A and /B families)
     /// Uses `min_reads=1` like fgbio; filtering happens at duplex level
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #759, opened 2026-08-14
fix(consensus): clip overlapping reads by query distance, not reference distance

### LINKED ISSUE #752: overlap clipping reproduces fgbio #1090's reference/query distance conflation
## Summary

fgumi's overlap-clip arithmetic reproduces the reference-space/query-space conflation described in fgbio issue [#1090](https://github.com/fulcrumgenomics/fgbio/issues/1090). fgumi is structurally protected from the destructive consequence fgbio suffers, but it still computes the wrong clip point, and it still loses the affected read pair — just quietly rather than catastrophically.

## Background

fgbio's `CodecConsensusCaller.scala:246` carries the comment `// TODO remove this branch once SamRecordClipper has been updated` directly above its `ClipOverlapFailed` rejection. That reason exists specifically because fgbio's clipper is known to be wrong; the branch is fgbio self-diagnosing.

#1090's thesis: `mateUnSoftClippedEnd` is a *reference* coordinate built by adding a *query* distance. With an indel near the read end plus adapter read-through, `readPosAtRefPos(..., returnLastBaseIfDeleted = false)` lands inside a deletion, returns 0, and the computed clip becomes the entire read.

## The same arithmetic in fgumi

`crates/fgumi-raw-bam/src/overlap.rs`, `bases_extending_past_mate`:

```rust
let bases_past = compute_bases_past_ref_pos(&cigar_ops, this_pos, mate_unclipped_end);
read_length.saturating_sub(bases_past)
```

with `mate_unclipped_end` derived in `mate_soft_unclipped_from_mc` (documented as matching fgbio's `mateUnSoftClippedStart`/`mateUnSoftClippedEnd`), and `read_pos_at_ref_pos(..., return_last_base_if_deleted)` returning nothing for a position inside a deletion (pinned by an existing test: "Position in deletion - should return None with `return_last_base_if_deleted=false`"). Same construction, same failure mode.

## Why fgumi does not blow up the way fgbio does

Two structural differences, neither of which computes a better clip point:

1. fgumi derives both mates' clips from the **original** records. fgbio clips a record in place and then clips its mate against the already-modified (possibly unmapped) record, so the damage propagates from one read to its mate.
2. fgumi **clamps** (`clip_amount.min(bases.len())`) where fgbio calls `makeReadUnmapped` on an over-clip.

Consequence: on a dataset where fgbio rejected 192,140 reads as `clip_overlap_failed`, fgumi rejected 40,344. In a 95-molecule sample enriched for the disagreement, fgbio's 77 rejected reads decomposed into 11 genuinely-affected mates and 66 unaffected reads taken down with them.

## But fgumi is still wrong on the affected pairs

Run on an affected pair alone, with everything downstream held identical:

- correct clip per #1090's stated algorithm → emits a 127-base consensus
- fgbio → rejects as `r1_r2_overlap_too_short`
- fgumi → rejects as `indel_error_between_strands`

Both tools lose the pair. Across the sample, the buggy formula and the correct one disagreed on 113 of 262 templates, and all 113 were the deletion case: the formula answers with the full read length every time, where the correct answer ranged 0-40 bases.

## Suggested fix

Implement #1090's algorithm in `bases_extending_past_mate`: find the last reference position the read and its mate share, count the **query** bases each has past it, and clip that many. That is a self-contained change to fgumi's own arithmetic and does not depend on fgbio's timeline.

Note this will move reads between rejection buckets and may change consensus output on affected templates, so it should be landed with counts measured before and after rather than assumed.

nh13:
Closes #752. Stacked on #755 — base branch is `749/nhomer/match-fgbio-rejection-attribution`, so review that one first.

## The defect

`bases_extending_past_mate` decided where to trim a read that extends past its mate by comparing the read against the mate's soft-only unclipped **reference** end — a coordinate built by adding the mate's trailing soft clip, which is a **query** distance, to the mate's alignment end. Query and reference distances only agree across an ungapped alignment, so a pair that sequenced through into the adapter with an indel near the read's 3' end was mis-clipped. This is the fgumi side of fulcrumgenomics/fgbio#1090.

Two failure shapes, both reproduced as tests before the fix:

- **Deletion at the boundary.** The reference coordinate lands inside the read's deletion, where the read has no read position at all. The lookup answered "no such position" and the caller subtracted that from the read length, so the clip became the read's **entire length**. A positive read `2S124M1D3M` against a negative `3S124M2S` one base to its left was answered with all 129 of its bases; the correct clip is 2.
- **Insertion before the boundary.** The boundary moved the other way and the clip fell short by the insertion's length. fgbio#1090's own worked example — `70M10I23M47S` against `50S70M30S` — was answered with 40; the correct clip is 50, which is the answer that issue states.

fgumi never produced fgbio's catastrophic outcome (it derives both mates' clips from the original records and clamps rather than unmapping), but it computed the wrong clip point and quietly lost the affected pair.

## The algorithm

Measure both distances in query bases from a reference position the two reads actually share:

1. take the last reference position both alignments cover — the first, for a negative-strand read, whose 3' end is its leftmost;
2. count the query bases each read has past it;
3. clip the read down to the mate's count.

A boundary inside a deletion stops being a special case: the deletion consumes reference without query, so the walk carries past it and every base beyond it counts as past. Reading the deleted base as the last one before it (htsjdk's `returnLastBaseIfDeleted`) is **not** a fix — it replaces one reference-space answer with another and still carries the mate's soft clip in the wrong space. On the worked pair that knob yields 4 where the answer is 2.

### When the two alignments share no reference position

There is no landmark to anchor on, so this case keeps the previous arithmetic: project the read's soft-clipped tail one query base per reference base against the mate's soft-only unclipped boundary.

Answering 0 here is the intuitive reading — "no overlap, no read-through" — and it is false. Both alignments lie *within* the insert, so a pair that read through into the adapter can still leave them disjoint: it happens whenever each alignment stops short of the other's, which soft clipping at a variant or a mis-called base is enough to produce. Two 100-base reads over a 39-base insert, aligned `20M80S` at 1000 and `80S20M` at 1019, share exactly one reference position and give up 61 bases each; move the mate one base right and they share none, the insert is one base longer, and the answer must be 60 rather than 0. The extrapolation is the analytic continuation of the query-space count, which reduces to exactly that expression as the shared span shrinks to one position.

It is also safe against the defect this PR fixes: neither alignment places a base between the two, so neither can place an indel between them. An indel in the *sample* that falls in that gap is invisible to both reads and shifts the estimate by its own length in either direction, but the result is capped by the read's own soft clipping, so it never takes an aligned base.

**A read whose 3' end points *away* from a mate it does not overlap is now left alone.** That is an outward-facing template rather than read-through. Measuring it against the anchor like any other read counted every one of its query bases as past the mate and took its aligned ones — a forward `20M` at 301-320 against a reverse mate `10M5S` at 101-110 gave up 15 of its 20 aligned bases — which is the destructive answer this change exists to remove, on a geometry that cannot be verified. It is reachable per strand by a different inconsistency: `is_fr_pair_raw`'s forward arm reads `TLEN` (htsjdk's per-record test, which htsjdk 5 has since changed to prefer the mate CIGAR when `MC` is present), and its reverse arm reads the mate position recorded on the record, which the mate record in hand can contradict.

Mechanically, the two entry points now hand the mate's CIGAR to the shared core instead of two derived reference boundaries, and the MC-tag path parses the tag into BAM-packed ops so both paths share one set of CIGAR walkers. The parser gains one rejection: a run length past the 28-bit field a BAM CIGAR operation can hold cannot describe a real alignment, and this module's contract is to fail closed on a malformed MC tag.

## Measured rejection-bucket movement

`simplex`, on a synthetic read-through family of three pairs carrying a 1-base deletion three bases from the positive read's 3' end (`--min-reads 3`), before → after:

| stat | before | after |
| --- | --- | --- |
| `raw_reads_considered` | 6 | 6 |
| `raw_reads_used` | 0 | **6** |
| `raw_reads_rejected` | 6 | **0** |
| `raw_reads_rejected_for_zero_bases_post_trimming` | 3 | **0** |
| `raw_reads_rejected_for_orphan_consensus` | 3 | **0** |
| `consensus_reads_emitted` | 0 | **2** |

Every R1 was clipped to nothing, which orphaned the R2 half and lost the whole family — silently, since nothing about the input is malformed. After the fix both strands emit a 127-base consensus.

The codec caller, two templates of the same shape (`2S124M1D3M` at 200 against `3S125M` at 200, where the mate's alignment end itself lands in the deletion): `indel_error_between_strands` 4 → **0**, with one consensus emitted where there was none.

**No count pinned by #755 moved.** `test_clip_overlap_failed_attribution_matches_fgbio` and `test_clip_overlap_failed_counted_and_labeled` use ungapped fixtures (`50M`, `102M`, `40M`), and the change is provably inert on those — see below.

## Scope of the behavior change

Across 448,000 randomly generated FR pairs (20-base reads, leading and trailing soft clips 0-6, offsets ±25, half carrying a 1-3 base indel), gated by the same FR test the code applies, the new and old answers are **identical for every ungapped pair** — zero differences. They differ only when an indel lies between the shared anchor and the read's end. That is the intended blast radius, and it is why the pinned codec counts are unchanged.

Separately, a set of 20,000 randomly generated FR pairs (100-base reads, soft clips 0-80, mate offsets −160 to +260) spanning overlapping, touching and disjoint geometries was run against [fulcrumgenomics/fgbio#1172](https://github.com/fulcrumgenomics/fgbio/pull/1172), the independent rework of the same arithmetic in Scala: **40,000 evaluations, zero disagreements** between the two implementations. On the same set the new answers differ from the superseded reference-space formula on 662 evaluations, every one of them on a pair carrying an indel.

## Tests

`crates/fgumi-raw-bam/src/overlap.rs`
- `test_deletion_at_mate_boundary_clips_the_query_distance` — the worked pair, through both entry points: 2, was 129.
- `test_deletion_at_mate_boundary_leaves_the_unaffected_mate_alone` — the negative read of the same pair was already right and stays at 2.
- `test_deletion_at_mate_boundary_does_not_consume_the_whole_read` — pins that the answer is strictly below the read length, so the downstream clamp is no longer what stands between this and a deleted read.
- `test_insertion_before_mate_boundary_is_not_under_clipped` — fgbio#1090's example: 50, was 40.
- `test_ungapped_overlap_is_unchanged` — control, both strands, both entry points.
- `test_num_bases_extending_past_mate_raw_read_entirely_past_mate` — the outward-facing guard on the forward strand, reached when the per-record `TLEN` orientation test disagrees with the MC tag. Was 15 (of 20 aligned bases); now 0.
- `test_num_bases_extending_past_mate_vs_mate_raw_read_entirely_before_mate` — the reverse-strand mirror, reached through the mate-record entry point when the recorded mate position disagrees with the mate in hand. Was the whole read; now 0.
- `test_disjoint_alignments_still_clip_their_read_through` — the continuity pair above (61 with one shared reference position, 60 with none), on both strands. It fails against an implementation that answers 0 for disjoint alignments.
- `test_query_bases_{past,before}_ref_pos` and `query_bases_partition_the_read` — the walkers, including a target inside a deletion and a leading soft clip at the alignment start (which is what makes the two modes mirror images rather than one offset by a base).
- `test_parse_mc_cigar_ops` / `mc_cigar_path_rejects_malformed` — the parser's validation, unchanged apart from the new out-of-range run length, plus that a rejected MC tag leaves the read unclipped.

`crates/fgumi-consensus/src/codec_caller.rs` — `test_indel_at_overlap_boundary_still_calls_a_consensus`: the family calls a consensus with no rejections, where it was wholly rejected.

`tests/integration/test_simplex_command.rs` — `test_simplex_indel_at_overlap_boundary_still_calls_a_consensus`: end-to-end through the CLI, pinning the **full consensus sequence** of both strands rather than their lengths, so a clip of the right size taken from the wrong end fails too.

## Out of scope

`SamRecordClipper::num_bases_extending_past_mate{,_raw}` and `record_utils::num_bases_extending_past_mate` in `fgumi-sam` carry the same arithmetic and serve `fgumi clip`. Fixing them means changing a public signature to take the mate's CIGAR rather than two derived reference boundaries, which is a separate change; filed as #760. The consensus callers — the subject of #752 — all route through `fgumi-raw-bam::overlap` and are fixed here.

Risk: consensus output changes, pinned by regression fixtures and unchanged ungapped/#755 counts; `unsafe` changes: none; memory, queue, and backpressure policy changes: none.

Fix: compute overlap clipping in query space with shared CIGAR walkers. The change handles indels, disjoint alignments, and unverifiable outward-facing reads without whole-read over-clipping.

- Validates packed BAM CIGAR operations from MC tags, including malformed input and 28-bit run-length overflow.
- Adds regression coverage for deletion and insertion boundaries, disjoint alignments, outward-facing reads, codec consensus, and the simplex CLI.
- Preserves query-space extrapolation for disjoint read-through pairs.
- Keeps outward-facing reads unclipped when geometry cannot verify read-through.
- Updates review guidance to use `mate_soft_unclipped_from_ops`.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -240,7 +240,7 @@ reviews:
         it, and integer accumulation over attacker-controlled CIGAR or tag values
         that can overflow `i32`/`u32`. DO NOT propose clamping unclipped
         coordinates. `unclipped_start_from_raw_cigar`, `unclipped_other_start` /
-        `unclipped_other_end`, and `mate_soft_unclipped_from_mc` produce
+        `unclipped_other_end`, and `mate_soft_unclipped_from_ops` produce
         template-coordinate SORT-KEY LANES, not genomic positions: they may
         legitimately go negative or exceed contig bounds, and a `.max(0)` or
         contig-floor clamp on one side desynchronizes the two mates of a template
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,6 +118,26 @@ All notable changes to this project will be documented in this file.
   exactly the templates it is meant to exclude. All three now count only emitted consensus
   reads, matching fgbio.
 
+- Overlap clipping — how far a read is trimmed back for extending past its mate, in `simplex`,
+  `duplex` and `codec` — now measures that distance in query bases rather than mixing query and
+  reference distances. The mate boundary it clipped against was a *reference* coordinate built
+  by adding the mate's trailing soft clip, a *query* distance, to the mate's alignment end.
+  Those two spaces only agree across an ungapped alignment, so a pair that sequenced through
+  into the adapter with an indel near the read's 3' end was mis-clipped: with a deletion at the
+  boundary the read had no position there at all and the computed clip became the read's
+  **entire length**, and with an insertion it under-clipped by the insertion's length. The clip
+  is now the number of query bases the read has past the last reference position it shares with
+  its mate, less the number the mate has, which is exact in the presence of indels and
+  unchanged for every ungapped overlap
+  ([#752](https://github.com/fulcrumgenomics/fgumi/issues/752), the fgumi side of
+  [fgbio#1090](https://github.com/fulcrumgenomics/fgbio/issues/1090)).
+
+  This changes consensus output on affected templates. A read that was clipped to nothing was
+  banked under `raw_reads_rejected_for_zero_bases_post_trimming` and could orphan its mate; on
+  a read-through family with a 1-base deletion three bases from the positive read's 3' end,
+  `simplex` went from 6 of 6 raw reads rejected and no consensus emitted, to 6 of 6 used and
+  both strands emitted.
+
 - `dedup --metrics` now reports the templates dropped by its template filter, broken out by
   reason. `dedup` is a read filter as well as a duplicate marker, but the drop counts were
   collected, merged across workers, and then discarded at the serialization boundary: a run
--- a/crates/fgumi-consensus/src/codec_caller.rs
+++ b/crates/fgumi-consensus/src/codec_caller.rs
@@ -3383,6 +3383,103 @@ mod tests {
         assert_rejected_whole_family(&caller, &output, record_count);
     }
 
+    /// Builds a CODEC family of `templates` read pairs that sequence through the insert and
+    /// into the adapter, with a 1-base deletion three bases from the positive read's 3' end —
+    /// the fulcrumgenomics/fgumi#752 shape.
+    ///
+    /// - positive: `2S124M1D3M` at 200 (reference span 200-327, the deletion at 324)
+    /// - negative: `3S125M`     at 200 (reference span 200-324)
+    ///
+    /// The negative read's soft-only unclipped *reference* end is 324 — squarely inside the
+    /// positive read's `1D`, where the positive read has no read position at all. Measuring the
+    /// clip in query space from the last reference position the two reads share (324) instead
+    /// gives the positive read 3 query bases past it against the negative read's 0, so it gives
+    /// up 3 and keeps 126.
+    fn indel_readthrough_family(templates: usize) -> Vec<RawRecord> {
+        (0..templates)
+            .flat_map(|i| {
+                create_fr_pair(
+                    &format!("t{i}"),
+                    200,
+                    200,
+                    129,
+                    35,
+                    &[
+                        (Kind::SoftClip, 2),
+                        (Kind::Match, 124),
+                        (Kind::Deletion, 1),
+                        (Kind::Match, 3),
+                    ],
+                    &[(Kind::SoftClip, 3), (Kind::Match, 125)],
+                    "mi",
+                    Some("ACC-TGA"),
+                    false,
+                    true,
+                )
+            })
+            .collect()
+    }
+
+    /// A read-through pair with an indel at the overlap boundary must survive to a consensus.
+    ///
+    /// The overlap clip is the first thing the caller computes, and the superseded
+    /// reference-space arithmetic answered with the positive read's *entire* 129 bases here
+    /// (the boundary landed inside its deletion, where the read has no position at all). The
+    /// read was clipped away to nothing and the whole family was banked as a rejection. With
+    /// the clip measured in query space it gives up 3 bases and the family calls normally.
+    ///
+    /// This is the end-to-end half of the fgumi#752 fix: the unit tests in
+    /// `fgumi_raw_bam::overlap` pin the arithmetic, this pins what it costs a family to get it
+    /// wrong.
+    #[test]
+    fn test_indel_at_overlap_boundary_still_calls_a_consensus() {
+        let options = CodecConsensusOptions {
+            min_reads_per_strand: 1,
+            min_duplex_length: 1,
+            ..Default::default()
+        };
+        let mut caller = CodecConsensusCaller::new("codec".to_string(), "RG1".to_string(), options);
+
+        let fixture = indel_readthrough_family(2);
+        let record_count = fixture.len();
+
+        let output =
+            caller.consensus_reads_from_sam_records(fixture).expect("the family calls a consensus");
+
+        let stats = caller.statistics();
+
+        // Pin the emitted consensus, not merely its existence. This PR changes the
+        // overlap-clip arithmetic; a regression that clips the wrong (still nonzero) number of
+        // query bases would keep `count > 0` and zero rejections intact while silently shifting
+        // the output. The family shares one MI, so it collapses to exactly one molecule; the
+        // query-space clip fixes both the 127-base length and the reference-derived bases below,
+        // and the superseded reference-space arithmetic would have produced a different record.
+        let records = ParsedBamRecord::parse_all(&output.data);
+        assert_eq!(output.count, 1, "the single molecule must emit exactly one consensus record");
+        assert_eq!(records.len(), 1, "exactly one consensus record must be serialized");
+        assert_eq!(
+            records[0].bases.len(),
+            127,
+            "the query-space overlap clip fixes the consensus length",
+        );
+        assert_eq!(
+            &records[0].bases[..],
+            b"ANGNNANTNNNTNANNNTNNNNNNNNTNNNNGTNNNNNNNNTANNAATTNNNNNNNANNNTNNNNANNNTNNNNNNNAATTNNTANNNNNNNNACNNNNANNNNNNNNANNNTNANNNANTNNCNTT",
+            "the consensus bases are fixed by the fixture; a mis-placed clip shifts them",
+        );
+
+        assert_eq!(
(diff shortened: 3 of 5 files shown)


nh13:
@coderabbitai pause

nh13:
Force-pushed an amendment. The query-space count is unchanged; two things moved.

**The disjoint-alignment branch was contested and is now settled in this PR's favour.** fgbio's independent rework of the same defect ([fulcrumgenomics/fgbio#1172](https://github.com/fulcrumgenomics/fgbio/pull/1172)) answered `0` there, on the reasoning that without a shared reference position neither read can have read through into the other's adapter. That reasoning is false — both alignments lie within the insert, so a read-through pair is left disjoint whenever each alignment stops short of the other's. Answering 0 is also discontinuous: `20M80S` at 1000 against `80S20M` at 1019 share one reference position and give up 61 bases, and moving the mate one base right must give 60. On a shared set of 20,000 pairs spanning overlapping, touching and disjoint geometries, the 0-answering revision differed from the superseded formula on 28 **indel-free** evaluations, which a fix for a reference/query conflation must not do. fgbio#1172 has been amended to match; the two implementations now agree on all 40,000 evaluations of that set.

**One case moved the other way.** A read whose 3' end points *away* from a mate it does not overlap was still being measured against the anchor, which counts every one of its query bases as past the mate and takes its **aligned** ones — a forward `20M` at 301-320 against a reverse mate `10M5S` at 101-110 gave up 15 of its 20 aligned bases. That is the destructive answer this PR exists to remove, resurfacing in the degenerate branch, on a geometry that describes an outward-facing template rather than read-through and that neither `is_primary_fr_pair_raw` nor fgbio's `isFrPair` accepts. It now returns 0.

It is reachable per strand by a different inconsistency, and both are now pinned:

- forward: `is_fr_pair_raw`'s forward arm reads `TLEN`, which the `MC` tag can contradict (`test_num_bases_extending_past_mate_raw_read_entirely_past_mate`, previously asserting 15);
- reverse: that arm is CIGAR-derived but reads the mate position recorded *on the record*, while `num_bases_extending_past_mate_vs_mate_raw` measures against the mate record in hand (`test_num_bases_extending_past_mate_vs_mate_raw_read_entirely_before_mate`, previously the whole read).

With that guard in place the function's answer is now strictly below the read length on every input, which is the property #752 was really about.

Also added `test_disjoint_alignments_still_clip_their_read_through`, the 61/60 continuity pair on both strands — it fails against an implementation that answers 0 for disjoint alignments.

Separately, and out of scope here: `is_fr_pair_raw` mirrors htsjdk's *per-record* orientation test as it was before htsjdk 5, which changed the forward arm to prefer the mate CIGAR when `MC` is present. Bringing it in line would close the forward-strand door properly rather than guarding behind it, but it changes FR classification everywhere. Noted on #760.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #763, opened 2026-08-15
codec caller test fixture stores the reverse read's SEQ in read orientation, so its consensus is almost all N


nh13:
`create_fr_pair`, the CODEC caller's read-pair test fixture in `crates/fgumi-consensus/src/codec_caller.rs`, stores a reverse-strand record's `SEQ` in *read* orientation. A BAM stores it in *reference* orientation, and the caller reads it that way, so every family built by this fixture has its two strands pointing in opposite directions.

## Symptom

The consensus these fixtures produce is almost entirely `N`: the two strands disagree at roughly three positions in four, which is what comparing a sequence against its own reverse complement gives. Nothing catches it, because no test built on this fixture asserts consensus *bases* — only lengths, counts and rejection reasons — and `max_duplex_disagreement_rate` defaults to 1.0, so a family that disagrees everywhere is still emitted.

## Reproduction

Call any `create_fr_pair` family and print the consensus. For a plainly overlapping pair — forward `2S127M` at 200, reverse `3S126M` at 200 — with `min_reads_per_strand = 1` and `min_duplex_length = 1`:

```
len=128 n_count=89
bases   = AANNNNNTTNNNNANNGTNNNNNNNNTNNN
REF[199..] = GTAACTTTTTGATAGTAGCGGGAGTAGGAG
```

Both reads were cut from the same reference bases, so the consensus should carry those bases. The few non-`N` positions are where the two orientations happen to coincide.

Deleting the two reverse-complements in `create_fr_pair`

```rust
let seq1: Vec<u8> = if strand1_reverse { reverse_complement(&seq1_fwd) } else { seq1_fwd };
let seq2: Vec<u8> = if strand2_reverse { reverse_complement(&seq2_fwd) } else { seq2_fwd };
```

gives, for the same family:

```
len=128 n_count=0
bases   = AAGTAACTTTTTGATAGTAGCGGGAGTAGG
```

— the reference sequence, behind the two soft-clipped placeholder bases.

## Blast radius of that change

96 of the 102 CODEC caller tests pass unchanged with the reverse-complements removed. The six that fail are the ones that assert duplex *disagreement*:

- `test_consensus_reads_typed_disagreement_rate`
- `test_consensus_reads_typed_disagreement_count`
- `test_high_duplex_disagreement_counted` (both cases)
- `test_high_duplex_disagreement_tracks_rejects`
- `test_not_emit_consensus_high_disagreement`

[...]

PR #767, opened 2026-08-15
fix(codec): clamp the duplex window to the region both strands align

### LINKED ISSUE #761: codec counts indel-free duplex families as indel_error_between_strands when the two strands' alignments dovetail
`fgumi codec` rejects duplex families under `indel_error_between_strands` when neither strand's alignment contains an indel.

## Symptom

A CODEC family whose two strands both align across the same insert, but whose alignments *dovetail* — the reverse read's alignment starting left of the forward read's, and/or the forward read's alignment ending right of the reverse read's — is rejected, and every one of its records is counted under `indel_error_between_strands`. This happens even when both CIGARs are pure matches.

Dovetailing is routine for CODEC rather than exotic: R1 and R2 cover the same molecule end to end, so the aligner regularly places one strand's terminal base as aligned where it soft-clipped the other strand's.

## Reproduction

Each row below is a single duplex template (one forward read, one reverse read) run through `CodecConsensusCaller` with `min_reads_per_strand = 1` and `min_duplex_length = 1`. Positions are 1-based leftmost alignment positions; both reads are 129 bases.

| # | forward | reverse | indel present? | observed |
| --- | --- | --- | --- | --- |
| 1 | `2S126M` @ 201 | `3S127M` @ 200 | no | `indel_error_between_strands`, no consensus |
| 2 | `2S127M` @ 201 | `3S124M2S` @ 200 | no | `indel_error_between_strands`, no consensus |
| 3 | `2S124M1D3M` @ 201 | `3S124M2S` @ 200 | yes — past the reverse read's aligned end | `indel_error_between_strands`, no consensus |
| 4 | `2S127M` @ 200 | `3S126M` @ 200 | no | one consensus, nothing rejected |

Case 1 is the smallest shape: the reverse read simply aligns one base further left than the forward read, and neither CIGAR has an indel. Case 4 is the same family without the dovetail, and it consenses.

The test suite already contains a family of this shape. `window_end_past_r2_fixture` in `crates/fgumi-consensus/src/codec_caller.rs` builds four records whose CIGARs are `100M`, `40M`, `30M` and `50M` — not an indel among them — and `test_window_end_past_r2_stays_indel_error` pins that the whole family is counted under `IndelErrorBetweenStrands`.

## Expected

`indel_error_between_strands` should count families whose two strands disagree about an indel *inside the region both strands align*. Cases 1 and 2 have no indel at all. Case 3's deletion sits outside the region the two strands share, so the strands are in register everywhere they overlap.

nh13:
Closes #761. Stacked on #759 — base branch is `752/nhomer/clip-overlap-in-query-space`, so review that one first.

## The defect

The codec caller took the duplex window to be `[negative.start, positive.end]`, mirroring `CodecConsensusCaller.scala:211-213`. That is the intersection of the two alignments only while the positive one opens first and closes first. CODEC breaks that assumption routinely: both strands cover the same insert end to end, so the aligner takes as matching a terminal base it soft-clipped on the other strand and the two alignments *dovetail* by a base or two.

The window then reached past one read or the other, and both of its consumers were answered about a reference position that read never aligned. `check_overlap_phase_raw` read the out-of-range boundary as htsjdk's `0` and failed; `compute_consensus_length_raw` had no position to anchor on at all. Both land the family in `indel_error_between_strands` — whether or not it contains an indel.

Three shapes, all reproduced as tests before the fix (one duplex template, `min_reads_per_strand = 1`, `min_duplex_length = 1`):

| forward | reverse | indel present? | before | after |
| --- | --- | --- | --- | --- |
| `2S126M` @ 201 | `3S127M` @ 200 | no | `indel_error_between_strands` | 128-base consensus |
| `2S127M` @ 201 | `3S124M2S` @ 200 | no | `indel_error_between_strands` | consensus |
| `2S124M1D3M` @ 201 | `3S124M2S` @ 200 | yes, past the reverse read's aligned end | `indel_error_between_strands` | 127-base consensus |

The first is the smallest: the reverse read aligns one base further left than the forward read, and neither CIGAR has an indel.

The suite already contained a family of this shape. `window_end_past_r2_fixture` builds four records whose CIGARs are `100M`, `40M`, `30M` and `50M` — not an indel among them — and `test_window_end_past_r2_stays_indel_error` pinned that the whole family was counted under `IndelErrorBetweenStrands`.

## The fix

Clamp the window to the intersection: `(neg_start.max(pos_start), pos_end.min(neg_end))`.

This is inert whenever the alignments do not dovetail, since `max`/`min` then select the same two boundaries. An empty intersection is not a special case either: it can only arise when the two alignments do not overlap, where the clamp again yields the same two boundaries, and `min_duplex_length` (at least 1, enforced by the command) rejects it as before.

Two consequences worth naming:

- **The out-of-range sentinel is no longer reachable from the caller.** Both boundaries now lie inside both reads, where every reference position resolves — deletions included, since the lookups pass `return_last_base_if_deleted`. The substitution stays in `check_overlap_phase_raw` because it is a predicate over two arbitrary boundaries and has no business panicking on one, and `test_check_overlap_phase_out_of_range_start_uses_fgbio_sentinel` still pins it as fgbio's predicate rather than a stricter one.
- **`--min-duplex-length` now measures the region both strands align**, not the span their alignments cover between them. For a dovetailed family those differ by the dovetail, so a threshold can newly reject one; `test_min_duplex_length_is_measured_over_the_shared_region` pins both sides of the boundary.

## Measured rejection-bucket movement

**A real CODEC library**, aligned to hs38DH and grouped with `group --strategy paired --edits 1`, then `codec --min-reads 1`, before → after:

| stat | before | after |
| --- | --- | --- |
| `raw_reads_considered` | 6,104,018 | 6,104,018 |
| `raw_reads_rejected_for_indel_error_between_strands` | 1,586,438 | **64,461** |
| `raw_reads_rejected_for_clip_overlap_failed` | 1,205 | **4,032** |
| `raw_reads_used` | 1,254,091 | **2,773,241** |
| `frac_raw_reads_used` | 0.2055 | **0.4543** |
| `consensus_reads_emitted` | 108,256 | **262,904** |

Every other reason row is unchanged, `r1_r2_overlap_too_short` included. The counts close exactly: 1,521,977 raw reads left `indel_error_between_strands`, 2,827 of them landed in `clip_overlap_failed`, and the remaining 1,519,150 are precisely the increase in `raw_reads_used`.

**The change is purely additive on that library.** Comparing the two consensus BAMs on (read name, `SEQ`, `QUAL`): 0 records present before and absent after, and 154,648 new ones. Every consensus read the pre-fix binary emitted is emitted identically.

**The codec caller unit fixtures**, per the table above: `IndelErrorBetweenStrands` 6 → 0 across the three geometries, with three consensus reads emitted where there were none.

## Divergence from fgbio

fgbio computes the same window and so has the same defect; this makes fgumi's codec output differ from fgbio's on dovetailed families until the equivalent change lands there. The divergence is confined to families whose two alignments dovetail — on every other family the clamp selects the same boundaries fgbio does, and the output is what it was.

`window_end_past_r2_fixture`'s family moves bucket as a result: `IndelErrorBetweenStrands` → `ClipOverlapFailed`. The verdict is unchanged (rejected, nothing emitted), and `ClipOverlapFailed` is the reason that is true of it — its consensus comes out 70 bases against a 100-base R1 single-strand consensus, which is exactly what fgbio's `CodecConsensusCaller.scala:245-247` labels. The test is renamed `test_r1_running_past_r2_is_not_an_indel_error` and pins that, plus `assert_rejected_whole_family` for the unchanged verdict.

## Tests

`crates/fgumi-consensus/src/codec_caller.rs`

- `test_dovetailed_starts_without_an_indel_call_a_consensus` — the smallest shape, one base of dovetail and no indel: a 128-base consensus with nothing rejected.
- `test_terminal_indel_outside_the_shared_region_calls_a_consensus` — #761's geometry, dovetailed at both ends with a deletion two bases past where the reverse read stops aligning: a 127-base consensus, nothing rejected.
- `test_indel_inside_the_shared_region_is_still_rejected` — the control. Same geometry with the deletion moved into the middle of the shared region, where the strands really are out of register: still `IndelErrorBetweenStrands`, still the whole family.
- `test_min_duplex_length_is_measured_over_the_shared_region` — 126 shared positions satisfy a threshold of 126 and not 127.
- `test_r1_running_past_r2_is_not_an_indel_error` — the renamed pin described above.
- `test_clip_overlap_failed_attribution_matches_fgbio` and `test_indel_at_overlap_boundary_still_calls_a_consensus` are unchanged controls; the clamp is provably inert on the first (the longest R1 lies wholly inside the longest R2, so the intersection is the whole of it) and on the second (the alignments start together).

## Out of scope

These tests assert consensus *lengths* rather than consensus *bases*. `create_fr_pair` stores a reverse-strand record's `SEQ` in read orientation where a BAM stores it in reference orientation, so the two strands of every family it builds point opposite ways and the consensus comes out mostly `N` — for all of them, not just these. Filed as #763, with the blast radius measured; it is not this change's to fix, and the lengths this change is about are unaffected by it.

Risk: consensus and emitted duplex metrics change for valid dovetailed families, pinned by shared-region windowing and regression tests; grouping, sort order, and corrected UMIs: none; `unsafe` changes: none, and CLAUDE.md allowlist changes: none; memory, queue, and thread/backpressure policy changes: none.

Fix CODEC duplex window calculation by restricting it to the intersection of both strand alignments. This prevents out-of-range lookups and limits `--min-duplex-length` to the shared aligned region.

Indels outside the shared region no longer cause `indel_error_between_strands`. Indels inside the shared region remain rejected. Tests cover dovetailed alignments, clipping, terminal and internal indels, minimum duplex length, consensus lengths, rejection attribution, and emitted-statistic handling when output construction fails.

### DIFF
--- a/crates/fgumi-consensus/src/caller.rs
+++ b/crates/fgumi-consensus/src/caller.rs
@@ -423,7 +423,8 @@ pub enum RejectionReason {
     InsufficientOverlap,
     /// Only one of R1/R2 consensus was generated (orphan)
     OrphanConsensus,
-    /// Overlap boundary lands in an indel between strands
+    /// The two strands disagree about an indel inside the region they both align, so their
+    /// query bases cannot be laid out on one axis (codec)
     IndelErrorBetweenStrands,
     /// Overlap clipping failed: the computed consensus length is shorter than a
     /// single-strand consensus, so the overlapping ends could not be clipped
--- a/crates/fgumi-consensus/src/codec_caller.rs
+++ b/crates/fgumi-consensus/src/codec_caller.rs
@@ -815,8 +815,32 @@ impl CodecConsensusCaller {
         let pos_ref_len =
             bam_fields::reference_length_from_cigar(&longest_pos.clipped_cigar) as usize;
         let pos_end = pos_start + pos_ref_len.saturating_sub(1);
+        let neg_ref_len =
+            bam_fields::reference_length_from_cigar(&longest_neg.clipped_cigar) as usize;
+        let neg_end = neg_start + neg_ref_len.saturating_sub(1);
 
-        let (overlap_start, overlap_end) = (neg_start, pos_end);
+        // The duplex window is the *intersection* of the two alignments: the reference
+        // positions at which both strands actually have an aligned base, and so the only
+        // positions at which they can be said to agree or disagree.
+        //
+        // fgbio takes it to be `[negative.start, positive.end]`
+        // (`CodecConsensusCaller.scala:211-213`), which is the intersection only while the
+        // positive alignment opens first and closes first. CODEC breaks that: both strands
+        // cover the same insert end to end, so the aligner routinely takes as matching a
+        // terminal base that it soft-clipped on the other strand, leaving the two alignments
+        // *dovetailed* by a base or two. The window then reaches past one read or the other,
+        // and every consumer of it is answered about a position that read never aligned:
+        // `check_overlap_phase_raw` reads the out-of-range boundary as htsjdk's `0` and the
+        // family fails a check about indels with no indel anywhere in it, and
+        // `compute_consensus_length_raw` has no position to anchor on at all
+        // (fulcrumgenomics/fgumi#761). Clamping to the intersection is inert whenever the
+        // alignments do not dovetail, since `max`/`min` then select the same two boundaries.
+        //
+        // An empty intersection is not a special case: it can only arise when the two
+        // alignments do not overlap, where these are still the same two boundaries fgbio
+        // computes, and `min_duplex_length` (at least 1, enforced by the command) rejects it
+        // just below.
+        let (overlap_start, overlap_end) = (neg_start.max(pos_start), pos_end.min(neg_end));
         let duplex_length = overlap_end as i64 - overlap_start as i64 + 1;
 
         if duplex_length < self.options.min_duplex_length as i64 {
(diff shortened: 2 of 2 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed the outside-diff finding on the emitted-base statistics (`codec_caller.rs`): `build_duplex_consensus_from_padded` no longer mutates `self.stats`. It now returns the duplex-base deltas, and the caller folds `consensus_duplex_bases_emitted` / `duplex_disagreement_base_count` into the stats only after `build_output_record_into` succeeds — alongside `consensus_reads_generated` / `consensus_bases_emitted`. A molecule that computes a consensus but fails to emit (e.g. an over-long input-derived read name rejected by `try_build_record`) now advances none of the emitted-base counters. Added `test_rejected_over_long_read_name_leaves_duplex_stats_unchanged` as a regression, which pins the read-name failure path and asserts both counters stay at zero.

nh13:
@coderabbitai review

PR #768, opened 2026-08-15
fix(codec): store the CODEC fixture's reverse read in reference orientation

### LINKED ISSUE #763: codec caller test fixture stores the reverse read's SEQ in read orientation, so its consensus is almost all N

nh13:
Closes #763.

`create_fr_pair` — the CODEC caller's shared read-pair test fixture — cut each read's bases straight out of `REF_BASES`, i.e. already in reference orientation, and then reverse-complemented the minus-strand read before storing it, under a comment asserting that "Negative strand reads are stored as revcomp in BAM". That comment is wrong: a BAM stores `SEQ` in reference orientation for **every** record, because the aligner has already applied the flip. Both strands of every family the fixture built therefore pointed in opposite directions.

## Confirming the defect

Removing the two `reverse_complement` calls and running the crate's suite gives exactly the split the issue predicts. Of the 93 runnable `codec_caller` tests (`cargo nextest list -p fgumi-consensus`; the issue's "102" counts something slightly different, but the failing set is identical), **87 pass unchanged and 6 fail** — and the 6 are precisely the duplex-*disagreement* tests:

```
test_consensus_reads_typed_disagreement_rate
test_consensus_reads_typed_disagreement_count
test_high_duplex_disagreement_counted::case_1_by_count
test_high_duplex_disagreement_counted::case_2_by_rate
test_high_duplex_disagreement_tracks_rejects
test_not_emit_consensus_high_disagreement
```

They passed because the fixture disagreed everywhere, not because any of them constructed a disagreement.

## The test that would have caught it (added first)

`test_fixture_consensus_reproduces_the_reference_bases` builds a family whose two reads align to exactly the same reference span — so every consensus position is a duplex position, with no single-strand tail that could legitimately be masked — and asserts the consensus **is** those reference bases, with no no-calls.

On the old fixture it fails with 25 of 30 positions no-called:

```
left:  "NTNNNNNTNNNNNNNNNNNNNNANNNNNAN"
right: "GTAACTTTTTGATAGTAGCGGGAGTAGGAG"
```

This is the important artifact here. Until now every test on this fixture checked consensus *length*, record counts or rejection reasons, so **a consensus of the correct size assembled from the wrong bases would have passed the entire CODEC suite.**

## The fix

`get_sequence` already returns reference-oriented bases, so both strands now store them verbatim. The comment is replaced with one that states the actual BAM contract.

## Repairing the six vacuous tests

`duplex_disagreement_fixture` now takes the number of disagreements to build. The pair (R1 forward at 1, R2 reverse at 11, both `30M`) overlaps over 20 reference positions; each requested disagreement substitutes one base of R2 *inside that overlap* for a base the reference does not carry there, leaving R1 alone. Both reads keep the same base quality, so the caller sees an equal-quality disagreement at each substituted position — the same shape a real strand mismatch has.

| Test | Disagreement introduced | Why it matches the test's intent |
| --- | --- | --- |
| `test_consensus_reads_typed_disagreement_count` | 1 substitution in the overlap | The test sets `max_duplex_disagreements: 0` with a permissive rate, so the count check is the only thing that can fire. One disagreement is the minimum that exceeds zero tolerance, and the assertion now pins `DuplexDisagreementCount { disagreements: 1 }` rather than accepting any count. |
| `test_consensus_reads_typed_disagreement_rate` | 1 substitution in the overlap | The test sets `max_duplex_disagreement_rate: 0.0` with `max_duplex_disagreements: usize::MAX`, so only the rate check can fire. One disagreement over 20 duplex positions is a rate of 0.05, and the assertion now pins that value rather than accepting any rate — which is also what keeps the fixture's stated geometry honest. |
| `test_high_duplex_disagreement_counted::case_1_by_count` | 1 substitution in the overlap | The case makes the rate permissive so the count is the only trigger; one disagreement exceeds its `max_duplex_disagreements: 0`. What the test actually asserts — the `HighDuplexDisagreement` rejection reason and the dedicated HDD counter — is unchanged. |
| `test_high_duplex_disagreement_counted::case_2_by_rate` | 1 substitution in the overlap | Same family, opposite trigger: `usize::MAX` count with a `0.0` rate, so the 0.05 rate is what rejects. Sharing one fixture across both cases is what makes them a genuine A/B on the trigger. |
| `test_high_duplex_disagreement_tracks_rejects` | 1 substitution in the overlap | The test's subject is that HDD-rejected records reach the `--rejects` output byte-for-byte. It only needs a real HDD reject to exist; one disagreement against its `max_duplex_disagreements: 0` produces one. The byte-identity assertion is unchanged and still snapshots the substituted records. |
| `test_not_emit_consensus_high_disagreement` | 6 substitutions in the overlap | This one is named for "a lot of disagreements", and its thresholds (`5` / `0.05`) only mean something if the family carries more than five. Six disagreements over 20 duplex positions is a count of 6 and a rate of 0.3, so the same family is emitted under the permissive thresholds (`100` / `1.0`) and rejected under the strict ones — the rejection is now attributable to the thresholds and to nothing else. The rejecting half also pins `DuplexDisagreementCount` instead of asserting only `is_err()`. |

Each of the six was verified to be non-vacuous by rebuilding the fixture with zero disagreements: all six fail, and only those six.

## Audit: other assertions weakened by the same cause

While the fixture produced a mostly-`N` consensus, no test on it could assert content, and several settled for something weaker. One is fixed here because its doc comment stated a falsehood; the rest are listed for follow-up rather than changed, to keep this diff to the defect.

Fixed:

- **`test_make_consensus_from_simple_reads`** asserted only `bases.len() == 40`, rationalised by a doc comment reading "Positions covered by only one strand get N bases ... so we don't check exact sequence content". That is not what the caller does — `build_duplex_consensus_from_padded`'s single-strand branch keeps the covering strand's base unless its quality is `MIN_PHRED` — and with the fixture fixed the consensus is the reference over the whole 40bp span, single-strand tails included. Now asserted in full, comment corrected.

Not changed, but weaker than the contract they claim:

- `test_make_consensus_r1_deletion` and `test_make_consensus_both_soft_clipped_same_end` assert only `!bases.is_empty()`. An all-`N` consensus of any length passes both.
- `test_make_consensus_r2_deletion` and `test_make_consensus_soft_clipping` assert only the consensus length. Both are about how a CIGAR operation shifts the *bases*, which is the part not checked.
- `test_emit_consensus_in_r1_orientation` asserts a 40bp length and the `REVERSE` flag, but never that the reverse-R1 consensus is the reverse complement of the forward-R1 one — which is the claim its name makes, and which is only assertable now.
- `test_mask_single_stranded_regions` asserts `has_low_qual || quals.is_empty()`; the second disjunct is an escape hatch that makes the assertion hard to falsify, and it never checks *which* positions were masked.
- `test_rejected_reads_tracking_enabled` asserts `!rejected_reads().is_empty() || reads_filtered > 0` — the same unfalsifiable-disjunction shape. Not caused by the fixture, but the same class.

## Base branch

Based on `main`. #755, #759 and #767 all touch `codec_caller.rs` but none of them touches `create_fr_pair` or any of the six disagreement tests — their test-side hunks are further down the file (#755 around the alignment-filter tests, #767 around the overlap-window tests). #759 and #767 are stacked on #755, and #767 is currently conflicting against its own base, so `main` is both the correct and the lowest-conflict base.

## Verification

`cargo ci-fmt`, `cargo ci-lint`, `cargo ci-test` (7495 passed), `cargo ci-doctest` — all green.

Risk: consensus test-fixture output changes are pinned by exact reference-base and disagreement assertions; `unsafe` changes are none; memory bounds, queue capacity, and thread/backpressure policy changes are none.

Fix CODEC reverse-strand fixtures to store `SEQ` in reference orientation. Add regression coverage for exact consensus bases and no no-calls. Update disagreement tests to use explicit overlap mismatches and verify exact count and rate errors.

### DIFF
--- a/crates/fgumi-consensus/src/codec_caller.rs
+++ b/crates/fgumi-consensus/src/codec_caller.rs
@@ -2365,16 +2365,15 @@ mod tests {
             seq
         };
 
-        let seq1_fwd = get_sequence(start1, cigar1);
-        let seq2_fwd = get_sequence(start2, cigar2);
-
-        // For a true FR pair sequencing the same DNA molecule from opposite strands:
-        // - Positive strand reads are stored as-is in BAM (matches reference)
-        // - Negative strand reads are stored as revcomp in BAM
-        //   After revcomp during SS building, they become the same as reference
-        // This ensures that both R1 and R2 SS consensuses agree in the overlap.
-        let seq1: Vec<u8> = if strand1_reverse { reverse_complement(&seq1_fwd) } else { seq1_fwd };
-        let seq2: Vec<u8> = if strand2_reverse { reverse_complement(&seq2_fwd) } else { seq2_fwd };
+        // A BAM stores `SEQ` in *reference* orientation for every record, whatever
+        // the strand — the aligner has already reverse-complemented a read that
+        // mapped to the minus strand. `get_sequence` reads straight out of
+        // `REF_BASES`, so the bases it returns are already in that orientation and
+        // must be stored verbatim for both strands. Reverse-complementing the
+        // minus-strand read here would point the two strands of a pair in opposite
+        // directions and make their consensus almost entirely `N` (issue #763).
+        let seq1 = get_sequence(start1, cigar1);
+        let seq2 = get_sequence(start2, cigar2);
 
         let qual1 = vec![base_quality; seq1.len()];
         let qual2 = vec![base_quality; seq2.len()];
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

Issue #746, opened 2026-08-14
codec exceeds its declared queue memory budget and is OOM-killed when the output device is slow


nh13:
## Summary

Under a slow or contended output device, `fgumi codec` grows well past the queue memory budget it reports at startup and is killed by the OOM killer. With the same input and the same flags on an uncontended device, it runs in a small, flat footprint.

## Observed

Host: AWS `c8g.4xlarge` (16 vCPU, 30 GiB, arm64), Amazon Linux 2023, 300 GB gp3 root volume. Input: a CODEC-grouped BAM of ~502M records (~30.7 GB). Identical command in both runs:

```
fgumi codec --input grouped.bam --output out.bam \
    --min-reads 1 --min-input-base-quality 10 --threads 16
```

Both runs logged the same budget at startup:

```
Queue memory budget: 12.0 GiB total (768.0 MiB/thread × 16 threads)
```

| run | output volume | peak RSS | outcome |
| --- | --- | --- | --- |
| A | contended — a concurrent ~21 GB download saturating a 125 MB/s gp3 volume | **31.3 GB** | OOM-killed at ~26M of 502M records |
| B | idle, same volume raised to 1000 MB/s | **3.28 GB** | completed in 128 s |

Kernel confirmation from run A:

```
Out of memory: Killed process 39102 (fgumi) total-vm:38919296kB, anon-rss:31339624kB, file-rss:0kB, shmem-rss:0kB
```

Progress in run A degraded from ~1M records / 5 s to ~1M records / 80 s shortly before the kill, which is consistent with the reader continuing to produce while the writer stalls.

## Why this matters

The 12 GiB figure reads as a bound, but peak RSS reached ~2.5× it and ~10× the uncontended steady state. Any deployment whose output path can stall — a network filesystem, a shared or throttled volume, an object-store-backed mount, or simply another process competing for the same disk — can hit this, and the failure mode is an OOM kill rather than backpressure or a slow-but-correct run.

## What I have not established

- The contention was induced artificially; I do not have a minimal reproducer.
- This is **not** shown to be a regression. fgumi 0.4.0 was never run under the same contention, so it may be long-standing.
- Only `codec` was exercised. Other consensus commands share pipeline machinery and may behave the same way, but I did not test them.

Build under test: `main` at `0f685c38`.

PR #772, opened 2026-08-15
fix(pipeline): enforce the queue memory budget at the FASTQ Read step

### LINKED ISSUE #766: FASTQ pipeline (extract) does not bound its queue memory, same gap as #746
## Summary

The FASTQ unified pipeline (`fgumi extract`) accepts `--max-memory`, logs a queue memory budget, and does not bound the bytes its queues hold. It is the same gap #746 reports for the BAM pipeline, in the sibling implementation.

## Reproduction

`src/lib/unified_pipeline/fastq.rs` mirrors `bam.rs`'s structure: the same options surface, the same `Queue memory budget: N GiB total (M MiB/thread x T threads)` log line, and the same slot-count bounds (`clamp(threads*16, 64, 256)`) on its inter-stage queues. What it does not have is a gate on the read stage keyed to total in-flight bytes.

Consequence: with a slow or stalled output sink, resident memory is a function of the input data rather than of the declared budget, exactly as measured for the BAM pipeline in #746 (peak RSS 8,512 / 9,068 / 8,133 MB for declared budgets of 256 MiB / 12 GiB / 32 GiB).

The BAM side is addressed in #764. This issue tracks the FASTQ side, which that PR deliberately does not touch.

## Impact

`fgumi extract` on a large FASTQ pair with a constrained or contended output device can exceed its declared memory budget without any signal that the budget is not being enforced.

nh13:
Closes #766. **Based on #764** — that PR does the same job for the BAM pipeline and lays the `refund_queue_bytes` groundwork this one reuses, so this branch targets it rather than `main`. Retarget to `main` once #764 lands.

## The defect holds, and it is wider than the BAM one

`fgumi extract --threads 8` logs `Queue memory budget: 6.0 GiB total (768.0 MiB/thread × 8 threads)` and then bounds almost none of it. `--max-memory` reaches exactly two places in the FASTQ pipeline: `ReorderBufferState::new` for the write reorder buffer (itself clamped to `BACKPRESSURE_THRESHOLD_BYTES`, 512 MiB) and `DeadlockState::new`. Everything between Read and Write is bounded by a *slot count* — `queue_capacity = (threads * 128).min(1024)`, four times larger than BAM's `clamp(threads * 16, 64, 256)` — with no bound on the bytes those slots hold.

Where FASTQ differs from `bam.rs`, it differs for the worse:

- **Q0 (raw chunks), Q1 (decompressed chunks) and the Pair buffer had no byte accounting at all.** Q1 in particular holds whole decompressed payloads.
- **Q3 (templates) had a counter that was pure fiction.** `groups_heap_bytes` was mutated only under `memory-debug`, and even then the push side added a hardcoded `0` (`let q4_heap: u64 = 0;`) while the pop side subtracted a real figure. With the feature off both were no-ops, so the counter read zero however much the queue held — and Q3 holds fully parsed `FastqTemplate`s, the single largest consumer when the writer stalls.
- **The accounting that did exist was not derived from the budget.** `Q2_BLOCK_PARSED_BACKPRESSURE_BYTES` (128 MiB), `PENDING_BACKPRESSURE_BYTES` (256 MiB) and `Q5_BACKPRESSURE_THRESHOLD_BYTES` (256 MiB) are hardcoded per-stage thresholds; none scales with `--max-memory`, and together they bound only three of ten stages.
- **Two stages park data that has already left a queue.** `BlockMergeState`'s pending maps (BGZF path) and `PairState`'s pending map (gzip path) hold chunks that Q1/Q2 have already refunded, so a naive port would lose sight of them entirely.

## Reproduction

3,000,000 synthetic 150 bp read pairs (8 bp UMI in R1), gzip, through `fgumi extract --threads 8 --output -` into a 2 MB/s sink, varying only the declared budget. The deadlock detector is disabled on both arms so both complete and the peaks are comparable — see the next section for what happens with it left on.

| declared budget | logged as | before | after |
| --- | --- | --- | --- |
| `--max-memory 256M --memory-per-thread false` | `244.1 MiB total` | 2,031 MB | **273 MB** |
| default (`768` per thread) | `6.0 GiB total` | 2,038 MB | 2,045 MB |
| `--max-memory 32G --memory-per-thread false` | `29.8 GiB total` | 2,039 MB | 2,059 MB |

Before, peak RSS is flat within 0.4% across a 128× range of declared budgets: it is a function of the data and the slot counts, not of the configuration. After, the tight budget is honored and the generous ones are untouched, because the gate never fires when the budget is far above what the queues hold.

Every run above exited 0 and wrote all 6,000,000 records (`samtools view -c`), and the three `after` runs produced byte-identical output sizes to each other.

**A side effect worth knowing:** with the deadlock detector at its default 10 s, the pre-fix binary does not merely use too much memory on this workload — it **aborts**. It swallows the entire input in ~12 s, then makes no queue progress at 2 MB/s, and the detector fires: `DEADLOCK DETECTED: No progress for 10s`, exit 1, 1,208,000 of 6,000,000 records written. The post-fix binary completes the same cell cleanly with the detector at its default, because reading and draining stay interleaved.

| `--max-memory 256M --memory-per-thread false`, detector at default | exit | records | peak RSS |
| --- | --- | --- | --- |
| before | **1** (deadlock abort) | 1,208,000 / 6,000,000 | 1,680 MB |
| after | 0 | 6,000,000 / 6,000,000 | 293 MB |

## The default path is unchanged

Same input, fast sink, default budget, three reps each:

| | before | after |
| --- | --- | --- |
| wall | 1.34 / 1.29 / 1.31 s | 1.29 / 1.28 / 1.28 s |
| peak RSS | 364 / 232 / 198 MB | 165 / 164 / 186 MB |

Wall time is unchanged. Peak RSS is noisy on an unbounded pipeline (198–364 MB across three identical runs) and settles tighter afterwards; there is no regression in either.

## The fix

Read is the pipeline's only source of new bytes and nothing downstream waits on it, so gating Read is both the one place a total can be enforced and a place where enforcing it cannot deadlock.

- Charge the unaccounted queues — Q0, Q1, Q3, the Pair buffer, and the merge step's pending maps. The merge maps get a lock-free mirror of `BlockMergeState::pending_heap_bytes` so the gate can read them without taking the merge lock; every mutation goes through one `charge_block_merge_pending` / `refund_block_merge_pending` pair, so the two cannot drift.
- Sum all eleven counters in `FastqPipelineState::queue_bytes_in_flight` and gate `fastq_try_step_read` on it, after the held-item advance so a held chunk can always still be placed.
- Refunds go through `refund_queue_bytes`, promoted from `bam.rs` to `base.rs` now that both pipelines use it.
- The deadlock snapshot now reports `queue_in_flight=NNNMB`. A run the gate has stopped looks idle in the queue lengths alone, which is exactly the wrong impression to give whoever reads that dump.

### Liveness: two exemptions, one of which is load-bearing

The Pair and `BlockMerge` steps release a batch only once every stream has delivered the matching index, or once every stream has reached EOF. Only Read can satisfy either condition, so `read_is_required_for_liveness` exempts two cases from the gate:

- **A stream that has fallen behind another.** What the assembly step is holding is waiting on exactly this stream's next batch. Exempting only the laggard bounds the extra read-ahead at the skew that already existed.
- **A stream whose sibling has already reached EOF.** Everything past the shorter stream's end is unreleasable until `read_done`, which needs *this* stream at EOF too. For equal-length inputs that covers the last batch or two.

The second is not defensive decoration. A mismatched-length FASTQ pair accumulates unreleasable surplus in the Pair buffer, that surplus counts against the budget, and the budget is then what prevents it ever being drained. `mismatched_stream_lengths_still_complete_under_a_tight_budget` hangs for the full 120 s timeout with that exemption removed, and passes with it. For malformed input this deliberately trades the bound for liveness — a run that cannot finish is worse than one that exceeds its budget.

### Charge before push, not after

Every charge now lands before its item becomes visible to a consumer (`push_charged`). The previous idiom throughout this file was

```rust
let result = queue.push(item);
if result.is_ok() { counter.fetch_add(heap_size, ...); }
```

which leaves a window in which a consumer pops the item and refunds bytes that were never charged. `refund_queue_bytes` saturates at zero — which is what keeps an unpaired refund from wrapping to `u64::MAX` — so when the counter is near zero, the normal state of a queue draining as fast as it fills, that refund is simply lost and the charge landing afterwards is never released. Each occurrence permanently inflates a counter that gates Read; enough of them and the pipeline stops reading with its queues empty, which cannot resolve. Charging first inverts the failure into a brief over-count during a rejected push.

This was latent before #746/#766 made these counters load-bearing, which is why it has never been observed. `a_failed_push_leaves_no_charge_behind` pins the contract.

## Tests

`tests/integration/test_fastq_pipeline_memory_backpressure.rs`, the sibling of #764's file, drives `run_fastq_pipeline` over a paired-end input with a `CountingReader` per stream and a writer that blocks until released:

- `stalled_writer_stops_the_fastq_reader_within_the_queue_memory_budget` — read-ahead must stay under half the input. Before: the reader consumed all 18,997,780 bytes.
- `fastq_read_ahead_under_a_stalled_writer_shrinks_with_the_queue_memory_budget` — the discriminating one. Input and slot counts fixed, only `--max-memory` varies; before, both budgets produced the identical 18,997,780 bytes.
- `fastq_reader_resumes_and_completes_after_the_writer_recovers` — every record reaches the output, in order, once the writer recovers.
- `mismatched_stream_lengths_still_complete_under_a_tight_budget` — liveness, with the same input run at `queue_memory_limit = 0` as the oracle so the budget is shown to change neither the outcome nor the ability to reach it.

Every one of the four checks the emitted read names against the input's, in order, rather than an exit status or a record count: throttling that dropped, duplicated or reordered records would leave both of those intact and change the sequence.

Unit tests in `fastq.rs` pin the gate's algebra (no-limit, at-budget, drained-below-budget, first-batch-always-admitted, both liveness exemptions), that every counter reaches the sum, and that each charge is refunded exactly on pop.

The integration tests exercise the gzip/plain path, which is where injectable readers are possible and which is what `.fastq.gz` inputs actually take. The BGZF-specific counters (Q2, merge pending) are covered by unit tests.

## Three findings on #764, two of them fixed here

Both of #764's failing CI jobs are caused by its own diff, and this branch inherits them, so both are fixed here. They should be fixed there too so that PR can go green on its own.

- **`docs` fails.** `q2b_boundaries`' new doc links `BamPipelineState::q2b_push` / `::q2b_pop` as intra-doc links; both are private, and `cargo ci-doc` runs under `RUSTDOCFLAGS=-D warnings`. Fixed by demoting the two to plain backticks.
- **`test` fails on a timeout, not an assertion.** Its three backpressure tests build a 600,000-record BAM in a debug build and run 29-42 s on an M-series laptop; on a GitHub runner they exceed the repo's standing `slow-timeout = 60s, terminate-after = 2` and are killed at 120 s, taking 3,652 unrun tests down with them. Scaled the fixture to 60,000 families against a 512 KiB budget — 15-21 s locally, and still a 7.6x input-to-budget ratio. Verified the retuned tests still discriminate: with `read_admission_allowed` removed from `try_step_read`, the reader again consumes all 3,996,131 input bytes and both budgets produce identical read-ahead.
- **`q2b_push` has the charge-after-push shape described above**, as do the two `q1_heap_bytes.fetch_add` sites in `try_step_read`. Left alone here — that is BAM's behavior and that PR's code — but it is the same latent permanent-over-count and deserves the same reorder.

## Known, out of scope, worth their own issues

- **`OutputPipelineQueues::process_output_push` / `serialize_output_push` in `base.rs` still charge after push**, and their pops use plain `fetch_sub`. They are unreachable from the FASTQ pipeline (it pushes those queues directly, and those two helpers are only called from `shared_try_step_process` / `shared_try_step_serialize`, which nothing calls), so they do not affect this gate — but they are the same shape and should be fixed or deleted.
- **`fgumi extract` drops templates from the surplus tail of a mismatched-length FASTQ pair** — 29,800 of 30,000 in the new test's fixture. Confirmed pre-existing: identical loss with the gate disabled entirely. The new test uses the ungated run as its oracle precisely so it pins the budget's neutrality without baking that defect in.
- **`ReorderBufferState::effective_limit`'s 512 MiB clamp is untouched**, as in #764.

Risk: output change none; `unsafe` change none and CLAUDE.md allowlist update none; memory bounds and read-stage backpressure change yes.

FASTQ extraction now enforces `--max-memory` across all in-flight queues and pending buffers. Charges occur before enqueue, and failed pushes or pops use saturating refunds.

The change adds Q2b accounting, aggregate memory diagnostics, liveness exemptions, and deadlock snapshot reporting. BAM queue sites now use the shared accounting helper.

FASTQ pairing and BGZF merge now reject mismatched input lengths and flush EOF residuals. Integration tests cover stalled writers, memory budgets, recovery, plain/gzip/BGZF input, accounting, and output ordering.

### DIFF
--- a/src/lib/commands/common.rs
+++ b/src/lib/commands/common.rs
@@ -1228,15 +1228,18 @@ fn resolve_memory_budget_with_total(
 /// a hard RSS cap — a single pathological position group is still processed
 /// whole, and each worker has transient working-set memory on top of the queue.
 ///
-/// In the **BAM** pipeline the budget is enforced at the Read step: the
-/// pipeline stops admitting input once the bytes queued between its stages
-/// reach it, so a slow or contended output device backs pressure up to the
-/// reader instead of filling every queue to its slot count. See
-/// `BamPipelineState::read_admission_allowed`.
+/// Both pipelines enforce the budget at the Read step: they stop admitting
+/// input once the bytes queued between their stages reach it, so a slow or
+/// contended output device backs pressure up to the reader instead of filling
+/// every queue to its slot count. See `BamPipelineState::read_admission_allowed`
+/// and `FastqPipelineState::read_admission_allowed`.
 ///
-/// The FASTQ pipeline (`fgumi extract`) shares these options and logs the same
-/// budget, but has no equivalent admission gate yet: there it still bounds only
-/// the reorder buffers, each capped at
+/// Memory held *outside* those aggregates is not covered: a worker's
+/// in-progress batch is bounded by the thread count rather than by this budget.
+/// The write reorder buffers, by contrast, *are* summed into the aggregates
+/// (the FASTQ aggregate counts its write reorder state and the BAM aggregate its
+/// input and write reorder states), so this budget bounds them through the Read
+/// gate; they additionally apply their own threshold, capped at
 /// [`BACKPRESSURE_THRESHOLD_BYTES`](crate::unified_pipeline::BACKPRESSURE_THRESHOLD_BYTES).
 #[derive(Debug, Clone, Args)]
 pub struct QueueMemoryOptions {
--- a/src/lib/unified_pipeline/bam.rs
+++ b/src/lib/unified_pipeline/bam.rs
@@ -33,7 +33,7 @@ use super::base::{
     QueueSample, RawBlockBatch, ReorderBufferState, SerializePipelineState, SerializedBatch,
     StepContext, WorkerCoreState, WorkerStateCommon, WritePipelineState,
     finalize_pipeline_with_buffers, generic_worker_loop, handle_worker_panic, join_monitor_thread,
-    join_worker_threads, shared_try_step_compress,
+    join_worker_threads, push_charged, refund_queue_bytes, shared_try_step_compress,
 };
 use super::deadlock::{
     DeadlockAction, DeadlockConfig, DeadlockState, QueueSnapshot, check_deadlock_and_restore,
@@ -74,19 +74,6 @@ impl MemoryEstimate for BoundaryBatch {
     fn estimate_heap_size(&self) -> usize {
         self.buffer.capacity() + self.offsets.capacity() * std::mem::size_of::<usize>()
     }
-}
-
-/// Release `bytes` from a queue byte counter, saturating at zero.
-///
-/// The counters summed by [`BamPipelineState::queue_bytes_in_flight`] gate the
-/// Read step, so an unpaired refund would not merely skew a statistic: a plain
-/// `fetch_sub` past zero wraps to `u64::MAX` and stops the pipeline reading for
-/// the rest of the run. Saturating turns that class of bug into a bounded
-/// accounting error rather than a hang.
-fn refund_queue_bytes(counter: &AtomicU64, bytes: u64) {
-    let _ = counter.fetch_update(Ordering::AcqRel, Ordering::Acquire, |current| {
-        Some(current.saturating_sub(bytes))
-    });
 }
 
 /// State for the `FindBoundaries` step (sequential).
@@ -942,21 +929,14 @@ impl<G: Send, P: Send + MemoryEstimate> BamPipelineState<G, P> {
     /// Returns the batch unchanged when Q2b is out of slots, so callers keep
     /// the existing held-item retry behaviour.
     fn q2b_push(&self, serial: u64, batch: BoundaryBatch) -> Result<(), (u64, BoundaryBatch)> {
-        // Charge before the push so the batch is never visible to a consumer
-        // while its bytes are still uncharged: a pop in that window would refund
-        // bytes never added and, because `q2b_heap_bytes` gates Read via
-        // `queue_bytes_in_flight`, permanently over-state the counter and close
-        // the Read gate. Refund on the out-of-slots path so a rejected push
-        // leaves the counter unchanged.
+        // `push_charged` charges before the push so the batch is never visible to
+        // a consumer while its bytes are still uncharged: a pop in that window
+        // would refund bytes never added and, because `q2b_heap_bytes` gates Read
+        // via `queue_bytes_in_flight`, permanently over-state the counter and
+        // close the Read gate. It refunds on the out-of-slots path so a rejected
+        // push leaves the counter unchanged.
         let heap_size = batch.estimate_heap_size() as u64;
-        self.q2b_heap_bytes.fetch_add(heap_size, Ordering::AcqRel);
-        match self.q2b_boundaries.push((serial, batch)) {
-            Ok(()) => Ok(()),
-            Err(returned) => {
-                refund_queue_bytes(&self.q2b_heap_bytes, heap_size);
-                Err(returned)
-            }
-        }
+        push_charged(&self.q2b_boundaries, &self.q2b_heap_bytes, heap_size, (serial, batch))
     }
 
     /// Pop a boundary batch from Q2b, refunding its heap bytes.
@@ -974,14 +954,7 @@ impl<G: Send, P: Send + MemoryEstimate> BamPipelineState<G, P> {
     /// counter unchanged. Callers record the deadlock-detector push on `Ok`.
     fn q1_push(&self, serial: u64, batch: RawBlockBatch) -> Result<(), (u64, RawBlockBatch)> {
         let heap_size = batch.estimate_heap_size() as u64;
-        self.q1_heap_bytes.fetch_add(heap_size, Ordering::AcqRel);
-        match self.q1_raw_blocks.push((serial, batch)) {
-            Ok(()) => Ok(()),
-            Err(returned) => {
-                refund_queue_bytes(&self.q1_heap_bytes, heap_size);
-                Err(returned)
-            }
-        }
+        push_charged(&self.q1_raw_blocks, &self.q1_heap_bytes, heap_size, (serial, batch))
     }
 
     /// Pop a raw block batch from Q1, refunding its heap bytes.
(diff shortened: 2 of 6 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
Outside-diff finding (src/lib/unified_pipeline/bam.rs `q2b_push` / `q1_push`): addressed. Moved `push_charged` into `base.rs` as `pub(crate)` next to `refund_queue_bytes`, and both BAM push sites now delegate to it instead of open-coding charge-then-refund, so the charge-before-publish ordering lives in one place. `fastq.rs` now imports the shared helper rather than defining its own.

nh13:
@coderabbitai review

nh13:
Addressed the outside-diff-range comment (reuse the shared `push_charged` helper in `q2b_push`/`q1_push`): the current code already routes both through `push_charged`, so the charge-before-publish ordering lives in one place. Also fixed a related unpaired-refund that surfaced after rebasing onto main's #776 — `flush_residual_suffixes_at_eof` was pushing to the groups queue via a direct uncharged `output.groups.push`, which left `q3_pop`'s refund unpaired; it now charges through `q3_push` like every other producer.

nh13:
@coderabbitai review

nh13:
Addressed the latest review round:

1. **Issue reference `#776` → `#773`** (test docstring) — fixed; the out-of-sync rejection is `#773` everywhere else (`#766` is the memory-budget work).

2. **Outside-diff: add EOF residual-suffix regression tests** — added three unit tests directly invoking `flush_residual_suffixes_at_eof` in `fastq.rs`: single-stream emission of a residual with no trailing newline (which also guards the `q3_push` charge — an uncharged push would trip `q3_pop`'s refund assertion), a paired one-sided residual left in the surplus for out-of-sync rejection, and the `q3_push`-full path that stores `worker.held_parsed`.

Separately, the `coverage` job's intermittent `attempt to add with overflow` in `queue_bytes_in_flight` is a pre-existing reorder-buffer `sub_heap_bytes` wrap (documented at base.rs:733), unrelated to this PR and unreproducible outside llvm-cov's scheduling jitter — filed as #810 with a fix to follow in a separate PR.

nh13:
@coderabbitai review

PR #774, opened 2026-08-15
test(codec): make the CODEC tests that cannot fail fail

### LINKED ISSUE #771: CODEC tests with assertions that cannot fail, including one that never reaches its subject
## Summary

A number of CODEC consensus tests contain assertions that cannot fail, or that pass for a reason unrelated to what the test is named for. Two have been confirmed by deliberately removing the behavior under test and observing that the test still passes.

## Confirmed by perturbation

**`test_not_emit_consensus_chimeric_pair`** passes with the chimeric mutation deleted. Its fixture is built with starts `100` and `135`, both `30M`, so R1 spans 100-129 and R2 spans 135-164 — the two reads do not overlap at all. The pair is rejected for insufficient overlap before `is_primary_fr_pair_raw`'s cross-chromosome check is consulted. Removing both `set_ref_id` calls leaves the test green, so the cross-chromosome path it exists to cover is untested.

**`test_check_overlap_phase_indel_mismatch`** discards its result (`let _result = ...`) under a comment stating the outcome is indeterminate. It is not: the fixture yields `29 != 24`, deterministically `false`. The sibling `test_check_overlap_phase_deletion_mismatch` already asserts exactly that, so the assertion is available and the comment is wrong.

## Also observed, not individually confirmed

- `codec_downsampling_retains_lowest_hashing_pairs` — its R2 claim is guaranteed by the `#[cfg(test)]` wrapper and states the opposite of production behavior.
- `test_downsample_pairs_no_limit` — returns before any production code runs.
- `test_clip_overlap_failed_counted_and_labeled` — its negative guard cannot fail.
- `test_clear_rejected_reads` — clears an already-empty buffer.
- `test_codec_statistics_tracking` — asserts `Default`.
- `test_mask_end_qualities` — asserts `<= 5` where the code assigns exactly 5, and never checks the interior.
- `test_build_clipped_info_clip_from_start_reverse` — asserts `> 100` where the value is exactly 103.
- `test_reverse_complement_ss_depths_errors_reversed` — uses the palindrome `ACGT`, so it cannot observe a reversal. That function has no non-palindromic test anywhere in the crate.

## Context

These surfaced while strengthening a different set of weak assertions (#769, PR #770) and fixing the shared fixture defect (#763, PR #768). The pattern is the same one that let #763 go unnoticed: an assertion that observes a property the code cannot violate, or that is satisfied before the code under test runs.

## Suggested approach

For each, establish what the test is named for, assert that, and prove the assertion can fail by perturbing the behavior it covers. Where a fixture cannot express the claim — as with the chimeric pair's non-overlapping reads — the fixture needs changing, not just the assertion.

Build under test: `main` at `3d83a55d`, with #768 and #770 applied.

nh13:
Closes #771.

Ten CODEC tests asserted something the code cannot violate, or were satisfied before the code under test ran. Each now asserts what its name claims, and **every new assertion was verified by removing the behavior it covers and observing the failure**. The perturbation table is below; nothing here rests on inspection alone.

## Base branch

Based on #770 (`769/nhomer/assert-codec-consensus-bases`), which is itself based on #768. Several of these tests need a family that actually consenses — `test_not_emit_consensus_chimeric_pair`'s control assertion, `codec_uncapped_family_contributes_every_read`'s depth tags — which is only true once #768 stores the fixture's reverse read in reference orientation. This cannot stand alone.

## The two confirmed defects

**`test_not_emit_consensus_chimeric_pair` needed a fixture, not an assertion.** It built its pair at starts `100`/`135` with `30M` each, so R1 spanned 100-129 and R2 spanned 135-164 and the two never overlapped. `duplex_length` came out negative and phase 4 rejected the family as `InsufficientOverlap` — long after phase 2 had already dropped it for the reason the test is named for. Deleting both `set_ref_id` calls left it green.

The pair now overlaps (R1 at 1, R2 at 11, both `30M`), a control asserts that the same family *does* consense while both reads are on one chromosome, and the rejection reason is asserted rather than only the absence of output.

Worth knowing for anyone perturbing this: **the cross-chromosome rejection is doubly guarded**, and removing either guard alone is undetectable. `is_primary_fr_pair_raw` compares the two records' `ref_id`s, and the `is_fr_pair_raw` it then calls on the reverse record compares that record's `ref_id` against its `mate_ref_id`. Removing only the first leaves the test passing; removing both makes the family emit a consensus and the test fail.

**`test_check_overlap_phase_indel_mismatch` discarded its result** under `// Result depends on whether the boundaries land in indels - just verify it runs`. It does not depend: over the overlap 100-129, R1's `30M` gives query offsets 1 and 30 while R2's `15M5D15M` gives 1 and 25, so the phases are `0` and `5` and the answer is deterministically `false`. It is asserted, the comment is replaced with that arithmetic, and a `30M` control pins that the `false` is attributable to the deletion rather than to a lookup that returned `None` — which also yields `false`.

## Per test

| Test | What it should prove | What it now asserts |
| --- | --- | --- |
| `test_not_emit_consensus_chimeric_pair` | A cross-chromosomal pair is rejected *by the cross-chromosome check* | Overlapping fixture; a control that consenses; `NotPrimaryFrPair: 2` and no other reason |
| `test_check_overlap_phase_indel_mismatch` | A one-sided indel puts the strands out of phase | `false`, plus a `30M` control that is `true` |
| `test_clip_overlap_failed_counted_and_labeled` | The reject is counted and labeled `ClipOverlapFailed`, not the indel reason | Drives the production `consensus_length < ss.bases.len()` branch end to end |
| `codec_downsampling_retains_lowest_hashing_pairs` | The cap keeps the lowest-*hashing* templates, not the first N | Both strands capped independently through `keep_indices_for_infos`; R2's list is reversed, so the same three names come back at different positions |
| `codec_uncapped_family_contributes_every_read` (was `test_downsample_pairs_no_limit`) | With no cap, every read contributes | `aD`/`bD` of 10 for a ten-template family, against a capped control of 3 |
| `test_clear_rejected_reads` | It empties a populated buffer, and touches nothing else | Buffer filled by a rejected family first; buffer emptied; statistics survive |
| `test_codec_statistics_tracking` | The counters report what the caller did | One consensing family and one rejected one; all three counters driven off zero separately |
| `test_mask_end_qualities` | `--outer-bases-qual` masks the **ends** and only the ends | Ends exactly at the masked quality; every interior position above it |
| `test_build_clipped_info_clip_from_start_reverse` | A three-base start clip advances the position by three | `adjusted_pos == 103` |
| `test_reverse_complement_ss_depths_errors_reversed` | Bases are reversed **and** complemented | Non-palindromic `AACG` → `CGTT` |

Two of these needed more than a stronger assertion.

**`test_clip_overlap_failed_counted_and_labeled` was not reaching its subject at all.** Its only interaction with the caller was a direct `reject_records_count(2, ClipOverlapFailed)`, so its stated regression guard — that the reject had not leaked into `IndelErrorBetweenStrands` — could not fire: nothing in the test could insert the other reason.

Its own doc comment claimed the production branch was unreachable from a synthetic fixture. **That turns out to be wrong**, and the mechanism is worth recording. The branch fires when the consensus is shorter than a single strand, and the overlap clipper measures the mate overhang in *soft*-clip unclipped coordinates. A **query-only** operation is therefore invisible to it: R2 at `20I30M` carries twenty query bases the clipper does not see and does not clip, the consensus length comes out at 40 in reference space, and R2's own strand consensus is 50. Putting the insertion ahead of the overlap keeps `check_overlap_phase_raw` satisfied, so the earlier `IndelErrorBetweenStrands` gate does not fire first. The CIGAR is degenerate rather than aligner-shaped, and deliberately so — a soft-clip-built fixture is normalised by the clipper to exactly `consensus_length == ss.bases.len()`, one short of the branch. That degeneracy is what the reject exists for.

**`codec_downsampling_retains_lowest_hashing_pairs`' R2 claim was a property of the test harness.** The `#[cfg(test)]` `downsample_pairs` wrapper derived one index vector from R1 and mapped it over R2, so "R2 must keep the same templates as R1" held by construction. Both strands now go through `keep_indices_for_infos` — the function `cap_infos_to_lowest_ranking`, and therefore `consensus_reads_raw`, calls — each capped from its own list. R2's list is built in the reverse of R1's, so the R2 arm now carries the "lowest-hashing, not first-N" half of the claim: an order-dependent rule would keep `{q0, q1, q2}` from R1 and `{q9, q8, q7}` from R2, and the two arms would disagree.

One qualification on the issue's wording there. The old assertion did not *state the opposite* of production for this fixture — production caps each strand on its own ranks, and mates share a read name and therefore a rank, so two independent caps over identically-named lists do land on the same templates. What was wrong was presenting that as the rule: the rule is per-strand independence, and `codec_caps_each_strand_from_its_own_filtered_set` is the test where the two strands deliberately retain *different* templates. The old assertion was unfalsifiable and its comment was misleading; it was not asserting a falsehood.

## Deletions

`downsample_pairs` (the `#[cfg(test)]` wrapper) and `test_downsample_pairs` are removed. Once the two tests above go through production, the only remaining caller of the wrapper was `test_downsample_pairs`, whose assertion (`ds_r1s.len() == 2`) is strictly weaker than `codec_downsampling_retains_lowest_hashing_pairs`' on the same rule — and the cap's boundary behaviour is separately pinned by `test_select_lowest_ranking`'s `at_cap_keeps_everything` / `below_cap_keeps_everything` cases. No coverage is lost.

This is the only non-test edit, along with the two doc comments that referred to the wrapper.

## Non-vacuity

Every strengthened assertion was verified against a perturbation of the behavior it pins. Each perturbation was also run against the pre-change tree, so the last column is measured rather than asserted.

| Perturbation | Fails now | Old test caught it |
| --- | --- | --- |
| Delete the chimeric mutation from the fixture | `not_emit_consensus_chimeric_pair` | no |
| Remove **both** same-reference checks (`is_primary_fr_pair_raw` **and** `is_fr_pair_raw`) | `not_emit_consensus_chimeric_pair` | — (new coverage) |
| Remove only `is_primary_fr_pair_raw`'s same-reference check | nothing — the second guard still holds | — |
| Overlap-phase check hardcoded to `true` | `check_overlap_phase_indel_mismatch` | no (result discarded) |
| Overlap-phase check hardcoded to `false` | `check_overlap_phase_indel_mismatch` (control) | no (result discarded) |
| `ClipOverlapFailed` site relabeled `IndelErrorBetweenStrands` — the pre-fix behavior | `clip_overlap_failed_counted_and_labeled` | no |
| An uncapped run silently caps at 3 | `codec_uncapped_family_contributes_every_read` | no |
| Name ranking replaced by input order | `codec_downsampling_retains_lowest_hashing_pairs` | yes |
| `clear_rejected_reads` body emptied | `clear_rejected_reads` | no |
| `clear_rejected_reads` delegates to `clear()` | `clear_rejected_reads` | no |
| `total_input_reads` never accumulated | `codec_statistics_tracking` | no |
| `consensus_reads_generated` never incremented | `codec_statistics_tracking` | no |
| `reads_filtered` never accumulated | `codec_statistics_tracking` | no |
| Outer-bases mask applied to every position | `mask_end_qualities` | no |
| Outer-bases mask assigns 4 instead of 5 | `mask_end_qualities` | no |
| Start-clip advances the position by one too many | `build_clipped_info_clip_from_start_reverse` | no |
| `reverse_complement_ss` reverses without complementing | `reverse_complement_ss_depths_errors_reversed` | yes |
| `reverse_complement_ss` complements without reversing | `reverse_complement_ss_depths_errors_reversed` | yes |
| `reverse_complement_ss` leaves bases untouched | `reverse_complement_ss_depths_errors_reversed` | no |
| Zero clip substitutes a different one-op CIGAR | `build_clipped_info_zero_clip_preserves_all` | no |
| Shorter pad target repads instead of returning unchanged | `pad_consensus_shorter_target` | no |

Sixteen of these were undetected by the test that owns them.

Two corrections to the issue while doing this. **`ACGT` does separate reverse-only from complement-only** — both give `TGCA`, which the old assertion rejected. What it could not catch is the function being a **no-op**, and that is the case that was untested. Relatedly, "no non-palindromic test anywhere in the crate" was true of *direct* tests only: the crate-wide blast radius of a no-op `reverse_complement_ss` is 13 tests, 12 of which come from #768's and #770's end-to-end base assertions. Every other unconfirmed claim in the issue held exactly as written.

## Sibling sweep

Two more of the same shape, adjacent to tests already being fixed, folded in here rather than filed:

- `test_build_clipped_info_zero_clip_preserves_all` asserted the CIGAR had **one op**, not which op. It now compares against `4M`.
- `test_pad_consensus_shorter_target` asserted a length where the claim in its own comment is "returned unchanged". It now compares bases, quals, depths and errors.

## Still weaker than the contract, not changed here

Listed rather than fixed, to keep the diff on the issue's set:

- `test_to_source_read_*_over_clip` (four assertions across two tests) check `bases.is_empty()` / `quals.is_empty()` after over-clipping. That is the right claim, but it is also what an unconditional `Vec::new()` returns; neither test pins that a *partial* clip keeps the right remainder.
- `test_codec_caller_creation` and `test_vanilla_to_single_strand_negative_strand` echo their own constructor arguments.
- `test_consensus_reads_typed_disagreement_count`'s `assert!(err.is_duplex_disagreement())` is implied by the `matches!` on the line above it.

## Verification

`cargo ci-fmt`, `cargo ci-lint`, `cargo ci-test` (7494 passed), `cargo ci-doctest` — all green.

## Local review

CodeRabbit's automatic reviews are paused on this repo, so a local sweep was run against `.coderabbit.yaml`'s test instruction — "flag assertions weaker than the contract". It caught this PR committing the same defect it fixes: `test_reverse_complement_ss_depths_errors_reversed` had gained two `assert_ne!`s (`!= "GCAA"`, `!= "TTGC"`) that are implied by the `assert_eq!(rc.bases, b"CGTT")` above them and cannot fail. They are gone; the reasoning they encoded lives in the doc comment. The sweep also flagged that `test_downsample_pairs_no_limit` no longer named anything that exists once the wrapper was removed — hence the rename — and produced the two sibling fixes above, which are the "check the file for its siblings and list them all" clause applied to `build_clipped_info` and `pad_consensus`.

Risk: command output changes: none; `unsafe` changes: none, and no `CLAUDE.md` allowlist update applies; memory bounds, queue capacities, and thread/backpressure policies: none.

- Strengthen CODEC tests so they execute production paths and assert the intended behavior.
- Cover chimeric-pair rejection, overlap-phase indels, clip-overlap rejection labels, downsampling, rejected-read clearing, statistics, quality masking, clipping, and reverse complementation.
- Correct and expand fixtures for reference-oriented bases, CIGAR variants, saturation behavior, and fgbio oracle results.
- Remove the obsolete test-only `downsample_pairs` wrapper and its test.
- Validate assertions through behavior perturbations.
- Update related documentation.

### DIFF
--- a/crates/fgumi-consensus/src/codec_caller.rs
+++ b/crates/fgumi-consensus/src/codec_caller.rs
@@ -1049,9 +1049,9 @@ impl CodecConsensusCaller {
     ///
     /// The one implementation of codec's cap rule: rank each read name with
     /// [`fgumi_raw_bam::hash::fgbio_read_name_rank`], then hand the ranks to
-    /// [`select_lowest_ranking`]. Production (`consensus_reads_raw`) and the `downsample_pairs`
-    /// test wrapper both go through here, so a test cannot pass against a ranking that
-    /// production does not use.
+    /// [`select_lowest_ranking`]. Every caller — `consensus_reads_raw` via
+    /// [`Self::keep_indices_for_infos`], and the cap tests via the same function — goes through
+    /// here, so a test cannot pass against a ranking that production does not use.
     ///
     /// Ranking here rather than stamping every `ClippedRecordInfo` at construction keeps the
     /// hash inside the caller's own `max_reads_per_strand` guard: an uncapped run — the default
@@ -1978,29 +1978,6 @@ impl CodecConsensusCaller {
         let info_pos = Self::build_clipped_info(pos_rec, 0, 0);
         let info_neg = Self::build_clipped_info(neg_rec, 1, 0);
         Self::compute_consensus_length_raw(&info_pos, &info_neg, overlap_end)
-    }
-
-    /// Test-only wrapper: downsample pairs using the `max_reads_per_strand` option.
-    ///
-    /// Delegates to [`Self::keep_indices_by_name_rank`], the same function `consensus_reads_raw`
-    /// uses, so a change to the ranking or selection rule cannot pass here while breaking
-    /// production. It differs from the production path only in operating on `RawRecord`s
-    /// directly rather than on already-clipped `ClippedRecordInfo`s.
-    pub fn downsample_pairs(
-        &self,
-        r1s: Vec<RawRecord>,
-        r2s: Vec<RawRecord>,
-    ) -> (Vec<RawRecord>, Vec<RawRecord>) {
-        let Some(max_reads) = self.options.max_reads_per_strand else {
-            return (r1s, r2s);
-        };
-        let indices = Self::keep_indices_by_name_rank(
-            r1s.iter().map(|r| RawRecordView::new(r.as_ref()).read_name()),
-            max_reads,
-        );
-        let new_r1: Vec<_> = indices.iter().map(|&i| r1s[i].clone()).collect();
-        let new_r2: Vec<_> = indices.iter().map(|&i| r2s[i].clone()).collect();
-        (new_r1, new_r2)
     }
 }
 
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #775, opened 2026-08-15
fix(pipeline): separate the per-stage high-water mark from --max-memory

### LINKED ISSUE #765: --max-memory above 512 MiB is silently clamped and has no effect
## Summary

`--max-memory` values above 512 MiB have no effect on the unified pipeline's reorder buffers. `--max-memory 32G` and `--max-memory 512M` configure the same limit.

## Reproduction

`ReorderBufferState::effective_limit` (`src/lib/unified_pipeline/base.rs:609-611`, and again at `:628-630`) returns:

```rust
self.limit_bytes.min(BACKPRESSURE_THRESHOLD_BYTES)
```

`BACKPRESSURE_THRESHOLD_BYTES` is 512 MiB. `--max-memory` reaches the pipeline only through `ReorderBufferState::new()` for the Q2, Q3 and write reorder buffers, so every declared budget above 512 MiB is clamped away before it is used. Q5 uses a separate hardcoded 256 MiB threshold that is not derived from the budget at all.

Observable effect: raising `--max-memory` above 512M does not change peak RSS, and the run log's `Queue memory budget: N GiB total` line reports a number the pipeline does not enforce.

## Impact

The knob silently under-delivers in the one direction users reach for it — a user who raises `--max-memory` to let the pipeline buffer more gets no change, with no warning.

Note that lifting the clamp *raises* memory use and so needs its own before/after measurement rather than being removed on sight. Filed separately from #746 for that reason: #746's gate binds the total when the pipeline backs up, which is the opposite direction.

nh13:
Closes #765. Stacked on #764 — this branch is based on `746/nhomer/bound-consensus-memory-under-writer-stall`, so only the last commit is new.

## What the clamp turned out to be

`ReorderBufferState::effective_limit` returns `min(budget, BACKPRESSURE_THRESHOLD_BYTES)`, and that constant is 512 MiB. The issue reads that as a plain bug. It is not — it is two ideas sharing one number.

A stage's **high-water mark** is a *trigger*: the depth at which one stage stops pulling in work so the rest of the pipeline can catch up. A **budget** is a *capacity*: how many bytes the pipeline may hold in total. Before #764 the mark was the only place `--max-memory` reached, so the two fused, and the capacity half did not exist at all. That is why `--max-memory 32G` looked inert: the only number it set was a trigger — and a trigger that is capped for good reasons.

#764 moved the capacity half to where it belongs, onto total in-flight bytes gated at Read. What is left in `effective_limit` is only the trigger.

## Recommendation: honour the budget as a total, keep the per-stage cap

The reasons, then the numbers that back them:

- **Raising a trigger buys no throughput.** The scheduler's response to `memory_high` is a priority reordering, not a stop. A higher mark only postpones the moment output steps are favoured, so the stage parks more bytes and the pipeline runs no faster.
- **Reorder capacity is bounded by serial skew, not by bytes.** The pre-Group reorder buffers absorb out-of-order completion between workers; that skew is bounded by the queue slot count. Bytes past the skew window are dead weight.
- **These marks are the only byte bound on Q2, Q3 and Q5 individually.** Uncapped, one stage parks the entire budget and the total gate fires before any other stage buffers anything.

So `--max-memory 32G` is honoured in the sense a user means it — the pipeline may hold 32 GiB in total — without becoming a 32 GiB allowance for each reorder buffer.

## Measurements

`group --strategy adjacency --threads 8 --compression-level 1` over a 24,033,782-record synthetic BAM (`fgumi simulate mapped-reads -n 4000000 --seed 42`), `--memory-per-thread false` so each budget is a total, peak RSS from `/usr/bin/time -l`. Two sinks: `fast` writes straight to disk; `slow` writes through a 60 MB/s rate limiter, roughly 4.5x slower than the pipeline produces.

Three builds: **base** = #764 as it stands; **uncapped** = an experimental build with the clamp lifted (`effective_limit` returns the budget); **this PR**.

| declared budget | base fast | uncapped fast | PR fast | base slow | uncapped slow | PR slow |
| --- | --- | --- | --- | --- | --- | --- |
| 64 MiB | 616 | 584 | 560 | 573 | 507 | 493 |
| 256 MiB | 1264 | 1264 | 1231 | 1227 | 1173 | 1088 |
| 512 MiB | 1673 | 1650 | 1606 | 1422 | 1543 | 1454 |
| 4 GiB | 2360 | **5433** | 2413 | 4453 | **7984** | 4377 |
| 32 GiB | 2602 | **7190** | 2480 | 4422 | **17117** | 4357 |
| default (6 GiB) | 2443 | **7565** | 2580 | 4359 | **10541** | 4398 |

Peak RSS in MB. Every run exited 0 and emitted all 24,033,782 records.

Two things fall out of that table.

**The knob is not inert.** Peak RSS rises monotonically with the declared budget from 64 MiB to 4 GiB on both sinks — that is #764's Read gate doing the work the budget was always supposed to do. Above 4 GiB the input itself becomes the ceiling, which is why 32 GiB does not exceed 4 GiB here.

**Lifting the clamp is not free, and the price lands on the default path.** At the shipped default the uncapped build peaks at 7,565 / 10,541 MB against the base's 2,443 / 4,359 — 3.1x and 2.4x. At 32 GiB with a slow sink it reaches 17,117 MB against 4,422. It is not buying anything for that: wall time is unchanged, and the output is byte-identical. This is the one-stage-monopolises-the-budget failure mode, measured.

**The default path does not move.** Three replicates each, same session, interleaved:

| cell | base | this PR |
| --- | --- | --- |
| default, fast | 2427, 2399, 2438 | 2449, 2575, 2662 |
| default, slow | 4391, 4405, 4424 | 4386, 4392, 4384 |
| 64 MiB, fast | 626, 615, 627 | 589, 591, 579 |
| 64 MiB, slow | 573, 568, 599 | 505, 480, 524 |

Default is unchanged (fast-sink spread overlaps; slow-sink figures agree to within 0.5%). The 64 MiB rows are the Q5 fix showing up: 6% lower on a fast sink and 15% lower on a slow one, with no overlap between the two sets. Wall time is unchanged everywhere; the slow-sink cells are all 42.0s because the sink, not the pipeline, sets the pace.

**Output identity.** Eight independent runs — all three builds at 64 MiB and 32 GiB, base and this PR at 256 MiB — produce the same record stream, digest `a5829f1a…` on `samtools view` output. The whole-file digests differ between builds because BGZF block framing follows batch boundaries; the records do not.

Caveat: a shared laptop, so wall times are noisy and cross-cell RSS carries roughly ±100 MB of session noise. That is why the decisive cells were replicated and interleaved rather than read off a single pass.

## What changed

- **`stage_high_water_mark(budget, default_mark)`** is the single home for the rule, and carries the argument above in its docs. `ReorderBufferState::effective_limit` and `MemoryTracker`'s two threshold checks route through it. No behaviour change.
- **Q5 is the one behavioural fix.** Its 256 MiB mark was not derived from the budget at all, so `--max-memory 64M` left the processed queue free to park four times the whole declared budget on its own. It is now `min(budget, 256 MiB)`, resolved once at construction. This can only lower memory: at any budget at or above 256 MiB the mark is byte-identical, which covers every default configuration.
- **A liveness test covers the regime that widens as a result.** With `mi_assign_fn` installed, batches drained out of Q5 stay charged to `processed_heap_bytes` while they sit in `mi_assign_reorder`, so a mark below one batch holds the Process step off at its backpressure check. Forward progress then rests entirely on the unconditional held-batch push ahead of that check, and nothing pinned it. The new test drives the pipeline at 4 KiB and at 8 MiB — the first binds Read to one batch at a time, the second leaves several in flight so the reorder buffer actually has out-of-order serials to park — and asserts every record comes out, in order, with the hook run once per item.
- **The startup line and `--max-memory`'s help stop overclaiming.** The line now reports the marks actually in force beside the total:

  ```
  Queue memory budget: 6.0 GiB total (768.0 MiB/thread × 8 threads); per-stage high-water marks 512.0 MiB (reorder-buffered stages), 256.0 MiB (processed queue)
  ```

  A smaller budget shows tightened marks, which is the case where they move.
- **Two stale doc claims corrected.** `BACKPRESSURE_THRESHOLD_BYTES` said the FASTQ pipeline reads its drain signal from a pre-Group `q2_reorder_heap_bytes`; there is no such counter — `FastqStepContext::get_backpressure` reads `output.write_reorder_state`, the post-Compress write reorder buffer. And `MemoryTracker` now says outright that nothing in this crate constructs one.
- **The review config** claimed the budget is consulted through `MemoryTracker`. It points at the real gates now.

Also clears three `RUSTDOCFLAGS="-D warnings"` failures: two private-item links on `q2b_boundaries` carried in from the parent branch, and the redundant explicit link targets this change introduced.

## Two things the issue got wrong

Neither changes the conclusion, but both are worth recording.

**The cited lines point at dead code.** The issue attributes the clamp to `ReorderBufferState::effective_limit` "at `base.rs:609-611`, and again at `:628-630`". On `main` at `3d83a55d` those lines are `MemoryTracker::is_at_limit` and `MemoryTracker::is_below_drain_threshold`, which contain the identical `.min()` expression — hence the mix-up. `MemoryTracker` has no constructor anywhere outside this module's own tests, so `--max-memory` never reaches it. The live site is `effective_limit` at `:813-819`. The substance of the report is right; only the citation is off.

**"Raising `--max-memory` above 512M does not change peak RSS" is no longer true on top of #764**, as the table shows. It was true on `main`, which is where the issue was filed. What survives on this base is narrower: the *per-stage* marks do not scale, which is deliberate, plus Q5's mark not deriving from the budget at all, which was not.

## Scope

The FASTQ pipeline's missing total gate is #766 and is untouched here. This PR does change the shared `OutputPipelineQueues::is_processed_memory_high`, so `fgumi extract` picks up the Q5 derivation too — the only overlap, and it is additive.

## Verification

`cargo ci-fmt`, `cargo ci-lint`, `cargo ci-test` (7,516 passed), `cargo ci-doctest`, and `RUSTDOCFLAGS="-D warnings" cargo ci-doc` all green. The Q5 test was confirmed to fail against the pre-change implementation before the fix was applied.

Risk: output changes none, pinned by full `RecordBuf` identity, ordering, and liveness tests; `unsafe` changes none, so CLAUDE.md allowlist changes none; memory bounds and backpressure policy change.

Fix: Separate the total `--max-memory` budget from capped per-stage high-water marks through `stage_high_water_mark`.

- Enforce total in-flight memory at Read admission.
- Derive Q5’s mark from the configured budget, capped at 256 MiB.
- Add liveness coverage for small-budget backpressure.
- Report total and per-stage limits at startup and in help text.
- Update documentation and review guidance.
- Fix documentation warnings.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -187,9 +187,12 @@ reviews:
         producer gating as liveness- and memory-critical. Require that no path
         lets a queue or reorder buffer grow without a byte bound, so steady-state
         memory stays a function of config rather than input size: the per-queue
-        bound is `OrderedQueue::limit_bytes` (`queue.rs`), and the shared
-        pipeline budget is consulted through `MemoryTracker::is_at_limit` /
-        `MemoryTracker::is_below_drain_threshold` (`base.rs`). Flag a cap
+        bound is `OrderedQueue::limit_bytes` (`queue.rs`); the pipeline total is
+        gated at Read by `BamPipelineState::read_admission_allowed` (`bam.rs`);
+        and each stage's own high-water mark comes from `stage_high_water_mark`
+        (`base.rs`), which caps the mark rather than scaling it with the budget.
+        `MemoryTracker` (`base.rs`) is NOT on that path — nothing in this crate
+        constructs one, so do not cite it as the live budget check. Flag a cap
         checked on only one sub-condition so another code path bypasses it; flag
         capping a consumer drain loop (consumers must stay unbounded — bound the
         producers instead); and flag byte accounting that measures logical length
--- a/docs/src/guide/performance-tuning.md
+++ b/docs/src/guide/performance-tuning.md
@@ -78,6 +78,28 @@ total budget to avoid OOM.
 With `--max-memory auto`, the total budget is instead `host RAM − reserve`
 (divided across threads when per-thread), so it never scales past the host.
 
+### What the Budget Bounds
+
+`--max-memory` is a **total**: the BAM pipeline stops admitting input once the
+bytes queued between its stages reach it, so a slow or contended output device
+backs pressure up to the reader instead of filling every queue.
+
+It is not a per-stage allowance. The reorder-buffered stages back off at
+512 MiB and the processed queue at 256 MiB, and those marks do not rise with
+the budget. A budget below one of them pulls it down; a budget above it leaves
+it where it is. Raising `--max-memory` therefore raises how much the pipeline
+may hold in total, not how much any one stage may hold, and the startup line
+reports both numbers:
+
+```
+Queue memory budget: 6.0 GiB total (768.0 MiB/thread × 8 threads); per-stage high-water marks 512.0 MiB (reorder-buffered stages), 256.0 MiB (processed queue)
+```
+
+Note also that `fgumi extract`'s FASTQ pipeline shares this option and logs the
+same line, but does not yet enforce the total — there only the per-stage marks
+bind. That gap is tracked in
+[fgumi#766](https://github.com/fulcrumgenomics/fgumi/issues/766).
+
 ## Compression Options
 
 ### Compression Level
--- a/src/lib/commands/common.rs
+++ b/src/lib/commands/common.rs
@@ -10,7 +10,10 @@ use std::sync::Arc;
 use crate::assigner::Strategy;
 #[cfg(feature = "simplex")]
 use crate::logging::OperationTimer;
-use crate::unified_pipeline::{BamPipelineConfig, SchedulerStrategy};
+use crate::unified_pipeline::{
+    BACKPRESSURE_THRESHOLD_BYTES, BamPipelineConfig, Q5_BACKPRESSURE_THRESHOLD_BYTES,
+    SchedulerStrategy, stage_high_water_mark,
+};
 use crate::validation::validate_input_exists;
 use bytesize::ByteSize;
 use clap::Args;
@@ -1241,9 +1244,20 @@ fn resolve_memory_budget_with_total(
 /// input and write reorder states), so this budget bounds them through the Read
 /// gate; they additionally apply their own threshold, capped at
 /// [`BACKPRESSURE_THRESHOLD_BYTES`](crate::unified_pipeline::BACKPRESSURE_THRESHOLD_BYTES).
+///
+/// The budget is a **total**, and it is not the same thing as what any one stage
+/// may hold. Stages back off at their own high-water marks —
+/// [`BACKPRESSURE_THRESHOLD_BYTES`](crate::unified_pipeline::BACKPRESSURE_THRESHOLD_BYTES)
+/// (512 MiB) for the queues and reorder buffers,
+/// [`Q5_BACKPRESSURE_THRESHOLD_BYTES`](crate::unified_pipeline::Q5_BACKPRESSURE_THRESHOLD_BYTES)
+/// (256 MiB) for the processed queue. A budget below one of those marks tightens
+/// it; a budget above it leaves it alone. So raising `--max-memory` raises how
+/// much the pipeline may hold in total, not how much any one stage may hold —
+/// see [`stage_high_water_mark`](crate::unified_pipeline::stage_high_water_mark)
+/// for why a per-stage trigger should not scale with the total (issue #765).
 #[derive(Debug, Clone, Args)]
 pub struct QueueMemoryOptions {
-    /// Maximum memory for the pipeline queues.
+    /// Maximum total memory for the pipeline queues.
     ///
     /// Default is "768" (768 MiB) per thread. Pass "auto" to detect host memory
     /// (cgroup-aware) and subtract --memory-reserve, so the budget shrinks to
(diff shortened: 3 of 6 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed the outside-diff-range comment: updated the stale `OutputPipelineQueues::groups_heap_bytes` doc in `base.rs`. It no longer claims "stats/reporting only" / `memory-debug`-gated, and now states the field is charged by the Group step, refunded by the Process step, and summed into `BamPipelineState::queue_bytes_in_flight` for Read admission — matching the `q1_heap_bytes`/`q2b_heap_bytes` wording.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai full review

PR #783, opened 2026-08-15
fix(pipeline)!: fail on data left unflushed at end of stream

### LINKED ISSUE #782: EOF flush paths discard buffered data instead of failing
## Summary

Several end-of-stream flush paths discard buffered data instead of failing, in some cases contradicting their own documentation. Two can produce silently wrong output; the rest weaken the completion checks that exist to catch exactly this.

These were found while fixing #773 (mismatched-length FASTQ pairs), which is the same shape: data parked in a stage buffer, and an EOF path that drops it rather than erroring.

## Sites

**1. `simulate/parallel_gzip_writer.rs:176-213` — silently writes corrupt output.**
The final flush emits leftover reorder blocks with no check for a gap in the serial sequence. A missing serial produces a valid gzip file with a hole in it: readable, complete-looking, and short. The sibling implementation `fgumi-sort/src/bgzf_io.rs:283-301` explicitly errors on this condition, so the correct behavior is already established in the codebase.

**2. `grouper.rs:551-569` — release-only data loss.**
`RecordPositionGrouper::finish` guards its flush with `debug_assert!`. In a release build the assertion compiles out, the buffered records are dropped, and `has_pending()` continues to report `true` afterwards.

**3. `bam.rs:256-302` `BoundaryState::finish` — contradicts its own doc.**
Silently drops a 1-3 byte record remainder at EOF. Its documentation promises an error for that case.

**4. `bam.rs:930-1035` `validate_completion` — does not cover the states it exists for.**
It checks neither `GroupState::pending_groups`, nor `grouper.has_pending()`, nor `boundary_state.leftover`, although `base.rs` names those three as precisely what completion validation is meant to catch. The FASTQ pipeline has the same gap.

**5. `simulate/common.rs:637-700` `RecordSink` — last batch lost on drop.**
If the sink is dropped without calling `finish()`, the final partial batch is discarded. There is no `Drop` implementation, and both call sites discard `finish()`'s boolean return.

**6. Any failed `extract` leaves a valid, complete-looking empty BAM** at `--output`, with a BGZF EOF marker and zero records. Pre-existing and command-wide rather than specific to one flush path, but it is the same failure to distinguish "finished" from "stopped".

## Requested change

Make these paths fail rather than discard. Where a flush cannot complete, return an error naming what was left unflushed; where a completion predicate exists, have it cover the states that can hold data. Site 1 has a working reference implementation in `bgzf_io.rs` to follow.

Site 6 may warrant separate treatment since it is about output-file lifecycle rather than a flush path.

nh13:
Closes #782.

## What was wrong

Five end-of-stream paths ended a run *successfully* while still holding data, so the loss looked exactly like a clean finish. Each site was re-verified against `main` before it was touched — the results are in the table, including the one the issue got wrong.

| # | Site | Real? | Failure it produced |
| --- | --- | --- | --- |
| 1 | `simulate/parallel_gzip_writer.rs` final flush | yes | a valid gzip file with a hole in it |
| 2 | `RecordPositionGrouper::finish` | yes | release-only: buffered records dropped, `has_pending()` still `true` |
| 3 | `BoundaryState::finish` | yes | a 1-3 byte record fragment discarded at EOF |
| 4 | `validate_completion` (BAM) | yes | none of the three internal buffers it documents were checked |
| 4b | `validate_completion` (FASTQ) | **no — already safe** | it already checks pair state, block-merge state and per-stream boundary leftovers |
| 5 | `simulate/common.rs` `RecordSink` | partly | last partial batch lost on drop; the two call sites' discarded `bool` is benign |
| 6 | failed `extract` leaves an empty BAM | out of scope | output-file lifecycle, not a flush path — see below |

**Site 1** wrote whatever reorder blocks were left, in serial order, with no gap check. `fgumi-sort/src/bgzf_io.rs` already rejects exactly this; the fix follows it, including its wording.

**Site 2** guarded its flush with `debug_assert!`. With assertions compiled out the `if let Some(key)` simply failed, `finish` returned `Ok(None)`, and the records stayed buffered forever.

**Site 3** scanned with `cursor + 4 <= leftover.len()`, so a 1-3 byte tail fell out of the loop: discarded with `Ok(None)` when no complete record preceded it, and left past the final offset when one did. Its own doc comment already promised an error. **This one has real reach**: the `--threads` path never calls `finish()` at all, so a truncated BAM's trailing bytes were dropped there too — now caught by site 4's `boundary_leftover` check.

**Site 4** — `base.rs` says completion validation covers "internal buffers ... (grouper state, boundary leftovers)". The BAM implementation checked none of the three. Two of them are not reachable from the pipeline state at all: the grouper and its pending groups sit behind the Group step's own mutex.

**Site 5** — `RecordSink` had no `Drop`, and dropping the sender is what signals end-of-stream, so an abandoned sink looked like a clean end of input and the sort wrote a short output. The issue also flags the two call sites discarding `finish()`'s `bool`; that turned out benign — `false` only happens when the sorter has already hung up, and `sort_records` surfaces that error itself. The call sites now say so instead of discarding silently.

### Corrections to the issue

- **The FASTQ side does not have the same `validate_completion` gap.** It already reports `pair_state`, `block_merge_state` and every stream's `boundary_leftover`. Its BGZF branch skips the boundary-leftover check, but `boundary_state` is only touched on the gzip/plain path (`fastq_try_step_find_boundaries` runs only when `!inputs_are_bgzf`), so there is nothing to miss. No change was needed there.
- **Site 5's "both call sites discard `finish()`'s bool" is not itself a loss.** The drop path is; the discarded bool is not.
- **Site 6 stayed out of scope**, as the issue itself suggests and as #776's agent decided. It is about when the output file is created and finalized, not about a flush path, and fixing it means an output-lifecycle change (temp-file + persist, or unlink-on-error) across every command. It did not fall out naturally: this PR makes the pipeline *return* an error where it used to succeed, which is the input side of the same problem, but the empty-BAM artifact is still written.

## One site the issue does not list

Sweeping for site 2's shape — *a `debug_assert!` standing in for a check whose failure is silent data loss in release* — found `ReorderBuffer::insert_with_size` (`crates/fgumi-bam-io/src/reorder.rs`), which asserted **both** of its preconditions that way. Compiled out, a duplicate serial overwrote the buffered item, dropping a whole batch of records while `count` and `heap_bytes` counted it twice; and a serial below the base underflowed the `u64` `seq - next_seq`, leaving the extend loop growing the buffer toward an unreachable index. `.coderabbit.yaml` calls reorder accounting "data-integrity-critical" for exactly this reason. It is a separate commit (`fix(bam-io): ...`) so it can be split out if preferred.

## The fix

- **Site 1** — the leftover drain walks `first_key_value()` and errors on a gap: `missing compressed block 0: next available is 1; the output would be silently truncated`.
- **Site 2** — the guard is unconditional and returns an error naming the record count, so it is observable in a release build (verified: `cargo nextest run --release` on that test).
- **Site 3** — trailing bytes below a `block_size` are an `UnexpectedEof` naming how many.
- **Site 4** — `validate_completion` reports `boundary_leftover (n)`; `finalize_pipeline_with_buffers` folds `GroupState::unflushed_buffers()` (`group_pending_groups (n)`, `grouper (partial group pending)`) into the same `PipelineValidationError`, so one error lists everything that still held data. The single-threaded path, which has no pipeline state, checks the grouper directly after `finish()`.
- **Site 5** — `Drop for RecordSink` sends the loss as the stream's terminal error, so the sort aborts rather than finishing short. `finish()` and `fail()` both leave the buffer empty, so it is a no-op after either.

## Tests

TDD — every test below was written first and observed failing against `main`:

```
the pipeline reported success (0 records) with the grouper still holding 30 record(s);
the output holds 0 of 30 input records
```

```
thread '...test_io_writer_loop_rejects_a_gap_in_the_serial_sequence' panicked:
a missing block must fail the write, not be skipped over: ()
```

- `test_io_writer_loop_rejects_a_gap_in_the_serial_sequence` — asserts the **emitted bytes**, not just the error: on `main` the sink decodes to `BBBBCCCC`, a valid gzip stream missing block 0. Its control, `test_io_writer_loop_drains_out_of_order_blocks`, asserts `AAAABBBBCCCC` from blocks fed 2-0-1.
- `test_record_position_grouper_finish_rejects_records_it_cannot_emit` — drives the state the `debug_assert!` claimed to rule out, asserts the error names the count *and* that the records are still held. `test_record_position_grouper_finish_emits_every_buffered_record` pins the happy path by **read name**, not count.
- `test_boundary_state_finish_rejects_a_sub_header_fragment` (fragment-only and record-then-fragment) — asserts the error kind and text, and that the complete records came through **byte-for-byte**. Control: `test_boundary_state_reassembles_a_record_split_across_calls`.
- `test_pipeline_rejects_a_grouper_still_holding_records` (1 and 4 threads) — a grouper that swallows records must fail the run; on a wrongly-successful run it reports how many of the 30 input records reached the output.
- `test_validation_detects_boundary_leftover`, `test_group_state_reports_unflushed_records_and_groups`, and three `merge_unflushed_buffers` cases (clean, already-failing, nothing-buffered).
- Four `RecordSink` tests asserting delivered records **by name** plus the terminal error, covering `finish`, drop-with-buffer, drop-after-finish, and `fail`.
- Two `ReorderBuffer` `#[should_panic]` tests for the duplicate and stale serials.

Gate: `cargo ci-fmt`, `cargo ci-lint`, `cargo ci-test` (7,515 passed), `cargo ci-doctest`.

## Base

Based on **`main`**. #764, #772 and #775 all touch `bam.rs`/`base.rs`, but in the queue-memory accounting and the `try_step_*` bodies; this PR touches `BoundaryState`, `validate_completion`, `GroupState`, and the finalize helper, and needs nothing from them. #776 touches `fastq.rs`, which this PR does not modify at all — the FASTQ side turned out already safe. The one point of contact is textual proximity in `bam.rs`, not a semantic conflict.

## Behavior change

Marked breaking on the same reasoning as #776: a BAM whose last record is truncated used to produce a short output and exit 0, and now exits non-zero. The other four sites fire only on an internal invariant a correct run never reaches, so no well-formed input changes behavior.

Risk: command output changes: none on successful runs; EOF data-loss cases now fail instead; `unsafe`: none; memory bounds, queue capacity, and thread/backpressure policy: none.

Fix: reject unflushed or invalid EOF state across BAM, gzip, grouper, and simulation record sinks.

- Reject missing gzip block serials during final output draining.
- Report grouper flush failures in release builds.
- Reject truncated BAM fragments, boundary leftovers, pending groups, and unflushed group buffers.
- Report dropped `RecordSink` buffers as terminal errors.
- Reject duplicate and stale reorder serials without relying on `debug_assert!`.
- Add tests for each failure mode.
- Leave FASTQ validation and failed `extract` output-file handling unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,23 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Bug Fixes
+
+- **Breaking:** end-of-stream flush paths now fail instead of discarding what they were holding,
+  so an input that used to produce a short output and exit 0 now exits non-zero. A BAM whose last
+  record is truncated is the case a user can hit: the boundary finder dropped a 1-3 byte record
+  fragment at EOF (though its own documentation promised an error), and in `--threads` mode
+  nothing checked its leftover bytes at all. The other four paths report an internal invariant
+  that a correct run never reaches: `simulate`'s parallel gzip writer emitted whatever blocks
+  were left without checking the serial sequence, so a block that never arrived produced a valid
+  gzip file with a hole in it; `RecordPositionGrouper::finish` guarded its flush with a
+  `debug_assert!`, so release builds dropped the buffered records while `has_pending()` went on
+  reporting them; the BAM pipeline's completion validation checked none of the three internal
+  buffers it is documented to cover (the grouper's partial group, groups never pushed to Q4, and
+  the boundary finder's leftover bytes); and `simulate`'s record sink lost its last partial batch
+  if it was dropped without `finish()`. Each now reports what was left unflushed
+  ([#782](https://github.com/fulcrumgenomics/fgumi/issues/782)).
+
 ## [0.6.0] - 2026-08-14
 
 ### Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,12 @@ All notable changes to this project will be documented in this file.
   if it was dropped without `finish()`. Each now reports what was left unflushed
   ([#782](https://github.com/fulcrumgenomics/fgumi/issues/782)).
 
+- The pipeline's reorder buffer now rejects a duplicate serial, and a serial below its base, in
+  every build rather than only in debug. Both were `debug_assert!`s, so a release build silently
+  overwrote the buffered batch in the first case — losing its records while double-counting them
+  in the memory accounting — and underflowed the `u64` index computation in the second
+  ([#782](https://github.com/fulcrumgenomics/fgumi/issues/782)).
+
 ## [0.6.0] - 2026-08-14
 
 ### Bug Fixes
--- a/crates/fgumi-bam-io/src/reorder.rs
+++ b/crates/fgumi-bam-io/src/reorder.rs
@@ -78,8 +78,8 @@ impl<T> ReorderBuffer<T> {
     ///
     /// # Panics
     ///
-    /// Panics in debug mode if an item with the same sequence number is already buffered,
-    /// or if the sequence number is before the current base.
+    /// Panics if an item with the same sequence number is already buffered, or
+    /// if the sequence number is before the current base.
     pub fn insert(&mut self, seq: u64, item: T) {
         self.insert_with_size(seq, item, 0);
     }
@@ -97,15 +97,20 @@ impl<T> ReorderBuffer<T> {
     ///
     /// # Panics
     ///
-    /// Panics in debug mode if an item with the same sequence number is already buffered,
-    /// or if the sequence number is before the current base.
+    /// Panics if an item with the same sequence number is already buffered, or
+    /// if the sequence number is before the current base.
+    ///
+    /// Both checks are unconditional, not `debug_assert!`s. Serial accounting
+    /// here is data-integrity-critical, and in a release build the compiled-out
+    /// assertions left two silent failures: a duplicate serial overwrote the
+    /// buffered item — dropping a whole batch of records while `count` and
+    /// `heap_bytes` counted it twice — and a serial below the base underflowed
+    /// `seq - next_seq`, so the extend loop below tried to grow the buffer to
+    /// an unreachable index. Neither is recoverable in place, so this fails
+    /// loudly; the pipeline's `catch_unwind` turns it into a reported error.
     #[allow(clippy::cast_possible_truncation)]
     pub fn insert_with_size(&mut self, seq: u64, item: T, heap_size: usize) {
-        debug_assert!(
-            seq >= self.next_seq,
-            "Sequence number {seq} is before base {}",
-            self.next_seq
-        );
+        assert!(seq >= self.next_seq, "Sequence number {seq} is before base {}", self.next_seq);
 
         let index = (seq - self.next_seq) as usize;
 
@@ -114,7 +119,7 @@ impl<T> ReorderBuffer<T> {
             self.buffer.push_back(None);
         }
 
-        debug_assert!(self.buffer[index].is_none(), "Duplicate sequence number: {seq}");
+        assert!(self.buffer[index].is_none(), "Duplicate sequence number: {seq}");
         self.buffer[index] = Some((item, heap_size));
         self.count += 1;
         self.heap_bytes += heap_size as u64;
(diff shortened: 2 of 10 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
Addressed the outside-diff finding in `src/lib/unified_pipeline/bam.rs`: the trailing-byte check made the `if cursor == 0 { return Ok(None); }` branch unreachable (after the early empty-leftover return, `cursor == 0` is always caught by the `cursor < len` check), so it is now a `debug_assert!(cursor > 0, ...)` documenting the invariant. Folded into the first commit.

PR #526, opened 2026-07-10
fix(consensus): emit @HD SO:unsorted to match fgbio

nh13:
## Summary

Unmapped consensus output (`simplex`/`duplex`/`codec`) wrote `@HD ... SO:unknown` while fgbio's consensus tools emit `SO:unsorted` (both with `GO:query`). This changes `create_unmapped_consensus_header` to emit `SO:unsorted`, matching fgbio exactly.

## Evidence (real-data validation)

Across every consensus cross-compare on real data — `idt-cfdna`/`SRR6109255`/`SRR6109273`/`schmitt-abl1` simplex & duplex and `CODEC` codec — consensus record content was already byte-identical to fgbio (content diffs: 0); the **sole** divergence was this one header field, `@HD SO: "unknown" vs "unsorted"`.

Confirmed with the fixed binary on `idt-cfdna` duplex: regenerated output header is now `@HD VN:1.6 SO:unsorted GO:query`, byte-identical to the fgbio baseline, and the record comparison remains EQUIVALENT (322/322 records, 0 grouping mismatches).

## Testing

- New regression test `test_create_unmapped_consensus_header_sort_order_matches_fgbio` asserts `SO == unsorted` and `GO == query`.
- `cargo nextest` (consensus_runner module), `cargo clippy --workspace --all-targets --features compare,simulate,profile-adjacency -- -D warnings -W clippy::pedantic`, and `cargo fmt --check` all clean.

## Notes

Independent of #508 (`fix(consensus): sort-order + error-rate guards`), which validates consensus **input** sort order (`checkSortOrder`) and does not touch the output `@HD SO` value.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated unmapped consensus SAM headers to use the standard `unsorted` sort order and `query` read-group order.
  * Improved compatibility with fgbio by matching its expected header values.

PR #784, opened 2026-08-16
fix(compare): treat spec-equivalent @HD SO/GO spellings as compatible

### LINKED ISSUE #778: compare bams: SO:unknown vs SO:unsorted is rejected as incompatible, though both mean unordered
`fgumi compare bams` refuses to compare two BAMs when one declares `@HD SO:unknown` and the other `@HD SO:unsorted`:

```
Error: input BAM headers are incompatible

Caused by:
    @HD SO/GO tags differ and neither input's sort order could be determined
```

Both values mean "these records are not in a defined sort order", so there is nothing to reconcile — the comparison the user asked for is well-defined and could proceed. `SO:unknown` is the SAM spec's *default* value, and `SO:unsorted` differs from it only in asserting that the lack of order is deliberate. Neither implies an ordering that the other contradicts.

## How it surfaced

fgumi's own output changed spelling between releases. Sampling `@HD` from the same rule's output across four runs of our benchmark suite:

| fgumi | `@HD` |
| --- | --- |
| v0.4.0 (`7adda90`) | `VN:1.6 SO:unknown GO:query` |
| v0.5.0 (`3df1022`) | `VN:1.6 SO:unsorted GO:query` |
| v0.6.0 (`cf41215`) | `VN:1.6 SO:unsorted GO:query` |

So comparing any v0.5.0-or-later output against any pre-v0.5.0 output is currently impossible, even though the records are directly comparable and `GO:query` matches on both sides. In our case this silently disabled a whole pipeline's equivalency checking: four methylation comparison cells failed as hard errors rather than producing a result, and because they errored rather than reporting a difference, they read as infrastructure breakage rather than as a signal about output.

## Suggested behaviour

Treat `unknown` and `unsorted` as compatible — both denote "no defined order" — and fall back to the order-insensitive path, exactly as if both had said `unsorted`. Alternatively, keep the error but only when the two sides genuinely disagree about an *ordering* (e.g. `coordinate` vs `queryname`), which is the case where a positional comparison really would be meaningless.

If the strictness is deliberate, a targeted flag (`--assume-unsorted`, or allowing `--ignore-order` to override the header check) would let callers proceed without hand-editing headers on multi-GB files.

## Workaround

`samtools reheader` to rewrite the older side's `SO:` value, which requires rewriting the whole file — 4 GB per input in our case, for a one-word header change.

nh13:
Fixes #778.

`fgumi compare bams` hard-errored when one input declared `@HD SO:unknown` and the other `SO:unsorted`:

```
Error: input BAM headers are incompatible

Caused by:
    @HD SO/GO tags differ and neither input's sort order could be determined
```

Both values denote "these records are in no defined order" — `unknown` is the SAM spec's default and `unsorted` differs from it only in asserting that the lack of order is deliberate. Neither implies an ordering the other contradicts, so the comparison the user asked for is well-defined.

## Root cause

`require_compatible_headers` already classified the pair correctly as the *benign* orderless case — `sort_order_from_header` returns `Err` for both, and `hd_so_claims_orderable` is `false` for both, so neither header claims an order the engine failed to verify — and then threw that classification away on a raw byte comparison of the two spellings in `compare_hd`. The tell is that `SO:unknown` vs `SO:unknown` already compares fine today; only the spelling mismatch failed.

This surfaced because fgumi's own output changed spelling: 60dd5e7f ("fix(consensus): emit @HD SO:unsorted to match fgbio", #526, first released in v0.5.0) switched unmapped consensus output from `SO:unknown` to `SO:unsorted` to match fgbio. Any pre-v0.5.0 vs v0.5.0+ comparison has been impossible since, without rewriting multi-GB files through `samtools reheader`.

## Fix

Canonicalize each `@HD` tag to the spec state it denotes before comparing:

- `SO`: absent | `unknown` | `unsorted` → one state (absent is the spec default, `unknown`)
- `GO`: absent | `none` → one state (the same class of bug, since absent means the spec default)

Every other value is compared verbatim, so `coordinate` vs `queryname` and `GO:query` vs `GO:reference` remain divergences. Diff messages still render the raw declared values, so a reported divergence names what each header actually declares.

`compare_hd` backs *both* the fatal `require_compatible_headers` gate and the non-fatal `compare_headers` diff (via `fold_header_diffs` in the `positional` and `sort_verify` engines), so fixing it there means the pair compares cleanly — fixing only the gate would have downgraded the hard error to a spurious `DIFFER`.

No new flag is needed; the equivalence is unconditional and spec-grounded, so `--assume-unsorted` (floated in the issue) is unnecessary.

## Testing

Written test-first — the 8 new equivalence cases failed with exactly the reported error before the fix, and the 3 boundary cases passed before and after.

- `spec_equivalent_orderless_hd_spellings_are_not_a_diff` — 4 cases over `compare_headers`
- `require_compatible_headers_accepts_spec_equivalent_orderless_spellings` — same 4 cases over the fatal gate, asserting `Ok(None)` (compatible, no verifiable order)
- `genuinely_differing_hd_tags_remain_a_diff` — the boundary the equivalence must not cross (`unknown` vs `coordinate`, `unsorted` vs `queryname`, `GO:query` vs `GO:reference`)

`cargo ci-test` (7205 tests), `cargo ci-fmt`, and `cargo ci-lint` all pass. End-to-end on real BAMs with the release binary:

| comparison | exit | result |
| --- | --- | --- |
| `SO:unknown GO:query` vs `SO:unsorted GO:query` | 0 | IDENTICAL (was exit 1, hard error) |
| `SO:coordinate` vs `SO:queryname` | 1 | headers incompatible |
| `GO:query` vs `GO:reference` | 1 | headers incompatible |
| equivalent spellings, differing record | 1 | DIFFER, 1 content diff |

## Out of scope

`SO:unknown GO:query SS:template-coordinate` vs `SO:unsorted GO:query SS:template-coordinate` still hard-fails, because `sort_order_from_header` recognizes only `unsorted` for template-coordinate and the pair therefore takes the `(Err, Ok)` arm. fgumi has never written `SS` alongside `SO:unknown`, so this does not affect the reported case; widening it would mean touching `sort_order_from_header` as well.

Risk: grouping, consensus, sort order, corrected UMIs, and metrics output changes none; `unsafe` changes none and CLAUDE.md requires no allowlist update; memory bounds, queue capacity, and thread/backpressure policy changes none.

Fix: Treat spec-equivalent `@HD SO` and `GO` spellings as compatible in BAM comparisons while preserving raw values in diff messages.

- Applies to fatal compatibility checks and non-fatal header diffs.
- Keeps genuine sort and group order conflicts incompatible.
- Adds coverage for equivalent values, real differences, and end-to-end comparisons.
- `cargo ci-test`, `cargo ci-fmt`, and `cargo ci-lint` pass.

### DIFF
--- a/src/lib/commands/compare/engines/header.rs
+++ b/src/lib/commands/compare/engines/header.rs
@@ -12,8 +12,13 @@
 //! fgbio (or between two independent tool invocations) even on functionally equivalent
 //! output:
 //!
-//! - **`@HD`** — the `SO` (sort order) and `GO` (group order) tags are compared
-//!   byte-for-byte. The `SS` (sub-sort) tag is deliberately **not** byte-compared here:
+//! - **`@HD`** — the `SO` (sort order) and `GO` (group order) tags are compared after
+//!   canonicalizing each to the SAM-spec *state* it denotes, so two headers that spell the
+//!   same state differently do not read as a content difference (`SO:unknown` vs
+//!   `SO:unsorted` — both "no defined order" — and an absent tag vs. its spec default; see
+//!   [`canonical_sort_order`]/[`canonical_group_order`], issue #778). Any genuine ordering
+//!   disagreement (`coordinate` vs `queryname`, `GO:query` vs `GO:reference`) is still a
+//!   divergence. The `SS` (sub-sort) tag is deliberately **not** byte-compared here:
 //!   fgumi writes the bare form (e.g. `SS:template-coordinate`) while fgbio/samtools write
 //!   the `<sort-order>:`-prefixed form (`SS:unsorted:template-coordinate`) for the same
 //!   sort order (`R2-HDR-01`), so a byte comparison would spuriously flag every
@@ -75,8 +80,9 @@ pub fn compare_headers(h1: &Header, h2: &Header) -> Option<Vec<String>> {
 /// [`SortOrder`] (see [`sort_order_from_header`] — this is the case for `sort`/`group`
 /// output and any coordinate/queryname-sorted BAM). Returns `Ok(None)` when **neither**
 /// header's raw `@HD SO` claims a recognized orderable order (`coordinate`/`queryname`) —
-/// i.e. both are genuinely orderless (absent `SO`, or `SO:unsorted`) — and the two headers'
-/// raw `@HD SO`/`GO` tags still agree byte-for-byte: this is the legitimate case for
+/// i.e. both are genuinely orderless (absent `SO`, `SO:unknown`, or `SO:unsorted`) — and the
+/// two headers' `@HD SO`/`GO` tags still agree once canonicalized to the state each denotes
+/// (see [`compare_hd`]; `unknown`/`unsorted`/absent are one state): this is the case for
 /// pre-sort pipeline stages (`extract`/`fastq`/`zipper`/unmapped `consensus` output all
 /// declare bare `SO:unsorted GO:query` with no `SS` — genuinely unordered data, not a
 /// recognized sort order, but still a valid same-format pair to content-compare). A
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #785, opened 2026-08-16
fix(consensus): count and route reads dropped by downsampling

### LINKED ISSUE #724: Reads dropped by consensus downsampling are counted as used and never reach --rejects

nh13:
## Summary

Reads discarded by the `--max-reads` consensus-downsampling cap were counted as *used* and never reached `--rejects`. Because `raw_reads_used = total_input_reads - filtered_reads` and only a recorded rejection increments `filtered_reads`, a family of 10 capped to 3 reported `raw_reads_used = 10` against a consensus of depth 3 (`cD:i:3`), and the 7 discarded reads were dropped silently.

Closes #724.

## What changed

A new `Downsampled` rejection reason (caller code `d`) is recorded at each site that actually *drops* reads under the cap, so `raw_reads_used` excludes them:

- **simplex** (`VanillaUmiConsensusCaller::process_group`): `downsample_reads` now returns `(kept, dropped)`; the dropped reads are counted and, when rejects tracking is enabled, routed to the `--rejects` output byte-for-byte in input order — matching how the vanilla caller already routes its other per-read rejections.
- **codec** (`CodecConsensusCaller`): the per-strand cap counts the discarded reads, mirroring how the codec caller already counts its other per-read rejections (e.g. `MinorityAlignment`).

## Scope notes

- **#727 already mooted the duplex/codec `consensus_call` path.** Since #727, that path retains the full uncapped `source_reads` (the cap shapes only the consensus bases/quals/depths), so no reads are dropped there and nothing needs routing. The remaining drop sites are the simplex group-level cap and the codec per-strand cap — the two this PR addresses. The issue text predates #727 and references stale line numbers.
- **Codec is count-only here.** The codec caller does not yet route *any* per-read rejection to `--rejects` (only whole-group failures); that machinery is added by #751. Routing the downsampled codec reads now would double-count on the whole-group-failure path and collide with #751, so this PR fixes the codec *count* (which is what `raw_reads_used` depends on) and leaves per-read codec routing to land with #751.
- **fgbio divergence.** `Downsampled` is fgumi-specific: fgbio has no equivalent metric yet (fulcrumgenomics/fgbio#1166, #1167). Under a cap, fgumi's `raw_reads_used` and the new `raw_reads_rejected_for_downsampled` row now differ from fgbio until fgbio#1167 lands. The new metric row is emitted only when non-zero, so runs without a cap are unchanged.

## Tests

- `vanilla_caller::downsampled_reads_are_rejected_and_routed_to_the_rejects_output` — a family of 10 capped to 3 records 7 `Downsampled` rejections, includes them in `filtered_reads`, and emits them to `--rejects` byte-for-byte in input order.
- `codec_caller::codec_downsampled_reads_are_counted_as_rejected` — six identical-alignment templates capped to 3 per strand record 6 `Downsampled` rejections.

Full workspace suite green (6682 tests); `cargo ci-fmt` and `cargo ci-lint` clean.

Risk: output changes for `--rejects` BAM and rejection metrics, pinned by `RejectionReason::Downsampled` and input order; unsafe changes none, so `CLAUDE.md` allowlist changes none; memory, queue, and thread/backpressure policy changes none.

Fix: Count reads discarded by consensus downsampling as `Downsampled` rejections.

- Add `Downsampled` rejection metadata and consensus metrics.
- Route discarded simplex reads to `--rejects`.
- Count discarded CODEC reads per strand.
- Preserve existing duplex and CODEC `consensus_call` behavior.
- Add regression tests for rejection routing and counts.

### DIFF
--- a/crates/fgumi-consensus/src/caller.rs
+++ b/crates/fgumi-consensus/src/caller.rs
@@ -436,6 +436,10 @@ pub enum RejectionReason {
     PotentialCollision,
     /// Template did not have a single primary FR pair of reads (codec)
     NotPrimaryFrPair,
+    /// Read was discarded by consensus downsampling (a deliberate `--max-reads` cap, not a
+    /// quality or alignment failure). Counted so `raw_reads_used` excludes it and routed to
+    /// the `--rejects` output (fgumi#724).
+    Downsampled,
     /// Other unspecified reason
     Other,
 }
@@ -464,6 +468,7 @@ impl RejectionReason {
             Self::HighDuplexDisagreement => 'H',
             Self::PotentialCollision => 'C',
             Self::NotPrimaryFrPair => 'R',
+            Self::Downsampled => 'd',
             Self::Other => 'O',
         }
     }
@@ -493,6 +498,7 @@ impl RejectionReason {
                 "Potential collisions (reads with the same MI but different strands)"
             }
             Self::NotPrimaryFrPair => "Template did not have a single primary FR pair of reads",
+            Self::Downsampled => "Downsampled to the maximum reads per consensus",
             Self::Other => "Other reason",
         }
     }
@@ -530,6 +536,9 @@ impl RejectionReason {
             Self::OrphanConsensus => CentralReason::OrphanConsensus,
             Self::PotentialCollision => CentralReason::DuplicateUmi,
             Self::NotPrimaryFrPair => CentralReason::NotPrimaryFrPair,
+            // fgumi-specific: fgbio has no downsampled rejection metric yet
+            // (fulcrumgenomics/fgbio#1166, #1167), so this diverges from fgbio under a cap.
+            Self::Downsampled => CentralReason::Downsampled,
         }
     }
 }
--- a/crates/fgumi-consensus/src/codec_caller.rs
+++ b/crates/fgumi-consensus/src/codec_caller.rs
@@ -756,8 +756,16 @@ impl CodecConsensusCaller {
                 );
                 return Ok(ConsensusOutput::default());
             }
-            Self::cap_infos_to_lowest_ranking(records, &mut r1_infos, max_reads);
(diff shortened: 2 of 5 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #787, opened 2026-08-16
fix(pipeline): admit the gap-filler serial to Decode under memory-high

nh13:
> **Draft, stacked on #764** (base = `746/nhomer/bound-consensus-memory-under-writer-stall`). It fixes a deadlock that #764's tests still hit — see below. Rebase onto #764 when that lands.

## The problem: #764's deadlock is not fixed, and green CI is masking it

#764's three `test_pipeline_memory_backpressure` tests **pass in CI but reliably hang on a loaded host**. The hang is a real, scheduling-sensitive deadlock in the Decode step, not flaky tests:

- The Decode step gates new work with a blanket `q3_reorder_state.is_memory_high()` check *before* popping Q2b (bam.rs, `try_step_decode`). It was copied from the Decompress step, whose comment argues the guard is deadlock-safe because "Q1 is FIFO, so `next_seq` for the reorder buffer has already been produced."
- That invariant does **not** hold for Decode. When a slow writer backs the pipeline up and the Q3 reorder buffer reaches its high-water mark, the guard also stops Decode from decoding **the very serial the reorder buffer is waiting on**. That serial never reaches Q3, Group starves on it, and the pipeline wedges with the reader gated — exactly the OOM-avoidance path #764 adds.

**Evidence (controlled A/B on the loaded host, same commit, only this guard toggled):**

| Decode Q3 `is_memory_high` guard | 4 backpressure tests |
| --- | --- |
| ON (as in #764) | 4/4 **TIMEOUT** (hang) |
| OFF | 4/4 pass |

CI's lighter scheduling doesn't trip the race, so #764 goes green — a false negative for this deadlock class.

## The fix

Replace the blanket guard with the reorder buffer's own per-serial `can_proceed`, applied *after* the pop:

- It **always admits the gap-filler** (`serial == next_seq`), even over the memory limit, so the serial Group is waiting on can never be starved.
- It backpressures only **future** serials, once the buffer is half full.
- A future batch rejected by `can_proceed` is **returned to Q2b** so another worker can still reach the gap-filler — avoiding the "all workers hold a non-`next_seq` batch and nobody can produce `next_seq`" deadlock the code hit when `can_proceed` was previously wired into Decompress (see that step's comment).

On the healthy path (memory below the high-water mark) `can_proceed` is always `true`, so throughput is unchanged — the new backpressure engages only under the exact memory-high condition that used to deadlock.

## Validation

- The four backpressure tests pass on the loaded reproducer across **3 repeated runs** (they reliably hang without this change).
- Full workspace suite: **6675/6675 pass**. `cargo ci-fmt` and `cargo ci-lint` clean.

## Caveat worth raising separately

CI cannot currently catch this deadlock class (it needs constrained scheduling to reproduce). A follow-up that reproduces it under load in CI would keep it from regressing silently.

Risk: output for grouping, consensus, sort order, corrected UMIs, and metrics: none; `unsafe`: none, and no CLAUDE.md allowlist update; memory bound, queue capacity, and thread/backpressure policy: changed for Decode under memory-high conditions.

Fix: Decode applies per-serial admission after popping Q2b. It requeues future serials and admits the required Q3 gap-filler serial. This prevents stalls while preserving serial order.

Validation:
- Four backpressure tests pass across three loaded-reproducer runs.
- Full workspace suite passes: 6675/6675 tests.
- Formatting and lint checks pass.

### DIFF
--- a/src/lib/unified_pipeline/bam.rs
+++ b/src/lib/unified_pipeline/bam.rs
@@ -966,6 +966,36 @@ impl<G: Send, P: Send + MemoryEstimate> BamPipelineState<G, P> {
         Some((serial, batch))
     }
 
+    /// Decide whether a boundary batch just popped from Q2b should be decoded
+    /// now or handed back to Q2b under per-serial memory backpressure (fgumi#746).
+    ///
+    /// The Q3 reorder buffer's own [`ReorderBufferState::can_proceed`] always
+    /// admits the serial Group needs next (the gap-filler), even when memory is
+    /// high, and backpressures only *future* serials once the buffer is half
+    /// full. Returns:
+    ///
+    /// - `Some((serial, batch))` — decode it now: either it is the gap-filler,
+    ///   or it is a future batch whose requeue lost the race to a full Q2b (a
+    ///   rare, bounded overshoot — never a lost or reordered-into-loss batch,
+    ///   since Q3 reassembles by serial).
+    /// - `None` — the batch was requeued to Q2b so another worker can still
+    ///   reach the gap-filler, avoiding the "all workers hold a non-`next_seq`
+    ///   batch and nobody can produce `next_seq`" deadlock.
+    fn admit_or_requeue_decode(
+        &self,
+        serial: u64,
+        batch: BoundaryBatch,
+    ) -> Option<(u64, BoundaryBatch)> {
+        if self.q3_reorder_state.can_proceed(serial) {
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #794, opened 2026-08-16
consensus: absent base qualities abort in fgbio but are silently rejected as zero-length in fgumi


nh13:
## Summary

fgumi and fgbio diverge on how they treat an input read with **absent base qualities** (BAM `QUAL` of `*`, encoded as `0xFF` per base). fgbio raises a fatal error and aborts; fgumi silently rejects the read as zero-length-after-trimming and continues. This is a pre-existing behavior difference in the vanilla consensus path (inherited by duplex and codec), surfaced while working #792.

## The two behaviors

**fgbio** — `UmiConsensusCaller.toSourceRead` (`UmiConsensusCaller.scala:267-270`) throws when `rec.quals` is empty:

```scala
val quals = if (rec.quals.nonEmpty) rec.quals.clone() else {
  throw new IllegalArgumentException(
    f"The input read is missing base qualities: ${rec.asSam.getSAMString}"
  )
}
```

**fgumi** — `VanillaUmiConsensusCaller::create_source_read` (`crates/fgumi-consensus/src/vanilla_caller.rs`) returns `None` for absent qualities, which the caller counts as `ZeroLengthAfterTrimming`:

```rust
// Per the BAM spec, absent quality scores are encoded as 0xFF per base. Reject
// such reads rather than treating 0xFF as genuine very-high-quality evidence
// (which would let unqualified reads dominate consensus).
if quals.iter().all(|&q| q == 0xFF) {
    return None;
}
```

So on the same input (a mapped read with `QUAL = *`), fgbio fails the run while fgumi drops the read and produces output — the `--stats` and record counts cannot be compared because one tool never finishes.

## Why it matters

The fgumi rejection is a defensible design choice (a comment explains it: treating `0xFF` as Phred 255 would let unqualified reads dominate a consensus). But it is an undocumented divergence from fgbio, and it is distinct from the genuine zero-length-after-trimming path that #792 reconciles — there fgbio *also* rejects-and-continues (`ZeroPostAfterTrimming`), so the two tools agree.

## Options to consider (not yet decided)

1. **Match fgbio** — error on a mapped read with absent qualities (behavior-breaking; callers relying on the current silent-drop would start failing).
2. **Keep the divergence, document it** — record it as an intentional difference in the parity notes and give absent-qualities its own rejection reason (e.g. `AbsentQualities`) instead of folding it under `ZeroLengthAfterTrimming`, so `--stats` distinguishes the two causes.
[...]

Issue #792, opened 2026-08-16
duplex silently drops zero-length-after-trimming reads from both --stats and --rejects, unlike the simplex path


nh13:
## Summary

In the duplex path, reads that `create_source_read` rejects (zero length after trimming, or absent qualities — it returns `None`) are dropped silently: they are counted in **neither** `--stats` nor written to `--rejects`. The vanilla (simplex) path handles the same `None` return by counting the read as `ZeroLengthAfterTrimming` and writing it to the rejects BAM. So `duplex --stats` under-counts `raw_reads_rejected` by exactly these reads, and they are absent from `--rejects` too.

This was flagged as out-of-scope in #758's "Not changed here" section. Because the reads are missing from both outputs *symmetrically*, they do not break the record-count-equals-`raw_reads_rejected` invariant #758 fixed — but they are still a distinct under-count. Filing separately.

## Mechanism

Duplex builds its `SourceRead` vectors with `filter_map`, which discards `None` returns with no accounting (`crates/fgumi-consensus/src/duplex_caller.rs:2026-2043`):

```rust
let x_sources: Vec<SourceRead> = x_raws
    .iter()
    .enumerate()
    .filter_map(|(i, r)| {
        let mate_clip = bam_fields::num_bases_extending_past_mate_raw(r);
        ss_caller.create_source_read(r, i, mate_clip)   // None -> silently dropped
    })
    .collect();
// ... same for y_sources
```

The vanilla path, by contrast, tracks the dropped indices and both counts and writes them (`crates/fgumi-consensus/src/vanilla_caller.rs:1348-1368`):

```rust
if let Some(sr) = self.create_source_read(raw.as_ref(), idx, mate_clip) {
    source_reads.push(sr);
} else {
    zero_length_indices.push(idx);
}
...
if !zero_length_indices.is_empty() {
    self.stats.record_rejection(RejectionReason::ZeroLengthAfterTrimming, zero_length_indices.len());
    if self.track_rejects {
        for &idx in &zero_length_indices { self.rejected_reads.push(group_reads[idx].to_vec()); }
    }
}
```

## Requested change

In the duplex path, when `create_source_read` returns `None`, record the read as `ZeroLengthAfterTrimming` and, when `--rejects` is set, write the raw record to the rejects buffer — matching the vanilla path. The reads must reach both `--stats` and `--rejects`.

## Acceptance / tests (TDD)

[...]

PR #795, opened 2026-08-16
fix(duplex): split whole-group rejection reasons first-writer-wins

### LINKED ISSUE #791: duplex whole-group rejection re-attributes single-strand MinorityAlignment reads to the duplex reason, so --stats reason rows diverge from fgbio
## Summary

On the duplex whole-group rejection path, reads that the single-strand layer already rejected as `MinorityAlignment` are re-attributed entirely to the duplex whole-group reason (e.g. `InsufficientReads`, `PotentialCollision`, `FragmentRead`). fgbio instead keeps the single-strand reason on those records (first-writer-wins) and applies the whole-group reason only to the survivors. The **totals** agree between the two tools — one record, one count — but the per-reason split in `--stats` does not, so the reason rows are not comparable on this path.

This was called out as out-of-scope in #758's "Not changed here" section, which pointed at #749 as its home. #749 is now closed (COMPLETED) and was scoped to *codec* attribution *ordering*, not the duplex whole-group split, so this work has no open home. Filing it as its own issue.

## Mechanism

`DuplexConsensusCaller::process_group` returns `DuplexGroupOutcome::RejectedWholeGroup(raw_records)` when the whole group is rejected. At the drain site in `consensus_reads` (`crates/fgumi-consensus/src/duplex_caller.rs:2524-2536`):

```rust
let ss_stats = self.ss_caller.take_statistics();
let ss_rejected_raw = self.ss_caller.take_rejected_reads();
match duplex_outcome {
    DuplexGroupOutcome::Kept => {
        self.stats.merge(&ss_stats);                       // ss reasons kept
        ...
    }
    DuplexGroupOutcome::RejectedWholeGroup(duplex_rejected_raw) => {
        if self.track_rejects {
            self.rejected_reads.extend(duplex_rejected_raw); // ss_stats DROPPED
        }
    }
}
```

On the whole-group path `ss_stats` is discarded entirely (deliberately, to keep the reject-count total correct), and the group's rejection was recorded under a single duplex reason by `record_rejection` at the whole-group sites (`duplex_caller.rs:1947`, `1984`, `1996`, `2296`, `2460`, plus the `FragmentRead` site at `2481`). So a read the sub-caller had already charged to `MinorityAlignment` is counted only under the duplex reason.

fgbio's `rejectRecords` is first-writer-wins (`recs.filterNot(_.contains(RejectReasonTag))`), so a record already tagged `MinorityAlignment` keeps that reason and the whole-group reason lands only on the survivors — see the comment at `DuplexConsensusCaller.scala:359-362`.

## Requested change

Preserve the single-strand rejection reason for records the sub-caller already rejected, and attribute only the remaining (survivor) records to the duplex whole-group reason, matching fgbio's first-writer-wins split. Do not change any accept/reject decision or the reject-count total — only which reason bucket each already-rejected read lands in.

This touches all five whole-group rejection sites and their reject payloads. The reason breakdown in `--stats` on the whole-group path should then match fgbio's for the same input.

## Acceptance / tests (TDD)

- A molecule that (a) drops one or more reads at the single-strand `MinorityAlignment` layer and (b) is then rejected as a whole group must, in `--stats`, report those reads under `MinorityAlignment` and the survivors under the whole-group reason — not the whole group under the duplex reason.
- The `raw_reads_rejected` total and the rejects-BAM record count must remain equal (the invariant #758 established must not regress).
- Compare the per-reason `--stats` split against fgbio on an input exercising this path.

nh13:
Closes #791.

## The bug

On the duplex whole-group rejection path, reads the single-strand layer had already rejected as `MinorityAlignment` were re-attributed *entirely* to the duplex whole-group reason (e.g. `InsufficientSupport`). fgbio instead keeps the single-strand reason on those records (first-writer-wins) and applies the whole-group reason only to the survivors. The reject-count **total** agreed either way — which is why #758's reconciliation held — but the per-reason split in `--stats` did not, so the reason rows were not comparable to fgbio on this path.

Mechanism: when the duplex layer rejects the whole group downstream of the single-strand filter, `process_group` counts the entire raw group (`N` reads) under one duplex reason, and the drain site dropped the single-strand caller's stats delta (the `M` `MinorityAlignment` reads) to avoid double-counting the total. So those `M` reads were counted only under the duplex reason, never under `MinorityAlignment`.

## The fix

The rejects BAM carries no per-record reason (records are byte-for-byte identical to their input, as #758's tests assert), so this is purely a `--stats` attribution fix — handled at the drain site, not the five whole-group sites.

New `reattribute_single_strand_rejections` moves the `M` single-strand-rejected reads out of the whole-group reason bucket and into their own (`MinorityAlignment`/`Unmapped`), matching fgbio's first-writer-wins. It is a **reason move, not a total change**: `filtered_reads` is decremented by `M` then the single-strand stats are folded back in, so the net total is unchanged and `--rejects` still reconciles with `raw_reads_rejected`. Pre-filter rejections leave the single-strand stats empty, so it is a no-op for them.

## Parity with fgbio

fgbio's `rejectRecords` is first-writer-wins (`recs.filterNot(_.contains(RejectReasonTag))`): a record already tagged `MinorityAlignment` in `filterToMostCommonAlignment` keeps that reason, and the whole-group `rejectRecords(groups.flatten, …)` lands only on the survivors (`DuplexConsensusCaller.scala:359-362`). This change makes the fgumi `--stats` split match.

## Tests

TDD; the unit assertions were confirmed to fail (`MinorityAlignment` `0` vs `4`) against the pre-fix code before implementing.

- `test_whole_group_rejection_preserves_single_strand_reasons` (unit, was `test_single_strand_rejections_are_not_double_counted_on_whole_group_rejection`) — a molecule met by `min_reads = 4` before the single-strand filter and missed after it, with a control run pinning the fixture is not vacuous. Now asserts the four minority reads keep `MinorityAlignment`, the twelve survivors carry `InsufficientReads`, the total is still every input read once, and the rejects buffer reconciles. The pre-existing no-double-count invariant is retained.
- `test_whole_group_rejection_counts_are_independent_of_rejects_tracking` (unchanged) — still passes: the reattribution reads the single-strand stats, which are independent of `--rejects` tracking, so the per-reason breakdown is identical with and without it.
- `test_duplex_whole_group_rejection_splits_stats_reasons` (integration, single-threaded and `--threads 2`) — the split in the actual `--stats` file: `raw_reads_rejected_for_minority_alignment = 4`, `raw_reads_rejected_for_insufficient_support = 12`, `raw_reads_rejected = 16`, and the rejects BAM record count equals `raw_reads_rejected`.

## Not changed here

The other "Not changed here" items from #758 (the `create_source_read` zero-length drop, tracked in #792/PR #793; the absent-qualities divergence in #794) are separate and untouched.

## Gate

`cargo ci-fmt`, `cargo ci-lint`, `cargo ci-test` all clean.

Risk: command output changes only in `--stats` reason attribution; grouping, consensus, sort order, corrected UMIs, and rejection totals are unchanged; no `unsafe` changes or CLAUDE.md allowlist update; no memory-bound, queue-capacity, or thread/backpressure changes.

Fix: Preserve first-writer-wins attribution for single-strand rejection reasons. Apply the duplex whole-group reason only to surviving reads.

- Track `ZeroLengthAfterTrimming` reads and retain them in rejects output order.
- Preserve rejection totals and `raw_reads_rejected` reconciliation.
- Add single-threaded and multithreaded integration coverage for mixed rejection reasons, ordering, deduplication, output counts, and `--stats`.

### DIFF
--- a/crates/fgumi-consensus/src/duplex_caller.rs
+++ b/crates/fgumi-consensus/src/duplex_caller.rs
@@ -1884,6 +1884,53 @@ impl DuplexConsensusCaller {
         );
     }
 
+    /// Reattributes single-strand rejections out of a whole-group reason, matching fgbio's
+    /// first-writer-wins reason assignment (#791).
+    ///
+    /// On a whole-group rejection downstream of the single-strand filter, `process_group` counted
+    /// the entire raw group (`N` reads) under one duplex reason, and `stats` has already had that
+    /// `group_stats` merged in. But the `M` reads the single-strand layer already rejected
+    /// (`MinorityAlignment`, `Unmapped`, …) should keep *their* reason; only the `N - M` survivors
+    /// belong to the whole-group reason. This moves those `M` reads from the whole-group reason
+    /// bucket into the single-strand reasons, leaving the reject-count *total* unchanged — so
+    /// `--rejects` still reconciles with `raw_reads_rejected` — while the per-reason split now
+    /// matches fgbio.
+    ///
+    /// `group_stats` holds exactly the one whole-group reason on this path (pre-filter rejections
+    /// leave `ss_stats` empty, so this is a no-op for them). The single-strand rejections are a
+    /// subset of the group, so `M <= N`.
+    fn reattribute_single_strand_rejections(
+        stats: &mut ConsensusCallingStats,
+        group_stats: &ConsensusCallingStats,
+        ss_stats: &ConsensusCallingStats,
+    ) {
+        let ss_rejected = ss_stats.filtered_reads;
+        if ss_rejected == 0 {
+            return;
+        }
+        debug_assert_eq!(
+            group_stats.rejection_reasons.len(),
+            1,
+            "a whole-group rejection records exactly one duplex reason"
+        );
+        if let Some((&whole_group_reason, &whole_group_count)) =
+            group_stats.rejection_reasons.iter().next()
+        {
+            debug_assert!(
+                ss_rejected <= whole_group_count,
+                "single-strand rejections are a subset of the whole group"
+            );
+            if let Some(count) = stats.rejection_reasons.get_mut(&whole_group_reason) {
+                *count = count.saturating_sub(ss_rejected);
+            }
+            stats.filtered_reads = stats.filtered_reads.saturating_sub(ss_rejected);
+        }
+        // Fold the single-strand reasons back in: this re-adds the `M` reads under
+        // `MinorityAlignment`/`Unmapped` and restores `filtered_reads`, so the net effect is a
+        // reason move, not a change to the total.
+        stats.merge(ss_stats);
+    }
+
     /// Processes a single UMI group to generate duplex consensus
     #[expect(clippy::too_many_arguments, reason = "group processing requires many parameters")]
     #[expect(
@@ -2544,9 +2591,13 @@ impl ConsensusCaller for DuplexConsensusCaller {
         //
         // When the duplex layer rejects the group it counts and returns *every* raw input
         // record, so the single-strand layer's rejections — a subset of those same records —
-        // are already covered. Adding them on top would report more rejected reads than the
-        // rejects BAM holds. Otherwise the group survived, and the single-strand drops are
-        // the only rejections it produced: they belong in both outputs.
+        // are already covered in the reject *count*. But fgbio attributes reasons
+        // first-writer-wins: a read the single-strand filter already rejected keeps its own
+        // reason, and only the survivors carry the whole-group reason. The whole-group site
+        // counted the entire raw group under one duplex reason, so reattribute the
+        // single-strand-rejected reads from that reason to their own. Otherwise the group
+        // survived, and the single-strand drops are the only rejections it produced: they belong
+        // in both outputs.
         let ss_stats = self.ss_caller.take_statistics();
         let ss_rejected_raw = self.ss_caller.take_rejected_reads();
         match duplex_outcome {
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

nh13:
Rebased onto #793 (`792/…zero-length-rejects`) so this branch carries both fixes, and added a composition test: `test_whole_group_rejection_splits_all_single_strand_reasons` (unit) pins that a whole-group rejection carrying **both** minority-alignment and zero-length reads splits all three reasons first-writer-wins (minority → `MinorityAlignment`, zero-length → `ZeroLengthAfterTrimming`, survivors → `InsufficientReads`). Base is now #793; it will need re-targeting to `main` once #793 merges.

nh13:
@coderabbitai review

nh13:
@coderabbitai full review

nh13:
@coderabbitai review

nh13:
Rebased onto the updated base (#793) to resolve the merge conflict — the stale duplicate of the "count and route zero length" commit is dropped, so the PR now contains only the whole-group rejection-reason split.

On the review findings:
- **\`process_subgroup\` helper reuse** (\`vanilla_caller.rs\`): resolved — \`process_subgroup\` now calls \`record_zero_length_after_trimming\` rather than duplicating the reject-count logic, inherited from the #793 base.
- The outside-diff findings on \`crates/fgumi-bgzf/src/reader.rs\` and the \`test_dedup_command.rs\` oracle are not in this PR's diff (this branch touches only the duplex/vanilla callers and \`test_duplex_command.rs\`); they belong to #798 and #793 respectively, where they are addressed.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #796, opened 2026-08-16
fix(consensus)!: error on reads with absent base qualities

### LINKED ISSUE #794: consensus: absent base qualities abort in fgbio but are silently rejected as zero-length in fgumi

nh13:
Closes #794.

> Stacked on #793 (`792/…zero-length-rejects`), since it changes the `create_source_read` signature that #793's duplex loop consumes. Base will need re-targeting to `main` once #793 merges.

## The divergence

A mapped read that reaches a quality-weighted consensus caller with **absent base qualities** (BAM `QUAL` of `*`, encoded as `0xFF` per base) is structurally invalid input — it cannot be weighted by the model. fgbio treats it as a hard error: `UmiConsensusCaller.toSourceRead` throws `IllegalArgumentException("The input read is missing base qualities")` (`UmiConsensusCaller.scala:267-270`). fgumi instead dropped the read **silently**, folding it under `ZeroLengthAfterTrimming`.

Silent dropping is the dangerous behavior: if an upstream tool strips `QUAL` from a lane, fgumi quietly degrades every affected molecule's depth — worst case for duplex/cfDNA rare-variant work — with no signal to the user. The right classification is *invalid input → abort*, not *biological filter → reject-and-count*.

## The fix

`create_source_read` now returns `Result<Option<SourceRead>>`:

- **`Err`** — absent qualities (all `0xFF`) or a quality string whose length does not match the sequence. Structurally invalid; the run aborts with a message naming the read.
- **`Ok(None)`** — the read trims to zero length (`ZeroPostAfterTrimming`). A legitimate biological filter, unchanged.
- **`Ok(Some)`** — a usable source read.

The two production callers — the vanilla `process_subgroup` and the duplex `process_group` X/Y source construction — propagate the error with `?`. Codec is unaffected: it builds source reads through its own `toSourceReadForCodec`, not this path.

Per the discussion on the issue, this is not gated behind a flag — absent qualities always error, matching fgbio.

## Tests

TDD; the driving test was confirmed to fail (`consensus_reads` returned `Ok`, silently dropping the read) before the fix.

- `test_absent_base_qualities_abort_consensus` (unit, vanilla) — a molecule with an absent-quality read must make `consensus_reads` return `Err`, driving the whole path rather than just `create_source_read`.
- `test_absent_base_qualities_abort_duplex_consensus` (unit, duplex) — the same through the duplex path, where `create_source_read` is reached via the single-strand caller.
- `test_reads_without_base_qualities` (unit) — the fgbio-port test, whose name always said "Exception when reads lack base qualities", now asserts `create_source_read` returns `Err` rather than `None`.
- The zero-length-after-trimming path (`Ok(None)`) is unchanged and still verified by #793's tests.

## Gate

`cargo ci-fmt`, `cargo ci-lint`, `cargo ci-test` all clean.

Risk: consensus output changes for reads with absent or invalid qualities, pinned by fgbio-compatible tests; unsafe changes none; memory, queue, and backpressure changes none.

- `create_source_read` now errors on absent or length-mismatched qualities.
- Zero-length records and reads trimmed to zero length remain valid rejection paths.
- Vanilla and duplex callers propagate errors and record zero-length rejects.
- Tests cover vanilla, duplex, fgbio compatibility, and legal zero-length records.


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
Addressed the outside-diff findings from the latest review (rebased onto the updated `792/...` base, so this is now a single commit):

- **Legal zero-length `SEQ=*` records** (`vanilla_caller.rs` `create_source_read`): now returns `Ok(None)` for `read_len == 0` before the quality checks, so an empty record routes as `ZeroLengthAfterTrimming` instead of tripping the length-mismatch bail (which would abort `--allow-unmapped` runs). Added a regression test.
- **Test-only `to_source_read_from_record`**: added the same all-`0xFF` absent-qualities guard as the production path so the bridge cannot treat absent-quality reads as high-quality evidence.
- **Stale contract comments**: updated `record_zero_length_after_trimming` (`vanilla_caller.rs`) and the X-partition source-read loop (`duplex_caller.rs`) to state `create_source_read` returns `None` only for reads that trim to zero length; absent qualities now return `Err`.

nh13:
@coderabbitai review

nh13:
Rebased onto the updated base (#793) to resolve the merge conflict — the stale duplicate of the "count and route zero length" commit is dropped, leaving just the absent-base-qualities change.

The rebase preserves the intended routing: \`create_source_read\` returns \`Err\` for absent base qualities (propagated via \`?\` in \`process_subgroup\`, so the run aborts) and \`Ok(None)\` for zero-length-after-trimming (counted through the inherited \`record_zero_length_after_trimming\` helper). Both paths stay covered by \`test_absent_base_qualities_abort_consensus\` and \`test_zero_length_record_is_not_treated_as_absent_qualities\`.

The outside-diff \`crates/fgumi-bgzf/src/reader.rs\` findings and the \`test_dedup_command.rs\` oracle are not in this PR's diff; they belong to #798 and #793, where they are addressed.

nh13:
@coderabbitai review

Issue #786, opened 2026-08-16
Tracker: adopt dupblaster IO/dedup optimizations into fgumi


nh13:
# Tracker: adopt dupblaster IO/dedup optimizations into fgumi

## Context

[dupblaster](https://github.com/fulcrumgenomics/dupblaster) and `fgumi dedup` solve different problems: dupblaster is a streaming, sort-free samblaster successor (coordinate-signature marking, no UMI), while `fgumi dedup` is a template-coordinate-sort-dependent, UMI-aware marker with base-quality representative scoring. They already share the `fgumi-raw-bam` crate. A review of dupblaster surfaced a handful of IO/serialization mechanics and one metrics gap worth adopting into fgumi. This tracks that work.

**A second deep-dive scan confirmed most other dupblaster cleverness is already present in fgumi** (fat LTO + `codegen-units=1`, single-pass CIGAR length/clip helpers, serde-struct TSV output, and — importantly — the family-size histogram, which `fgumi dedup` already emits via `--family-size-histogram` using the same `FamilySizeMetrics` type as `fgumi group`). Two candidate items were dropped for the same reason a sort-free marker was dropped: they are dupblaster's job, not fgumi's.

**Working principle: benchmark-gate every performance change.** Each perf-touching change is implemented on its own branch, measured against baseline, and only merged if it demonstrably wins; a change that does not win is recorded as a negative result and dropped. Correctness-affecting changes are gated on parity instead (see Validation).

### In scope

- [ ] **1. Stop zero-filling the record body in `fgumi-raw-bam::read_raw_record`.** `read_raw_record` (`crates/fgumi-raw-bam/src/raw_bam_record.rs`) does `record.0.resize(block_size, 0)` and then fully overwrites those bytes with `read_exact` — gigabytes of wasted zero-stores on large inputs. Replace with `clear()` + read into spare capacity (optionally a `MaybeUninit` spare-capacity read to skip the fill entirely). Behavior-neutral; lands in the crate dupblaster also depends on, so both tools benefit. Prove: read-throughput microbenchmark.

[...]

PR #798, opened 2026-08-17
perf(bgzf): skip BGZF CRC verification on trusted stdin input by default (#786)

nh13:
Part of #786 (dupblaster→fgumi optimizations).

BGZF decode verifies each block's CRC32. For a freshly-piped aligner stream (`bwa-mem3 ... | fgumi ... -i /dev/stdin`) that check is redundant — any corruption there is an upstream bug, not data at rest — while a file on disk may have been archived, transferred, or copied, which is exactly what CRC32 exists to catch. Following dupblaster's policy, CRC verification now defaults **on for file input, off for stdin**, with `--check-crc` / `--no-check-crc` explicit overrides (mutually exclusive; explicit always wins).

This is a default-behavior change, so it is **non-silent**: every run logs a `CRC verify: on/off …` line stating what actually happened. The flags are surfaced on the shared `BamIoOptions`, so every pipeline-backed command inherits them; `verify_crc` is set in one place (`build_pipeline_config`) so a command can't log one setting and apply another. The `--check-crc`/`--no-check-crc` help spells out the commands/modes that don't honor the flag: `downsample` never does, and `clip`/`codec`/`duplex`/`simplex`/`correct`/`group` only do with `--threads N` (their single-threaded fast path decodes via noodles-bgzf, which always verifies and has no skip knob).

## Benchmark

Added `benches/stored_block_decode.rs` (c8g.4xlarge): skip 252 µs vs verify 441 µs, ~43% on near-incompressible synthetic blocks — an **upper bound** (the CRC delta implies hardware CRC is active; on real BGZF that does actual LZ/Huffman work the win is smaller, on the order of the ~6–10% of decode CRC represents per #686). Direction is conclusive. Output is byte-identical by construction (verified by the crate's unit/Miri/roundtrip tests).

## Scope

`fgumi-sort` and `extract` have their own reader stacks and are **not** wired here — filed as separate follow-ups.

## Verification

Full workspace suite green; new `effective_check_crc` truth-table unit test + corrupted-CRC end-to-end integration tests (default rejects a corrupted file; `--no-check-crc` accepts it).

Risk: output changes none; `unsafe` changes none and `CLAUDE.md` allowlist changes none; memory, queue, and thread/backpressure policy changes none.

Fix: CRC verification is enabled for file input and disabled for stdin by default. Use `--check-crc` or `--no-check-crc` to override the default. Explicit flags take precedence.

- Added configurable CRC verification to BGZF and pipeline readers.
- Preserved decompressed-size validation when CRC verification is disabled.
- Added effective-setting logs and truth-table coverage.
- Added direct slice-API coverage and corrupted-CRC integration coverage for dedup.
- `downsample` always verifies CRC.
- `fgumi-sort` and `extract` remain out of scope.
- A benchmark reports approximately 43% faster decoding when CRC verification is skipped on synthetic near-incompressible blocks.


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
Addressed the outside-diff finding on `crates/fgumi-bgzf/src/reader.rs`: `decompress_block_into` and `decompress_block_slice_into` now delegate to their `_opts` counterparts with `verify_crc = true`, so the EOF/empty and short-input/`uncompressed_size == 0` guards live in a single body per sibling pair. Also strengthened `test_dedup_no_check_crc_accepts_corrupted_crc_on_file_input` to a record-identity oracle: it now dedups an uncorrupted copy and compares the decoded records element-for-element against the corrupted-input run, replacing the count-only check.

nh13:
@coderabbitai review

nh13:
Both outside-diff-range findings on \`crates/fgumi-bgzf/src/reader.rs\` are addressed:

- **Delegate non-\`_opts\` entry points:** already in place — \`decompress_block_into\` / \`decompress_block_slice_into\` delegate to their \`_opts\` siblings with \`verify_crc = true\`, so the EOF/empty and short-input/\`uncompressed_size == 0\` guards live in a single body each.
- **Direct \`decompress_block_slice_into_opts\` coverage:** added \`decompress_slice_opts_skips_crc_but_still_checks_size\`, the slice-API twin of the existing \`RawBgzfBlock\` test — asserts \`verify_crc = false\` skips the CRC32 compare (corrupted CRC still decodes to the original bytes) but still catches a corrupted-ISIZE size mismatch, pinning the flag-forwarding through \`decompress_and_verify\`. Folded into the feat commit.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #801, opened 2026-08-17
feat(bam-io): unify the single-threaded raw-BAM reader onto fgumi-bgzf (#800)

### LINKED ISSUE #800: Unify the raw-BAM reader onto fgumi-bgzf (retire the second noodles-bgzf decode path)
## Summary

`fgumi-bam-io` has two BGZF decoder stacks, and the single-threaded / non-pipeline read paths use the slower, less-capable one.

- **noodles-bgzf** (`BgzfReaderEnum`, `crates/fgumi-bam-io/src/reader.rs`): used by `create_raw_bam_reader` / `create_raw_bam_reader_with_opts` and by BAM-header parsing. This is what the single-threaded fast paths of `clip`/`codec`/`duplex`/`simplex`/`correct`/`group`, plus `downsample` (always), `fastq`, and `merge` decode through.
- **fgumi-bgzf** (the in-house crate): used only by the multi-threaded unified pipeline decode sites.

noodles-bgzf is correct, so this is not a correctness bug — it's a missed-optimization / inconsistency:

1. **CRC policy can't reach those paths.** #798 added a `--check-crc`/`--no-check-crc` policy backed by `fgumi-bgzf`'s CRC-skip. noodles-bgzf has no knob to disable CRC, so the single-threaded / downsample modes always verify regardless of the flag. This is also why `PipelineReaderOpts.verify_crc` is currently a pure-carrier field (set but never consumed on these paths).
2. **They miss fgumi-bgzf's performance work** — the stored-block fast path and the raw-byte decode mode (the code notes ~15% CPU vs the noodles decode/encode roundtrip).
3. **Two decoders to reason about** — divergent behavior/logging between single-threaded and `--threads N` modes for the same command.

## Proposed change

Route the raw-BAM reader path through `fgumi-bgzf` instead of noodles-bgzf: rework `open_bgzf_reader` / `create_raw_bam_reader[_with_opts]` / `RawBamReader` (and the header-parse step) to decode with `fgumi-bgzf`, honoring `PipelineReaderOpts.verify_crc` (making that field live). Once unified, `clip`/`codec`/`duplex`/`simplex`/`correct`/`group` single-threaded modes and `downsample` honor `--check-crc`/`--no-check-crc`, and the per-command "only with `--threads`" help caveat added in #798 can be dropped.

## Acceptance / validation

- Byte-identical decode vs the current noodles path (roundtrip + equivalence tests).
- `--no-check-crc` honored (and a corrupted-CRC file rejected by default) in a single-threaded command, not just the pipeline.
- Full workspace suite green; benchmark the decode change (expect the stored-block/raw-byte win on the affected commands).

## Out of scope / follow-ups

- `fgumi-sort` and `extract` have their own separate reader stacks; wire those in their own issues if not covered here.

Relates to #786, #798.

nh13:
Closes #800. Follow-up to #798.

> **Stacked on #798** (`786/nhomer/perf-bgzf-crc-skip`) — it uses that PR's `decompress_block_into_opts`. Review/merge #798 first; this PR's base will retarget to `main` once #798 lands.

## Problem

`fgumi-bam-io` had two BGZF decoder stacks. The single-threaded raw-BAM reader (`create_raw_bam_reader[_with_opts]`) decoded via **noodles-bgzf**, which always verifies CRC32 and has no skip knob — so the `--check-crc`/`--no-check-crc` policy from #798 (backed by our own `fgumi-bgzf`) never reached the single-threaded fast paths, and `PipelineReaderOpts.verify_crc` was a dead carrier field.

## Change

Add `FgumiBgzfReader`, a safe streaming `Read`/`BufRead` adapter over `fgumi-bgzf`'s `read_raw_blocks` + `decompress_block_into_opts` (no `unsafe`; one block decoded per refill, so a CRC/size error is tied to the bytes the caller actually reads), and route the single-threaded raw reader through it via a new `BgzfReaderEnum::Fgumi` arm. `PipelineReaderOpts.verify_crc` is now live.

Coverage after this change (`--check-crc`/`--no-check-crc`):
- **`clip`/`codec`/`duplex`/`simplex`/`downsample`** — honored single- or multi-threaded (their single-threaded reader now decodes through fgumi-bgzf).
- **`correct`/`group`** — honored only with `--threads N`; their single-threaded mode reads through `create_bam_reader_for_pipeline_with_opts` + a manual noodles decode, so they keep the noodles path (help text says so).
- `threads > 1` still uses noodles' multithreaded decoder (a threaded fgumi-bgzf raw reader is out of scope).

`fgumi-sort` and `extract` have their own reader stacks and remain follow-ups.

## Benchmark (`benches/bgzf_decode.rs`, same BGZF payload, single-threaded)

| decoder | time | thrpt |
| --- | --- | --- |
| fgumi verify_on (file default) | 70.5 ms | ~454 MiB/s |
| fgumi verify_off (stdin default / CRC-skip) | 67.6 ms | ~474 MiB/s |
| noodles (always verifies) | 68.9 ms | ~465 MiB/s |

fgumi-bgzf is at parity with noodles single-threaded (verify_on ~2% slower but within overlapping CIs; verify_off ~2% faster). The deliverable is bringing the CRC-skip policy to these paths, not a raw decode speedup.

## Tests

TDD: the reader-level `verify_crc` test (corrupted-CRC file accepted with `verify_crc=false`, rejected with `true`) was written first and confirmed red against the noodles path before implementing. Plus a multi-block roundtrip test (exercises block-boundary buffering) and three end-to-end `downsample` CRC tests (`--no-check-crc` accepts a corrupted file; default and `--check-crc` reject it). Full workspace suite green (2546).

Risk: output changes are limited to CRC acceptance or rejection and are pinned by CRC, round-trip, multi-block, partial-read, and downsample record-identity tests; no `unsafe` changes and no CLAUDE.md allowlist update; no memory-bound, queue-capacity, or thread/backpressure policy changes.

Fix: Route single-threaded raw-BAM decoding through `FgumiBgzfReader` and honor `PipelineReaderOpts.verify_crc`.

- Add configurable CRC verification for single-threaded `clip`, `codec`, `duplex`, `simplex`, and `downsample`.
- Preserve existing `correct` and `group` paths and all multithreaded decoding.
- Add BGZF buffering, lazy decoding, partial-read support, sticky decode failures, and EOF-marker handling.
- Update BGZF decompression to verify non-EOF zero-size blocks when CRC checking is enabled.
- Add CRC, multi-block, partial-read, corrupted-EOF, and downsample integration tests.
- Add the `bgzf_decode` benchmark.

### DIFF
--- a/crates/fgumi-bam-io/Cargo.toml
+++ b/crates/fgumi-bam-io/Cargo.toml
@@ -35,3 +35,7 @@ tempfile = { workspace = true }
 rstest = { workspace = true }
 proptest = { workspace = true }
 criterion = { workspace = true }
+
+[[bench]]
+name = "bgzf_decode"
+harness = false
--- a/crates/fgumi-bam-io/benches/bgzf_decode.rs
+++ b/crates/fgumi-bam-io/benches/bgzf_decode.rs
@@ -1,0 +1,82 @@
+//! Decode-throughput benchmark for the single-threaded raw-BAM reader unify
+//! (#800): the fgumi-bgzf streaming decoder (`FgumiBgzfReader`) — with CRC32
+//! verification on and off — versus noodles' single-threaded BGZF reader, all
+//! decoding the *same* BGZF byte buffer. This isolates the decode-path change
+//! the unify makes on the single-threaded fast paths.
+
+use std::hint::black_box;
+use std::io::{Cursor, Read};
+
+use criterion::{Criterion, Throughput, criterion_group, criterion_main};
+use fgumi_bam_io::FgumiBgzfReader;
+use fgumi_bgzf::InlineBgzfCompressor;
+
+/// Build a moderately compressible payload of `len` bytes (BAM-like: a mix of a
+/// small alphabet so blocks actually compress, deterministic so runs compare).
+fn make_payload(len: usize) -> Vec<u8> {
+    let mut out = Vec::with_capacity(len);
+    let mut state: u32 = 0x1234_5678;
+    while out.len() < len {
+        // xorshift for cheap deterministic pseudo-randomness
+        state ^= state << 13;
+        state ^= state >> 17;
+        state ^= state << 5;
+        // Fold into a small-ish alphabet so the deflate stage has real work.
+        out.push((state & 0x3F) as u8 + b' ');
+    }
+    out.truncate(len);
+    out
+}
+
+/// Compress `payload` into a complete BGZF byte stream (blocks + EOF marker).
+fn make_bgzf(payload: &[u8], level: u32) -> Vec<u8> {
+    let mut compressor = InlineBgzfCompressor::new(level);
+    compressor.write_all(payload).expect("compress payload");
+    compressor.flush().expect("flush compressor");
+    let mut out = Vec::new();
+    compressor.write_blocks_to(&mut out).expect("write blocks");
+    out.extend_from_slice(&fgumi_bgzf::BGZF_EOF);
+    out
+}
+
+fn bench_bgzf_decode(c: &mut Criterion) {
+    let payload = make_payload(32 * 1024 * 1024); // 32 MiB uncompressed
+    let bgzf = make_bgzf(&payload, 6);
+
+    let mut group = c.benchmark_group("bgzf_decode");
+    group.throughput(Throughput::Bytes(payload.len() as u64));
+
+    let mut scratch = Vec::with_capacity(payload.len());
+
+    group.bench_function("fgumi_verify_on", |b| {
+        b.iter(|| {
+            scratch.clear();
+            let mut reader = FgumiBgzfReader::new(Box::new(Cursor::new(bgzf.clone())), true);
+            reader.read_to_end(&mut scratch).expect("decode");
+            black_box(scratch.len());
+        });
+    });
+
+    group.bench_function("fgumi_verify_off", |b| {
+        b.iter(|| {
+            scratch.clear();
+            let mut reader = FgumiBgzfReader::new(Box::new(Cursor::new(bgzf.clone())), false);
+            reader.read_to_end(&mut scratch).expect("decode");
+            black_box(scratch.len());
+        });
+    });
+
+    group.bench_function("noodles", |b| {
+        b.iter(|| {
+            scratch.clear();
+            let mut reader = noodles_bgzf::io::Reader::new(Cursor::new(bgzf.clone()));
+            reader.read_to_end(&mut scratch).expect("decode");
+            black_box(scratch.len());
+        });
+    });
+
+    group.finish();
+}
+
+criterion_group!(benches, bench_bgzf_decode);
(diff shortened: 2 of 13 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed all four review findings:

- **Honor `--no-check-crc` during pipeline header parsing (Major):** the header parse in `read_header_and_replay` runs through noodles' `BgzfReader`, which has no CRC-skip knob, and threading fgumi-bgzf in would break the exact-byte `TeeReader` replay (its readahead batches blocks). Took the documented-limitation option the review offered: documented on `read_header_and_replay` and the `--check-crc`/`--no-check-crc` help that the **header block is always CRC-verified** and `--no-check-crc` skips the record body only, and added `test_pipeline_reader_always_verifies_the_header_block`, which corrupts block 0 and asserts the pipeline reader rejects it even with `verify_crc: false`.
- **Verify decoded record identity, not only count (Minor):** the downsample `--no-check-crc` test now collects `io::Result<Vec<_>>` (surfacing decode errors) and pins the complete `read_0..read_2999` name set, so drops-and-replacements, duplicates, or error items can no longer satisfy it.
- **Buffer the async source before framing BGZF blocks (outside-diff, Major):** `open_bgzf_reader` now wraps **both** branches in `BufReader::with_capacity(BGZF_INPUT_BUFFER_SIZE, …)`, keeping the `PrefetchReader` inside — the frame reader issues 18-byte header + body reads, so the unbuffered async path paid per-read (plus cross-thread handoff) overhead. Corrected the stale doc rationale.
- **Verify CRC for non-EOF zero-size blocks (outside-diff, Major):** `decompress_block_into_opts` / `decompress_block_slice_into_opts` now skip only the **exact** `BGZF_EOF` marker; every other zero-size block (e.g. a CRC-corrupted EOF whose bytes no longer match) runs through `decompress_and_verify`, so `verify_crc: true` catches it. Added `decompress_opts_verifies_a_corrupted_eof_marker` covering both APIs. (The non-opts entry points delegate to these, so they inherit the fix.)

Note: the previous run's `test` failure was a **flaky** hit on the pre-existing `queue_bytes_in_flight` checked-add overflow (bam.rs:1012, "attempt to add with overflow") — the exact bug #811 fixes with saturating debits; this branch is stacked on #798, not #811, so it inherits it. It reproduced 0/1 locally and every other job passed; `sort-correctness` was again the ~6h "Install samtools" apt stall (cancelled, not failed — #825 hardens it). This push re-triggers CI. Full local gate green (fmt / clippy-pedantic / 7712 tests).

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #805, opened 2026-08-17
feat(bam-io): honor --check-crc on correct/group single-threaded paths (#800)

nh13:
Part of #800. Stacked on #801 (base branch `800/nhomer/unify-raw-reader-fgumi-bgzf`).

#801 unified the single-threaded *raw* reader (`clip`/`codec`/`duplex`/`simplex`/`downsample`) onto fgumi-bgzf so `--check-crc`/`--no-check-crc` took effect there. This PR closes the two remaining single-threaded gaps `#800` called out: the `correct` and `group` fast paths, which decoded through noodles' always-verify BGZF reader and so silently ignored `--no-check-crc`.

## What changed

- **`fgumi-bam-io`**: new `create_raw_bam_reader_from_stream_with_opts`, which decodes an already-opened, still-compressed BGZF byte stream through `FgumiBgzfReader` (honoring `verify_crc`) and wraps it in a `RawBamReader`. This is the stream `correct`/`group` open once up front (required for stdin, which cannot be re-opened). The header-parse-and-wrap tail is factored into a shared `finish_raw_bam_reader` used by both this and the from-path `create_raw_bam_reader_with_opts`.
- **`correct` / `group`**: single-threaded fast paths now decode via the new factory instead of a hand-rolled `noodles::bam::io::Reader` + header skip, threading `verify_crc` through from `pipeline_reader_opts()`.
- **`common.rs`**: removed `log_effective_check_crc_for_fast_path` — its only purpose was to log that the single-threaded path always verifies, which is no longer true for its only two callers. `correct`/`group` now log the real effective policy. Updated the `--check-crc` help text: honored on every command's input decode in both modes, with the single remaining exception of `clip --threads N`.

## Coverage after this PR

| command | single-threaded | multi-threaded |
| --- | --- | --- |
| `downsample`, `codec`, `simplex`, `duplex` | ✅ | ✅ (input decode is always single-threaded) |
| `correct`, `group` | ✅ (this PR) | ✅ (unified pipeline, #798) |
| `clip` | ✅ | ❌ still verifies (`--threads N` uses noodles' multithreaded reader) |

The remaining `clip --threads N` gap — and the multithreaded raw reader generally — is the last piece of #800 and is tracked as a follow-up.

## Tests

- `fgumi-bam-io`: `test_raw_reader_from_stream_honors_verify_crc` — corrupted CRC32 rejected with verify on, accepted with it off (the case that could not pass against noodles).
- `correct` / `group`: `--no-check-crc` accepts a CRC-corrupted file; default and `--check-crc` reject it — mirroring the existing `downsample` trio.

Full suite green (2552 passed); fmt / clippy / tag-literals / publish-order clean.

Risk: command output changes—none; `unsafe` changes—none, and `CLAUDE.md` needs no allowlist update; memory bounds, queue capacity, and thread/backpressure policy changes—none.

- Added stream-based BAM reading with configurable `verify_crc`.
- Applied CRC options to single-threaded `correct` and `group` paths.
- Shared header parsing between reader paths.
- Removed obsolete fast-path CRC logging and updated help text.
- Added corrupted-BGZF CRC tests for `correct`, `group`, and `clip --threads N`.
- Full suite passes with 2552 tests.

### DIFF
--- a/crates/fgumi-bam-io/src/lib.rs
+++ b/crates/fgumi-bam-io/src/lib.rs
@@ -33,8 +33,8 @@ pub use reader::{
     BamReaderAuto, BgzfReaderEnum, ChainedReader, FgumiBgzfReader, PipelineReaderOpts,
     RawBamReaderAuto, TeeReader, create_bam_reader, create_bam_reader_for_pipeline,
     create_bam_reader_for_pipeline_with_opts, create_bam_reader_with_opts, create_raw_bam_reader,
-    create_raw_bam_reader_with_opts, make_bgzf_reader, open_normalized_input,
-    read_header_and_replay, read_prefix,
+    create_raw_bam_reader_from_stream_with_opts, create_raw_bam_reader_with_opts, make_bgzf_reader,
+    open_normalized_input, read_header_and_replay, read_prefix,
 };
 pub use reorder::{DrainReady, ReorderBuffer};
 pub use writer::{
--- a/crates/fgumi-bam-io/src/reader.rs
+++ b/crates/fgumi-bam-io/src/reader.rs
@@ -537,18 +537,60 @@ pub fn create_raw_bam_reader_with_opts<P: AsRef<Path>>(
 ) -> Result<(RawBamReaderAuto, Header)> {
     let path_ref = path.as_ref();
     let bgzf_reader = open_bgzf_reader(path_ref, opts, threads, "raw BAM reader")?;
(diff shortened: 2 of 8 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #808, opened 2026-08-17
test(codec): assert the bases the codec_template consensus tests are named for

### LINKED ISSUE #807: test(codec): codec_template tests assert consensus length, not bases, under a stale #763 rationale
## Summary

#768 (which closed #763, "store the CODEC fixture's reverse read in reference orientation") fixed `create_fr_pair` so that `codec_template`-built CODEC families now produce real, reference-derived consensus **bases**. But two tests added with #767 still assert only the consensus **length**, and the `codec_template` doc comment still claims the bases are unusable — a rationale that #768 has made false.

## The stale rationale

In `crates/fgumi-consensus/src/codec_caller.rs`, on the `codec_template` helper:

> The tests below assert consensus *lengths* rather than consensus *bases*, because `create_fr_pair` stores the reverse read's `SEQ` in read orientation where a BAM stores it in reference orientation — so its two strands point opposite ways and the consensus comes out mostly `N` for every family the fixture builds, not just these. Tracked as fulcrumgenomics/fgumi#763; the length is what this fix is about, and it is unaffected.

`#763` is closed (fixed by #768). The consensus is no longer mostly `N`; it is the reference sequence over the aligned span. This was verified while merging #770: `test_indel_at_overlap_boundary_still_calls_a_consensus` now pins the exact reference-derived bases (0 mismatches against `REF_BASES`, with the deletion skipped).

## Length-only assertions that could assert bases

- `test_dovetailed_starts_without_an_indel_call_a_consensus` — asserts `consensus[0].bases.len() == 128`
- `test_terminal_indel_outside_the_shared_region_calls_a_consensus` — asserts `consensus[0].bases.len() == 127`

Both emit a consensus and assert only its length. This is exactly the weakness issues #769 / #771 removed elsewhere in the CODEC tests ("assert length or non-emptiness where they should assert bases").

## Proposal

Strengthen those two tests to assert the emitted consensus bases (verified against `REF_BASES`), and correct/remove the now-false `codec_template` doc comment. `test_min_duplex_length_is_measured_over_the_shared_region` and `test_indel_inside_the_shared_region_is_still_rejected` are appropriate as-is (threshold and rejection tests, respectively).

Test-only; no production change.

nh13:
Closes #807.

## What

Two CODEC tests added with #767 assert only the consensus **length**, under a `codec_template` doc comment that still claimed the bases were unusable and pointed at #763. #768 (the fix for #763) stores the reverse CODEC read in reference orientation, so `codec_template` families now consense to the reference sequence over the region the two strands share — not the mostly-`N` output the `create_fr_pair` orientation bug used to produce. This makes those tests assert the bases they are named for, and corrects the stale doc.

## Changes

- `test_dovetailed_starts_without_an_indel_call_a_consensus` — now asserts the 128 emitted consensus bases (was length-only).
- `test_terminal_indel_outside_the_shared_region_calls_a_consensus` — now asserts the 127 emitted consensus bases (was length-only).
- Corrected the `codec_template` doc comment (removed the false "consensus comes out mostly `N` … Tracked as #763" rationale).

The two sibling `codec_template` tests are appropriate as-is: `test_min_duplex_length_is_measured_over_the_shared_region` is a threshold test and `test_indel_inside_the_shared_region_is_still_rejected` rejects (no bases to assert).

## Verification

Each new expected literal was verified byte-for-byte against `REF_BASES`, not merely captured from the code:

- **dovetail** (128): the body after the 2-base soft-clip edge equals `REF_BASES[200..326]` exactly (0 non-`N` mismatches, 0 `N`s in the body).
- **terminal-indel** (127): the body equals the reference over the shared span with the terminal deletion skipped (0 non-`N` mismatches); the lone `N` is the single-strand edge past the reverse read's close.

Test-only; no production change. `cargo nextest run -p fgumi-consensus` (1267 passed), `cargo fmt --check`, and `cargo clippy -p fgumi-consensus --all-targets` all clean.

### DIFF
--- a/crates/fgumi-consensus/src/codec_caller.rs
+++ b/crates/fgumi-consensus/src/codec_caller.rs
@@ -4050,12 +4050,11 @@ mod tests {
     /// Both reads are cut from the same shared reference bases, so the alignment geometry is
     /// the only variable between them.
     ///
-    /// The tests below assert consensus *lengths* rather than consensus *bases*, because
-    /// [`create_fr_pair`] stores the reverse read's `SEQ` in read orientation where a BAM
-    /// stores it in reference orientation — so its two strands point opposite ways and the
-    /// consensus comes out mostly `N` for every family the fixture builds, not just these.
-    /// Tracked as fulcrumgenomics/fgumi#763; the length is what this fix is about, and it is
-    /// unaffected.
+    /// With the reverse read stored in reference orientation (fulcrumgenomics/fgumi#763, fixed
+    /// by #768), a family with no strand disagreement consenses to the reference sequence over
+    /// the region the two strands share — so the tests below assert the consensus *bases*, not
+    /// merely its length. Positions only one strand covers — the soft-clip read-through edges —
+    /// resolve to `N`.
     fn codec_template(
         pos_start: usize,
         pos_cigar: &[(Kind, usize)],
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #811, opened 2026-08-18
fix(pipeline): saturate queue-byte debits so queue_bytes_in_flight cannot overflow

### LINKED ISSUE #810: Intermittent u64 overflow in queue_bytes_in_flight: reorder-buffer sub_heap_bytes can wrap past zero
### Summary

Under coverage CI (`cargo llvm-cov nextest`, a debug build with overflow checks), `test_pipeline_memory_backpressure::reader_resumes_and_completes_after_the_writer_recovers` intermittently panics inside `BamPipelineState::queue_bytes_in_flight`:

```
thread '<unnamed>' panicked at src/lib/unified_pipeline/bam.rs: attempt to add with overflow
... "Worker thread 0 panicked: attempt to add with overflow"
```

### Root cause

`queue_bytes_in_flight()` sums nine byte counters with checked `+`. Most are the `*_heap_bytes` atomics, which are only ever debited through `refund_queue_bytes` → `saturating_refund` (saturates at zero, cannot underflow). The three reorder-buffer counters (`q2_reorder_state`, `q3_reorder_state`, `write_reorder_state`) are debited through `ReorderBufferState::sub_heap_bytes`, which is a **raw `fetch_sub`**:

```rust
pub fn sub_heap_bytes(&self, bytes: u64) {
    self.heap_bytes.fetch_sub(bytes, Ordering::AcqRel);
}
```

If a `sub_heap_bytes` is ever unpaired or races ahead of its matching `add_heap_bytes`, the atomic wraps past zero to ~`u64::MAX`. The subtraction itself does not panic (atomics wrap silently), but the next `queue_bytes_in_flight` sum then overflows on the checked `+` in a debug build. The wrap hazard is already documented at `base.rs:733` ("`fetch_sub` past zero wraps to `u64::MAX` and stops the pipeline reading …").

### Why it's rarely seen

Reproduces only under llvm-cov's instrumentation, whose per-branch counters reschedule threads enough to expose the interleaving. It did **not** reproduce in 73 normal debug-build runs (25 via `cargo test` + 48 via the compiled binary under CPU stress). It is timing-dependent, so a plain re-run of the coverage job may pass.

### Impact

- Intermittently reddens the `coverage` CI job.
- In release builds (overflow checks off) the sum wraps silently, so a transiently-underflowed reorder counter can momentarily read as a huge in-flight total and **close the Read admission gate** until the counter is corrected — a latent correctness issue for the memory-budget backpressure, not just a test flake.

### Suggested fix

Two layers:
1. **Defensive:** make `sub_heap_bytes` saturating (mirror `refund_queue_bytes` / `saturating_refund`), with a `debug_assert!` that the sub is paired — so an unpaired sub is caught in tests but cannot wrap in production.
2. **Root cause:** find the unpaired/racy `add_heap_bytes` / `sub_heap_bytes` pairing in the reorder-buffer lifecycle (charge-before-publish vs debit-on-pop ordering across the Read/Decode/Group/Write stages) and fix the pairing itself.

### Notes

Pre-existing on `main`; surfaced while re-running CI for #772, which does not touch the reorder-buffer accounting. Same "pre-existing intermittent CI failure exposed by #772's re-run" family as the flaky `read_ahead_under_a_stalled_writer_shrinks_with_the_queue_memory_budget` (#809).

nh13:
## Summary

Fixes #810 — an intermittent `attempt to add with overflow` panic in `BamPipelineState::queue_bytes_in_flight`, seen under coverage CI (`cargo llvm-cov nextest`, a debug build with overflow checks).

## Root cause

`queue_bytes_in_flight` sums nine heap-byte counters that gate the Read step. Credits are generally taken before a batch is published, but several debits use a raw `fetch_sub`:

- the three reorder buffers via `ReorderBufferState::sub_heap_bytes` (the wrap hazard is already noted in that struct's docs), and
- the serialized / compressed output counters via `q5_track_pop` / `q6_track_pop` — and `q6_push` charges the counter *after* publishing the batch, so a consumer can pop-and-refund before the charge lands.

When a debit races ahead of its credit it subtracts past zero and wraps the atomic to `u64::MAX`. The atomic op doesn't panic, but the next `queue_bytes_in_flight` overflows the checked `+` (debug/coverage) or, in release, reads a bogus huge in-flight total that spuriously closes the admission gate.

## Fix

- `ReorderBufferState::sub_heap_bytes` floors at zero (saturating `fetch_update`).
- BAM `q5_track_pop` / `q6_track_pop` route through the existing saturating `refund_queue_bytes` helper.
- `queue_bytes_in_flight` folds with `saturating_add` as a backstop, since the nine loads are not one atomic snapshot.

In a balanced run these are arithmetically identical to the previous code; they diverge only in the underflow case that caused #810.

## Scope / follow-up

Observed on the BAM pipeline, so this hardens the BAM aggregate and the shared `ReorderBufferState` (which also covers the FASTQ reorder buffers). The FASTQ pipeline's own `queue_bytes_in_flight` and per-counter debits are left untouched here to avoid overlapping the in-flight #766 charging rework in fastq.rs; a follow-up can extend the same `saturating_add` backstop there.

## Testing

- `cargo ci-fmt` and `cargo ci-lint` clean; full `cargo test` suite green locally.
- New regression test `test_reorder_buffer_sub_heap_bytes_saturates_at_zero` (an over-subtraction floors at zero instead of wrapping to `u64::MAX`).
- The overflow only reproduces under llvm-cov's scheduling jitter (it did not reproduce in 73 normal-build runs), but the fix removes it by construction: saturating arithmetic makes the wrap impossible regardless of thread interleaving.

Risk: command output changes: none; `unsafe` changes: none, and the CLAUDE.md allowlist is unchanged; memory-bound, queue-capacity, and thread/backpressure policy changes: none.

Fix: BAM queue-byte accounting now uses saturating debits and aggregate additions to prevent underflow and overflow during concurrent updates.

- Reorder-buffer, serialized-output, compressed-output, and processed-byte debits saturate at zero.
- Output charges occur before publication and are refunded when pushes fail.
- Regression tests cover debit saturation, charge ordering, refunds, and rejected pushes.

### DIFF
--- a/src/lib/unified_pipeline/bam.rs
+++ b/src/lib/unified_pipeline/bam.rs
@@ -1036,15 +1036,25 @@ impl<G: Send, P: Send + MemoryEstimate> BamPipelineState<G, P> {
     /// stalls.
     #[must_use]
     pub fn queue_bytes_in_flight(&self) -> u64 {
-        self.q1_heap_bytes.load(Ordering::Acquire)
-            + self.q2_reorder_state.get_heap_bytes()
-            + self.q2b_heap_bytes.load(Ordering::Acquire)
-            + self.q3_reorder_state.get_heap_bytes()
-            + self.output.groups_heap_bytes.load(Ordering::Acquire)
-            + self.output.processed_heap_bytes.load(Ordering::Acquire)
-            + self.output.serialized_heap_bytes.load(Ordering::Acquire)
-            + self.output.compressed_heap_bytes.load(Ordering::Acquire)
-            + self.output.write_reorder_state.get_heap_bytes()
+        // Saturating sum: every debit that feeds these counters now floors at
+        // zero, but the nine loads are not one atomic snapshot, so an in-flight
+        // debit/credit interleaving could still transiently read a counter high.
+        // Fold with `saturating_add` so the aggregate can never overflow the
+        // checked `+` in a debug/coverage build even under that interleaving —
+        // the gate is conservative for an instant rather than panicking.
+        [
+            self.q1_heap_bytes.load(Ordering::Acquire),
+            self.q2_reorder_state.get_heap_bytes(),
+            self.q2b_heap_bytes.load(Ordering::Acquire),
+            self.q3_reorder_state.get_heap_bytes(),
+            self.output.groups_heap_bytes.load(Ordering::Acquire),
+            self.output.processed_heap_bytes.load(Ordering::Acquire),
+            self.output.serialized_heap_bytes.load(Ordering::Acquire),
+            self.output.compressed_heap_bytes.load(Ordering::Acquire),
+            self.output.write_reorder_state.get_heap_bytes(),
+        ]
+        .into_iter()
+        .fold(0u64, u64::saturating_add)
     }
 
     /// Whether the Read step may admit another batch under the queue memory
@@ -1378,7 +1388,10 @@ impl<G: Send + 'static, P: Send + MemoryEstimate + 'static> OutputPipelineState
     }
 
     fn q5_track_pop(&self, heap_size: u64) {
-        self.output.serialized_heap_bytes.fetch_sub(heap_size, Ordering::AcqRel);
+        // Saturating (via `refund_queue_bytes`): this counter is summed into the
+        // overflow-checked `queue_bytes_in_flight`, so an under-subtraction must
+        // floor at zero rather than wrap to `u64::MAX`.
+        refund_queue_bytes(&self.output.serialized_heap_bytes, heap_size);
     }
 
     fn q6_pop(&self) -> Option<(u64, CompressedBlockBatch)> {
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
Addressed the review feedback:

- **Q5/Q6 publish-before-charge race (Major):** `serialize_output_push` and `q6_push` now charge their heap-byte counters *before* publishing to the queue and refund the charge if the push fails, so a consumer can no longer pop-and-refund before the producer's charge lands. Applied the same ordering to the two sibling serialized-push sites in `try_step_serialize` (held-batch and priority-5 paths), which had the identical charge-after-publish shape. Updated the now-stale `q6_track_pop` comment.
- **Outside-diff accounting (Major):** verified the reorder-buffer add/remove path — `add_heap_bytes` and `sub_heap_bytes` both run under the same `write_reorder_buffer().lock()`, so unlike the lock-free Q5/Q6 `ArrayQueue`s there is no charge-before-publish window there; no change needed. While auditing, found `processed_heap_bytes` (also summed into `queue_bytes_in_flight`) was the one remaining debit still using a raw `fetch_sub`, contradicting this PR's new "every debit floors at zero" comment — converted its three debit sites to the saturating `refund_queue_bytes` so the claim holds.
- **`Read` identifier backticks (Minor):** backticked `Read` in the two new doc comments (`base.rs` reorder-buffer sub and the #810 regression test).

Added two regression tests: `serialize_output_push_charges_before_publish_and_refunds_on_pop` and `q6_push_charges_on_publish_and_refunds_a_rejected_push` (the latter fills the bounded queue and asserts a rejected push leaves no phantom charge). Full local gate green (fmt / clippy-pedantic / 7657 tests).

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #812, opened 2026-08-18
fix(consensus)!: cap simplex --max-reads per end, not per group (#723)

### LINKED ISSUE #723: simplex --max-reads caps the whole MI group instead of each end

nh13:
Closes #723.

## Problem

`fgumi simplex --max-reads` applied the cap to the entire MI group *before* splitting it into Fragment/R1/R2 subgroups, whereas fgbio's `CallMolecularConsensusReads` caps each end independently. A cap of N therefore admitted N reads *total* across both ends in fgumi, but N reads *per end* in fgbio.

Because the retained reads were split arbitrarily between R1 and R2, the effective per-end depth was roughly halved, and families whose split left one end empty produced no consensus pair at all. On a real IDT cfDNA grouped BAM (~77k molecules, `--max-reads 2` / `-M 1`), fgumi emitted 726,854 consensus reads (all `cD:1`) against fgbio's 734,536 (`cD` 1 and 2) — the 7,682 missing reads being families the group-level cap collapsed to a single end.

## Fix

Apply the cap **per end**, inside `process_subgroup`, and **after** filtering to the most common alignment. This matches fgbio's `consensusCall` on two axes:

- **Per-end, not per-group** — each of Fragment/R1/R2 gets its own cap of N.
- **After the alignment filter** — fgbio downsamples the alignment-filtered reads and ranks only those. Capping the raw group instead (the pre-fix behavior) ranks reads fgbio never considers and can under-fill the consensus when a kept read subsequently fails the alignment filter.

The cap still routes the reads it discards to the `--rejects` output and counts them as `Downsampled` rejections so `raw_reads_used` excludes them (preserving the accounting added in #724). The per-end selector preserves input order, because the retained reads also feed order-sensitive tag extraction (cell-barcode / methylation-strand read the first retained read); the duplex/codec `consensus_call` cap remains rank-ordered as before. Duplex and codec do not route through `process_group`, so they are unaffected.

The `simplex --max-reads` help text previously documented the old whole-group behavior and cited this issue as an open caveat; it is updated to describe the per-end cap.

This changes output for anyone setting `--max-reads` on `simplex`, so it is a breaking change with a CHANGELOG note.

## Tests

- `max_reads_caps_each_end_independently` — an `#[rstest]` table over caps 1/2/4/8 on a four-template paired family; each end must be capped independently. The **cap-of-1** case is the regression that previously orphaned an end and emitted no consensus.
- `max_reads_is_applied_to_fragment_subgroup` — the Fragment subgroup is capped independently.
- `per_end_cap_keeps_lowest_ranking_reads_in_input_order` — the per-end selector keeps the htsjdk-`Murmur3(42)` lowest-ranking reads (q0, q2, q3) but in **input** order, distinguishing it from the rank-ordered duplex/codec selector.
- `downsampled_reads_are_rejected_and_routed_to_the_rejects_output` — #724 guard, now exercising the after-filter cap path end to end.
- `consensus_call_cap_is_independent_of_simplex_path` — locks in the duplex/codec cap path.

The obsolete whole-group downsampling tests are removed. The existing `test_consensus_downsampling` cross-execution-mode determinism suite (`--threads 1` == `--threads 8` == no-threads) still passes, as does the full workspace suite (7682 tests).

Risk: `fgumi simplex` output changes are pinned by alignment filtering, per-subgroup caps, deterministic hash ranks, and input-order preservation; `unsafe` changes: none, so no `CLAUDE.md` allowlist update; memory, queue, and backpressure policy changes: none.

`--max-reads` now applies independently to Fragment, R1, and R2 subgroups after alignment filtering. Discarded reads go to `--rejects` and count as `Downsampled`.

The change preserves duplex and codec behavior. It updates help text and adds a changelog note for the breaking behavior change.

Tests cover per-end and fragment caps, deterministic ranking, input order, mate consistency, rejection routing, and duplex/codec behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Breaking Changes
+
+- **fgumi simplex:** `--max-reads` now caps each end (Fragment, R1, R2) independently rather
+  than the whole MI group ([#723](https://github.com/fulcrumgenomics/fgumi/issues/723)). A family
+  with 6 R1 + 6 R2 reads and `--max-reads 3` previously produced a single consensus with depth 3;
+  it now produces two consensus records (one per strand), each with depth 3. This matches fgbio's
+  `CallMolecularConsensusReads`, which applies the cap independently to each end of the family.
+
 ### Bug Fixes
 
 - **Breaking:** end-of-stream flush paths now fail instead of discarding what they were holding,
--- a/crates/fgumi-consensus/src/vanilla_caller.rs
+++ b/crates/fgumi-consensus/src/vanilla_caller.rs
@@ -721,8 +721,9 @@ impl VanillaUmiConsensusCaller {
         // fulcrumgenomics/fgbio#1166 — previously `shuffle.take(maxReads)`).
         // The duplex and codec callers reach the single-strand consensus through
         // `consensus_call`, so without this cap `--max-reads-per-strand` is silently ignored
-        // on those paths (DUPLEX3-01). The vanilla path caps raw records earlier in
-        // `process_group` and does not route through here, so it is unaffected.
+        // on those paths (DUPLEX3-01). The vanilla path caps alignment-filtered `SourceRead`s
+        // in `process_subgroup` (per read-type, via `downsample_filtered_source_reads`) and
+        // does not route through here, so it is unaffected.
         //
         // The cap shapes ONLY the consensus bases/quals/depths: the FULL (uncapped)
         // `source_reads` are retained on the output below. fgbio caps inside `consensusCall`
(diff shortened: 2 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
Addressed the review comment:

- **Preserve rejected-record input order (Major):** `process_group` now tags every input record with its group-input position up front, threads those positions through `filter_reads`/`subgroup_reads`/`process_subgroup`, and collects all rejects — secondary/supplementary filter, each end's min-reads/zero-length/alignment/downsample filters, and the orphan-consensus survivors — into one positioned buffer that a new `flush_group_rejects` sorts by input position before draining into `rejected_reads`. That fixes the "all R1 rejects, then all R2 rejects" reordering (R1 and R2 are processed in separate passes) so the `--rejects` secondary output honors `SimplexProcessedBatch`'s input-order contract. Added `test_rejected_reads_preserve_interleaved_group_input_order`, which feeds an interleaved a/R1, a/R2, b/R1, b/R2 family (min_reads=3, so each 2-read end is rejected) and asserts the rejects stay interleaved rather than grouped by end.

Note: the previous run's `sort-correctness` was **cancelled**, not failed — again the ~6h "Install samtools" apt stall (#825 hardens it); this push re-triggers CI. Full local gate green (fmt / clippy-pedantic / 7683 tests).

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #814, opened 2026-08-18
perf(sort): give read-ahead depth to the file the merge is actually on

nh13:
Stacked on the instrumentation PR. The changes that moved wall clock, plus four pieces of hygiene found along the way — and an honest account of how much is left, because it turns out to be the more useful result.

**Every number below is a paired control and arm in one boot on one host.** `c7g.4xlarge`, `1kg-wgs-HG00096`, coordinate → template-coordinate, `--max-memory 512M --max-temp-files 1024 --temp-compression 1 --compression-level 1`, 779,820,469 records. `--max-memory` is per thread, so t8 spills 89 runs and t16 spills 44 — different merges, whose absolute walls are not comparable to each other.

## Result

Control is `main`, which now contains #806.

| | merge wall | consumer CPU | park | worker util | peak RSS |
| **t16** `main` | 174.6s | 74.3s | 97.8s | 51% | 9.599 GB |
| **t16** this PR | **157.7s (−9.7%)** | 113.0s | 44.2s | 57% | 10.568 GB |
| **t8** `main` | 197.2s | 75.4s | 121.6s | 89% | 4.813 GB |
| **t8** this PR | **189.8s (−3.8%)** | 187.8s | 1.4s | 93% | 9.344 GB |

`fgumi compare bams --command sort` reports **IDENTICAL** on the full 779,820,469 records — which is what licenses reading these gaps as overhead removed rather than work skipped.

**These are marginal figures for this PR alone.** An earlier draft of this work quoted −42.5% at t8; that was cumulative from before #806, which did the heavy lifting (326.5s → ~200s). This PR is the next 4.3% at t8 and 10.8% at t16.

## What it costs

**Peak RSS grows, and at t8 it nearly doubles: 4.813 → 9.344 GB (+94%). At t16 it is +10% (9.599 → 10.568 GB).** Phase 2 read-ahead is not charged against `--max-memory`, so this is real footprint on top of the budget, and it scales with the number of spill files rather than with the budget — t8 has 89 files against t16's 44, which is why the same change costs so much more there.

At t16 the growth is attributable: disabling only the predictive read-ahead returns RSS to baseline (9.594 GB against an uninstrumented 9.572 GB, measured on #806's pre-merge head). **So top-k is essentially all of the t16 growth — for −4.6% of wall.** The t8 split is not isolated; if a reviewer wants it attributed before merge, say so and I will run it.

Anyone sizing a host on `--max-memory` alone should know this. It is the hazard the sort sweep's own documentation warns about: a tighter budget spills more runs, so the widest merges land on the hosts least able to absorb them.

## The one finding worth carrying forward

**Targeted depth wins; uniform depth fails.** Fifteen interventions were measured. Three paid, and all three gave depth to the single file the consumer was blocked on or about to reach:

| change | measured |
| --- | --- |
| read ahead 4 MiB rather than 2 MiB on the blocked file | −2.8% t8, −5.2% t16 |
| let the consumer decompress a block itself instead of parking | park 122.2 → 2.8s at t8 |
| let the loser tree name the *next* source so its read starts early | **−4.6% t16**, isolated |

Everything that spread depth uniformly failed: decompress cap 32/128 (+2.3%/+3.4%), raw cap 16/32 (~0%/−0.6%), filling the cap on every claim (+36%). Everything that tried to signal workers better also failed: wake targeting (−0.9%), priority reorder (0.06%), scan-start redirect (0%).

Both failure modes share one cause. 98-99% of residual park time is `starved` — nothing buffered, nothing in flight, nothing read at all — while the device runs at 323 of 1000 MB/s provisioned. Uniform depth offers somewhere to put work that does not exist yet; better signalling wakes a worker to do work that does not exist yet. Only naming *which* file to start reading changes anything.

## Headroom left, stated plainly

This is the part I would want to read first as a reviewer, and it is not flattering to the PR.

**At t8 the merge is now at its floor — but the floor moved.** Merge 189.8s = consumer serial CPU 187.8s (98.9%) + park 1.4s, with a worker floor of 1410.5s ÷ 8 = 176.3s. Recoverable without doing less work: **2.0s, 1.1%.**

That is worth reading carefully, because it is not the same as "we captured the available headroom". On `main` at t8 the merge was *worker*-bound: loop 197.2s against a 175.4s worker floor, 21.8s recoverable. This PR lowered the loop by 7.4s and simultaneously **raised the consumer floor from 75.4s to 187.8s**, because self-decompression moves work onto the serial thread. The merge is now consumer-bound where it was worker-bound. **~13.5s (7%) sits between the current 189.8s and the 176.3s worker floor, and it is reachable only by giving that decompression back to the workers without reintroducing the park.** That is the clearest next target this work produces.

**At t16, 44.7s (28%) is recoverable**, all of it consumer park, 99% `starved`, with workers 43% idle and a worker floor of only 89.8s. That is the largest identified prize. But a control says most of it is a property of the *input shape*: the same records sorted queryname → coordinate merge in **126.7s** with the consumer 97% busy and park ~3s. Sorting an already-`SO:coordinate` file into template-coordinate is what produces the starving run structure — and that is also the production path for dedup, so it is worth attacking; it is just not general engine headroom.

**And the number that actually binds has never been decomposed.** Consumer serial CPU is the floor in all three configurations — 186.5s at t8 (239 ns/record), 112.1s at t16 (144 ns/record), 123.2s on the queryname→coordinate control (158 ns/record). Same records, same block count, near-identical worker-side decompress, and a 66% spread in per-record cost that is unexplained. A loser-tree microbenchmark accounts for ~5% of it. The other ~95% is unattributed, and it is larger than everything this PR delivers.

This PR also demonstrates why: at t8 it removed **119.4s of park** and gained **8.2s of wall**, because park converted almost 1:1 into consumer serial CPU (74.6 → 186.5s). The work was not removed, it was relocated onto the critical path. That is a real win where the consumer had idle capacity and nearly a no-op where it did not.

## Reading order

**The three wins:** `3d08ea8d` (read-ahead 4 MiB), `8b66d5d1` (consumer self-decompress), `5f52205c` (top-k prediction). `dca835cf` batches a per-record atomic off the consumer path (−1.5% at t8, null at t16).

**Hygiene, near-zero on wall, kept for correctness or clarity:** `ff7824ad` makes the zstd spill frame size a knob *and fixes a latent bug it exposed* — the spill writer pre-flushed at a hardcoded 64 KiB rather than the configured frame size, so the knob was inert and a frame-size sweep measured a no-op as a 4% regression. `bc0ab690` squashes three scheduling tidy-ups that measured 0%, −0.9% and 0.06%, kept because each removes a wrong behaviour the next reader would otherwise re-derive.

## Notes for review

- **`losers[1]` is not the runner-up**, which is the tempting O(1) shortcut in the top-k commit. After a replay that node holds whoever lost *that* replay's final comparison, not the global second-smallest. `runner_up()` walks the losers along the winner's root-to-leaf path instead, O(log k). There is a test on exactly the case the shortcut gets wrong, and it fails against the shortcut.
- **Self-decompress happens only on the way into a park, never in place of merging.** The consumer is the serial bottleneck; taking delegable work while it still has records to emit would be strictly worse. Bounding it to one block per park instead of draining costs +7.6%.
- **`runner_up()` is called 25,003,410 times**, once per ~31 records, not once per run as an earlier draft claimed. At 11.63 ns/call that is ~0.29s of serial consumer time — 0.19% of merge wall, so the design holds, but with less margin than "once per run" implied.
- **The two validity counters are not decoration.** `Consumer served itself` and `Next-source predictions published` each guard a silent failure mode: the mechanism can fail to run at all and still report a plausible wall time. During this work the gate function was briefly defined but never called, so both were dead-code eliminated — and neither `clippy -D warnings` nor rustc's `dead_code` reported it. Grepping the built binary is what caught it.
- A byte-budgeted uniform read-ahead floor was developed and **left out**: it showed −3.6% once and nothing at all on an identical repeat in the same boot, and the sweep meant to settle it had clamped two of three arms to the same value.

## Verification

`cargo ci-fmt`, `ci-lint` clean; `cargo ci-test` **7559 passed / 30 skipped**; every commit builds independently. Correctness gated on `compare bams --command sort` throughout — IDENTICAL every time, 0 sort-order violations, 0 run-multiset mismatches.

Risk: output changes: none; identical sorted output and sort-order checks pin behavior. `unsafe` changes: none; CLAUDE.md allowlist update: none. Memory and scheduling policy changes: yes; targeted read-ahead increases peak RSS outside `--max-memory`.

Fix: steer read-ahead to blocked and predicted merge sources, and let the consumer decompress queued blocks before parking.

- Add codec-specific BGZF and Zstd spill frame sizing.
- Add loser-tree runner-up prediction.
- Increase targeted read-ahead to 4 MiB with byte bounds.
- Prioritize starving and predicted sources.
- Batch progress updates every 4096 records.
- Add tests for frame sizing, decompression, scheduling, prediction, and progress.
- Report self-served parks and predictions.

Merge wall time decreases by 9.7% at t16 and 3.8% at t8. Peak RSS increases by 10% at t16 and 94% at t8. Correctness, formatting, lint, and CI checks pass.

### DIFF
--- a/crates/fgumi-sort/src/bgzf_io.rs
+++ b/crates/fgumi-sort/src/bgzf_io.rs
@@ -15,8 +15,37 @@ use std::collections::BTreeMap;
 use std::io::{BufWriter, Write};
 use std::sync::Arc;
 
-/// Padding beyond `BGZF_MAX_BLOCK_SIZE` for the staging buffer capacity.
+/// Padding beyond the frame size for the staging buffer capacity.
 const STAGING_PADDING: usize = 4096;
+
+/// Uncompressed bytes per zstd **spill** frame.
+///
+/// BGZF mandates blocks of at most 64 KiB; zstd has no such limit, and spill
+/// files are a private temporary format that nothing outside this crate reads.
+/// Pinning zstd frames to the BGZF ceiling was therefore vestigial, and it set
+/// the merge's block count -- 5,368,249 blocks on a measured 89-way merge, of
+/// which the consumer takes roughly one per round trip. Fewer, larger frames cut
+/// round trips proportionally.
+///
+/// Raising this costs memory in two places, both worth watching: the per-worker
+/// zstd decompress buffer ([`crate::worker_pool::zstd_decomp_cap`]) and the
+/// per-file reorder buffer, which holds up to `PHASE2_DECOMP_CAP` *uncompressed*
+/// frames for each of K files. Read-ahead itself does not scale with it, because
+/// the refill allowance is sized in bytes rather than blocks.
+pub(crate) const SPILL_FRAME_BYTES: usize = BGZF_MAX_BLOCK_SIZE;
+
+/// Bytes a staging buffer accumulates before it submits a compress job.
+///
+/// Keyed on the codec so BAM output stays inside the BGZF format: output always
+/// uses [`SpillCodec::Bgzf`], so it cannot pick up a spill-only frame size by
+/// accident.
+#[must_use]
+pub(crate) const fn spill_frame_bytes(codec: SpillCodec) -> usize {
+    match codec {
+        SpillCodec::Bgzf => BGZF_MAX_BLOCK_SIZE,
+        SpillCodec::Zstd => SPILL_FRAME_BYTES,
+    }
+}
 
 /// Per-block position notification emitted by the I/O writer loop when index
 /// generation is enabled.
@@ -76,7 +105,7 @@ impl StagingBuffer {
     ) -> Self {
         Self {
             pool,
-            buf: Vec::with_capacity(BGZF_MAX_BLOCK_SIZE + STAGING_PADDING),
+            buf: Vec::with_capacity(spill_frame_bytes(codec) + STAGING_PADDING),
             next_serial: 0,
             result_tx,
             permit_pool,
@@ -96,7 +125,7 @@ impl StagingBuffer {
     /// Returns true if the staging buffer has reached the BGZF block size threshold.
     #[inline]
     pub(crate) fn is_full(&self) -> bool {
-        self.buf.len() >= BGZF_MAX_BLOCK_SIZE
+        self.buf.len() >= spill_frame_bytes(self.codec)
     }
 
     /// Current uncompressed length of the pending (not-yet-flushed) block.
@@ -136,8 +165,9 @@ impl StagingBuffer {
         self.permit_pool.acquire()?;
 
         let data = std::mem::replace(&mut self.buf, self.pool.buffer_pool.checkout());
-        if self.buf.capacity() < BGZF_MAX_BLOCK_SIZE + STAGING_PADDING {
-            self.buf.reserve(BGZF_MAX_BLOCK_SIZE + STAGING_PADDING - self.buf.capacity());
+        let want = spill_frame_bytes(self.codec) + STAGING_PADDING;
+        if self.buf.capacity() < want {
+            self.buf.reserve(want - self.buf.capacity());
         }
 
         let serial = self.next_serial;
@@ -174,7 +204,7 @@ impl StagingBuffer {
     pub(crate) fn write_chunked(&mut self, data: &[u8]) -> anyhow::Result<()> {
         let mut remaining = data;
         while !remaining.is_empty() {
-            let space = BGZF_MAX_BLOCK_SIZE.saturating_sub(self.buf.len());
+            let space = spill_frame_bytes(self.codec).saturating_sub(self.buf.len());
             let n = remaining.len().min(space);
             self.buf.extend_from_slice(&remaining[..n]);
             remaining = &remaining[n..];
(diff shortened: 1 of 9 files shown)


nh13:
Stack: based on #813 (instrumentation only, no behaviour change), so the diff shown here is just the behavioural half. Review #813 first; this will retarget to `main` once it merges.

nh13:
@coderabbitai pause

nh13:
Stack now has a third: #823 adds the merge floor line (which of three limits binds, and what is recoverable), the clock-overhead correction for the sampled sub-phases, and demotes the breakdown to `debug!`. It ships no speedup — it ships the measurement that says the 8-thread merge is at 99.4% of its serial floor and that the 16-thread gap is not worker capacity.

It also records the arms that failed against that gap, so they are not re-litigated: five that moved decompression off the serial consumer (slope −0.373 s/s, r = −0.975), an adaptive regime gate (+12.5%, and its discriminant moved under its own action), a consumer-side read of the starved block (fired on 0.6% of starved parks), and finer publish granularity on the deep read (0.1%).

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #818, opened 2026-08-18
perf(sort): honor --check-crc in the fgumi-sort reader stack (#817)

### LINKED ISSUE #817: perf(sort): honor --check-crc / skip BGZF CRC in the fgumi-sort reader stack
## Summary

`fgumi sort` decodes its input BGZF through its own reader stack (`crates/fgumi-sort/src/reader.rs` `RawBamRecordReader`, and the parallel decode workers in `crates/fgumi-sort/src/worker_pool.rs`), separate from the raw-BAM reader unified in #800/#801. That stack calls the always-verify fgumi-bgzf primitives (`decompress_block` / `decompress_block_into`) and exposes no way to skip the per-block CRC32, so the `--check-crc` / `--no-check-crc` policy added in #798 never reaches `fgumi sort`.

This is the `fgumi-sort` follow-up #800 explicitly deferred ("`fgumi-sort` and `extract` have their own separate reader stacks; wire those in their own issues").

## Proposed change

- Thread a `verify_crc` flag through the sort reader stack — `RawBamRecordReader` and the `worker_pool` decode workers — switching the decode calls to the CRC-skip-capable `_opts` variants (`decompress_block_into_opts` / `decompress_block_slice_into_opts`), honoring the flag (decompressed-size check always runs).
- Add `--check-crc` / `--no-check-crc` to `fgumi sort`, matching the semantics established in #798: default verifies for file input and skips for trusted stdin; explicit flags override; log the effective policy at startup.

## Acceptance / validation

- Byte-identical decode vs the current always-verify path (roundtrip tests).
- A corrupted-CRC input is rejected by default and accepted under `--no-check-crc`, and rejected under `--check-crc`, for both the single-threaded and worker-pool decode paths.
- Full workspace suite green; benchmark the sort read+decompress phase (expect the ~6% `verify_decompression` CPU saved on the skip path, no regression when verifying).

Relates to #786, #798, #800.

nh13:
Closes #817. Stacked on #798 (base `786/nhomer/perf-bgzf-crc-skip`) — review after #798.

## What

`fgumi sort` decoded its input BGZF through its own reader stack — the worker-pool decode workers (`worker_pool.rs`) and the single-threaded `RawBamRecordReader` (`reader.rs`) — using the always-verify fgumi-bgzf primitives, so the `--check-crc` / `--no-check-crc` policy from #798 never reached it. This routes both paths through the CRC-skip-capable `_opts` primitives and adds the flag to the command. It's the `fgumi-sort` follow-up #800 explicitly deferred.

## How

- **Worker pool (main sort path).** A `verify_crc` bit lives in `SharedPipelineState` as an `AtomicBool`, set via `SortWorkerPool::set_verify_crc` right after pool construction so the already-spawned workers observe it — this keeps `SortWorkerPool::new`'s signature (and its ~30 test call sites) unchanged. It's wired from a new `RawExternalSorter::verify_crc()` builder option. Only the **Phase 1 input** decode honors it; **spill-chunk** decode is deliberately left always-verifying, since fgumi's own temp files are a separate integrity concern from the input.
- **Single-threaded path (`--verify`).** `RawBamRecordReader` gains a `verify_crc` field (`new_with_opts`), and `open_raw_bam_record_reader_with_header_opts` threads it.
- **Command.** `fgumi sort` gains `--check-crc` / `--no-check-crc` with the #798 policy (verify on for files, off for trusted stdin; explicit flags override), logged as a `CRC verify:` line at startup for both sort and `--verify` mode.

## Tests

- `RawBamRecordReader` (single-threaded) and `RawExternalSorter` (worker pool) each honor `verify_crc` on a CRC-corrupted multi-block BAM: rejected by default, read clean with it off. Both are dependency-free (no samtools).
- The command's `effective_check_crc` policy truth table (file/stdin × the two flags).

Full workspace suite green (2552 passing), clippy clean, fmt/tag-literals/publish-order clean.

## Benchmark

The CRC-skip decode primitive (`decompress_block_into_opts`) was benchmarked at parity vs always-verify in #798/#801, and the incremental change here is a like-for-like allocation at one call site. The expected ~6% *decode* saving on trusted input is a small fraction of total sort CPU and below this dev host's wall-time noise floor, so a dedicated decode-bound sort hot-path benchmark is deferred to a quiet/EC2 host rather than reported from noisy local numbers.

### DIFF
--- a/crates/fgumi-sort/src/external.rs
+++ b/crates/fgumi-sort/src/external.rs
@@ -2192,6 +2192,11 @@ pub struct RawExternalSorter {
     initial_capacity: Option<usize>,
     /// When true, wrap input in a `PrefetchReader` for async I/O.
     async_reader: bool,
+    /// Verify each input BGZF block's CRC32 while decompressing it. Defaults to
+    /// `true` (the safe, pre-existing behavior); the sort command lowers it per
+    /// the `--check-crc` / `--no-check-crc` policy. Applies to input decode only,
+    /// never to spill chunks.
+    verify_crc: bool,
     /// Which optional template-key lanes to retain (template-coordinate only).
     ///
     /// Defaults to [`KeyTypesSpec::Auto`], which provisions the narrowest key
@@ -2300,6 +2305,7 @@ impl RawExternalSorter {
             cell_tag: None,
             initial_capacity: None,
             async_reader: false,
+            verify_crc: true,
             key_types: KeyTypesSpec::default(),
         }
     }
@@ -2510,6 +2516,17 @@ impl RawExternalSorter {
     #[must_use]
     pub fn async_reader(mut self, enabled: bool) -> Self {
         self.async_reader = enabled;
+        self
+    }
+
+    /// Set whether to verify each input BGZF block's CRC32 while decompressing.
+    ///
+    /// Defaults to `true`. Pass `false` to skip verification on trusted input
+    /// (the decompressed-size check always runs). Applies to the Phase 1 input
+    /// decode only; spill chunks are always verified.
+    #[must_use]
+    pub fn verify_crc(mut self, enabled: bool) -> Self {
+        self.verify_crc = enabled;
         self
     }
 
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai pause

PR #822, opened 2026-08-19
docs(pipeline): fix redundant explicit intra-doc links breaking the docs CI job

nh13:
## What

The `docs` CI job is currently **failing on `main`** (HEAD, from #775). `RUSTDOCFLAGS=-D warnings` promotes `rustdoc::redundant_explicit_links` to an error, and `QueueMemoryOptions`' doc comment in `src/lib/commands/common.rs` links three pipeline constants with explicit `crate::unified_pipeline::…` targets that are identical to what rustdoc already resolves from the backticked link text:

- `BACKPRESSURE_THRESHOLD_BYTES` (twice)
- `Q5_BACKPRESSURE_THRESHOLD_BYTES`
- `stage_high_water_mark`

## Fix

Collapse the four links to shortcut intra-doc links (`[`Foo`]`), so the destination is inferred. The rendered documentation is unchanged; only the redundant explicit targets are removed.

## Verification

`RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --workspace` now completes cleanly.

This unblocks the `docs` job on `main` and on every open PR rebased onto it (e.g. #793).

### DIFF
--- a/src/lib/commands/common.rs
+++ b/src/lib/commands/common.rs
@@ -1243,17 +1243,17 @@ fn resolve_memory_budget_with_total(
 /// (the FASTQ aggregate counts its write reorder state and the BAM aggregate its
 /// input and write reorder states), so this budget bounds them through the Read
 /// gate; they additionally apply their own threshold, capped at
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #823, opened 2026-08-19
perf(sort): name which floor the merge is on, and stop shouting the detail

nh13:
Third in the stack: #813 (instrumentation) → #814 (the wins) → this.

## What this adds

**A floor line.** From exact counters, the merge now reports which of three limits binds — the consumer's serial CPU, worker capacity (busy ÷ threads), or coordination — and how much is recoverable without doing less work. It converts "is my sort slow?" into "which of three limits am I on?", and the three imply unrelated fixes.

Measured verdicts on the reference cell (`1kg-wgs-HG00096`, coordinate → template-coordinate, `--max-memory 512M --max-temp-files 1024`, 779,820,469 records, caches dropped, `compare bams` IDENTICAL):

| | loop | consumer serial | worker capacity | verdict | recoverable |
| 8 threads | 188.0s | 186.9s | — | consumer serial CPU | ~1s |
| 16 threads | 156.4s | 113.2s | 89.7s (16t) | coordination | 43.4s (28%) |

That table is the most useful thing to come out of this work. It is also what says the 8-thread merge is finished — it sits at 99.4% of its serial floor — and that the 16-thread gap is not worker capacity, which is 43% idle.

**A clock correction.** The sub-phase timings are sampled with an `Instant::now()`/`elapsed()` pair whose cost lands *inside* the interval — 15–35 ns against segments of 2–100 ns. Uncorrected, the segments summed to 321.5s of a 189.3s loop (−70%). The overhead is now calibrated per merge and subtracted per sample, and the partition prints a signed residual so the next such error shows up as an unattributed line rather than as plausible-looking rows.

**Quieter output.** A plain `fgumi sort` prints ~99 diagnostic lines at INFO, none of it opt-in. The consumer sub-phase rows and headroom detail move to `debug!`; the three floor lines stay at INFO because they are the part a user can act on. Its own commit, because it touches rows older than this PR — `log_consumer_rows` was extracted here and carries lines that predate it, and splitting the block by authorship would leave half a table at each level. ~72 `info!` calls elsewhere in this file deserve the same treatment and are left alone; a follow-up should put all of it behind a `--sort-stats` flag, paired with the benchmark harness that greps these lines.

## Headroom, stated plainly

This PR ships no speedup. It ships the measurement that says where the remaining time is and, more usefully, where it is not.

The 16-thread gap is 43.4s (28%), and it is precisely localized: 97% of park time is `waiting for a worker` while the work itself measures 40 µs against 1,325 µs of waiting; 89.4% of that has a worker asleep; 0.29% of block pulls carry 28% of merge wall at 2.9 ms each; the awaited file is 98% `starved` by time.

**Nineteen interventions against that picture have all measured worse or neutral.** Five arms that moved decompression off the serial consumer lost wall clock in proportion — slope **−0.373 s of merge per second of consumer fetch, r = −0.975**. A regime gate that adapted on self-serves-per-park cost +12.5% and was refuted twice over: its discriminant moved under its own action (9.2/park ungated → 31/park gated, against a threshold of 32), and the counter it divided by counts *calls* rather than blocks. Letting the consumer read the starved block itself fired on 0.6% of starved parks — `try_lock` on the reader failed ~99% of the time, because a worker was already reading exactly what the consumer wanted. Publishing that read every 16 blocks instead of every 404 moved wall clock 0.1%.

So the shipped configuration is a local optimum in every dimension tested: decompress cap (non-monotonic, optimum at the current 8), publish granularity (neutral), consumer-side read (neutral), prediction window (negative), regime gate (negative), wake targeting (`0% recoverable` by its own counter). It is not bandwidth — 52.6 GB over 156s is ~337 MB/s against a 1000 MB/s volume.

I would not spend more on the 16-thread merge on this evidence.

## Notes for review

- `merge_headroom.rs` is new and self-contained: `MergeFloors`, `ConsumerSample`, `LoopPartition`, and the clock calibration. 22 tests, all mutation-verified — two tests initially survived removing the code they covered and were rewritten.
- `LoopPartition::unattributed_secs` is deliberately **signed**. It is what caught the −70% error above; an unsigned or clamped residual would have hidden it.
- 7,745 tests pass. Instrumentation cost measured twice against a proper in-boot control at −0.34% / +0.15% / −0.66% — opposite signs, inside noise.

Risk: sort diagnostics and merge-floor metrics change output, pinned by calibrated timing, consumer samples, and floor classification; unsafe: none; CLAUDE.md allowlist: unchanged; memory bounds, queue capacity, and thread/backpressure policy: none.

- Add merge-floor reporting for consumer serial CPU, worker capacity, and coordination limits.
- Report recoverable wall-clock headroom and binding classifications.
- Partition consumer-loop timing into prediction, presentation, writing, source advancement, and loser-tree work.
- Calibrate and subtract clock overhead from sampled timings.
- Report signed residuals for unattributed loop time.
- Move detailed consumer and lifecycle diagnostics to `debug!`.
- Keep actionable floor verdicts and summary metrics at `INFO`.
- Add tests for calibration, correction, binding, residual accounting, clamping, scaling, and diagnostic reporting.

### DIFF
--- a/crates/fgumi-sort/src/external.rs
+++ b/crates/fgumi-sort/src/external.rs
@@ -4596,10 +4596,10 @@ impl RawExternalSorter {
         let (samples_taken, records_merged) = sampling;
         let est_tree = scaled.tree;
         let est_read = scaled.advance;
-        info!(
+        debug!(
             "  Consumer (main thread; {samples_taken} samples of {records_merged} records, scaled {scale:.0}x)"
         );
-        info!("    Fetch next record: {est_read:.1}s  (includes waiting on decompressed blocks)");
+        debug!("    Fetch next record: {est_read:.1}s  (includes waiting on decompressed blocks)");
         let MergeConsumerDiag {
             backpressure_secs: blocked_secs,
             backpressure_waits: blocked_waits,
@@ -4609,22 +4609,22 @@ impl RawExternalSorter {
         if blocked_waits > 0 {
             #[allow(clippy::cast_precision_loss, reason = "wait counts stay far below 2^52")]
             let mean_us = blocked_secs * 1e6 / blocked_waits as f64;
-            info!(
+            debug!(
                 "    Output backpressure: {blocked_secs:.1}s over {blocked_waits} waits \
                  (mean {mean_us:.0} us, exact) -- the consumer blocked for an output permit"
             );
         } else {
-            info!("    Output backpressure: none -- the compressors always had a permit ready");
+            debug!("    Output backpressure: none -- the compressors always had a permit ready");
         }
         if borrowed + reassembled > 0 {
             #[allow(clippy::cast_precision_loss, reason = "record counts stay far below 2^52")]
             let pct = 100.0 * reassembled as f64 / (borrowed + reassembled) as f64;
-            info!(
+            debug!(
                 "    Record presentation: {borrowed} borrowed zero-copy, {reassembled} \
                  reassembled across a block boundary ({pct:.2}%, exact)"
             );
         }
-        info!("    Loser tree:        {est_tree:.1}s");
+        debug!("    Loser tree:        {est_tree:.1}s");
         // Reconcile the sampled segments against the loop they are meant to
         // partition. Three of these rows were reported for a long time without
         // ever being summed against the loop, so there was no way to tell whether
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #825, opened 2026-08-19
ci: harden samtools install against apt mirror stalls

nh13:
## Problem

The `sort-correctness` job installs samtools with a bare `sudo apt-get update && sudo apt-get install -y samtools`. When an upstream (Azure) apt mirror stalls, `apt-get` hangs indefinitely. This step normally finishes in ~14s, but on 2026-08-19 it was observed stuck for **40+ minutes** on both a PR run and a concurrent `main` run — the same step, same time — pointing squarely at a transient mirror stall rather than any repository change.

No job in `check.yml` sets `timeout-minutes`, so a hung step runs up to GitHub's **6-hour default** before it's killed. That's why a stalled install looks like a job that just "runs forever" and has to be cancelled by hand.

## Change

Scoped to the `sort-correctness` job:

- **`timeout-minutes: 20`** on the job, so an infrastructure stall fails fast and obviously instead of masquerading as a slow job.
- **Bound each apt call** with `Acquire::Retries=3` and 30s per-request HTTP/HTTPS timeouts, so a stuck mirror errors out in seconds rather than hanging.
- **3-attempt retry loop** with 15s backoff around `update` + `install`, so a transient mirror failure self-heals within the run.
- **`DEBIAN_FRONTEND=noninteractive`** to guard against any post-install config prompt.

`actionlint` (including its shellcheck pass) is clean on the result.

## Notes

The hang was not caused by any code change — the samtools gating (#577) has run unchanged for a month and the step succeeded ~22h before the incident. This PR only makes CI resilient to the failure mode; it does not touch test or product code.

Risk: command output changes—none; `unsafe` changes—none, and no `CLAUDE.md` allowlist update is needed; memory bounds, queue capacity, and thread/backpressure policy changes—none.

- Adds a 20-minute timeout to the `sort-correctness` CI job.
- Makes `samtools` installation noninteractive, bounded, and retryable.
- Retries installation up to three times with 15-second backoff.
- Does not modify product or test code.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -281,6 +281,9 @@ jobs:
 
   sort-correctness:
     runs-on: ubuntu-latest
+    # A stalled apt mirror on the "Install samtools" step has hung this job for
+    # 30+ minutes (up to GitHub's 6-hour default). Cap it so an infrastructure
+    # stall fails fast and is obvious, instead of masquerading as a slow job.
     timeout-minutes: 20
     # This job only reads the repository -- it builds and runs tests, and never
     # writes back to GitHub. Drop the default GITHUB_TOKEN scopes to read-only so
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #827, opened 2026-08-19
test(pipeline): stabilize read-ahead backpressure assertion (#809)

### LINKED ISSUE #809: Flaky test: read_ahead_under_a_stalled_writer_shrinks_with_the_queue_memory_budget
### Summary

`test_pipeline_memory_backpressure::read_ahead_under_a_stalled_writer_shrinks_with_the_queue_memory_budget` (in `tests/integration/test_pipeline_memory_backpressure.rs`) is intermittently failing. It failed 1 of 3 local reps and has failed in CI, independent of any particular change (it reproduces on `main`).

### Symptom

```
read-ahead was 3169724 bytes under a 2097152-byte budget and 4578134 bytes under a 4194304-byte one; the gap is smaller than one tight budget, so the budget is not bounding the queues
```

The tight-budget arm is stable (~3169724 bytes across runs); the generous-budget arm varies with scheduler timing (observed 4578134 and 4929843 bytes). When the generous arm reads less far ahead, the tight-vs-generous gap dips below the assertion threshold:

```rust
assert!(
    generous.saturating_sub(tight) >= TEST_BUDGET_BYTES, // TEST_BUDGET_BYTES = 2 MiB
    ...
);
```

### Why it's flaky

The assertion requires the read-ahead gap between the two budgets to span at least one tight budget (2 MiB). How far the reader runs ahead of a stalled writer depends on scheduler timing and block boundaries, so the generous arm's read-ahead is not deterministic. On a slow/contended run it can land close enough to the tight arm that the gap falls under 2 MiB even though the budget IS bounding the queues.

### Suggested fix

Stabilize the discriminating assertion so it tolerates scheduler jitter without weakening what it proves, e.g.:
- average each arm's read-ahead over a few reps before comparing, or
- tie the required gap margin to the observed rep-to-rep variance rather than a fixed 2 MiB, or
- assert a proportional relationship (generous meaningfully exceeds tight) with a margin sized from measured spread.

### Notes

Surfaced while rebasing #772 onto `main`; the PR does not touch this test or the BAM pipeline code it exercises.

nh13:
Closes #809.

## Problem

`read_ahead_under_a_stalled_writer_shrinks_with_the_queue_memory_budget` compared a single read-ahead sample per budget arm. It still flakes on `main` — reproduced 1 failure in ~20 uncontended reps locally, and ~5% under CPU contention.

The failing shape is always the same: the generous (3 MiB) arm reads anomalously little and lands close to the tight (1 MiB) arm, so `generous - tight` dips under the `budget_difference / 2` threshold. Example failure: `tight = 1,145,011`, `generous = 1,585,185`, gap `440,174 < 1,048,576`.

## Root cause

The reader's budget-bounded stopping point is an *upper* bound it approaches from below, and the scheduler noise on it is one-sided and downward. If the reader is starved of CPU for a full second mid-read, `wait_for_consumption_to_settle` (which calls an arm settled after 4×250 ms of no observed progress) declares it settled at a *partial* read-ahead — well short of where the budget would actually stop it. The generous arm reads ~3× further than the tight arm, so it is far more exposed to that rare downward tail. Characterization confirmed both arms are well-separated and stable in the common case (tight ≈ 1.06–1.19 MB, generous ≈ 2.60–2.99 MB); the failure is a rare downward outlier on the generous arm, not the budget failing to bound the queues.

## Fix

Measure each arm's read-ahead as the **maximum over a few reps** instead of a single sample. The tail never lifts an arm above its true ceiling, so the maximum recovers that ceiling and is immune to it — without weakening what the test proves (the assertion, the sub-saturation budgets, and the `budget_difference / 2` margin are unchanged). Three reps drive the odds of every rep tailing at once low enough to be irrelevant while keeping the test cheap.

## Verification

- 40/40 passes under 2× CPU oversubscription (the contention that reproduced the tail), where a single-sample run failed ~5%.
- Full `test_pipeline_memory_backpressure` module green; `cargo ci-fmt` and `clippy --all-features --tests -D warnings` clean.
- Test-only change; no pipeline/production code touched.

Risk: command output changes: none; `unsafe` changes: none, so the `CLAUDE.md` allowlist is unchanged; memory bounds, queue capacity, or thread/backpressure policy changes: none.

The test now repeats each budget measurement three times and uses the maximum read-ahead value. This reduces scheduler-related false failures without weakening the existing assertion or budget settings. The nextest timeout for this slow test increases to four 60-second periods.

Verification passed 40/40 runs under 2× CPU oversubscription, the memory backpressure test module, formatting, and clippy.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,2 +1,13 @@
 [profile.default]
 slow-timeout = { period = "60s", terminate-after = 2 }
+
+# `read_ahead_under_a_stalled_writer_shrinks_with_the_queue_memory_budget` measures
+# each budget arm's read-ahead as the maximum over READ_AHEAD_REPS stalled-writer
+# runs (issue #809), so it spawns the stalled pipeline several times back to back,
+# each with its own settle wait. That is legitimately slow (~30s locally, and 2-3x
+# that under llvm-cov instrumentation plus CI's 2x CPU oversubscription), enough to
+# trip the default 120s (60s x 2) terminate-after. Give this one test a longer leash
+# rather than dropping reps and weakening the flake-suppression the maximum provides.
+[[profile.default.overrides]]
+filter = 'test(=test_pipeline_memory_backpressure::read_ahead_under_a_stalled_writer_shrinks_with_the_queue_memory_budget)'
+slow-timeout = { period = "60s", terminate-after = 4 }
--- a/tests/integration/test_pipeline_memory_backpressure.rs
+++ b/tests/integration/test_pipeline_memory_backpressure.rs
@@ -445,6 +445,19 @@ fn read_ahead_under_a_stalled_writer_shrinks_with_the_queue_memory_budget() {
     /// Read batch small enough that read-ahead tracks the byte budget instead of
     /// the ~1 MiB production batch quantum.
     const FINE_READ_BLOCKS: usize = 4;
+    /// Reps per arm whose maximum read-ahead is taken as that arm's measurement.
+    ///
+    /// The budget-bounded stopping point is an *upper* bound the reader
+    /// approaches from below, and the only scheduler noise on it is one-sided and
+    /// downward: if the reader is starved of CPU for a full second mid-read,
+    /// [`wait_for_consumption_to_settle`] can call the arm settled at a partial
+    /// read-ahead — well short of where the budget would actually stop it. That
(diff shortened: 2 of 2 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #829, opened 2026-08-19
perf(sort)!: cut the Phase 1 ingest thread's per-record cost

nh13:
Stacked on #826. Four per-record costs removed from `fgumi sort`'s serial Phase 1 ingest thread, measured end to end on a whole-genome sort.

## Why this thread

Phase 1 (read → in-memory sort → spill) is **60% of a spill-heavy sort's wall clock**, and **91% of it is one thread**. Measured on `1kg-wgs-HG00096` (43.1 GB, 779,820,469 records) at 16 threads on a `c7g.4xlarge`, caches dropped, with a per-thread `/proc` sampler:

| Phase 1 floor | value |
| --- | --- |
| phase wall | 240.7s |
| **consumer serial CPU** | **219.5s (91%)** ← binds |
| worker capacity (16 threads) | 65.8s |
| coordination (main thread idle) | 21.2s |

The other 15 cores average 4.4 busy. `perf` on that thread put 39.5% in `extract_template_key_inline`, 28.2% in the record memcpy, 6.1% in `core::hash::sip::Hasher::write`, 3.4% in `__aarch64_ldadd8_relax`, and 2.1% in `core::str::converts::from_utf8`.

## What changed

1. **`memchr` for aux value terminators.** Every `Z`/`H` aux value is NUL-terminated and finding that NUL is how the scan advances, so it runs once per such tag. It was a scalar `iter().position(|&b| b == 0)`; a new `nul_offset` helper wraps `memchr` and replaces all 11 sites. Per-aux-byte scan cost falls **0.400 → 0.061 ns**.
2. **`ahash` for the RG lookup.** `LibraryLookup::rg_to_ordinal` was a `HashMap<Vec<u8>, u32>` on std's default hasher, so every record paid SipHash over a ~40-byte RG id. The keys come from the header of a BAM the caller chose to sort, so there is no adversarial-input exposure that argues for SipHash here.
3. **`MC` carried as bytes.** `TemplateAuxTags::mc` and `AuxStringTags::mc` were `Option<&str>` built with `str::from_utf8(value).ok()` per record, and every consumer's first act was `.as_bytes()`. The `cigar` entry points now take `&[u8]`. **Breaking** — see the behavior note below.
4. **Batched progress counter.** `log_if_needed(1)` per record is a relaxed `fetch_add`, which on aarch64 is an outline-atomics call. `BatchedProgress` already existed for exactly this and the merge loops already used it; all three ingest loops now do.

## Measured

**399.0 → 385.2s (−13.8s, −3.5%)** on the cell above, both arms in one boot with caches dropped before each.

| | pre | fix | delta |
| --- | --- | --- | --- |
| Read + decompress | 157.1s | **144.0s** | −13.1s |
| In-memory sort | 18.9s | 17.5s | −1.4s |
| Spill write | 64.9s | 65.3s | +0.4s |
| K-way merge | 156.4s | 155.7s | −0.7s |
| **Total wall** | **399.0s** | **385.2s** | **−13.8s (−3.5%)** |
| Peak RSS | 10.587 GB | 10.573 GB | −14 MB |

From the per-thread sampler, which is where the mechanism is: **main-thread CPU 218.8 → 197.7s (−21.1s)**, main-thread busy share **91% → 87%**, merge wall **unchanged to a tenth of a second** — a clean control that the changes are confined to Phase 1.

Two boots agree to **0.15%** on total wall (399.6 / 399.0s), so the delta is ~23× the run-to-run spread.

**Only 64% of the removed CPU converts to wall, and the reason is worth knowing for whoever picks this up next.** `try_read_input_blocks` takes `input_file.try_lock()`, so exactly one worker reads at a time. That step is a fixed 125.9s over 319,651 batches, unchanged by this PR, so inside a shrinking ingest span its duty cycle rises from **79% to 86%**. The next tranche of serial CPU would convert at a worse rate than this one did. Its per-block cost (24.3 µs to frame an 8.4 KB block) is not explained by transfer — 43.1 GB at the measured 605 MB/s single-stream ceiling is ~71s against 125.9s of occupancy — and is the next thing to measure.

## Correctness

`fgumi compare bams --command sort` on the two 48 GB outputs: **IDENTICAL**, 779,820,469 records, zero sort-order violations in either arm, zero run-multiset mismatches. Record counts, spill volume (52,614.5 MB) and spill-run count (44) are identical across every arm. The comparison binary was built from the pre-change tree so it cannot be biased toward these changes.

`cargo ci-fmt`, `cargo ci-lint`, `cargo ci-test`: **7754 passed**, 30 skipped.

## Behavior change (one, deliberate)

A non-UTF-8 `MC` tag used to be discarded, leaving the mate position unadjusted; it now reaches the clip parsers. The observable difference is narrow: the parsers read a leading run of ASCII digits and stop at the first byte they cannot interpret, so a value that is garbage from its first byte still yields zero clips — the same answer discarding it gave. Only a value with a *valid CIGAR prefix* followed by invalid bytes changes, and it changes toward parsing the prefix, which is what samtools does.

Both byte-level accessors (`find_mc_tag`, `find_mc_tag_in_record`) drop the gate as well, so the two extraction paths stay in step. That matters because `src/lib/grouper.rs` cross-checks them: `test_mc_presence_matches_key_shape` and the `validate_mc_tag` doc both state that duplicate `MC` entries are the **only** case where the byte scan and the decoded key shape may disagree. Gating one path and not the other would have made a non-UTF-8 `MC` a second, undocumented divergence and changed `fgumi group`'s key for that record class.

New tests pin all of it: `test_extract_template_key_mate_lane_comes_from_mc` asserts the MC-derived mate lane equals the key of a record whose mate is already reported at the unclipped position, and `test_extract_template_key_mate_lane_parses_a_non_utf8_mc_prefix` pins the new behavior at the key level, with an `assert_ne!` against what discarding the tag would have produced.

## How to reproduce

```
fgumi sort -i 1kg-wgs-HG00096.bam -o out.bam \
    --order template-coordinate --threads 16 --merge-threads 16 \
    --max-memory 512M --max-temp-files 1024 --temp-compression 1 \
    --compression-level 1 --tmp-dir /var/tmp/work --sort-stats
```

`cargo bench -p fgumi-sort --bench template_key` sizes the key path alone. The bench builds records to the aux layout the measured sample actually carries, with a second arm carrying the long `XA:Z` tag so the per-aux-byte slope falls out of the pair — the reason its numbers are trustworthy is that the baseline it reports (77.9 ns/record on an M2 Max) sits at the expected microarchitectural ratio to the 111 ns/record the `c7g` profile implies. It is also what stopped the memchr change being oversold: the profile share implied ~25 ns/record, the bench measured 7.7 ns weighted, because memchr's per-call setup eats most of the gain on four short `Z` values and `XA` is on only 1.2–2.0% of real records.

Risk: output changes: none for grouping, consensus, sort order, corrected UMIs, or metrics; `unsafe` changes: none, and CLAUDE.md allowlist changes: none; memory bounds, queue capacity, and thread/backpressure policy changes: none.

Reduced serial Phase 1 ingest cost with `memchr`, `ahash`, byte-based `MC` handling, batched progress updates, and borrowed indexed records.

Added Phase 1 timing diagnostics and a template-key benchmark. Removed per-record merge-consumer atomic updates.

Valid CIGAR prefixes in non-UTF-8 `MC` tags now parse without UTF-8 validation. Formatting, linting, and tests passed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -849,6 +849,7 @@ dependencies = [
  "criterion",
  "fgumi-dna",
  "fgumi-tag",
+ "memchr",
  "noodles",
  "proptest",
  "rstest",
--- a/crates/fgumi-raw-bam/Cargo.toml
+++ b/crates/fgumi-raw-bam/Cargo.toml
@@ -13,6 +13,7 @@ anyhow = { workspace = true, optional = true }
 bytemuck = { workspace = true }
 fgumi-dna = { workspace = true }
 fgumi-tag = { workspace = true }
+memchr = { workspace = true }
 wide = { workspace = true }
 
 [features]
--- a/crates/fgumi-raw-bam/src/cigar.rs
+++ b/crates/fgumi-raw-bam/src/cigar.rs
@@ -295,7 +295,7 @@ pub fn unclipped_5prime_raw(bam: &[u8], pos: i32, is_reverse: bool) -> i32 {
 /// For reverse strand mate: `unclipped_end`
 #[inline]
 #[must_use]
-pub fn mate_unclipped_5prime(mate_pos: i32, mate_reverse: bool, mc_cigar: &str) -> i32 {
+pub fn mate_unclipped_5prime(mate_pos: i32, mate_reverse: bool, mc_cigar: &[u8]) -> i32 {
     if mate_reverse {
         unclipped_other_end(mate_pos, mc_cigar)
     } else {
@@ -311,7 +311,7 @@ pub fn mate_unclipped_5prime(mate_pos: i32, mate_reverse: bool, mc_cigar: &str) 
 pub fn mate_unclipped_5prime_1based(
     mate_pos_0based: i32,
     mate_reverse: bool,
-    mc_cigar: &str,
+    mc_cigar: &[u8],
 ) -> i32 {
     let mate_pos_1based = mate_pos_0based.saturating_add(1);
     if mate_reverse {
@@ -335,8 +335,8 @@ pub fn mate_unclipped_5prime_1based(
 /// `samtools sort --template-coordinate`.
 #[inline]
 #[must_use]
-pub(crate) fn unclipped_other_start(mate_pos: i32, mc_cigar: &str) -> i32 {
-    mate_pos.saturating_sub(parse_leading_clips(mc_cigar.as_bytes()))
+pub(crate) fn unclipped_other_start(mate_pos: i32, mc_cigar: &[u8]) -> i32 {
+    mate_pos.saturating_sub(parse_leading_clips(mc_cigar))
 }
 
 /// Calculate mate's unclipped end from MC tag CIGAR string.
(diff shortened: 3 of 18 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #830, opened 2026-08-19
perf(zipper): probe tag membership with a 256×256 bitset

nh13:
## Summary

The per-record tag-copy loop in `merge_raw` (zipper) probed three `HashSet<String>` sets (remove/reverse/revcomp) once per aux tag per mapped read — each lookup paying a SipHash of the tag name plus a UTF-8 conversion of the two raw tag bytes. This replaces the hot-path probe with a direct bit test on the two tag bytes via a new private `TagBitset` (256×256 bit table), built once per run into a `ZipperTags` struct and reused for every template.

## Approach

- `TagBitset` — a 256×256 bit table (1024 `u64` words) keyed by the two raw tag bytes; `contains` is a single bit test, no hash and no UTF-8 conversion. Only two-byte tag names are representable (all SAM tags are two bytes), so longer/shorter filter strings are dropped at construction — mirroring the prior `tag_str.len() == 2` guard.
- `ZipperTags` — precomputes the three bitsets plus the two-byte `remove_list` (for the Step-2 tag removal) and `has_transforms`, once per run.
- `merge_raw(&TagInfo)` stays as a thin public wrapper that builds `ZipperTags` and delegates to a new internal `merge_raw_with`; the hot per-template path in `process_raw` builds the lookups once and calls `merge_raw_with` directly. The public API and the `fgumi-umi` `TagInfo` type are unchanged.

## Measured impact (M3 Ultra, aarch64)

- Isolated probe: **29.6 ns/tag → 1.1 ns/tag** (populated Consensus sets), ~27× faster. The bitset also decisively beats a `HashSet<[u8;2]>` (which still hashes).
- Real per-record tag-copy loop (18-tag consensus record, ~3.5 KB aux, real `remove_tag`+append, neg strand): **−23%** (three runs, 20–26%). The loop is otherwise dominated by `remove_tag`+append over the consensus aux, so the probe is ~a quarter of it.
- The win scales with tags/record on the single-threaded tag-copy path; it is a **Consensus-config / many-tag** win. On the light default path (empty tag sets, few tags) the loop is BGZF-I/O-bound and the change is neutral (see below) — not a regression.

## Behavior

Tags are matched by their raw two bytes. For all spec-conforming SAM tags (ASCII `[A-Za-z][A-Za-z0-9]`) this is identical to the old `from_utf8`-based probe. The only divergence is for a non-UTF-8 tag byte pair (reachable only from a malformed BAM) combined with an empty-string filter: the old path aliased it to `""` and could match an empty filter, whereas here the raw bytes are matched directly. This is the intended behavior; it is documented on `TagBitset` and pinned by a test.

## Validation

- Full suite (7710 tests) + clippy `--pedantic -D warnings` + the bare-tag-literal check all green. Added unit tests for `TagBitset` membership (case-sensitivity, full 0..=255 byte range, two-byte filtering, empty-filter semantics) and `ZipperTags::from_tag_info`.
- **Real-data parity:** old vs new binaries on a 721,681-record real zipper run (`SRR6109273.aligned.bam` + `.extract.fgbio.bam`, `-r hs38DH`, `-t 1`) produce **byte-identical records** — the decompressed streams differ in exactly 46 of 393,397,066 bytes, reproducibly, and those 46 bytes are entirely the `@PG VN:` build commit hash. (Zipper output is deterministic: two runs of the *same* binary to the *same* output path are byte-identical — an earlier "1-byte per-run wobble" note was a measurement artifact, the differing byte being the `-o` filename recorded verbatim in the `@PG CL:` field.)
- **Real-data wall (default config, I/O-bound):** old 2.32/2.32/2.43 s vs new 2.32/2.35/2.34 s — neutral, within noise, as expected for empty tag sets and ~2 tags/record.

Risk: command output changes — none; `unsafe` changes — none, and the `CLAUDE.md` allowlist is unchanged; memory bounds, queue capacity, and thread/backpressure policy changes — none. The raw-byte `TagBitset` preserves existing two-byte tag filtering and transformation behavior. Tests cover membership, byte ranges, invalid-length filters, and lookup construction.

### DIFF
--- a/src/lib/commands/zipper.rs
+++ b/src/lib/commands/zipper.rs
@@ -403,6 +403,103 @@ fn collect_mapped_indices(mapped: &Template, is_first_segment: bool) -> Vec<usiz
     indices
 }
 
+/// A membership set over two-byte SAM tag names, backed by a 256×256 bit table.
+///
+/// Replaces a `HashSet<String>` on the zipper hot path. `merge_raw` probes tag
+/// membership once per aux tag per mapped read; a `HashSet<String>` lookup there
+/// costs a `SipHash` of the tag name plus a UTF-8 conversion of the two raw tag
+/// bytes. A direct bit test on those two bytes removes both. Measured at ~23% of
+/// the per-record tag-copy loop on a consensus-config zipper (M3 Ultra), where
+/// the bitset also decisively beats a `HashSet<[u8; 2]>` (which still hashes).
+///
+/// All SAM tags are exactly two bytes, so only two-byte names are representable;
+/// any longer or shorter filter string can never equal a real tag and is dropped
+/// at construction — matching the pre-existing `tag_str.len() == 2` guard.
+///
+/// Membership is tested against the tag's **raw two bytes**. The prior code
+/// converted the bytes with `str::from_utf8(..).unwrap_or("")` before a
+/// `HashSet<String>` lookup; for spec-conforming tags (ASCII `[A-Za-z][A-Za-z0-9]`)
+/// the two are identical. They differ only for a non-UTF-8 tag byte pair (reachable
+/// only from a malformed BAM): the old path aliased it to the empty string and so
+/// could match an empty (`""`) filter, whereas here the raw bytes are matched
+/// directly and an empty filter — being length ≠ 2 — is never inserted. This is the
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #831, opened 2026-08-19
perf(filter): batch the interim progress counter out of the per-record path

nh13:
## Summary

`fgumi filter`'s single-read pipeline mode did `progress.fetch_add(1, Relaxed)` on an `Arc<AtomicU64>` **once per record**, and that atomic is shared across all pipeline worker threads — so it was real cacheline ping-pong, not just an outline-atomics call. This moves the interim "Processed N records" heartbeat into `serialize_fn` (which already runs once per batch and updates the metrics slots), and drops the per-record `fetch_add` from the single-read `process_fn`, the per-batch `fetch_add` from the template `process_fn`, and the now-unused `ProcessCaptures.progress` field.

## Why it's safe

The progress counter is **log-only**. The authoritative record total is aggregated from each batch's `records_count` and logged at completion (`Processed N reads; kept … rejected …`), independent of this counter. So the change alters only the interim heartbeat's cadence — now emitted per batch in output order instead of per record — and nothing in the output. Full suite (7703 tests) + clippy-pedantic + docs + tag-literals green.

## Measured impact (M3 Ultra, aarch64)

Cost of a contended per-record `fetch_add` on one shared atomic vs the batched form, ns/record:

| worker threads | contended | batched |
| --- | --- | --- |
| 1 | 1.8 | ~0 |
| 4 | 6.0 | ~0 |
| 8 | ~15 | ~0 |
| 16 | ~60 | ~0 |

The cost is cacheline ping-pong that **scales with worker count** — negligible single-threaded, but ~15 ns/rec at 8 workers and ~60 ns/rec at 16. On a many-worker `filter` run this is a real per-record tax that batching removes entirely. (The template mode already batched at the sibling site; this brings the single-read mode in line and unifies both into one place.)

Risk: output for all commands is unchanged; `unsafe` is none and `CLAUDE.md` changes are none; memory bounds, queue capacity, and thread/backpressure policy changes are none.

Fix: Batch progress updates in `serialize_fn` with a shared `AtomicU64`. Remove per-record and per-template atomic updates and the unused `ProcessCaptures.progress` field. Add heartbeat accumulation tests across progress boundaries.

Validation: Full tests, clippy-pedantic, documentation checks, and tag-literals checks passed.

### DIFF
--- a/src/lib/commands/filter.rs
+++ b/src/lib/commands/filter.rs
@@ -283,7 +283,6 @@ struct FilterProcessCaptures {
     min_conversion_fraction: Option<f64>,
     methylation_mode: fgumi_consensus::MethylationMode,
     ref_names: Arc<Vec<String>>,
-    progress: Arc<AtomicU64>,
     header: Header,
 }
 
@@ -445,7 +444,6 @@ impl Filter {
                 self.methylation_mode,
             ),
             ref_names: Arc::new(ref_names),
-            progress: Arc::clone(&setup.progress_counter),
             header: header.clone(),
         }
     }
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #832, opened 2026-08-19
perf: avoid three per-record heap allocations (group / downsample / overlapping consensus)

nh13:
## Summary

Three independent per-record heap allocations, each a short buffer built only to be compared once and dropped. Removing them is a small, zero-risk cleanup (measured at ~9–10 ns/record each under mimalloc — the allocator's per-alloc bookkeeping, not the copy). One commit per site; suggested reading order below.

## Commits

1. **`perf(group)`** — `group_by_name_and_build` allocated a QNAME `Vec` (`read_name().to_vec()`) for every record just to compare it against the current group's name. Now compares via a borrow and copies the owned name only on a group change (the borrow must end before `extract` consumes the decoded record) — mirroring the sibling in `TemplateGrouper::add_records`.
2. **`perf(downsample)`** — `FamilyIterator::next_family` allocated a `String` per record's MI tag (`get_mi_tag`) only to compare it against the family key. New `mi_tag_equals` compares in place: a Z-typed MI compares bytes directly; the legacy integer form falls back to rendering. The family-key `String` is still built once per family. A test cross-checks `mi_tag_equals` against `get_mi_tag` for the Z, integer, and missing-MI cases.
3. **`perf(consensus)`** — `apply_overlapping_consensus` built its read-pair map with `entry(read_name().to_vec())`, allocating a QNAME `Vec` for every record and discarding it whenever the key already existed (the second read of each pair). Now looks the name up by borrow and copies an owned key only when inserting a new pair.

## Safety

All three are behavior-preserving. (1) and (2) keep the exact comparison/error semantics — (2)'s test pins parity with `get_mi_tag`. (3)'s map is read via `.values()` only, so iteration order is never observed and the `get_mut`/`insert` split is equivalent to the old `entry().or_insert()`. Full suite (7704 tests) + clippy-pedantic + docs + tag-literals green.

## Scale

Each site is ~10 ns/record — individually minor, but zero-risk and on default-on paths (grouping every template; downsample per family member; overlapping consensus with `--consensus-call-overlapping-bases`, which defaults to `true`). Part of a per-record-hot-path audit; the larger wins (zipper tag probe, filter progress atomic) ship separately.

Risk: command output changes: none; `unsafe` changes: none; memory bounds, queue capacity, and thread/backpressure policy changes: none.

- Avoids per-record QNAME allocation during template grouping and overlapping-consensus pairing.
- Avoids per-record MI `String` allocation during downsampling.
- Preserves grouping, pairing, consensus, hash-collision checks, UTF-8 validation, and missing-tag errors.
- Adds MI comparison tests for matching, mismatching, prefix, integer, and error cases.

### DIFF
--- a/crates/fgumi-consensus/src/overlapping.rs
+++ b/crates/fgumi-consensus/src/overlapping.rs
@@ -641,11 +641,25 @@ pub fn apply_overlapping_consensus(
             continue;
         }
 
-        let name = view.read_name().to_vec();
+        // Look the name up by borrow and only copy it into an owned key when a
+        // new pair is inserted. `entry(name.to_vec())` would allocate a QNAME
+        // `Vec` for every record and then discard it whenever the key already
+        // exists (i.e. on the second read of each pair).
+        let name = view.read_name();
         if flg & fgumi_raw_bam::flags::FIRST_SEGMENT != 0 {
-            read_pairs.entry(name).or_insert((None, None)).0 = Some(idx);
+            match read_pairs.get_mut(name) {
+                Some(pair) => pair.0 = Some(idx),
+                None => {
+                    read_pairs.insert(name.to_vec(), (Some(idx), None));
+                }
+            }
         } else if flg & fgumi_raw_bam::flags::LAST_SEGMENT != 0 {
-            read_pairs.entry(name).or_insert((None, None)).1 = Some(idx);
+            match read_pairs.get_mut(name) {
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #833, opened 2026-08-19
fix(clip): preserve non-UTF-8 string tag bytes when clipping

nh13:
## Summary

`SamRecordClipper` clipped a length-changing `Z` (string) tag by round-tripping the value bytes through `&str`, losing any non-UTF-8 bytes:

- `clip_extended_attributes` used `from_utf8(&bytes[start..end]).unwrap_or("")` — a non-UTF-8 tag value was **silently replaced with an empty string**.
- `upgrade_all_clipping` (soft→hard) used `String::from_utf8_lossy(..)` — non-UTF-8 bytes were **mangled to U+FFFD**.

noodles' `Value::String` is a byte string (it round-trips arbitrary bytes, e.g. `Value::String(vec![0xFF, 0xFE].into())`), so both sites now slice the bytes directly and preserve the clipped value verbatim.

## Scope

- Two sites, same class — fixed together. The production raw-byte path (`RawRecordClipper`, used by `fgumi clip`) was already byte-safe; this is the noodles-backed `SamRecordClipper`, which is **exported public API** (`pub use clipper::SamRecordClipper`) but currently has no in-tree production caller — so this is a latent correctness fix rather than a live bug.
- Regression tests added for both the 5'/auto-clip path and the soft→hard `upgrade_all_clipping` path: a non-UTF-8 tag whose length matches the read is clipped and asserted byte-for-byte (the old code would have emptied / mangled it).

## Validation

Full suite (7704 tests) + clippy-pedantic + docs + tag-literals green.

Risk: command output changes for clipped `Z` tags, pinned by regression tests; `unsafe` changes: none; memory bounds, queue capacity, and thread/backpressure policy: none.

`SamRecordClipper` now clips `Z` tags by byte range. This preserves non-UTF-8 bytes during 5′/auto clipping and `upgrade_all_clipping`. Regression tests cover both paths. Full validation passed.

### DIFF
--- a/crates/fgumi-sam/src/clipper.rs
+++ b/crates/fgumi-sam/src/clipper.rs
@@ -176,8 +176,13 @@ impl SamRecordClipper {
                 let (start, end) = if from_start { (remove, old_length) } else { (0, new_length) };
                 let new_value = match value {
                     Value::String(s) => {
+                        // Preserve the clipped bytes verbatim. The previous
+                        // `from_utf8(..).unwrap_or("")` silently replaced a
+                        // non-UTF-8 tag value with an empty string; noodles'
+                        // `Value::String` holds a byte string, so slice it
+                        // directly instead of round-tripping through `&str`.
                         let bytes: &[u8] = s.as_ref();
-                        Value::from(std::str::from_utf8(&bytes[start..end]).unwrap_or(""))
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #838, opened 2026-08-19
feat(sort): split the input reader's time between the disk and framing

nh13:
Stacked on #837.

`ReadInputBlocks` is Phase 1's second serial resource — `exclusive_step_for` gives worker 0 sole ownership, so one thread frames the whole input — and its cost could not be attributed to anything.

## Why the timer has to sit below the buffer

The 2 MiB refill happens **inside** `read_raw_blocks`, on whichever block exhausts the buffer. Timing the step therefore measures framing and I/O fused together, and no per-block average separates them.

`TimedReader` wraps the input file *inside* the `BufReader` and times every underlying `read()` with a byte count. That is what makes `ReaderReport::framing_secs` — the batch span minus its refills — mean anything.

## The answer

`1kg-wgs-HG00096` (43.1 GB, 779,820,469 records, template-coordinate, t16, caches dropped):

```
Phase 1 input reader: 125.4s over 319651 batches (24.5 us/block, 5114408 blocks)
  refill reads: 120.6s over 20569 calls (5.86 ms each, 358 MB/s, 43.1 GB)
  framing:      4.4s (0.9 us/block)
  dispatch:     0.3s (0.1 us/block)
  unattributed: 0.1s (0% of the step)
```

**96% of the step is inside `read()`**, and the residual closes to 0.1s. This refutes every remaining idea about the reader's own code — removing the per-block `Vec` entirely would buy ~0.3s of 125.4s.

Independently confirmed by the `read_ladder` harness in the parent PR: raw `read()` with nothing downstream is 358 MB/s / 120.38s against the production reader's 358 MB/s / 120.6s — identical to three digits, so there is no backpressure loss either.

## Details worth knowing

- Batches are recorded **past** the empty check, so the partition covers exactly the calls `record_step` charges to the step: it is timed on `StepResult::Success` alone and the EOF-detecting call returns `InputEmpty`. Recording it would put time in the parts that is not in the total and drive the residual negative for no reason.
- Only the synchronous input paths are wrapped. `--async-reader` moves the read onto a prefetch thread, so a zero refill row on an async run is correct rather than a measurement failure.
- Timed exactly rather than sampled: ~320k batches and ~21k refills against a ~30 ns clock is under 30 ms of measurement on a 125s step, five orders of magnitude below the quantity.
- `framing_secs` and `residual_secs` are **signed**. A small negative framing figure is a known benign case — the header parse refills once before Phase 1 starts, so a few ms of refill can sit outside every timed batch. A large negative means something else and should not be explained away.

The report is gated behind `--sort-stats`, like the rest of the Phase 1 diagnostics.

Risk: adds `--sort-stats` timing output, pinned by reader timing and report calculations; no `unsafe` changes; no memory bound, queue capacity, or thread/backpressure policy changes.

Fix: none required. Phase 1 reader timing now separates disk refills, framing, dispatch, and residual time. `TimedReader` instruments synchronous readers, while async readers remain unwrapped because they read on the prefetch thread. `ReaderReport` exposes throughput and per-block metrics through `--sort-stats`. Tests cover timing partitioning, signed residuals, throughput, and failed reads.

### DIFF
--- a/crates/fgumi-sort/src/external.rs
+++ b/crates/fgumi-sort/src/external.rs
@@ -360,7 +360,65 @@ impl SortPhaseTimer {
                 ingest.spill_waits
             );
         }
+        Self::log_reader_partition(phase1.reader);
         self.log_ingest_partition(phase1);
+    }
+
+    /// What the input reader's exclusively-owned thread spent its time on.
+    ///
+    /// The reader is Phase 1's *second* serial resource and the one the floor
+    /// line cannot see: `phase1_input_busy_secs` folds it in with decompression
+    /// and divides by the thread count, which is right for a step any worker may
+    /// run and wrong for one only worker 0 ever runs. So a reader at its limit
+    /// disappears into a worker-capacity figure that looks comfortable.
+    ///
+    /// The split that matters is refill against framing, because they have
+    /// nothing in common: refill time is the disk (compare it to the volume's
+    /// measured ceiling) and framing time is header parse, per-block allocation
+    /// and the body copy (compare it to the `raw_block_read` bench). Reporting
+    /// only the step total gives a number -- 24.3 us/block on the production
+    /// cell -- that is consistent with either being the whole cost.
+    #[allow(clippy::cast_precision_loss, reason = "call and byte counts stay below 2^52")]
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #842, opened 2026-08-20
feat(bam-io): unify the single-threaded raw-BAM reader onto fgumi-bgzf (#800)

### LINKED ISSUE #800: Unify the raw-BAM reader onto fgumi-bgzf (retire the second noodles-bgzf decode path)

nh13:
Closes #800. Supersedes #801 (auto-closed when its base branch `786/nhomer/perf-bgzf-crc-skip` was deleted by the #798 squash-merge; GitHub cannot reopen a PR whose base branch was deleted). Same head branch, rebased onto `main` — the one now-redundant doc commit dropped as already upstream, leaving the single substantive change.

## Problem

`fgumi-bam-io` had two BGZF decoder stacks. The single-threaded raw-BAM reader (`create_raw_bam_reader[_with_opts]`) decoded via **noodles-bgzf**, which always verifies CRC32 and has no skip knob — so the `--check-crc`/`--no-check-crc` policy from #798 (backed by our own `fgumi-bgzf`) never reached the single-threaded fast paths, and `PipelineReaderOpts.verify_crc` was a dead carrier field.

## Change

Add `FgumiBgzfReader`, a safe streaming `Read`/`BufRead` adapter over `fgumi-bgzf`'s `read_raw_blocks` + `decompress_block_into_opts` (no `unsafe`; one block decoded per refill, so a CRC/size error is tied to the bytes the caller actually reads), and route the single-threaded raw reader through it via a new `BgzfReaderEnum::Fgumi` arm. `PipelineReaderOpts.verify_crc` is now live.

Coverage after this change (`--check-crc`/`--no-check-crc`):
- **`clip`/`codec`/`duplex`/`simplex`/`downsample`** — honored single- or multi-threaded (their single-threaded reader now decodes through fgumi-bgzf).
- **`correct`/`group`** — honored only with `--threads N`; their single-threaded mode reads through `create_bam_reader_for_pipeline_with_opts` + a manual noodles decode, so they keep the noodles path (help text says so).
- `threads > 1` still uses noodles' multithreaded decoder (a threaded fgumi-bgzf raw reader is out of scope).

`fgumi-sort` and `extract` have their own reader stacks and remain follow-ups.

## Tests

TDD: the reader-level `verify_crc` test (corrupted-CRC file accepted with `verify_crc=false`, rejected with `true`) was written first and confirmed red against the noodles path before implementing. Plus a multi-block roundtrip test (exercises block-boundary buffering) and three end-to-end `downsample` CRC tests (`--no-check-crc` accepts a corrupted file; default and `--check-crc` reject it). The reader also gains a sticky-poison test: after a mid-stream decode failure both `read` and `fill_buf` stay in error. All 228 `fgumi-bam-io` + `fgumi-bgzf` tests green on the rebased tree.

### DIFF
--- a/crates/fgumi-bam-io/Cargo.toml
+++ b/crates/fgumi-bam-io/Cargo.toml
@@ -35,3 +35,7 @@ tempfile = { workspace = true }
 rstest = { workspace = true }
 proptest = { workspace = true }
 criterion = { workspace = true }
+
+[[bench]]
+name = "bgzf_decode"
+harness = false
--- a/crates/fgumi-bam-io/benches/bgzf_decode.rs
+++ b/crates/fgumi-bam-io/benches/bgzf_decode.rs
@@ -1,0 +1,82 @@
+//! Decode-throughput benchmark for the single-threaded raw-BAM reader unify
+//! (#800): the fgumi-bgzf streaming decoder (`FgumiBgzfReader`) — with CRC32
+//! verification on and off — versus noodles' single-threaded BGZF reader, all
+//! decoding the *same* BGZF byte buffer. This isolates the decode-path change
+//! the unify makes on the single-threaded fast paths.
+
+use std::hint::black_box;
+use std::io::{Cursor, Read};
+
+use criterion::{Criterion, Throughput, criterion_group, criterion_main};
+use fgumi_bam_io::FgumiBgzfReader;
+use fgumi_bgzf::InlineBgzfCompressor;
+
+/// Build a moderately compressible payload of `len` bytes (BAM-like: a mix of a
+/// small alphabet so blocks actually compress, deterministic so runs compare).
+fn make_payload(len: usize) -> Vec<u8> {
+    let mut out = Vec::with_capacity(len);
+    let mut state: u32 = 0x1234_5678;
+    while out.len() < len {
+        // xorshift for cheap deterministic pseudo-randomness
+        state ^= state << 13;
+        state ^= state >> 17;
+        state ^= state << 5;
+        // Fold into a small-ish alphabet so the deflate stage has real work.
+        out.push((state & 0x3F) as u8 + b' ');
+    }
+    out.truncate(len);
+    out
+}
+
+/// Compress `payload` into a complete BGZF byte stream (blocks + EOF marker).
+fn make_bgzf(payload: &[u8], level: u32) -> Vec<u8> {
+    let mut compressor = InlineBgzfCompressor::new(level);
+    compressor.write_all(payload).expect("compress payload");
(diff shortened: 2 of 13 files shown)


nh13:
@coderabbitai pause

Issue #715, opened 2026-08-05
Output-collision detection is approximate, and --stats/--metrics are unguarded


nh13:
PR #689 added `reject_colliding_outputs`, which refuses `--output` and `--rejects` when both name stdout or both resolve to the same regular file. Three things it deliberately left open, plus one duplicate it uncovered.

## 1. Output collision is approximated, not established

`reject_colliding_outputs` compares `(canonicalized parent, file name)`. That catches `out.bam` against `./out.bam` and `sub/../out.bam`, and it is the most that can be resolved before the file exists — `fs::canonicalize` fails with `ENOENT` on a path that does not exist yet, which is the usual case for an output.

It does not catch:

- two symlinks pointing at one target, or one symlink and its target (only the *parent* is canonicalized, not the final component)
- two hard links to one inode
- `Out.bam` against `out.bam` on a case-insensitive filesystem (macOS APFS default)

A true `dev` + `ino` comparison is the only thing that closes these, and it needs both files to exist — by which point both `File::create` calls have already truncated. Closing the gap properly means moving the check into the writer layer, where the handles are: have `open_output_writer` return something carrying its `dev`/`ino`, and reject the second open that matches an already-open output. That is an API change across the five `create_*_bam_writer` helpers plus the pipeline, which is why it was not done in #689.

## 2. `--stats` / `--metrics` have the same collision, unguarded

`reject_colliding_outputs` is only called with `--rejects`. The same defect exists one flag over and is arguably worse, because the write happens *after* the BAM is complete:

- `simplex.rs:481` finishes the output BAM
- `simplex.rs:496-500` then `DelimFile::write_tsv`s the stats to `stats_path`

So `fgumi simplex -o out.bam --stats out.bam` writes a correct BAM and then truncates it to a TSV. Same for the other `StatsOptions` / metrics consumers (`common.rs:580`, `filter.rs`, `group.rs`, `clip.rs`, `dedup.rs`).

Confirmed by source ordering; not yet reproduced end to end. The fix is presumably to call the same guard for every secondary output path a command accepts, not just BAM ones — which may be better expressed as "collect every output path this command will open, and reject duplicates" rather than a pairwise check per flag.

## 3. `open_pipeline_output` duplicates the stdout dispatch without the block-buffering

[...]

PR #845, opened 2026-08-20
fix(cli): reject colliding output paths across every command output

nh13:
## Problem

PR #689 added `reject_colliding_outputs`, but it only compared `--output` against `--rejects`, and only in the consensus commands. The same collision — two of a command's outputs resolving to one destination — was unguarded for `--stats`, `--metrics`, histograms, and the `--metrics PREFIX` expansions. It is arguably worse there, because the secondary write happens *after* the BAM is complete:

```
fgumi simplex -o out.bam --stats out.bam   # writes a correct BAM, then truncates it to a TSV
```

The inventory across the CLI found ~11 commands with 2–5 output paths and no guard across the other pairs (e.g. `group`'s `-M` prefix files landing on `--output`, `dedup`'s three metrics files, `downsample`'s two histograms).

Addresses **#715 (items 2 and 3)**.

## What changed

**Generalized the guard (item 2).** `reject_colliding_outputs(output, secondary, flag)` becomes a set-based `reject_output_collisions(&[(path, flag)])`: given every output a command will open as `(path, flag-label)` pairs, it rejects any two that resolve to the same destination. stdout may be named by at most one target; the null device stays exempt; the identity comparison and its error messages reuse the existing pre-open `(canonical parent, file name)` logic, with wording neutralized for non-BAM outputs.

**Wired every multi-output command** to hand the guard its full output list before any writer opens:

| command | outputs guarded |
| --- | --- |
| simplex / duplex / codec / filter | `--output`, `--rejects`, `--stats` |
| correct | `--output`, `--rejects`, `--metrics` |
| downsample | `--output`, `--rejects`, `--histogram-kept`, `--histogram-rejected` |
| group | `--output`, `-f`, `-g`, and the three `--metrics PREFIX` files |
| dedup | `--output`, `--metrics`, `--family-size-histogram`, `--duplication-ladder` |
| clip | `--output`, `--metrics` |

**Fixed `open_pipeline_output` (item 3).** It re-implemented the stdout dispatch and returned a `LineWriter`-backed `std::io::stdout()`, which tears every BGZF flush at each `0x0a`. It now delegates to `open_output_writer` — the single place the `-` convention is honoured for BAM output, which returns a block-buffered stdout handle.

## Tests

- **Unit** (`common.rs`): the existing output-vs-rejects cases, migrated to the set signature, plus new cases for a collision between *any* two outputs (`--stats` on `--output`, `--metrics` on `--rejects`), an all-distinct set, an empty set, and more-than-one-stdout.
- **End-to-end** (`test_streaming_output.rs`): a new case per command (simplex/duplex/codec/filter `--stats`; correct/dedup/clip `--metrics`; group `--family-size-histogram`) runs `-o out.bam <flag> out.bam` and asserts it is rejected, the error names the flag and the path, and the output BAM is **never created** (proving the guard fires before `File::create` truncates).

Verified: full suite (`cargo ci-test`) 3532 passed / 0 failed; `cargo ci-lint` (`-D warnings`) and `cargo ci-fmt` clean.

## Deliberately out of scope

- **Item 1 (dev/ino).** Moving identity into the writer layer (`open_output_writer` returns `dev`/`ino`; reject a second open matching an already-open inode) to catch symlink / hardlink / case-insensitive collisions a pre-open path comparison cannot. The author deferred it from #689 as an API change across the five `create_*_bam_writer` helpers plus the pipeline; it belongs in its own PR, so **#715 stays open** for it.
- **`simulate` commands.** `fastq-reads` (`-1`/`-2`/`--truth`) and `grouped-reads` (`-o`/`--truth`) are feature-gated test tooling where a collision yields a bad test file rather than silent production data loss. Easy to add if wanted.

### DIFF
--- a/src/lib/commands/clip.rs
+++ b/src/lib/commands/clip.rs
@@ -499,6 +499,14 @@ impl ClipParams {
 impl Command for Clip {
     #[allow(clippy::too_many_lines)]
     fn execute(&self, command_line: &str) -> Result<()> {
+        // Reject two outputs resolving to one destination before any writer opens.
+        let mut outputs: Vec<(&std::path::Path, &str)> =
+            vec![(self.io.output.as_path(), "--output")];
+        if let Some(path) = &self.metrics {
+            outputs.push((path.as_path(), "--metrics"));
+        }
+        crate::commands::common::reject_output_collisions(&outputs)?;
+
         // Validate the input exists (stdin paths are exempt).
         self.io.validate()?;
         validate_file_exists(&self.reference, "Reference FASTA")?;
--- a/src/lib/commands/codec.rs
+++ b/src/lib/commands/codec.rs
@@ -25,11 +25,12 @@ use fgumi_bam_io::{
     create_bam_reader_for_pipeline_with_opts, create_bam_writer, create_optional_bam_writer,
     create_raw_bam_reader_with_opts,
 };
+use std::path::Path;
 
 use super::common::{
     AllowUnmappedOptions, BamIoOptions, CompressionOptions, ConsensusCallingOptions,
     QueueMemoryOptions, ReadGroupOptions, RejectsOptions, SchedulerOptions, StatsOptions,
-    ThreadingOptions, build_pipeline_config, reject_colliding_outputs, serialize_raw_bam_records,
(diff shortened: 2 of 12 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai pause

PR #846, opened 2026-08-20
perf(sort): overlap key extraction, reuse arena pages, and read with real queue depth

nh13:
Five optimisations to `fgumi sort`, measured against `1kg-wgs-HG00096` (43.13 GB, template-coordinate, 16 threads, 779,820,469 records, `read_ahead_kb` left at its 128 default). Together they take the sort from **~377s to ~268s, about -29%**, with peak RSS unchanged, thread count unchanged, and output byte-identical.

Each change is paired against its own control in a single boot; the combined figure spans two boots, so treat it as approximate (in-boot noise is ±0.7%, cross-boot ±2.3%).

Stacked on #838. Suggested reading order is commit order — each one only makes sense given the last, and the later numbers are only interpretable once the earlier costs are gone.

## Reading order

**1. `0152e316` — extract template sort keys on the pool, not the ingest thread.** 93.6s of `extract_template_key_inline` at ~122 ns/record moves off the serial ingest thread onto the worker pool as a new `SortStep::ExtractKeys`, scheduled last everywhere so it can never displace the step that feeds the ingest thread. `SegmentedBuf` splits into sealed `Arc` segments plus a live one, so a worker can read a finished segment while ingest appends to a later one — safe by construction, no `unsafe`. Ingest serial goes 137.3s to 57.2s with 96.9% of keys overlapped and a 0.2s barrier wait.

**2. `bfa4d1d5` — keep the arena's pages across chunks instead of re-faulting them.** `SegmentedBuf::reset_for_reuse` replaces `clear()` at the three spill sites, and `release_retained` gives the segments back once ingest is over. -13.9s, minor faults 23.6M to 12.7M, and 1.16 GB of the retention cost recovered.

**3. `63dd566e` — read input and spill files with concurrent positional reads.** The bulk of the change; see below.

**4. `6ffe651c` — keep a fill in flight while the framer frames.** The scattered reader was demand-driven, so the device idled for as long as framing took. `fill` splits into `issue` and `collect`, and the phase-1 reader issues the next fill as soon as one is collected. Read span 69.4s to 66.9s, wall 271.8s to 267.7s, paired in one boot against a ±1.9s noise floor. Depth is a knob and **one is the answer** — the sweep is recorded next to the constant.

**5. `89d1f66b` — choose the stream count from a one-stream probe.** `--read-streams` gains `auto`, now the default. Whether concurrency helps depends entirely on the storage: EBS gp3 wants four (28% faster), a direct-attached instance-store SSD wants one (forcing eight is 1.8% slower). Auto measures the device's single-stream throughput over the first eight fills — costing nothing, since that *is* one stream — and picks `ceil(1.2 GB/s ÷ measured)`, capped at eight. The measurement is taken once and shared through the pool's `FetchQueue`, so a merge's 44 spill readers don't each measure a device they're contending for and over-provision. **Validated paired on both devices**: gp3 auto 273.7s vs pinned-four 273.3s (a dead heat, probed 351 MB/s → 4); instance SSD probed 1920 MB/s → 1, its best. The `~377s → ~268s` headline is the gp3 case; a fast-NVMe host sees far less, because there the read was never the bottleneck.

## The third commit

Both phases read a file through one blocking sequential `read()` at a time — phase 1 behind the exclusive `ReadInputBlocks` step, phase 2 behind each spill file's reader mutex. That is queue depth 1, and this device serves one stream far worse than several: **358 MB/s against 1177 MB/s for four concurrent streams**, with buffer size making no difference at all (2, 8 and 32 MiB all landed within 0.1%). A bigger single request is not more concurrency.

`read_ahead_kb=4096` fixes both, but it is root-only and system-wide, so a tool cannot rely on it. `posix_fadvise` was measured as the per-file substitute and does not work — `SEQUENTIAL` 359 MB/s and `WILLNEED` 348-350 MB/s against 360 with no hint, with `strace` confirming the calls land and the kernel simply declines. So `--read-streams N` cuts a fill into slices read at once and serves them in file order behind a plain `Read`. BGZF blocks and zstd frames are variable-length, so block N's offset is unknowable without scanning from block 0 — but framing is ~4% of the reader's time, so parallelising the byte fetch and leaving the framer, the block serials and the reorder buffer untouched is enough.

**The slices run on the existing worker pool, as a new `SortStep::FetchBytes` — not on threads of the reader's own.** I built the dedicated-thread version first and measured it: at four streams a `--threads 16` sort ran **38 threads instead of 34** and doubled involuntary context switches (1.39M to 2.81M). It was faster only because the host had idle cores to lend, and it silently broke what `--threads` means.

The filling thread reads the first slice itself and then **reclaims** any slice no worker has started, rather than waiting on one. Submitting and blocking would deadlock the moment every worker is a filler waiting for slices nobody is left to run; reclaim makes the worst case "the single sequential read we do today" instead of a wedge. In practice reclaim almost never fires — **70,211 of 70,310 slices (99.8%) were run by workers**, because slice 0 is a real disk read and that is a wide window.

## Measurements

| arm | wall | phase 1 read | k-way merge | peak threads | involuntary ctx sw |
| `--read-streams 1` | 377.7s | 141.8s | 155.1s | 34 | 1,390,118 |
| `--read-streams 4` | **271.2s** | **69.9s** | **120.2s** | **34** | **1,345,637** |

And the lookahead commit, paired separately in a later boot:

| lookahead depth | wall | phase 1 read span | peak RSS |
| --- | --- | --- | --- |
| 0 | 271.8s | 69.4s | 11,253 MB |
| **1** | **267.7s** | **66.9s** | 11,281 MB |
| 2 | 269.4s | 66.7s | 11,559 MB |
| 4 | 268.5s | 66.4s | 11,290 MB |

Depths past one move the read span by under a second and the wall clock not at all, and depth 2 cost 277 MB of peak RSS for it. One is enough because **the span is not fetch-bound**: the fetch runs at 1327 MB/s — above the 1177 MB/s the four-stream ladder measured — and takes 32.5s of a 66.4s span against an ingest-serial floor of 53.0s. What remains is coordination between the reader, the decompress workers and the ingest thread, which read-ahead does not touch.

That settles the phase-1 read path. At device speed and half the span, the reader is no longer the constraint there; the remaining headroom is the ingest thread's own serial cost and the coordination gap, not I/O.

The merge's own reads fall from 169.3s of worker-busy time to 79.5s, consumer park from 42.3s to 1.1s, and the merge reaches its consumer-serial floor: 118.0s of serial consumer inside a 120.1s loop, so 2% is all that remains recoverable there without doing less work per record. Peak RSS is unchanged at 10.7 GB. Total CPU rises 1.4%, all of it system time from roughly double the read syscalls.

`tricorder --trace` confirms no arm ever saturates the box: cores in use peak at 13.9 of 16 and no tick reaches 15.2, so the 34 threads are mostly idle rayon and writer threads rather than contention. Worth knowing that a `--threads 16` sort has always run ~34 threads.

## Correctness

`compare bams --command sort` reports **IDENTICAL** between one and four streams, across both spill codecs, and against the pre-change binary's output on the full 779,820,469-record sort. Locally, a spilling sort (11 spill runs) compares identical across bgzf/zstd x 1/4 streams.

The lookahead commit is separately verified IDENTICAL against the pre-change binary on the same full sort.

The unit tests cover in-order delivery at 1/2/4/8 streams, a short final slice, an empty file, 13-byte reads across slice seams, buffer recycling, offset starts (the zstd `ZSPILL_MAGIC` case), a failed fill reporting rather than hanging, that exactly one caller ever runs a given slice under an 8-way race, that a fill completes when **no** worker ever takes a slice, that lookahead keeps the depth it was given in flight and never runs past EOF, that **a failed fill rewinds the fetch offset** so a retry re-reads the failed range instead of skipping it, and — Linux only — that scattered reading spawns no threads of its own. Each was checked by mutation; the thread-ceiling one fails if a single thread is spawned.

The four sort arms of the `--async-reader` integration test now vary `--read-streams` instead, with spilling forced, so the merge's spill reads are covered end to end and not only at unit level.

## Two things to weigh

**`--async-reader` is removed from sort.** It was a hidden prototype whose whole job was decoupling the read from the block reader, which `--read-streams` now does without spawning a thread outside `--threads`. `extract` and `group` keep their own, which is unrelated. The one capability genuinely lost is prefetch on **piped stdin** — scattered reads need offsets, so a pipe falls back to the plain buffered reader. I think that is right, since the pool's `ReadInputBlocks` step already decouples that read from the consumer, but it is a removal rather than a refactor.

**The default is `auto`** (commit 5). The `c7gd` local-NVMe sweep that motivated it is done and is why: 1/2/4/8 streams there gave 248.4/249.4/251.4/252.9s — monotonically *worse*, because one stream already sustains 2214 MB/s and the ingest thread can only absorb ~810. So a fixed 4 would have been a 1.8% regression on instance storage while winning 28% on gp3; auto gets both by measuring. The one axis not measured is a CPU-saturated host — but auto only spawns extra streams on measurably slow storage, so the fetch/compression contention is only ever paid where it earns 28%.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -187,6 +187,16 @@ of records) and a safe rewrite measurably regresses sort throughput.
   `RawQuerynameKey::new`).
 - **`crates/fgumi-sort/src/radix.rs`** — internal radix-sort helpers; see file
   comments for the `SAFETY:` invariants.
+- **`crates/fgumi-sort/src/phase1_keys.rs`** — one `#[allow(unsafe_code)]` site in
+  `prefetch_read_l1`, a software-prefetch hint issued while a deferred key batch
+  scans record bodies out of a sealed arena segment. Those bytes were written up
+  to a segment ago and are cold by the time a worker reads them, so the scan is
+  latency-bound on cache misses. SAFETY: both `prfm pldl1keep` (aarch64) and
+  `_mm_prefetch` (x86_64) are *non-faulting hints* — they never read or write
+  observable memory and never trap, even on an unmapped address — and the
+  argument is a live `&u8`, so the pointer is valid to name. A no-op on other
+  architectures. Approved for the same reason as the other sort hot paths: it
+  runs once per record over hundreds of millions of records.
 
 ### Approved natural-order comparator (fgumi-raw-bam)
 
--- a/crates/fgumi-sort/examples/read_ladder.rs
+++ b/crates/fgumi-sort/examples/read_ladder.rs
@@ -96,6 +96,57 @@ enum Pattern {
     Interleaved,
 }
 
+/// What the reader tells the kernel about the access pattern it is about to make.
+///
+/// The question these exist to answer: a blocking `read()` keeps roughly one
+/// read-ahead window outstanding at the device, and on this volume that is ~3x
+/// too little to saturate it (358 MB/s single-stream against a ~1050 MB/s cap).
+/// Four threads fix that by carrying four windows. These advise calls try to fix
+/// it *without* extra threads, by asking the kernel to keep the queue primed on
+/// a single ordered stream -- which, unlike parallel streams, cannot deliver
+/// blocks out of order.
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum Advice {
+    /// No hint: the default read-ahead window and nothing else.
+    None,
+    /// `POSIX_FADV_SEQUENTIAL` once per stream. Linux doubles the read-ahead
+    /// window for the file; a 2x lever at most.
+    Sequential,
+    /// `POSIX_FADV_WILLNEED` for a window ahead of the cursor, reissued as the
+    /// cursor advances. Not bounded by the read-ahead window, so this is the one
+    /// that can actually fill the device queue.
+    WillNeed,
+}
+
+/// Issue one `posix_fadvise` call, or do nothing where it does not exist.
+///
+/// Errors are deliberately ignored: `fadvise` is a *hint*, and a kernel that
+/// declines it (memory pressure, an unsupported filesystem) must not fail the
+/// read. A hint that was dropped shows up as a throughput number that did not
+/// move, which is exactly the result the experiment is testing for.
+#[cfg(target_os = "linux")]
+fn advise(file: &File, offset: u64, len: u64, advice: nix::fcntl::PosixFadviseAdvice) {
+    let offset = i64::try_from(offset).unwrap_or(i64::MAX);
+    let len = i64::try_from(len).unwrap_or(i64::MAX);
+    let _ = nix::fcntl::posix_fadvise(file, offset, len, advice);
+}
+
+#[cfg(target_os = "linux")]
+fn advise_sequential(file: &File, offset: u64, len: u64) {
+    advise(file, offset, len, nix::fcntl::PosixFadviseAdvice::POSIX_FADV_SEQUENTIAL);
+}
+
+#[cfg(target_os = "linux")]
+fn advise_willneed(file: &File, offset: u64, len: u64) {
+    advise(file, offset, len, nix::fcntl::PosixFadviseAdvice::POSIX_FADV_WILLNEED);
+}
+
+#[cfg(not(target_os = "linux"))]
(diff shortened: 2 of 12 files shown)


nh13:
@coderabbitai pause

PR #848, opened 2026-08-20
ci: run checks in the merge queue and cap job runtimes

nh13:
Groundwork for enabling a branch merge queue on `main` (mirroring the ferro-hgvs setup). This is step 1 of 3: it lands the workflow changes the queue depends on, ahead of any ruleset change. No queue is enabled by this PR.

## What changes

- Adds a `merge_group` trigger to the **Check and Test** workflow, so all 13 jobs run against each group the queue assembles.
- Adds `timeout-minutes: 20` to every job in the workflow.

## Why the trigger has to land first

GitHub evaluates `merge_group` workflows from the **target branch** (`main`), not from the PR sitting in the queue. So the trigger must already be on `main` before the queue can require these checks — otherwise every queued PR would wait forever on checks that never start. The queue's required-check list will be exactly these job names.

## Why the timeouts

In a merge queue a hung check blocks every entry behind it until the check-response timeout evicts the whole group. An infrastructure stall — e.g. the stalled apt mirror already seen on the `sort-correctness` job — must fail fast rather than idle at GitHub's 6-hour default. This makes the 20-minute bound uniform across the workflow; the healthy path completes in ~4 minutes, so it's ~5× headroom.

## Not in this PR

- `miri.yml` is intentionally left PR-only (it's `paths`-filtered to `fgumi-raw-bam`; a required check that doesn't run on every group would deadlock the queue, so it stays off the required list).
- The branch ruleset (required status checks + `merge_queue` rule) — step 2, in a follow-up once this is on `main`.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -5,6 +5,10 @@ on:
     branches:
       - main
   pull_request:
+  # Run in the merge queue too. GitHub evaluates `merge_group` workflows from the
+  # target branch (main), so these checks gate every group the queue assembles;
+  # the branch ruleset lists these job names as the queue's required checks.
+  merge_group:
 
 # Every job here only checks out the repo and runs read-only build/test/lint
 # steps (the coverage upload authenticates with CODECOV_TOKEN, not the workflow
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #854, opened 2026-08-22
refactor(sam): remove dead SamRecordClipper, port its tests to RawRecordClipper

### LINKED ISSUE #843: SamRecordClipper is never constructed in production (superseded by RawRecordClipper) — remove or document
## Summary

`SamRecordClipper` (the `RecordBuf`-based clipper in `crates/fgumi-sam/src/clipper.rs`) is **never constructed in production code** — every `SamRecordClipper::new` / `SamRecordClipper::with_auto_clip` call in the workspace is inside its own `#[cfg(test)] mod tests`. The struct is re-exported (`crates/fgumi-sam/src/lib.rs`: `pub use clipper::{ClippingMode, RawRecordClipper, SamRecordClipper}`) but has no non-test caller anywhere; the live `fgumi clip` command and every consensus caller use the raw-byte `RawRecordClipper`.

This was surfaced while working #760 (which removed `SamRecordClipper`'s now-redundant *past-mate* methods): the broader struct is fully superseded by `RawRecordClipper`, the codebase's raw-byte clipper.

## Evidence

- `rg 'SamRecordClipper::(new|with)' crates src` → all hits in `clipper.rs` at lines ≥ 2479; the `#[cfg(test)] mod tests` begins at `clipper.rs:2471`.
- `rg 'SamRecordClipper' crates src | grep -v clipper.rs` → only a doc-comment references and the `lib.rs` re-export; no production construction or method call.

## Proposal / decision needed

Decide the fate of `SamRecordClipper`:

- **Remove it** (struct + its ~100 `#[cfg(test)]` tests + `RecordBuf`-only helpers), dropping it from the `fgumi-sam` public API. This is the natural end-state of the RecordBuf → raw-byte migration; `RawRecordClipper` + the cross-check tests already cover the behavior. Largest cleanup; a public-API removal, so worth a dedicated PR.
- **Or** keep it deliberately as public API for external `fgumi-sam` consumers and document that it is not used internally.

Not urgent and out of scope for #760 — filing so the decision is tracked rather than lost.

nh13:
## Summary

`SamRecordClipper` (the `RecordBuf`-based clipper) had no production caller left — `fgumi clip` and every consensus caller use the raw-byte `RawRecordClipper`, and the only `SamRecordClipper::new`/`with_auto_clip` calls in the workspace were in its own tests. It was the tail of the `RecordBuf` → raw-byte migration. This removes it (struct, impl, the `array_len`/`slice_array` macros, and the `reorient_strand_tags` helper it alone used) and drops it from the `fgumi-sam` public API.

## Preserving coverage

Its ~110 behavioral tests were the bulk of the clipping coverage; `RawRecordClipper` was largely verified only by cross-checking against it. To avoid losing that coverage, every test is ported to `RawRecordClipper` via a test-only `RawClipperOnBuf` adapter that runs the raw clipper through the `RecordBuf` API (`RecordBuf` → raw → clip → `RecordBuf`), so each test's setup and assertions run verbatim against the real raw clipper. The two non-UTF-8 tag tests — which a noodles `RecordBuf` cannot represent — are rewritten to build the raw record and tag directly. The former cross-check module now verifies the buf↔raw round-trip preserves CIGAR/pos/seq/qual (both sides run the same `RawRecordClipper`, so it no longer cross-checks an independent implementation; the concrete-expectation tests in `mod tests` are the correctness oracle). The `create_test_record` helper now normalizes SEQ to the CIGAR's query length so the records are valid BAM for the raw encoder without changing any assertion.

## Reading order

The change is a large deletion (−1066 net); start with the `RawClipperOnBuf` adapter (new `clip_test_adapter` module) to see the porting mechanism, then the `lib.rs` export removal, then the doc-reference fixes in `crates/fgumi-raw-bam/src/cigar.rs`.

Full suite (154 clipper tests, all present), clippy, fmt, and rustdoc pass.

Closes #843.

Risk: command output none; `unsafe` changes none, and CLAUDE.md allowlist changes none; memory bounds, queue capacity, and thread/backpressure policy changes none.

Removes the unused `SamRecordClipper` public re-export and updates documentation to use raw BAM clipping APIs.

### DIFF
--- a/crates/fgumi-raw-bam/src/cigar.rs
+++ b/crates/fgumi-raw-bam/src/cigar.rs
@@ -387,7 +387,7 @@ pub fn alignment_start_from_raw(bam: &[u8]) -> Option<usize> {
 ///
 /// Computes clipped CIGAR ops and the number of reference bases consumed by clipping,
 /// without modifying any record. This is the raw-byte equivalent of
-/// `SamRecordClipper::clip_start_of_read` / `clip_end_of_read` with Hard clipping mode.
+/// `RawRecordClipper::clip_start_of_read_raw` / `clip_end_of_read_raw` with Hard clipping mode.
 ///
 /// Like the Clipper, this first accounts for existing H+S clips at the relevant end.
 /// If `clip_amount` <= existing clips, only soft clips are upgraded to hard clips
@@ -664,7 +664,7 @@ fn parse_ref_len_and_trailing_clips(cigar: &[u8]) -> (i32, i32) {
 
 /// Upgrade existing soft clips to hard clips without changing alignment.
 ///
-/// Matches `SamRecordClipper::upgrade_clipping` in Hard mode.
+/// Matches `RawRecordClipper::upgrade_clipping_raw` in Hard mode.
 /// Used when `clip_amount` <= existing H+S clips.
 fn upgrade_clipping_raw(
     cigar_ops: &[u32],
--- a/crates/fgumi-sam/src/clipper.rs
+++ b/crates/fgumi-sam/src/clipper.rs
@@ -3,47 +3,7 @@
 //! This module provides functionality for clipping reads in various ways (soft, hard, etc.)
 //! and is used by tools like `clip` and consensus calling tools.
 
-use anyhow::Result;
-use noodles::core::Position;
-use noodles::sam::alignment::RecordBuf;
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #855, opened 2026-08-22
refactor(sort): make --sort-stats run-scoped instead of a process-global static

### LINKED ISSUE #849: refactor(sort): make --sort-stats run-scoped instead of a process-global static
`SORT_STATS` is currently a process-global mutable static gating the `stat!` diagnostics macro (added in #826). Because `fgumi_lib` is a library, two sorts run concurrently in one process with different `--sort-stats` settings clobber each other's flag — the same global-mutable-state hazard the `umask` probe had to wrap in a `Mutex`. Test isolation also relies on nextest's process-per-test model; plain `cargo test` (thread-per-test) races the flag (worked around in #826 by serializing the two flag tests with a mutex).

Long-term fix: make the flag run-scoped by hanging it off state that already flows through the sort run. The sort threads a config/context (`Sort` → external-sort/pipeline → `worker_pool`), and the `stat!` logging sites all live in `external.rs`, which already holds `self` with access to that pipeline state. Add a `sort_stats: bool` (or a small `Copy`/`Arc` `Diagnostics` handle) to the run context / pipeline struct — one field, not a bool threaded through ~20 signatures, because it rides a struct those sites already have — rework `stat!` to consult that scoped handle (e.g. `self.diag.enabled()`) instead of the static, and delete the global. This removes both the library-concurrency correctness bug and the plain-`cargo test` race, and lets the test mutex be removed.

nh13:
## Summary

`SORT_STATS` was a process-global `AtomicBool` gating the `stat!` diagnostics macro. Because fgumi-sort is a library, two sorts in one process with different `--sort-stats` settings clobbered each other's flag, and the two flag tests had to be serialized with a mutex to survive plain `cargo test`'s thread-per-test model.

## What changed

The flag now hangs off the sort's own state instead of a global. `sort_stats` is a field on `RawExternalSorter` (the sort-options struct, set via a `.sort_stats(bool)` builder from the CLI) and is copied into the `SortPhaseTimer` and `SortWorkerPool` that sorter builds — the two objects the `stat!` emitters already hold as `self`/`pool`. The macro takes the flag as its first argument (`stat!(sort_stats, ...)`); each emitter reads it from `self.sort_stats` / `pool.sort_stats()`, or takes a `sort_stats: bool` parameter where it is a free function. The global static, its `set_sort_stats`/`sort_stats_enabled` functions, and the `SORT_STATS_LOCK` test mutex are removed; the flag test is rewritten to check the run-scoped default and builder.

This removes both the library-concurrency hazard (concurrent sorts now keep independent settings) and the plain-`cargo test` race. Verified end-to-end: `fgumi sort --sort-stats` still emits the phase-timing diagnostics and a plain sort emits none. Full suite, clippy, and fmt pass.

Closes #849.

Risk: Sort diagnostics change when `--sort-stats` is enabled; grouping, consensus, sort order, corrected UMIs, and metrics output are none; `unsafe` changes and CLAUDE.md allowlist updates are none; memory bounds, queue capacity, and thread/backpressure policy changes are none.

Fix: Scope `sort_stats` to each `RawExternalSorter` run instead of process-global state. Propagate the setting through `SortPhaseTimer` and `SortWorkerPool`, remove the global state and test mutex, and add builder and end-to-end tests for enabled and default-disabled diagnostics.

### DIFF
--- a/crates/fgumi-sort/src/bgzf_io.rs
+++ b/crates/fgumi-sort/src/bgzf_io.rs
@@ -387,7 +387,7 @@ mod tests {
     /// without the `ZSPILL_MAGIC` prefix (which a real chunk writer would write
     /// before invoking the loop).
     fn roundtrip_data(data: &[u8], codec: SpillCodec) -> Vec<u8> {
-        let pool = Arc::new(SortWorkerPool::new(2, 1, 6, codec));
+        let pool = Arc::new(SortWorkerPool::new(2, 1, 6, codec, false));
         let (result_tx, result_rx) = pool.compress_result_channel();
         let buffer_pool = pool.buffer_pool.clone();
         let permit_pool = make_permit_pool(&pool);
@@ -425,7 +425,7 @@ mod tests {
     #[case(SpillCodec::Bgzf)]
     #[case(SpillCodec::Zstd)]
     fn test_staging_buffer_flush_empty_is_noop(#[case] codec: SpillCodec) {
-        let pool = Arc::new(SortWorkerPool::new(1, 1, 6, codec));
+        let pool = Arc::new(SortWorkerPool::new(1, 1, 6, codec, false));
         let (result_tx, _result_rx) = pool.compress_result_channel();
         let permit_pool = make_permit_pool(&pool);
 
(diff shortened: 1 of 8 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #856, opened 2026-08-22
docs(group): clarify the paired strategy never actually uses the index

nh13:
## Summary

`PairedUmiAssigner::uses_index` is documented as "the paired strategy indexes at every `--edits` value — N-gram for k=1, BK-tree for k>1", and the size gate it reads does return `true` above the threshold. But a paired canonical form always carries the `-` separator between its halves, and `-` is a non-`ACGT` byte, so `NgramIndex::new` and `BkTree::from_umis` — which encode only `ACGT` — invariably decline it. The paired strategy therefore always falls back to its reverse-aware linear scan, at every pool size: the index is never actually built for paired UMIs. The docs implied the opposite, which is misleading about both performance (large paired pools get no index acceleration) and correctness.

## What changed

Corrects the `uses_index` doc and the sibling test doc to say the paired strategy is index-*eligible* by size but never realizes it, and adds `test_paired_forms_are_never_indexed_because_of_the_dash` pinning that `NgramIndex::new` / `BkTree::from_umis` decline dash-delimited forms.

That test is a deliberate guard, not just a characterization: the indexed candidate search (`find_within`) is forward-Hamming only, so a future change that taught the index to accept dash-delimited forms (e.g. stripping the dash for a perf win on large paired pools) would silently drop the reverse-orientation (`A-B` == `B-A`) edges the paired matcher draws. Such a change must make the indexed path reverse-aware first, and this test forces that reckoning.

## Context

Surfaced while investigating (and dismissing) a CodeRabbit finding on #852 that claimed the parallel paired assigner diverges from the sequential one at ≥100 unique forms — it doesn't, precisely because the sequential paired path never indexes. This documents that invariant so the next person doesn't have to rediscover it. Docs + test only; no behavior change.

Risk: command output changes—none; `unsafe` changes—none, and the `CLAUDE.md` allowlist is unchanged; memory bounds, queue capacity, and thread/backpressure policy changes—none.

Fix: Clarify paired UMI index eligibility and document fallback to reverse-aware linear matching.

- Add regression coverage showing that dash-delimited paired UMIs are rejected by `NgramIndex` and `BkTree`.
- Document paired index eligibility for all edit distances.
- No runtime behavior changes.

### DIFF
--- a/crates/fgumi-umi/src/assigner.rs
+++ b/crates/fgumi-umi/src/assigner.rs
@@ -2165,14 +2165,22 @@ impl PairedUmiAssigner {
     ///
     /// As for [`AdjacencyUmiAssigner::uses_index`], an [`IndexThreshold::MinUmis`] is a
     /// MINIMUM: `0` indexes every group, and [`IndexThreshold::Never`] is what turns the
-    /// index off. Unlike the adjacency and edit strategies, the paired strategy indexes at
-    /// every `--edits` value — N-gram for k=1, BK-tree for k>1 — so only the threshold
-    /// applies. This reads the very gate `build_adjacency_graph` runs, not a copy of it.
+    /// index off. Unlike the adjacency and edit strategies, the paired strategy is index-
+    /// *eligible* at every `--edits` value — N-gram for k=1, BK-tree for k>1 — so only the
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

Issue #839, opened 2026-08-20
duplex/vanilla consensus under-trims read-through on dovetail FR pairs (per-record is_fr_pair_raw)


nh13:
`num_bases_extending_past_mate_raw` (fgumi-raw-bam `overlap.rs`), used by the simplex/duplex consensus callers to trim read-through past the mate before consensus, guards on the **per-record** `is_fr_pair_raw`. That predicate's forward-strand branch derives the mate's 5′ position from TLEN and misclassifies dovetail FR pairs — the same htsjdk asymmetry fixed for CODEC in #505 (via the symmetric `is_primary_fr_pair_raw`) and upstream by samtools/htsjdk#1771 (htsjdk 5.0.0). CODEC and `fgumi clip` (#760) both use the symmetric gate; the general simplex/duplex consensus path was never migrated.

## Effect

On a dovetail FR pair, the forward read is misclassified NOT-FR, so `num_bases_extending_past_mate_raw` returns 0 and its read-through / adapter bases past the mate are **not trimmed** — they leak into the simplex/duplex consensus. The bug is asymmetric: the reverse read is unaffected because its branch of `is_fr_pair_raw` is CIGAR-derived.

## Reproduction

A dovetail pair (fgumi-raw-bam `overlap.rs`): forward `50M50S` @ 1-based 101 with `TLEN = -90` and `MC = 100M`; reverse `100M` @ 1-based 61. Calling the two functions on the forward read:

```
is_fr_pair_raw(fwd)                                = false   # per-record check misclassifies the forward read
is_fr_pair_raw(rev)                                = true
num_bases_extending_past_mate_raw(fwd)             = 0       # MC-tag / per-record path (simplex/duplex) — BUG: no trim
num_bases_extending_past_mate_vs_mate_raw(fwd,rev) = 40      # symmetric path (CODEC / clip) — correct
is_primary_fr_pair_raw(fwd,rev)                    = true    # correct classification
```

40 query bases of read-through go untrimmed on the forward read.

## Fix sketch

[...]

PR #857, opened 2026-08-22
fix(clip): clip dovetail FR overlaps and make the TLEN FR check crate-internal

nh13:
## Summary

`RawRecordClipper::clip_overlapping_reads` (the `fgumi clip` overlap path) gated on the per-record `is_fr_pair_raw` for both reads. That check's forward-strand arm derives the mate 5' from `TLEN`, so it misclassifies dovetail FR pairs whose aligned ends coincide (htsjdk/samtools#1771): the forward read reads as non-FR, the guard returns `(0, 0)`, and read-through/adapter is left unclipped on an otherwise valid FR pair. This is the same class of bug #839 fixes for the MC-tag consensus path; this consumer was not covered.

Both records are in hand at this call site, so it now gates on the symmetric, order-independent `is_primary_fr_pair_raw(r1, r2)` instead — the same gate the sibling `clip_extending_past_mate_ends` already uses.

## Consolidation

To stop the per-record TLEN check from being grabbed as a pair-classification gate again:

- `is_fr_pair_raw` is demoted to `pub(crate)` and dropped from the crate's public re-export. It remains the reverse-arm building block for `is_primary_fr_pair_raw` and `is_fr_pair_with_mate_cigar_raw`.
- The test-only `CodecConsensusCaller::is_fr_pair` wrapper (and its test) are removed; the per-record FR/RF/FF coverage already lives in `overlap.rs`.

## Testing

- New `clip_overlapping_reads` dovetail-forward regression test (RED without the gate change, GREEN with it).
- Full `cargo ci-test` (7930 tests), `ci-fmt`, `ci-lint`, and `cargo doc -D warnings` pass.

## Stacking

Stacked on #839 (`839/nhomer/consensus-dovetail-fr-trim`); please merge after that PR — the FR-classification primitives this builds on land there.

Risk: command output changes for overlap clipping, pinned by symmetric FR-pair classification and regression tests; `unsafe`: none; memory bounds, queue capacity, and thread/backpressure policy: none.

- Fix dovetail FR-pair clipping in `RawRecordClipper::clip_overlapping_reads`.
- Use `is_primary_fr_pair_raw(r1, r2)` for symmetric pair classification.
- Add regression coverage for dovetail-forward pairs and inconsistent unmapped pairs.
- Make `is_fr_pair_raw` crate-private.
- Remove its public re-export and test-only wrapper.
- Update integration-test documentation for TLEN-independent FR detection.

### DIFF
--- a/crates/fgumi-consensus/src/codec_caller.rs
+++ b/crates/fgumi-consensus/src/codec_caller.rs
@@ -1933,11 +1933,6 @@ impl CodecConsensusCaller {
     reason = "test-only methods use owned values and numeric casts"
 )]
 impl CodecConsensusCaller {
-    /// Test-only wrapper: check if a `RawRecord` is part of an FR pair.
-    pub fn is_fr_pair(&self, rec: &RawRecord) -> bool {
-        bam_fields::is_fr_pair_raw(rec)
-    }
-
     /// Test-only wrapper: filter `RawRecord`s to most common alignment,
     /// returning the surviving records (index-stable relative to the input).
     pub fn filter_to_most_common_alignment(&mut self, recs: Vec<RawRecord>) -> Vec<RawRecord> {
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #861, opened 2026-08-24
feat(codec): add --legacy-overlap-window for fgbio-parity output

### LINKED ISSUE #1173: CallCodecConsensusReads drops read pairs whose alignments dovetail, including pairs with no indels
## Symptom

`CallCodecConsensusReads` rejects a read pair with `IndelErrorBetweenStrands` whenever the two alignments of the pair dovetail — i.e. whenever each alignment ends past the far end of the other in reference space, rather than one being contained within the span of the other. The pair is dropped and counted under `raw_reads_rejected_for_indel_error_between_strands`.

This fires even when neither cigar contains an indel at all, so no consensus is emitted for a pair that in fact has complete duplex overlap and nothing out of phase about it.

It also fires for pairs in which one strand's alignment runs through a short deletion that lies *beyond* the aligned end of the other strand's alignment — a geometry where the two alignments agree exactly over every reference position that both of them are aligned over, and the deletion is supported by only one strand because the other soft-clipped there.

## Reproduction

Both cases are read pairs added to `CodecConsensusCallerTest`'s `RefBases`; `RefBases.substring(a, b)` is used below to name reference bases (0-based, half-open) so the two reads of each pair carry the same insert.

### 1. A dovetailed pair with no indels

Two reads that sequenced the same 40 base insert, where the aligner soft-clipped a different end of each. R2 starts before R1 starts *and* R1 ends after R2 ends:

```scala
val insert  = RefBases.substring(100, 140)
val builder = new SamBuilder(readLength=insert.length, baseQuality=35)
val caller  = new CodecConsensusCaller(readNamePrefix="codec", minReadsPerStrand=1, minDuplexLength=1)
val raw     = builder.addPair(
  contig=0, start1=111, start2=101, cigar1="10S30M", cigar2="30M10S",
  bases1=insert, bases2=insert, attrs=Map(("RX", "ACC-TGA"), ("MI", "hi"))
)
caller.consensusReadsFromSamRecords(raw)
```

Expected: one 40 base consensus read whose bases are `insert` — the two reads carry identical bases and are in register at every position.

Actual: no consensus reads; both records are counted under `raw_reads_rejected_for_indel_error_between_strands`, despite neither cigar containing an indel.

### 2. A terminal deletion beyond the other strand's aligned end

Two reads that sequenced the same 127 base insert plus two bases of read-through that `clipExtendingPastMateEnds` removes. The positive strand alignment runs through a single base deletion that sits past the aligned end of the negative strand alignment, which soft-clipped its last two bases instead:

```scala
val posBases = RefBases.substring(98, 224) + RefBases.substring(225, 228)
val negBases = RefBases.substring(96, 224) + RefBases.substring(225, 226)
val builder  = new SamBuilder(readLength=posBases.length, baseQuality=35)
val caller   = new CodecConsensusCaller(readNamePrefix="codec", minReadsPerStrand=1, minDuplexLength=1)
val raw      = builder.addPair(
  contig=0, start1=101, start2=100, cigar1="2S124M1D3M", cigar2="3S124M2S",
  bases1=posBases, bases2=negBases, attrs=Map(("RX", "ACC-TGA"), ("MI", "hi"))
)
caller.consensusReadsFromSamRecords(raw)
```

After clipping, both reads are 127 bases, carry identical sequence, and are in register; the deletion is at a reference position that only the positive strand read is aligned over.

Expected: one 127 base consensus read whose bases are `RefBases.substring(98, 224) + RefBases.substring(225, 226)`.

Actual: no consensus reads. On `main` this pair is rejected earlier still, as `R1R2OverlapTooShort`, because `SamRecordClipper` un-maps the positive strand read; with #1172 applied the clipping is correct and the pair reaches — and fails — the phase check.

## Notes

Non-dovetailed geometries are unaffected: for every pair whose positive strand alignment starts at or before the negative strand alignment's start and ends at or before its end, the region the caller evaluates is already the intersection of the two aligned spans.

nh13:
## Summary

The corrected duplex overlap window (#761) admits dovetailed FR pairs that fgbio's `[negative.start, positive.end]` window rejects. On short-insert CODEC data — where the two strands routinely dovetail (each aligns a base to tens of bp past the far end of the other, from asymmetric terminal soft-clipping) — this recovers ~2.4× more consensus reads (108K → 263K on a real CODEC library). Those recovered pairs are legitimate short-insert molecules, not artifacts: their overhangs are all bounded by read length (median 26 bp, none > read length, so no mismapped/chimeric admission), they still pass the indel/phase check over the corrected window, and their realigned substitution error is ~1.9% vs reference — *below* the baseline consensus reads' ~3.3% (three methods agree, incl. fgbio's own `ErrorRateByReadPosition`).

This adds `--legacy-overlap-window` (off by default) so a run can reproduce fgbio's current-release CODEC consensus **output**, where those dovetailed pairs are rejected. fgbio's own fix for the same defect is in flight as fulcrumgenomics/fgbio#1174 (fixes fulcrumgenomics/fgbio#1173); until it lands and a new fgbio release ships, this flag is the escape hatch for anyone needing byte-for-byte fgbio parity.

## Honest stats labeling

Under `--legacy-overlap-window`, the rejected dovetails are reported with a new `Dovetail` rejection reason (`raw_reads_rejected_for_dovetail`) instead of the misleading `indel_error_between_strands` bucket fgbio mislabels them under (there is no indel in the CIGARs). This makes visible exactly how many pairs fgbio-parity is costing — flip the flag off to recover them.

Consequence, called out deliberately: the `--stats` breakdown under the flag is **not** byte-identical to fgbio's (fgbio buckets these under `indel_error_between_strands`), though the consensus BAM is. A pair that both dovetails **and** carries a real indel in the region both strands cover still fails the intersection check and is counted as `IndelErrorBetweenStrands`.

## Scope

- The flag gates only the reference window fed to `check_overlap_phase_raw` / `compute_consensus_length_raw`. The default path is unchanged: the `legacy_overlap_window &&` guard short-circuits, so record output and rejection reasons are byte-identical to `main` when the flag is off.
- New `Dovetail` reason threaded through the caller (`RejectionReason`) and metrics (`fgumi_metrics::rejection::RejectionReason`) enums — char code, `tsv_key` `raw_reads_rejected_for_dovetail`, emitted only when non-zero (mirroring `Downsampled`).
- Tests pin all three arms of the classifier: the default window still consenses the dovetail; the legacy window rejects the indel-free dovetail as `Dovetail` (not `IndelErrorBetweenStrands`); and a dovetail that also carries a real shared-region indel stays `IndelErrorBetweenStrands`.

Risk: consensus BAM output changes only when `--legacy-overlap-window` is enabled; corrected alignment intersection remains the default. Unsafe changes: none. Memory bounds, queue capacity, and thread/backpressure policy: none.

Add the `--legacy-overlap-window` option to reproduce legacy fgbio consensus output by rejecting dovetailed FR pairs.

- Keep the option disabled by default.
- Add `Dovetail` rejection classification with code `w`.
- Add `raw_reads_rejected_for_dovetail` metrics.
- Preserve `IndelErrorBetweenStrands` for pairs with a shared-region indel.
- Propagate the option through threaded and single-threaded CODEC paths.
- Add regression coverage for default acceptance, legacy rejection, and indel classification.

### DIFF
--- a/crates/fgumi-consensus/src/caller.rs
+++ b/crates/fgumi-consensus/src/caller.rs
@@ -441,6 +441,12 @@ pub enum RejectionReason {
     /// quality or alignment failure). Counted so `raw_reads_used` excludes it and routed to
     /// the `--rejects` output (fgumi#724).
     Downsampled,
+    /// The two strands' alignments dovetail (each runs past the far end of the other), so the
+    /// legacy `[neg.start, pos.end]` overlap window rejects them with no indel present. Only
+    /// produced by the codec caller under `--legacy-overlap-window`; the corrected default
+    /// window admits these pairs. Replaces the misleading `IndelErrorBetweenStrands` label
+    /// fgbio uses for the same pairs (fgumi#761, fulcrumgenomics/fgbio#1173).
+    Dovetail,
     /// Other unspecified reason
     Other,
 }
@@ -470,6 +476,7 @@ impl RejectionReason {
             Self::PotentialCollision => 'C',
             Self::NotPrimaryFrPair => 'R',
             Self::Downsampled => 'd',
+            Self::Dovetail => 'w',
             Self::Other => 'O',
         }
     }
@@ -500,6 +507,7 @@ impl RejectionReason {
             }
             Self::NotPrimaryFrPair => "Template did not have a single primary FR pair of reads",
             Self::Downsampled => "Downsampled to the maximum reads per consensus",
+            Self::Dovetail => "Alignments dovetail (rejected only under the legacy overlap window)",
             Self::Other => "Other reason",
         }
     }
@@ -540,6 +548,10 @@ impl RejectionReason {
             // fgumi-specific: fgbio has no downsampled rejection metric yet
             // (fulcrumgenomics/fgbio#1166, #1167), so this diverges from fgbio under a cap.
             Self::Downsampled => CentralReason::Downsampled,
+            // fgumi-specific: only produced under `--legacy-overlap-window`. fgbio buckets
+            // these dovetail rejections under `indel_error_between_strands`; fgumi labels them
+            // honestly instead, so the legacy-window stats diverge from fgbio here (fgumi#761).
+            Self::Dovetail => CentralReason::Dovetail,
         }
     }
 }
--- a/crates/fgumi-consensus/src/codec_caller.rs
+++ b/crates/fgumi-consensus/src/codec_caller.rs
@@ -166,6 +166,17 @@ pub struct CodecConsensusOptions {
     /// Minimum duplex overlap length (in bases)
     pub min_duplex_length: usize,
 
+    /// Use fgbio's legacy `[negative.start, positive.end]` duplex overlap window instead of the
+    /// corrected intersection `[max(starts), min(ends)]` (fgumi#761).
+    ///
+    /// The legacy window equals the intersection only when one alignment nests within the span
+    /// of the other. On CODEC's short inserts the two strands routinely *dovetail* (each aligns
+    /// past the far end of the other), and the legacy window then overshoots past one read and
+    /// rejects the pair as [`CallerRejectionReason::Dovetail`] even with no indel. `false` (the
+    /// default) admits those pairs; `true` reproduces fgbio's consensus output for the current
+    /// fgbio release (fulcrumgenomics/fgbio#1173).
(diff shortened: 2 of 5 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #865, opened 2026-08-26
perf(umi): use fixed-seed ahash on hot per-item maps to remove RandomState global-counter contention

nh13:
UMI grouping creates many small `AHashMap`s per thread (per-position-group, per-assign call). `AHashMap::new()`/`::default()`/`::with_capacity()`/`.collect()` all derive their state from `ahash::RandomState::new()`, which bumps a **process-global atomic counter** on every construction. At 16 threads that counter's cache line ping-pongs across cores and dominates CPU — a profile put `gen_hasher_seed` at ~28% of `group` CPU.

This gives each hot per-item map a **fixed seed** via a shared `deterministic_state()` helper (`ahash::RandomState::with_seeds(<4 consts>)`, `const fn`, no atomic), applied only at genuinely-hot per-item sites. The seed table lives in exactly one place (`fgumi-dna`); `fgumi_sort::cb_hasher` and the per-crate re-exports all delegate to it.

**Measured (WGS, 53,139,936 records, 16 threads, Graviton4, CPU-seconds):**

| `group` strategy | baseline | fixed-seed | throughput |
|---|---|---|---|
| adjacency | 117.7 | 86.1 | **+37%** |
| edit | 109.3 | 86.7 | **+26%** |
| paired | 99.6 | 88.6 | **+12%** |

`filter` +5.5%. `dedup`/`simplex` flat — expected: dedup creates maps less densely and spends ~21% between `assign()` calls in BGZF compression, so the counter never becomes the bottleneck (confirmed via a 2×2 build×workload matrix + sampling).

**Output is byte-identical.** This is a throughput fix, not a determinism fix: every seeded map is either sorted before molecule-ID assignment, or mints IDs in raw-input order, or is a lookup/accumulator whose iteration order never reaches output. Verified three ways — output histograms are MD5-identical before/after for all three `group` strategies (`db4dd08d` / `4c49dc67` / `a1fc4c9a`), the full suite (9467 tests) stays green and non-flaky (it would flake if any of these were order-dependent), and `with_seeds` (not `with_seed(0)`) makes the seed fixed across runs, not merely per-process.

Internal UMI/molecule maps over bounded, trusted per-template data — not attacker-controlled input — so fixing the seed carries no HashDoS exposure. Once-per-run counters, setup, and test maps are deliberately left unseeded.

Risk: output remains pinned by fixed hash seeds; no `unsafe` changes and no CLAUDE.md allowlist update; no memory-bound, queue-capacity, or thread/backpressure changes.

Fix: use a shared deterministic `ahash::RandomState` for hot per-item maps. This removes global seed-counter contention while preserving byte-identical output. Validation includes fixed-seed runs, MD5 comparisons, and 9,467 tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -820,6 +820,7 @@ dependencies = [
 name = "fgumi-dna"
 version = "0.7.0"
 dependencies = [
+ "ahash",
  "rstest",
 ]
 
@@ -901,6 +902,7 @@ dependencies = [
  "crossbeam-utils",
  "fgumi-bam-io",
  "fgumi-bgzf",
+ "fgumi-dna",
  "fgumi-raw-bam",
  "fgumi-sam",
  "fs4",
--- a/crates/fgumi-consensus/src/hashing.rs
+++ b/crates/fgumi-consensus/src/hashing.rs
@@ -1,0 +1,6 @@
+//! Deterministic, contention-free hashing for hot per-item maps.
+//!
+//! Re-exports the shared fixed-seed hasher so the seed table lives in exactly one place; see
+//! [`fgumi_dna::hashing::deterministic_state`] for the full rationale.
+
+pub(crate) use fgumi_dna::deterministic_state;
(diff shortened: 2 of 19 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #871, opened 2026-08-27
feat(retag): add --threads and route through the unified pipeline

nh13:
## What

Adds `--threads N` to `fgumi retag` and routes it through the existing unified 7-step BAM pipeline, matching how `filter` and `clip` already work. Without `--threads`, the original single-threaded serial loop runs unchanged.

`retag` is a pure per-record SAM aux-tag rewriter (copy/move/delete ops applied left-to-right). It previously hardcoded a 1-reader / 1-writer serial loop with no way to opt into parallelism, while every other BAM-processing command exposed `--threads`.

## How

- `execute()` keeps all up-front guards (`io.validate`, output-collision, input-aliasing via canonicalize, and now `log_effective_check_crc`) *before* the mode branch, so both paths enforce them identically.
- No `--threads` → the serial loop, extracted verbatim into `run_single_threaded`.
- `--threads N` → `run_threaded`, using `SingleRawRecordGrouper` (one record per group — retag needs no template batching) and applying the ops in the pipeline's `process_fn`.
- Per-operation `OpCounts` are `u64` counters accumulated into a `PerThreadAccumulator<Vec<OpCounts>>` and summed after the pipeline drains. Because addition is commutative and the counters are `u64`, the aggregated counts — and the `--metrics` TSV — are identical between serial and threaded modes at any thread count.
- `--metrics` deliberately keeps working under `--threads` (unlike `clip`, whose per-read metrics are not summable and are rejected under `--threads`).
- Sets `group_key_config` to the raw-no-cell config to skip a per-record cell-barcode scan.

The flattened `ThreadingOptions` / `SchedulerOptions` / `QueueMemoryOptions` (i.e. `--threads`, `--scheduler`, `--max-memory`/`--memory-reserve`/`--memory-per-thread`) mirror `filter`/`clip`. `--compression-level` and the CRC/async flags already existed.

Also a small drive-by fix: `retag` now calls `log_effective_check_crc()`, finally emitting the per-run `CRC verify:` line its `--check-crc` doc already promised.

## Correctness

Output order is preserved by the pipeline, so threaded output matches serial record-for-record. Verified via the real binary (serial vs `--threads 4`): decoded records + `--metrics` TSV byte-identical; output headers differ only in the `@PG CL:` field (which correctly records the different argv); `--compression-level` honored in threaded mode; the zero-match warning fires in threaded mode; `--threads 0` is rejected cleanly.

## Tests

All in `retag.rs`:

- `threaded_output_matches_single_threaded` (`--threads 1` and `--threads 4`): asserts decoded-record equality, header equality, and byte-identical `--metrics` TSV against the serial path, over a 50-record corpus spanning the classes where a decode divergence could hide — mapped/paired reads with real CIGAR + `MC`/`RG` tags (some secondary), records that pre-carry `BX` so `dst_overwritten` is exercised, and a zero-match `delete`.
- `threaded_preserves_input_record_order` (`--threads 4`).
- `threads_flag_coexists_with_positional_operations` (clap parse: `--threads` alongside the positional `SRC::op::DST` args).

`cargo ci-test`, `cargo ci-fmt`, `cargo ci-lint` all pass.

Risk: output changes for threaded `retag`, pinned by serial-equivalent metrics, ordering, validation, and compression tests; `unsafe` changes: none; memory and thread policy changes: `--threads`, scheduler, and queue-memory options add bounded threaded execution.

Fix: use the threaded path only when `--threads` is set. Keep the existing serial path unchanged.

- Add `--threads N` to `fgumi retag`.
- Route threaded execution through the unified seven-step BAM pipeline.
- Preserve output order and aggregate per-operation `u64` counts.
- Apply validation and collision checks in both execution modes.
- Skip unnecessary cell-barcode scanning.
- Log effective CRC verification settings at execution start.
- Add tests for output and metrics equivalence, ordering, positional operation parsing, compression, warnings, count aggregation, and invalid thread counts.
- CI tests, formatting, and linting pass.

### DIFF
--- a/src/lib/commands/retag.rs
+++ b/src/lib/commands/retag.rs
@@ -25,22 +25,34 @@
 //! in `fgumi extract`.
 
 use std::fmt;
+use std::io;
 use std::path::PathBuf;
 use std::str::FromStr;
+use std::sync::Arc;
 
 use anyhow::{Result, bail, ensure};
 use clap::Parser;
-use fgumi_bam_io::{ProgressTracker, create_raw_bam_reader_with_opts, create_raw_bam_writer};
+use fgumi_bam_io::{
+    ProgressTracker, create_bam_reader_for_pipeline_with_opts, create_raw_bam_reader_with_opts,
+    create_raw_bam_writer,
+};
 use fgumi_raw_bam::{RawRecord, append_raw_tag, remove_tag};
 use log::{info, warn};
+use noodles::sam::Header;
 use serde::{Deserialize, Serialize};
 
 use crate::commands::command::Command;
 use crate::commands::common::{
-    BamIoOptions, CompressionOptions, add_pg_record, ensure_hd_record, reject_output_collisions,
+    BamIoOptions, CompressionOptions, QueueMemoryOptions, SchedulerOptions, ThreadingOptions,
+    add_pg_record, build_pipeline_config, ensure_hd_record, reject_output_collisions,
+    serialize_raw_bam_records,
 };
+use crate::grouper::SingleRawRecordGrouper;
 use crate::logging::OperationTimer;
+use crate::per_thread_accumulator::PerThreadAccumulator;
+use crate::read_info::LibraryIndex;
 use crate::sam::SamTag;
+use crate::unified_pipeline::{GroupKeyConfig, Grouper, run_bam_pipeline_from_reader};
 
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #873, opened 2026-08-27
feat(copy-umi): add command to copy the read-name UMI into the RX tag

### LINKED ISSUE #869: Support reading UMIs from BAM read names
Hey @nh13, ran into this while migrating a workflow from fgbio to fgumi. I think having the UMI affixed to the read name is pretty common, do you think it'd be worth supporting this in fgumi?

> [!NOTE]
> Authorship: the content below was drafted by an AI coding agent and filed via gh under @msto's account. It reflects the agent's analysis, not a statement authored by @msto.

## Problem

fgbio's `CopyUmiFromReadName` extracts the UMI from a BAM read name (last `:` field) and writes it to the `RX` tag. Pipelines migrating from fgbio to fgumi still need this step, since `fgumi group` reads UMIs from `RX` (hardcoded) and `fgumi extract` operates on FASTQs, not BAMs.

This is the only remaining fgbio dependency in our pipeline after porting GroupReadsByUmi, CallMolecularConsensusReads, FilterConsensusReads, and ZipperBams to their fgumi equivalents.

### Option A: Standalone `fgumi copy-umi` command

A new command that:
- Reads a BAM
- Extracts the UMI from the read name (configurable delimiter, default `:`, last field)
- Writes it to the `RX` tag
- Handles multi-segment UMIs (e.g. `+` or `-` delimited)
- Optionally removes the UMI from the read name

Composable — works in a pipe between `samtools fixmate` and `fgumi sort`.

### Option B: `--umi-from-read-name` flag on `fgumi group`

Add a flag to `fgumi group` that reads the UMI directly from the read name instead of requiring the `RX` tag. This avoids a separate pre-processing step entirely and is closer to how many pipelines work (UMIs are in the read name from the sequencer, never explicitly copied to a tag before grouping).

This would be the simpler migration path for pipelines coming from fgbio, since `CopyUmiFromReadName` + `GroupReadsByUmi` collapses into a single `fgumi group` call.

Either option would allow complete migration away from fgbio for standard consensus calling pipelines.

nh13:
Closes #869.

Adds `fgumi copy-umi`, the port of fgbio's `CopyUmiFromReadName`: a non-destructive, streaming BAM→BAM command that copies the UMI embedded in each read name into the `RX` tag so downstream fgumi tools (which read UMIs only from `RX`) can consume it. This was the last remaining fgbio dependency for pipelines whose UMI arrives only in the read name (e.g. after `samtools fixmate`), and it composes in a pipe between alignment/fixmate and `fgumi sort`.

## What it does

- The UMI is the **last `--field-delimiter` (default `:`) field** of the read name, under fgbio's lenient last-field rule — no field-count requirement, unlike `extract`'s strict ≥8-field Illumina rule.
- It is normalized to the canonical SAM form via the shared normalizer: dual UMIs joined with `+` in the name become `-`-joined; `r`-prefixed segments are reverse-complemented back to the forward strand by default (`--reverse-complement-r-umis false` strips the marker instead, matching UMI-tools/fgpyo); the result is upper-cased and validated against `ACGTN-`.
- It runs on the unified BAM pipeline with `SingleRawRecordGrouper` (modeled on `filter`'s per-record path), so it honors `--threads` and the shared compression/scheduler/queue-memory options. It works on `RawRecord`, touching only the QNAME and the aux block, so alignments, `SEQ`, and unrelated tags pass through untouched.
- `--remove-umi` trims the UMI (and its delimiter) off the read name. `RX` is overwritten by default (`--fail-if-tag-present` errors instead), with an optional `-M` metrics report.
- Fail-fast on malformed names: too few fields, an empty last field, an invalid UMI, a field that normalizes to empty (e.g. a bare `r`), or a `--remove-umi` that would leave an empty read name.

## Design notes

- **Standalone command, not folded into `retag`.** `retag` is deliberately tag↔tag only ("never touches read names"); read-name handling stays out of it. `copy-umi` is the sanctioned home, mirroring fgbio's own split between `FastqToBam` (FASTQ) and `CopyUmiFromReadName` (BAM).
- **`RX` and the in-name `+` delimiter are hardcoded**, matching fgumi's opinionated tag vocabulary; `retag` handles alternate destination tags.
- **Empty/degenerate handling diverges from fgbio deliberately**: fgbio writes an empty `RX` for an empty field; `copy-umi` errors instead, since an empty `RX` only defers the failure to `group` (mixed UMI lengths) with a worse message.

## Suggested reading order

1. `refactor(umi): …` — a pure move lifting `normalize_read_name_umi` out of `extract` into `src/lib/umi/read_name.rs` as a shared free function, with no behavior change. Start here; it is small.
2. `feat(copy-umi): …` — the command itself, its tests, and the CLI/matrix wiring.

Note the parity fix called out in the second commit message: consolidating the normalizer also corrects a pre-existing fgbio-parity gap for an `r`-prefixed UMI field ending in a trailing `+` (e.g. `…:rACGT+`), which changes `extract`'s output for that narrow case toward fgbio parity (drops a spurious trailing `-`).

## Follow-up (separate PR)

A follow-up will bring `extract`'s FASTQ read-name path to the same lenient last-field parity, so the two commands share both the normalization and the field-location policy. The strict-vs-lenient default for `extract` (which affects `FastqToBam` parity) is the open decision to settle there.

## Testing

Full suite (`cargo ci-test`), `cargo ci-fmt`, `cargo ci-lint` (`-D warnings`), and the tag-literal check all pass. New coverage includes fgbio-parity vectors (RC on and strip-only, `+`→`-`, trailing/internal-empty split semantics), the non-destructive contract (SEQ/CIGAR/POS/QUAL and unrelated tags survive), `--field-delimiter`, `--threads`, overwrite/fail-if-present, metrics, write-target guards, and discriminating assertions on every error path.

Risk: outputs change for `RX`, read names, and `copy-umi` metrics, pinned by normalization, delimiter, overwrite, trimming, and metrics options; `unsafe`: none, and no `CLAUDE.md` allowlist update applies; memory/thread/backpressure policy: configurable queue memory, threading, and scheduler options are added.

Fix: Add `fgumi copy-umi` to extract normalized UMIs from BAM read names, write them to `RX`, and optionally trim read names.

- Add streaming, parallel BAM processing with overwrite protection and metrics.
- Share read-name UMI normalization with `extract`.
- Reject malformed or empty UMIs.
- Preserve alignment data and unrelated tags.
- Add CLI, README, I/O matrix, fixtures, and end-to-end tests.
- Enable migration from fgbio `CopyUmiFromReadName` before `fgumi sort`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,7 @@ Enable with: `cargo build --release --features <feature>`
 | `review` | Review consensus variants | `fgbio ReviewConsensusVariants` |
 | `downsample` | Downsample BAM by UMI family | N/A |
 | `retag` | Rewrite SAM tags (copy/move/delete) | N/A |
+| `copy-umi` | Copy the UMI from the read name into the RX tag | `fgbio CopyUmiFromReadName` |
 | `compare <cmd>` | Compare files (feature-gated) | N/A |
 | `simulate <cmd>` | Generate test data (feature-gated) | N/A |
 
--- a/src/lib/commands/copy_umi.rs
+++ b/src/lib/commands/copy_umi.rs
@@ -1,0 +1,410 @@
+//! `copy-umi` — copy the UMI embedded in a BAM read name into the `RX` tag.
+//!
+//! Ports fgbio's `CopyUmiFromReadName`: the UMI is the last field of the read
+//! name (split on `--field-delimiter`, default `:`), normalized to the canonical
+//! SAM form (dual UMIs `-`-joined, `r`-prefixed segments reverse-complemented by
+//! default), and written to the `RX` tag. It is a non-destructive, streaming
+//! BAM→BAM pass: alignments, `SEQ`, and other tags pass through untouched, and
+//! only the read name (optionally, with `--remove-umi`) and the `RX` tag change.
+//!
+//! Unlike `extract` (which parses FASTQ read names during unmapped-BAM creation
+//! and uses fgbio's strict ≥8-field rule), `copy-umi` uses fgbio
+//! `CopyUmiFromReadName`'s lenient rule: the UMI is simply the last delimited
+//! field, with no field-count requirement. The two commands share the
+//! normalization step (`crate::umi::read_name::normalize_read_name_umi`) but keep
+//! their own field-location logic.
+
+use std::io;
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
+
+use anyhow::{Context, Result, bail, ensure};
+use clap::Parser;
+use log::{info, warn};
+use noodles::sam::Header;
+
+use crate::commands::command::Command;
+use crate::commands::common::{
+    BamIoOptions, CompressionOptions, QueueMemoryOptions, SchedulerOptions, ThreadingOptions,
+    add_pg_record, build_pipeline_config, ensure_hd_record, reject_output_collisions,
+    serialize_raw_bam_records,
+};
+use crate::grouper::SingleRawRecordGrouper;
+use crate::logging::OperationTimer;
+use crate::per_thread_accumulator::PerThreadAccumulator;
+use crate::sam::SamTag;
+use crate::umi::read_name::normalize_read_name_umi;
+use crate::unified_pipeline::{Grouper, MemoryEstimate, run_bam_pipeline_from_reader};
+use fgumi_bam_io::create_bam_reader_for_pipeline_with_opts;
+use fgumi_raw_bam::{RawRecord, update_string_tag};
+
+/// Copy the UMI from a BAM read name into the RX tag.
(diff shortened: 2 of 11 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #875, opened 2026-08-27
perf(pipeline): skip per-record group-key computation for key-agnostic groupers

nh13:
## What

The unified BAM pipeline computed a `GroupKey` for **every** record during decode (coordinate + library-hash + UMI-position scan), even when the grouper never reads it. `SingleRawRecordGrouper` — used by `filter`'s single-read mode — calls `into_raw_bytes()` and discards the whole decoded record, so that work was pure overhead.

This threads the pipeline's `Option<GroupKeyConfig>` through to `decode_records` instead of unwrapping it to a header-built default at the three `run_bam_pipeline` entrypoints, and skips `compute_group_key_from_raw` when it is `None`, attaching a default placeholder key instead. `filter`'s single-read mode is wired to pass `None`; template mode stays `Some`.

## Why it's safe

`None` is only ever valid with a grouper that never reads the key. I verified the invariant holds for every path:

- Every key-consuming command sets `Some` explicitly: `group`, `dedup`, `clip`, `codec`, `correct`, `duplex`, `simplex`, and `filter`'s **template** mode.
- `TemplateGrouper` reads `decoded.key.name_hash` as a QNAME fast-path pre-check, so it must keep `Some` — and it does. (Its byte-level QNAME comparison is authoritative, so even a mistaken `None` would only lose the fast-path, not mis-group; but it stays `Some` regardless.)
- `SingleRawRecordGrouper` discards the decoded record entirely, so it never reads the key.
- Output order is preserved by the `ReorderBuffer` keyed on batch serial, **not** by `GroupKey`, so the placeholder key cannot affect ordering.

Note this does not introduce a new hazard for a hypothetical future misuse (key-consuming grouper + `None`): the previous `None → default-from-header` path was *already* wrong for such a grouper (default library index → incorrect cross-library grouping); skipping the key instead makes that misuse fail visibly in a test rather than silently mis-group. The `None`-means-skip contract is documented on `decode_records` and the `group_key_config` field.

## Testing

- Unit test `test_decode_records_skips_group_key_when_none`: `None` yields the default placeholder key and **byte-identical** record bytes, while `Some` computes a real coordinate key — proving the compute is skipped without disturbing passthrough.
- End-to-end: wiring `filter`'s single-read mode to `None` means the entire `filter` integration suite now exercises the `None` path. Full suite (`cargo ci-test`, 8031 tests), `cargo ci-fmt`, `cargo ci-lint` (`-D warnings`), tag-literal check, and `cargo doc` all pass.

## Impact

`filter --threads N` single-read filtering skips the unused per-record key work immediately. Profiled at ~0.6% of CPU on a 20M-record streaming pass (the workload is ~83% BGZF-bound, so this is a small but free win). PR #873 (`copy-umi`) also benefits automatically once merged — it already leaves `group_key_config` unset (`None`).

Risk: output changes apply only to grouping-related key computation, with template mode retaining `QNAME`/`LibraryIndex` keys and batch serials preserving record order; `unsafe` changes: none, and no `CLAUDE.md` allowlist update applies; memory bounds, queue capacity, and thread/backpressure policy: none.

- Single-read `filter` mode now skips unnecessary `GroupKey` computation.
- Template filtering and other key-consuming commands retain configured group-key computation.
- Optional `GroupKeyConfig` now propagates through BAM decoding and pipeline entry points.
- Decoding uses a placeholder key when no grouper consumes the key.
- Added coverage for skipped key computation and unchanged raw record bytes.
- Updated integration callers for the optional configuration.
- Reported validation passes for integration, formatting, lint, documentation, and full test suites.
- Measured improvement: approximately 0.6% CPU on a 20-million-record streaming workload.

### DIFF
--- a/src/lib/commands/filter.rs
+++ b/src/lib/commands/filter.rs
@@ -392,8 +392,15 @@ impl Filter {
             num_threads,
         )?;
 
-        let library_index = LibraryIndex::from_header(header);
-        pipeline_config.group_key_config = Some(GroupKeyConfig::new_raw_no_cell(library_index));
+        // Template mode groups by QNAME and reads the key's `name_hash` fast-path
+        // (see `TemplateGrouper`), so it needs the key. Single-read mode uses
+        // `SingleRawRecordGrouper`, which discards the decoded record entirely and
+        // never reads the key — so skip its per-record computation with `None`.
+        pipeline_config.group_key_config = if self.filter_by_template {
+            Some(GroupKeyConfig::new_raw_no_cell(LibraryIndex::from_header(header)))
+        } else {
+            None
+        };
 
         let config = Arc::new(FilterConfig::new(
             &self.min_reads,
--- a/src/lib/unified_pipeline/bam.rs
+++ b/src/lib/unified_pipeline/bam.rs
@@ -17,10 +17,10 @@ use std::time::{Duration, Instant};
 
 use crate::bgzf_reader::{BGZF_EOF, decompress_block_into_opts, read_raw_blocks};
 use crate::bgzf_writer::InlineBgzfCompressor;
+#[cfg(test)]
 use crate::sam::SamTag;
 use fgumi_bam_io::ProgressTracker;
 use fgumi_bam_io::ReorderBuffer;
-use noodles::sam::alignment::record::data::field::Tag;
 
 use super::base::{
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #870, opened 2026-08-26
feat(pipeline): typed-step pipeline foundation

nh13:
Lands the **typed-step pipeline foundation** onto `main`, rebased on v0.7.0. Each unit below was developed and reviewed as its own sub-PR on a shared integration branch; this brings them to `main` together so the workspace stays buildable at every crate boundary.

The domain-crate split (`fgumi-dna`, `fgumi-umi`, `fgumi-consensus`, `fgumi-sort`, `fgumi-bgzf`, `fgumi-bam-io`, `fgumi-raw-bam`, …) already shipped in v0.7.0 and is **not** introduced here — this PR adds the pipeline framework and a few supporting leaf crates on top of it.

**New crates (5):** `fgumi-fmt`, `fgumi-cli-common`, `fgumi-cli-macros` — format/CLI leaf crates (#672, #674); `fgumi-pipeline-core`, `fgumi-pipeline-io` — the typed-step pipeline framework (#697, #732).

**Typed-step pipeline foundation** (`src/lib/pipeline/**` + the two pipeline crates): the typed-step pipeline tree and foundational steps (#735), read-side source steps (#736), grouping and closure-driven mid-steps (#743), the UMI-correction step (#821), per-stage CLI option projection (#744), and the shared grouping/library-lookup domain types (#734, #741). The chain-builder layer that *consumes* this framework is not part of this PR; the framework is dormant on every production command path here.

**Sort-engine additions to the existing `fgumi-sort` crate:** the unified arena sort engine for all four sort orders (#720), the phase-2 merge handoff slot (#718), and BGZF slice decompression with caller-driven buffer recycling in `fgumi-bgzf` (#714).

Constituent PRs: #672, #674, #697, #714, #718, #720, #722, #732, #734, #735, #736, #741, #743, #744, #745, #821. Behavior is unchanged from what those merged; the only new work here is the v0.7.0 rebase and the crate-version reconciliation in `Cargo.lock`.

## Regression validation (v0.7.0 → this branch, t=16)

Output is **identical** and there is **no speed regression** — measured directly, v0.7.0 binary vs this branch:

- **`sort`** — byte-identical output (3,689,310 and 8,224,560 records).
- **`group`** (adjacency / edit / paired) — MI-equivalent (~86K molecules).
- **`dedup`** — content-identical (2,586,791 records).
- **`simplex`** — content-identical (81,768 consensus reads).
[...]

PR #876, opened 2026-08-27
feat(pipeline): wire dedup --threads onto the declarative chain builder

nh13:
Makes `dedup --threads N` a **live** chain path, the second command (after `group`) wired onto the declarative chain-builder layer. Stacked on #872 — base is `nh/chain-builder` and will be retargeted to `main` (and rebased) once #872 merges.

`dedup --threads N`'s hand-rolled unified-pipeline construction is replaced with `ChainSpec::single_stage(Stage::Dedup) → build_for → run`, reusing the shared helpers #872 introduced. dedup is the closest sibling to `group` (same template-coordinate input, same molecule machinery), so this mirrors the group pilot (`b01e18ea` in #872) closely.

## Key difference from the group pilot

Unlike `group`, `dedup` has no separate single-threaded fast path — its whole `execute()` is the unified pipeline scaled by thread count. This PR still **gates on `--threads`** and leaves the no-`--threads` path untouched, so that path stays as the **in-process parity oracle** the new tests diff against. Replacing all of `execute()` (the one-engine cleanup) is a deliberate follow-up, not this PR.

## What's here (reading order, oldest → newest)

1. **`thread the duplication-ladder recorder through the chain dedup MiAssign step`** — the dormant chain dedup builder carried `--metrics` and `--family-size-histogram` but not `--duplication-ladder`; repointing dedup would have silently dropped the ladder. The ladder is order-sensitive (it samples a saturation curve at cumulative-template intervals), so it is recorded in the **serial `MiAssign` step** — the same coordinate-order seam the non-chain path uses — not the parallel serialize step. `MiAssign` is `Serial` + `ByItemOrdinal`, so batches arrive in input-record order and groups within a batch are coordinate-ordered, reproducing the non-chain per-group stream exactly. The finalize hook reads the recorder through the lock (not `Arc::try_unwrap`), holds the `(path, recorder)` pair in one `Option` so the invariant is in the type, and now also emits the `"Filtered out N templates before marking"` diagnostic the non-chain path emits (it must not be `--metrics`-only). Promotes `DuplicationLadderRecorder`/`write_duplication_ladder` to `pub(crate)` and adds `Clone` to `MarkDuplicates`.
2. **`repoint dedup --threads onto the chain builder`** — the pilot: `execute_chain` + an early `--threads` dispatch placed after the reader-free pre-flight validations (which run on both paths) but before the timer/banner/reader (which `add_dedup` re-emits, and pre-opening would consume stdin, breaking stdin + `--threads`). The `--no-umi` override log is gated to the non-chain path so `--threads` doesn't double-log it. Adds the parity tests below.

## Validation

The pilot adds tests that pin behaviors the non-chain oracle cannot see:

- **`test_dedup_chain_matches_single_threaded`** — `--threads 1` and `--threads 4` output record-for-record identical to the non-chain path.
- **`test_dedup_chain_matches_non_chain_across_knobs`** — the CLI-default `adjacency`, `edit`, `--remove-duplicates`, `--no-umi`, and `--min-map-q` (filtering the fixture's mapq-10 subfamily) all match the non-chain path, on **mixed-UMI input** so adjacency/edit cluster non-trivially rather than collapsing to identity; the `@HD` sort-order header is compared too (the `@PG` command-line field legitimately differs between the two invocations).
- **`test_dedup_mixed_umi_fixture_distinguishes_adjacency_from_identity`** — a non-vacuity guard: adjacency must produce different output than identity on that fixture, else the adjacency parity cases pass vacuously.
- **`test_dedup_threaded_crc_policy`** — `--no-check-crc` / `--check-crc` / default CRC policy honored on the chain path; the accept case asserts record identity against an intact-file baseline, not merely non-empty.
- **`test_dedup_threaded_duplication_ladder_parity`** — the order-sensitive `--duplication-ladder` (plus `--metrics`, `--family-size-histogram`) file is byte-identical between the chain and non-chain paths. Runs at `--threads 4` with several hundred **non-uniform** position groups so a reordering regression in the ladder recording actually diverges the file (a `--threads 1` or uniform-size test could not catch it).

Beyond the suite (9,481 tests green), the `dedup` output was checked at scale against the non-chain engine on a 2.59M-record template-coordinate BAM (1.29M templates, 95% duplicate rate, 558K filtered): single-threaded **and** `--threads 8`, identity **and** default adjacency — output records identical (2,586,791 each, re-sorted to coordinate/mi then content-compared), and the duplication ladder, metrics, and family-size histogram byte-identical in every case; the restored filtered-templates diagnostic was confirmed emitted byte-identically on both paths at scale. The branch also went through two rounds of adversarial multi-lens review plus a scoped CodeRabbit-style pass; the most substantive finding was the dropped filtered-templates diagnostic (restored and extracted into a shared helper both paths call, so they cannot drift), and the second round's findings were all test-coverage gaps, now closed by the cases above.

## Deferred / follow-ups (not in this PR)

- **One-engine cleanup** — delete dedup's old unified-pipeline path entirely once every command is wired.
- **Per-command wiring PRs** — the remaining dormant builders (clip, codec, correct, duplex, simplex, filter, sort, zipper, align, extract, fastq) each get their own PR.

## Merge gate

Same release gate as #870/#872: the fgumi-benchmarks AWS run (fgbio equivalency + WES/WGS scale). This PR changes only the `dedup --threads` path's internal construction (proven record-identical at scale) and adds tests.

Risk: threaded `dedup` output changes, but parity and byte-identical tests pin grouping, corrected UMIs, metrics, histograms, duplication ladders, and BAM output; `unsafe`: none, with no allowlist update; memory, queue capacity, and thread/backpressure policy: none.

Fix: route `dedup --threads N` through `ChainSpec` while preserving the serial path.

- Record duplication ladders in coordinate-ordered `MiAssign`.
- Finalize ladders through the chain finalize hook.
- Preserve CRC handling, diagnostics, metrics, histograms, and output behavior.
- Add parity and non-vacuity tests for deduplication options, mixed-UMI input, mapping-quality filtering, CRC policies, and output artifacts.

### DIFF
--- a/src/lib/commands/dedup.rs
+++ b/src/lib/commands/dedup.rs
@@ -1327,9 +1327,13 @@ impl Command for MarkDuplicates {
             bail!("--no-umi cannot be used with --strategy paired");
         }
 
-        // Handle --no-umi mode: force identity strategy
+        // Handle --no-umi mode: force identity strategy. The effective-strategy
+        // computation itself must run on both paths (validate_index_threshold
+        // below consumes it), but the override log is gated to the non-chain
+        // path: the --threads chain re-emits it via add_dedup, so logging here
+        // too would double-log.
         let (effective_strategy, no_umi_edits_override) = if self.no_umi {
-            if !matches!(self.strategy, Strategy::Identity) {
+            if !matches!(self.strategy, Strategy::Identity) && self.threading.threads.is_none() {
                 info!("--no-umi mode: overriding strategy to identity");
             }
             (Strategy::Identity, true)
@@ -1356,6 +1360,19 @@ impl Command for MarkDuplicates {
 
         // Validate the input exists (stdin paths are exempt).
         self.io.validate()?;
+
+        // --threads N: run the dedup stage on the declarative chain builder.
+        // Dispatch here — after the reader-free pre-flight validations above
+        // (output collisions, strategy/min-umi combos, index-threshold,
+        // input existence), which must run for both paths — but BEFORE the
+        // timer/banner/reader below. `execute_chain` → `add_dedup` re-emits the
+        // timer, banner, and threading log lines and opens its own source, so
+        // running them here too would double-log and pre-consume stdin
+        // (breaking stdin + --threads). The no-`--threads` path keeps the
+        // hand-rolled unified pipeline below.
+        if self.threading.threads.is_some() {
+            return self.execute_chain(command_line);
+        }
 
         let min_mapq: u8 = self.min_map_q.unwrap_or(0);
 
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #880, opened 2026-08-29
fix(pipeline): fall back to a name-only group key on an unreadable aux offset

nh13:
`compute_group_key_from_raw` (`src/lib/unified_pipeline/bam.rs`) sliced `&raw[aux_offset..]` where `aux_offset = aux_data_offset_from_record(raw).unwrap_or(raw.len())`. The `unwrap_or` guards a *missing* offset (`None`) but not an out-of-range `Some(offset)`. `aux_data_offset_from_record` derives the offset from the record's own `n_cigar_op` / `l_seq` header fields, so a malformed or truncated record can report an offset past the record body and panic the slice while computing its group key. `validate_record_for_decode` does not reject this offset, so a corruption-controlled input reaches it.

## Fix

Clamp with `.min(raw.len())` at both aux-slice sites — the primary path and the secondary/supplementary (`tc`-keyed) path — matching the empty-aux fallback that `fgumi_raw_bam`'s sibling `aux_data_slice` helper already applies for a missing offset. An out-of-range record now yields an empty aux slice (no tags, and therefore no UMI position) and a name-only key, instead of panicking.

## Test

Adds `compute_group_key_from_raw_survives_out_of_range_aux_offset` with primary and secondary `#[rstest]` cases. Each builds a valid raw record then inflates `l_seq` so `aux_data_offset_from_record` reports an offset past the body (CIGAR/position left intact so both slice sites are reached). Without the clamp both cases panic at the slice; with it they return a name-only key.

Surfaced during CodeRabbit review of #872, but the code lives on `main` (it merged via #870), so it is fixed here rather than in that PR.

Risk: grouping output changes for malformed records, pinned by primary and secondary regression tests; `unsafe` changes: none, and the `CLAUDE.md` allowlist is unchanged; memory bounds, queue capacity, and thread/backpressure policy changes: none.

Clamps auxiliary-data offsets to `raw.len()` in both primary and secondary paths. Malformed records now produce a name-only group key instead of panicking.

Adds regression tests for out-of-range offsets.

### DIFF
--- a/src/lib/unified_pipeline/bam.rs
+++ b/src/lib/unified_pipeline/bam.rs
@@ -492,7 +492,15 @@ pub fn compute_group_key_from_raw(
         // UNKNOWN position that relies on the read sorting adjacent to its
         // primary. Library/cell are extracted the same way as primaries so the
         // position key matches (they share their template's RG/CB).
-        let aux_offset = fgumi_raw_bam::aux_data_offset_from_record(raw).unwrap_or(raw.len());
+        //
+        // `.min(raw.len())` guards a malformed record: `aux_data_offset_from_record`
+        // computes the offset from the record's own `n_cigar_op` / `l_seq` fields, so
+        // a corrupt or truncated record can report an offset past the body. Left
+        // unclamped the slice below would panic; clamp it to an empty aux slice
+        // instead — the same fallback `aux_data_offset_from_record`'s companion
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #789, opened 2026-08-16
test: harden pipeline-backpressure FAMILIES fixture read-ahead margin against BGZF block-size drift


nh13:
The pipeline backpressure integration tests in `tests/integration/test_pipeline_memory_backpressure.rs` size their input via a `FAMILIES` constant. During the 2026-08-16 review sweep, the read-ahead margin was measured at ~2.46 MiB sitting just under `input_len / 2` (~2.7 MiB) — limited headroom. A future toolchain or dependency change to BGZF block sizing could materially shift how much the reader buffers and break the margin, making these tests flaky or wrong.

The most brittle assertion (`settled.bytes < input_len / 2` in `stalled_writer_stops_the_reader_within_the_queue_memory_budget`) was already made budget-relative (`unread >= TEST_BUDGET_BYTES`) as part of PR #764's coverage fix. This issue tracks the remaining hardening: make the fixture sizing / any remaining fraction-based margins robust to BGZF block-size drift rather than relying on the current tight headroom.

### Task
- Audit `FAMILIES`/fixture sizing and every margin assertion in `test_pipeline_memory_backpressure.rs`.
- Make margins budget-relative (assert the reader left ≥ one budget unread) or size the fixture with clear headroom, without regressing the llvm-cov coverage-job runtime (the fixture was deliberately shrunk to avoid a >120s timeout — don't reintroduce that).

### Placement note
The fixture and these tests are introduced by **PR #764** (branch `746/nhomer/bound-consensus-memory-under-writer-stall`) and are **not in `main` yet**. The fix therefore belongs in #764 (folded into that PR), not a separate branch off `main`. This issue exists for tracking; link it from #764.

### Notes
- Found during the multi-PR review sweep on 2026-08-16; see the session retro (`reports/2026-08-16-retro.md`, item #7).

PR #881, opened 2026-08-29
test: speed up the slow-test tail without weakening invariants

nh13:
## What

Speeds up the slowest tests in the suite without weakening the invariants they protect (per the slow-test audit). Four independent clusters, one commit each; the whole `test_pipeline_memory_backpressure` module drops from ~78s to ~16s wall, the consensus error-bound cases from ~5s to ~0.3s each, and the sort soak block from ~143 CPU-s to seconds under coverage — with every guarantee intact.

## Cluster by cluster

**1. Backpressure read-ahead — observe the real gate instead of polling timing.** The read-ahead test measured where the reader parks by polling for wall-clock quiescence, a heuristic that false-settles when the reader is descheduled mid-fill — which is why #809 had to average each arm over `READ_AHEAD_REPS=3` samples, and why the flagship was the slowest test in the suite (~100s under coverage). This exposes the pipeline's actual gate condition: `PipelineConfig` gains an optional `queue_bytes_probe: Option<Arc<AtomicU64>>` (`None` in every production path — a single predictable branch, no behaviour change) that `read_admission_allowed` publishes the in-flight-bytes total into, and the test waits on `in_flight >= queue_memory_limit` directly. A descheduled reader shows in-flight *below* the limit, so the wait keeps going rather than declaring a partial read-ahead settled — immune to the #809 tail by construction, so each arm takes a single measurement (no reps). The shared fixture also shrinks 175k→143k families (still ~3.3 budgets, above the #789 drift-headroom floor that `shared_bam_bytes` asserts), and the now-obsolete per-test nextest slow-timeout override is dropped.

**2. Consensus ln-posterior error bound — stride the deep-depth sweep.** The test ran `add()` to depth 4000 for every one of ~94×94 `(post, obs)` pairs, though the worst case lives at depth 1 (on the always-run dense `1..=60` sweep). The deep spot-checks `[100..=4000]` now run only on a coarse stride of the grid; the gap's per-observation increment can't change sign with depth, so this re-confirms the deep regime broadly without repeating it thousands of times. Asserted bound and 5.0 ratio ceiling unchanged.

**3. Simulate-sort & group-determinism — shrink oversized fixtures.** Simulate's large case 10k→4k molecules and spill-parity cases 2k→1.5k (both still spill to disk and k-way merge). The determinism guard's mixed-orientation fixture 40→24 position groups, keeping `n_runs=6` and 256 UMIs — the unsorted-`AHashMap`-iteration bug is detected per position group across hash-seed-varying runs, so two dozen independent groups keep detection effectively certain. Assertions unchanged.

**4. Sort soak — collapse iterations under coverage only.** The soak tests (already `stress-tests`-gated, so coverage + nightly only) repeated their full iteration counts in the coverage job purely to line-cover ~200 lines. The count is now gated on `cfg(coverage)` (which `cargo llvm-cov` sets, registered like the existing `loom` cfg): the coverage job runs a minimal pass, and the full race-hunting soak still runs on the nightly `ci-test-stress` job, which does not set the cfg. No quarantine — the nightly soak is untouched.

## Verification

- Full suite green: `cargo ci-test` → 9524 passed; `cargo ci-fmt` and `cargo ci-lint` (incl. `--features stress-tests` and a `--cfg coverage` build) clean.
- Backpressure: both gate-break scenarios (gate always-admits; in-flight forced to zero) still make the read-ahead and stalled-writer tests **fail**; stress-tested 10/10 pass under CPU oversubscription with zero gap-flakes.
- Consensus: the documented worst case (`base=G post=Q35 obs=Q56 depth=1`) still lands on the dense shallow sweep, so the ratio ceiling stays meaningful.
- Soak `cfg(coverage)` path exercised: the full 14-case soak set passes under `--cfg coverage` in ~1s; the nightly (non-coverage) build still runs the full iteration counts.

## Notes / follow-ups

- `test_fastq_pipeline_memory_backpressure` is a sibling that still uses the wall-clock-quiescence helper. Giving `FastqPipelineConfig` the same probe and sharing one gate-observing helper is a natural follow-up (not done here to keep this change focused; the FASTQ test is unchanged from `main`).

Risk: command output changes: none; `unsafe` changes: none, and `CLAUDE.md` allowlist updates: none; memory bounds, queue capacity, and thread/backpressure policy changes: none.

- Speeds slow tests without changing assertions or protected invariants.
- Uses `queue_bytes_probe` for direct backpressure-gate detection instead of wall-clock polling.
- Strides deep consensus checks while retaining dense shallow coverage.
- Reduces sort, determinism, and simulation-sort fixtures while preserving spill, merge, and hash-order coverage.
- Reduces sort soak iterations only under coverage instrumentation. Nightly stress runs retain full iterations.
- Extends the stalled-writer test timeout to four 60-second periods.
- Verification reports 9,524 passing tests, clean formatting and lint checks, successful coverage builds, and stable backpressure stress tests.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,13 +1,2 @@
 [profile.default]
 slow-timeout = { period = "60s", terminate-after = 2 }
-
-# `read_ahead_under_a_stalled_writer_shrinks_with_the_queue_memory_budget` measures
-# each budget arm's read-ahead as the maximum over READ_AHEAD_REPS stalled-writer
-# runs (issue #809), so it spawns the stalled pipeline several times back to back,
-# each with its own settle wait. That is legitimately slow (~30s locally, and 2-3x
-# that under llvm-cov instrumentation plus CI's 2x CPU oversubscription), enough to
-# trip the default 120s (60s x 2) terminate-after. Give this one test a longer leash
-# rather than dropping reps and weakening the flake-suppression the maximum provides.
-[[profile.default.overrides]]
-filter = 'test(=test_pipeline_memory_backpressure::read_ahead_under_a_stalled_writer_shrinks_with_the_queue_memory_budget)'
-slow-timeout = { period = "60s", terminate-after = 4 }
--- a/crates/fgumi-consensus/src/base_builder.rs
+++ b/crates/fgumi-consensus/src/base_builder.rs
@@ -2357,9 +2357,14 @@ mod tests {
     /// `obs_qual` sweep their *entire* representable range (`0..=93`, not a sparse
     /// list), and `depth` densely covers `1..=60` — deliberately starting at `1`, not
     /// some higher floor, because the observed worst case (`base=G, post=Q35,
-    /// obs=Q56, depth=1`) is at the *shallowest* possible depth, plus a handful of
-    /// deeper spot-checks (up to `4000`) to retain the deep-depth coverage this test
-    /// has always had. `pre` is deliberately fixed at Q45 (not swept): `w`, `l`, and
+    /// obs=Q56, depth=1`) is at the *shallowest* possible depth. Deeper spot-checks
+    /// (up to `4000`) retain the deep-depth coverage this test has always had, but —
+    /// since running `add()` to depth 4000 for *every* `(post, obs_qual)` pair is what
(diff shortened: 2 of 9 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #883, opened 2026-08-30
feat(sort): inline BAI indexer on the arena sink

nh13:
Replaces the arena chain sort sink's post-pass BAI re-read (`IndexBamFinalizeHook` → `noodles::bam::fs::index`) with an inline indexer built during the BGZF write: per-record alignment context and per-physical-block compressed sizes are extracted in the parallel `BgzfCompress` step, ride on each `BgzfBlock` as a manifest, and are joined into a `BaiBuilder` owned by the serial `WriteBgzfFile`, which writes the `.bai` after the BGZF EOF. No re-read, no new thread or channel, and zero overhead on the non-indexed path.

Correctness is proven by samtools equivalence on the same BAM: the inline `.bai` matches `samtools index` on region-query records and `idxstats` across placed, placed-but-unmapped, and truly-unplaced reads over multiple references, plus a multi-block case with a record that straddles a physical BGZF block. The 65280-byte block-stride identity between the compressor and the index resolver is relied on explicitly.

This is the first PR of the sort→chain campaign. It does not route the `fgumi sort` command onto the chain — the inline indexer is exercised chain-directly via `SinkSpec::BamWithIndex`; the CLI cutover (and its wall-clock perf gate) follows in the next stacked PR.

## Suggested reading order

1. `feat(bam-io): make extract_alignment_context public and named`
2. `feat(bam-io): BaiBuilder record_with_context, prune_below, pub pending`
3. `feat(pipeline-io): add BamIndexManifest sidecar to BgzfBlock`
4. `feat(pipeline): derive BAM index manifest in BgzfCompress`
5. `feat(pipeline-io): inline BAI indexing in WriteBgzfFile`
6. `feat(chains): inline BAI on the arena sink; drop re-read hook`
7. `test(chains): samtools-equivalence and byte-identity gate for the inline BAI arena sink`
8. `docs(pipeline-io): note coordinate-order precondition on inline BAI indexer`

Risk: sort BAM output remains unchanged; the new `.bai` is pinned by `BamIndexManifest` offsets and `samtools` equivalence tests; `unsafe`: none, with no `CLAUDE.md` update; memory: adds bounded manifest storage with pruning, but changes no queue capacity, thread, or backpressure policy.

- Builds BAI indexes inline during BGZF compression and serial `WriteBgzfFile` output.
- Removes the post-pipeline BAM reread and finalize hook.
- Preserves the non-indexed path without manifest overhead.
- Validates mapped, unmapped, unplaced, multi-reference, multi-block, and straddling-record cases.
- Tests the 65280-byte block-stride assumption and deterministic BAM/BAI output.

### DIFF
--- a/crates/fgumi-bam-io/Cargo.toml
+++ b/crates/fgumi-bam-io/Cargo.toml
@@ -27,12 +27,12 @@ bstr = { workspace = true }
 bytes = { workspace = true }
 crossbeam-channel = { workspace = true }
 log = { workspace = true }
+tempfile = { workspace = true }
 
 [target.'cfg(target_os = "linux")'.dependencies]
 nix = { workspace = true }
 
 [dev-dependencies]
-tempfile = { workspace = true }
 rstest = { workspace = true }
 proptest = { workspace = true }
 criterion = { workspace = true }
--- a/crates/fgumi-bam-io/src/lib.rs
+++ b/crates/fgumi-bam-io/src/lib.rs
@@ -44,8 +44,8 @@ pub use reader::{
 };
 pub use reorder::{DrainReady, ReorderBuffer};
 pub use writer::{
-    BaiBuilder, BamWriter, BgzfWriterEnum, IndexingBamWriter, RawBamWriter, bai_sidecar_path,
-    create_bam_writer, create_indexing_bam_writer, create_optional_bam_writer,
-    create_raw_bam_writer, open_output_writer, write_bai_index, write_bai_sidecar,
-    write_bam_header,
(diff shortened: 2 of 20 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #884, opened 2026-08-30
perf(pipeline): don't fuse single-thread chains that declare Detached steps

nh13:
The sort-command cutover regressed `fgumi sort --threads 1` by ~8-11% wall: at one thread a fusible chain fuses onto the calling thread, which collapses the arena sort's Detached+Parallel overlap (the "N+2" model — a dedicated off-pool thread for serial coordination/I/O overlapping the pool's (de)compression) onto a single core. `should_fuse_single_thread` now declines fusion for any chain that declares a `StepKind::Detached` step (a policy gate, distinct from the structural `is_fusible_chain` capability), so such a chain takes the scheduled path even single-threaded — restoring the overlap. Fusion capability is unchanged (the fused driver can still run a Detached step inline; that's pinned by a test).

Validated: t1 non-spill −2.6%, spill +0.5% vs the owned engine (was +10.9%/+8.2%); `mean_load` 96%→111%.

Documented caveat (no production chain hits it, pinned by a test): a chain with a Detached step plus ≥2 Exclusive steps now fails `NotEnoughThreads` at `--threads 1` instead of fusing.

Stacked on #883.

Risk: command output changes: none; `unsafe` changes: none; memory bounds, queue capacities, and thread/backpressure policy: single-thread scheduling changes for chains with `StepKind::Detached`, including `NotEnoughThreads` for some chains.

Fix: route single-thread chains with detached steps through the scheduled path to preserve overlap with pool decompression work.

- Keep structural fusion support and direct fused execution unchanged.
- Update tests for detached-step completion and ordinal delivery.
- Add coverage for detached plus two exclusive steps returning `PipelineError::NotEnoughThreads`.
- Align the dropped-lane failure test with production by using a `Serial` sink.
- Reported performance improved by 2.6% without spills and 0.5% with spills versus the owned engine.

### DIFF
--- a/crates/fgumi-pipeline-core/src/builder.rs
+++ b/crates/fgumi-pipeline-core/src/builder.rs
@@ -3215,12 +3215,17 @@ mod tests {
         assert_received_every_ordinal_once(&sink, 0);
     }
 
-    /// L2.3 step 7: at `--threads 1` the fusible linear chain runs the Detached
-    /// step **inline** in the fused single-thread driver — no dedicated thread
-    /// is spawned (the fused path returns before `extract_detached_steps`), yet
-    /// the Detached step still drives to completion.
+    /// L2.3 step 7: a chain declaring a Detached step drives to completion at
+    /// `--threads 1` with every ordinal delivered exactly once. End-to-end this
+    /// asserts *correctness*, not which path ran: the fused and scheduled paths
+    /// produce identical output, so a run cannot distinguish them (that is the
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #885, opened 2026-08-30
feat(sort): route the sort command onto the declarative chain builder

nh13:
Routes `Sort::execute_sort` onto the declarative chain/arena engine unconditionally (hand-built `ChainSpec`; `SinkSpec::Bam`/`BamWithIndex` per `--write-index`), keeping the owned `RawExternalSorter` in-tree as a parity oracle (its deletion is a later PR). Closes parity gaps: threading default, memory budget sized by `max(threads, sort_threads)`, `FGUMI_TMP_DIRS` env fallback, owned-style `Spill runs:` summary wording.

Proves byte-identical output vs a saved baseline binary AND samtools (header-normalized) across all four sort orders + `--write-index`, spill and in-memory. Multi-thread: t4 −21%, t8 −15% vs owned (t16 was +3.4% here — the regression the stacked run-extension PR fixes).

**Behavior change beyond sort:** the shared source opener now rejects a records-only SAM with no `@` header (previously accepted with an empty header) across every chain command — fail-closed on malformed input, since records reference reference IDs an empty header can't resolve.

Stacked on the t1-fix PR. Reviewed with a multi-agent gauntlet + a CodeRabbit-style pass; fixes in the final commit.

Risk: sort output changes are pinned by byte-level parity tests against a saved baseline and samtools; `unsafe` changes are none and the `CLAUDE.md` allowlist is unchanged; memory, thread, and backpressure policies change through `max(threads, sort_threads)` and bounded decoded-record output.

- Route `Sort::execute_sort` through the declarative chain/arena engine.
- Preserve `RawExternalSorter` as a parity oracle.
- Support BAM output to stdout and `SinkSpec::BamWithIndex`.
- Classify SAM input by content and reject records-only SAM with `InvalidData`.
- Preserve temporary-directory fallback, spill behavior, CRC checks, and parse error context.
- Add `DecodedRecordBatchToRecordBatch` for bounded, ordered sort ingestion with backpressure.
- Add parity tests for sort orders, indexing, spilling, memory modes, flags, and invalid inputs.
- Report spill counts as `Spill runs`.

PR #888, opened 2026-08-31
fix(sort): thread --max-memory into the pipeline queue budget

nh13:
The sort-command cutover (#885) hardcoded `queue_memory: QueueMemoryOptions::default()` (768 MiB/thread) when building the chain, so `fgumi sort --max-memory` bounded only the sorter's in-memory record buffer, not the bytes held in the inter-stage pipeline queues — a small `--max-memory` (e.g. `-m 64K -@ 8`) shrank the sort buffer while the queues could still use gigabytes, so peak RSS barely moved.

This projects the command's single memory knob (`--max-memory` / `--memory-reserve` / `--memory-per-thread`) onto the queue budget via a new `queue_memory_options` helper, so one flag bounds both budgets. The two totals scale by different thread counts (the sorter by `max(threads, sort_threads)` since the sort phase fills the buffer, the queues by `threads` — the width of the ingest/output plumbing); that asymmetry is intentional and documented on the flag.

## Deliberate, benign consequences of sharing the flag

- **Default queue budget** now follows `--max-memory`'s default (`"768M"` = 768 MB decimal via the size parser), marginally below the former hardcoded 768 MiB, aligning it with the sorter's own default rather than leaving the two at different units.
- **A large or `auto` `--max-memory` does not balloon queue memory.** Raising the queue *total* does not raise the per-stage backpressure marks (512 MiB / 256 MiB, issue #765), which are what actually bound in-flight queue bytes; only a value *below* them tightens the queues (the point of a small `--max-memory`).

## Verification

- `cargo ci-fmt`, `cargo ci-lint` (pedantic), and the no-default-features check are clean.
- 147 sort unit tests (including two new cases that guard the `ChainSpec` wiring against a silent revert) and 16 `test_sort_cutover_parity` integration tests pass.
- Empirical smoke test on 602,870 records: `fgumi sort -m 64KiB -@ 8` completes with no deadlock, output record count equal to input, correct sort order, and peak RSS 248 MB vs 441 MB at the default (the queue budget is genuinely honored). Peak RSS is flat (~380–410 MiB) across `-m 768M/auto/4GiB/8GiB` at `-@ 8` on a 64 GiB host, confirming the per-stage marks bound queue memory rather than the total.

The `ChainSpec` construction was extracted into a unit-tested `build_sort_chain_spec` so the wiring (queue budget, stages, threading, write-index sink) is guarded directly, not only exercised through a full run where a dropped knob is invisible.

Risk: output changes none; `unsafe` changes none and the `CLAUDE.md` allowlist is unchanged; memory and queue budgets change, while queue backpressure policy remains unchanged.

`--max-memory` now controls both the sorter buffer and inter-stage queue budget. `queue_memory_options` applies distinct thread scaling to each budget and preserves the 768M decimal default.

`build_sort_chain_spec` centralizes chain construction and has unit tests for queue configuration, stages, threading, CRC verification, scheduler settings, and the write-index sink. Tests cover low-memory behavior and chain wiring. Formatting, linting, feature checks, 147 sort unit tests, 16 integration tests, and a smoke test pass.

### DIFF
--- a/src/lib/commands/sort.rs
+++ b/src/lib/commands/sort.rs
@@ -32,13 +32,15 @@ use fgumi_sort::{
 };
 
 use log::{debug, info, warn};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use crate::commands::command::Command;
 use crate::commands::common::{
-    CompressionOptions, MaxTempFiles, MemoryLimit, MemoryReserve, parse_max_temp_files,
-    parse_memory, parse_memory_reserve, resolve_memory_budget,
+    CompressionOptions, MaxTempFiles, MemoryLimit, MemoryReserve, QueueMemoryOptions,
+    SchedulerOptions, ThreadingOptions, parse_max_temp_files, parse_memory, parse_memory_reserve,
+    resolve_memory_budget,
 };
+use crate::pipeline::chains::{ChainSpec, SinkSpec, SourceSpec, Stage, StageOptionsBag, build_for};
 
 /// Sort order for BAM files.
 ///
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #889, opened 2026-09-01
feat(sort): restore --read-streams on the arena chain via a concurrent scatter reader

nh13:
## What & why

PR #885's unified-engine cutover left `fgumi sort --read-streams` a dead `warn!` shim: the flag still parsed but the arena chain ignored it. This restores the mechanism of PRs #838/#846 — concurrent positional reads that raise the device's read queue depth — on the current chain source.

The win in #846 was **disk queue depth**, not decode: a single outstanding `read()` leaves most of a deep-queue device's bandwidth idle (EBS gp3 sustains ~358 MB/s at queue-depth-1 vs ~1177 MB/s with four reads in flight). The arena chain's BGZF decode is already parallel (`InflateToArena`) and a single `PrefetchReader` is still queue-depth-1, so only genuine N-way concurrent reads recover it.

## Mechanism

New `crates/fgumi-bam-io/src/scatter_reader.rs`: a `ScatterReader<S: PositionalSource>` that, per 4 MiB fill window, slices the byte range into N sub-ranges, fetches them with independent concurrent positional reads (`pread` / `FileExt::read_at`, which takes `&self` — no shared file offset), reassembles them **in file order**, and presents an ordinary forward `Read` stream, so the downstream BGZF framer is untouched.

- **No new `unsafe`.** Each slice reads into its own `Vec<u8>` owned by the scoped thread that fills it (`std::thread::scope`); `read_at` on `&self` needs no synchronization. `#![forbid(unsafe_code)]` is preserved and the module is Miri-clean.
- **`Auto` (the default)** reads the first `AUTO_PROBE_FILLS` at a single stream, then commits once to `ceil(1.2 GB/s ÷ measured)`, capped at `MAX_STREAMS = 8`. Device-adaptive: gp3 → 4, fast NVMe → 1. An explicit `Fixed(n)` is clamped to the same `[1, MAX_STREAMS]` ceiling.
- **`Fixed(1)` is the plain reader** — no threads, byte-identical to the old direct `File` read. `PipelineReaderOpts`/`ChainSpec` default to `Fixed(1)` for **every** command except `sort`, so scatter probing never turns on codebase-wide (pinned by a regression test).
- Scatter engages only for seekable regular files; stdin, pipes, and non-Unix fall back to the sequential/async reader, warning only when the user *explicitly* asked for `Fixed(n>1)` that can't be honored (`Auto` falls back silently).

## Correctness

Output is byte-identical across `--read-streams 1`/`4`/`auto` — the count changes only *how* the input is read, never *what*. Proven by:
- `scatter_reader` proptests `byte_identical_to_source` / `byte_identical_under_auto` (random data × stream counts × fill/slice sizes), a fail-closed `UnexpectedEof` test for truncation, error-prefix ordering, and a positive test that `Fixed(4)` actually fans reads across threads.
- The `cutover_read_streams_produce_identical_output` integration test (coordinate / queryname-natural / template-coordinate, forced spill), asserting byte-identical sorted output across 1/4/auto.
- Miri over the concurrency tests.

## Benchmark (EBS gp3, validated)

38.1 GB coordinate sort, `c8g.4xlarge`, gp3 300 GB / 3000 IOPS / **500 MB/s**, cold page cache, `-@16 -m256M`, 2 reps each:

| `--read-streams` | mean wall | vs baseline |
|---|---|---|
| 1 (baseline) | 207.6 s | — |
| 4 | 179.2 s | **−13.7%** |
| auto | 180.8 s | **−12.9%** (resolved to 4 streams, as designed) |

All six runs exit 0 and produced 254,093,850 records (== input); `samtools view` md5 is byte-identical across all three arms. The gain is smaller than #846's −28% on the old owned engine because the arena chain's parallel decode already shrank the ingest share — read-queue-depth recovers the remaining fraction.

## Note on the default

Because `--read-streams` defaults to `auto`, a plain `fgumi sort <regular-file>` now runs an 8-fill single-stream throughput probe and may then read with a few concurrent streams. Output is unchanged (byte-identical, per above); on a fast device the probe resolves to a single stream. This is the intended, measured behavior — the flag was always documented as `auto` by default.

Risk: Sort output remains byte-identical because `ScatterReader` reassembles reads in file order; `unsafe` changes: none; memory and concurrency policy: the reader adds up to eight per-window read threads with a fixed 4 MiB fill window, while downstream processing remains unchanged.

Fix: Route `--read-streams` through the BAM input path and use sequential fallback for unsupported inputs.

- Add `ReadStreams::{Auto, Fixed(usize)}` to the shared BAM reader API.
- Add `ScatterReader` for concurrent positional reads on seekable regular files.
- Preserve sequential behavior for non-sort commands and `Fixed(1)`.
- Clamp explicit and auto-selected concurrency to eight streams.
- Keep spill-file reads separate from input BAM reads.
- Warn when explicit stream concurrency cannot apply to stdin or non-regular inputs.
- Add unit, property, integration, and Miri coverage for byte identity, EOF, errors, and stream selection.

### DIFF
--- a/crates/fgumi-bam-io/src/lib.rs
+++ b/crates/fgumi-bam-io/src/lib.rs
@@ -23,6 +23,7 @@ pub mod progress;
 pub mod reader;
 pub mod reorder;
 pub mod sam_input;
+pub mod scatter_reader;
 pub mod writer;
 
 pub(crate) mod vendored;
--- a/crates/fgumi-bam-io/src/lib.rs
+++ b/crates/fgumi-bam-io/src/lib.rs
@@ -37,7 +37,7 @@ pub use paths::{is_stdin_path, is_stdout_path};
 pub use progress::ProgressTracker;
 pub use reader::{
     BamReaderAuto, BgzfReaderEnum, ChainedReader, FgumiBgzfReader, PipelineReaderOpts,
-    RawBamReaderAuto, TeeReader, create_bam_reader, create_bam_reader_for_pipeline,
+    RawBamReaderAuto, ReadStreams, TeeReader, create_bam_reader, create_bam_reader_for_pipeline,
     create_bam_reader_for_pipeline_with_opts, create_bam_reader_with_opts, create_raw_bam_reader,
     create_raw_bam_reader_from_stream_with_opts, create_raw_bam_reader_with_opts, make_bgzf_reader,
     open_normalized_input, read_header_and_replay, read_prefix,
--- a/crates/fgumi-bam-io/src/reader.rs
+++ b/crates/fgumi-bam-io/src/reader.rs
@@ -236,6 +236,49 @@ pub type BamReaderAuto = noodles::bam::io::Reader<BgzfReaderEnum>;
 /// Type alias for a raw BAM reader that supports both single and multi-threaded BGZF.
 pub type RawBamReaderAuto = RawBamReader<BgzfReaderEnum>;
 
+/// Read-stream policy for a seekable BAM/SAM source: how many concurrent
+/// positional reads to issue per fill window.
+///
+/// Ports the semantics of fgumi v0.7.0's `fgumi sort --read-streams` flag. The
+/// mechanism (concurrent positional reads that raise the device's read queue
+/// depth) lives in the `scatter_reader` module; on a slow, deep-queue device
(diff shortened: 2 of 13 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #890, opened 2026-09-01
refactor(sort): delete the CLI's throwaway sorter construction

nh13:
## What

Removes the CLI-side throwaway sorter construction from `fgumi sort`. Since #885 routed the sort command onto the declarative arena chain, `execute_sort` no longer runs the owned engine — but it still built a full `RawExternalSorter` via a private `Sort::build_sorter` helper purely to read three numbers off it for the startup banner (Phase-1/Phase-2 thread counts and the max-temp-files limit). This deletes that dead construction and sources those numbers directly from the command's own options.

Concretely, in `src/lib/commands/sort.rs`:

- **Adds** `Sort::phase1_threads()` / `Sort::phase2_threads()` — small helpers that resolve the effective per-phase thread counts without constructing a sorter. They are deliberate, exact mirrors of `RawExternalSorter::phase1_threads()` / `phase2_threads()` (a two-line `sort_threads.unwrap_or(threads).max(1)` formula); duplicating rather than exporting a cross-crate function is intentional, because after this change the engine's own methods have only one internal caller (`SortBuffer::from_sorter`). The duplication is pinned against drift by a new `test_phase_threads_match_engine`, which constructs a `RawExternalSorter` directly and asserts the two implementations agree across representative inputs (including the zero-override clamp).
- **Sources** the max-temp-files banner line from `self.resolved_max_temp_files(soft_nofile)` directly — an exact identity, since `build_sorter` set the sorter's limit via that same call and the banner read it back.
- **Deletes** `Sort::build_sorter` and the already-`#[allow(dead_code)]` `Sort::auto_initial_capacity` helper, plus the tests that only existed to exercise them.

## What is NOT changed

`RawExternalSorter` / `SortWorkerPool` remain live production code — used by `fgumi merge`, `fgumi simulate`, and the arena chain's own per-chunk in-memory sort (`SortBuffer::from_sorter`). `crates/fgumi-sort` is untouched. This PR only removes the CLI's throwaway *construction* of a sorter for banner values, not the engine itself.

## Correctness

The startup banner text is unchanged: the new helpers reproduce the exact numbers the throwaway sorter reported (`test_sort_thread_logging` pins the literal banner line, and passes byte-identically). Output is unchanged: the `fgumi sort` cutover parity suite (`test_sort_cutover_parity`) is untouched and still passes byte-for-byte against the saved pre-cutover baseline binary across all four sort orders (coordinate, queryname, queryname-natural, template-coordinate), plus the spill/consolidation record-identity and stdin/CRC/write-index cases.

Test coverage is preserved: the deleted wiring/agreement tests were redundant with retained ones (`test_memory_budget_threads` keeps its full formula table; the new `test_phase_threads_match_engine` and the renamed `test_resolved_max_temp_files` assert the same properties without constructing a throwaway sorter), and the `--max-temp-files` consolidation-record-identity test was rewritten to build the engine directly rather than via the deleted helper.

No new `unsafe`. Full gate green (fmt, clippy `-D warnings -W pedantic`, doc `-D warnings`, test suite).

Risk: command output changes: none; existing chain sorting and parity tests pin output. `unsafe` changes: none; CLAUDE.md allowlist unchanged. Memory bounds, queue capacity, and thread/backpressure policy changes: none; helpers mirror existing engine formulas.

Fix: Removed the CLI-only `RawExternalSorter` construction from `fgumi sort`.

- Added sorter-free phase-thread and max-temp-file resolution helpers.
- Updated banner values to use command options.
- Removed obsolete `Sort::build_sorter` and `Sort::auto_initial_capacity`.
- Added tests for engine parity, consolidation, banner values, and record identity.

### DIFF
--- a/src/lib/commands/sort.rs
+++ b/src/lib/commands/sort.rs
@@ -27,9 +27,7 @@ use anyhow::{Result, bail};
 use bytesize::ByteSize;
 use clap::Parser;
 use fgumi_bam_io::is_stdout_path;
-use fgumi_sort::{
-    KeyTypesSpec, QuerynameComparator, RawExternalSorter, SortOrder, verify_sort_order,
-};
+use fgumi_sort::{KeyTypesSpec, QuerynameComparator, SortOrder, verify_sort_order};
 
 use log::{debug, info, warn};
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #891, opened 2026-09-01
fix(sort): resolve --max-temp-files once and bake it into the chain spec

nh13:
## What

Resolves `fgumi sort --max-temp-files` **once** and bakes the result into the chain spec, instead of resolving `Auto` twice from two independent `RLIMIT_NOFILE` snapshots.

Before this change, `execute_sort` resolved `--max-temp-files Auto` for the startup banner (`self.resolved_max_temp_files(soft_nofile)`), while `build_sort_chain_spec` baked the *raw* `MaxTempFiles::Auto` into `SortOptions` — so the chain builder re-resolved it against a second, independent `RLIMIT_NOFILE` snapshot. The two snapshots agree in practice (the descriptor limit is stable within a process), but they are logically independent.

`build_sort_chain_spec` now takes a `resolved_max_temp_files: usize` and bakes `SortOptions.max_temp_files = MaxTempFiles::Fixed(resolved)`. `execute_sort` passes the value it already computed for the banner, so the banner and the chain share a single resolution. The chain builder's own `Auto` branch is unchanged (other callers keep it); `sort` now always hands it a `Fixed` value.

The banner's `auto → N` framing is intentionally unchanged (it still receives the raw `self.max_temp_files` for framing plus the resolved number).

## Correctness

- Output is unchanged: the cutover parity suite (`test_sort_cutover_parity`) passes **byte-identical** (19/19, all four sort orders) against the saved pre-cutover baseline binary, confirming the baked-`Fixed` value matches what the builder resolved before.
- A new unit assertion pins the fix: with `--max-temp-files auto`, the built spec carries `MaxTempFiles::Fixed(resolved)`, not `Auto` — proving the chain is handed a single resolution.
- No new `unsafe`. Full gate green (fmt, clippy `-D warnings -W pedantic`, doc `-D warnings`, test suite).

Addresses a CodeRabbit finding raised on #890.

Risk: command output changes: none, pinned by the cutover parity suite; `unsafe` changes: none, so the `CLAUDE.md` allowlist is unchanged; memory, queue, thread, and backpressure policy changes: none.

Fix: Resolve `Auto` once in `execute_sort` and store the result as `MaxTempFiles::Fixed`. The startup banner and chain builder now use the same `RLIMIT_NOFILE` snapshot.

Tests verify the resolved value in the chain specification. Formatting, linting, documentation, and test gates pass.

### DIFF
--- a/src/lib/commands/sort.rs
+++ b/src/lib/commands/sort.rs
@@ -559,14 +559,27 @@ impl Sort {
     /// the `FGUMI_TMP_DIRS` fallback applied); it is baked into the `SortOptions`
     /// verbatim because `add_sort` reads `SortOptions.tmp_dirs` and never consults
     /// the environment.
+    ///
+    /// `resolved_max_temp_files` is the already-resolved `--max-temp-files`
+    /// value (see [`Self::resolved_max_temp_files`]), computed once by the
+    /// caller from a single `RLIMIT_NOFILE` snapshot and baked in here as
+    /// `MaxTempFiles::Fixed`. This avoids the chain builder re-resolving
+    /// `MaxTempFiles::Auto` itself from a second, independent snapshot that
+    /// could disagree with the one used for the startup banner.
     fn build_sort_chain_spec(
         &self,
         output: &Path,
         resolved_tmp_dirs: Vec<PathBuf>,
+        resolved_max_temp_files: usize,
         command_line: &str,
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #892, opened 2026-09-01
feat(filter): route the filter command onto the declarative chain builder

nh13:
## What

Route `fgumi filter` onto the declarative chain builder — the first of the R3 command cutovers (following the merged group/dedup/sort cutovers).

`filter --threads N` now runs through `execute_chain` → `ChainSpec::single_stage(Stage::Filter, …)` → `build_for(spec)?.run()`. The no-`--threads` path keeps its existing unified-pipeline implementation unchanged as an **in-process parity oracle**; a follow-up PR removes it once parity is trusted (mirroring the sort PR A→B split). Both paths call the same shared `process_record_raw`, so the chain path reuses the filtering logic rather than reimplementing it.

## Correctness fixes folded in

- **Query-grouped input guard on the chain path.** `add_filter` now calls `require_query_grouped` (FILT3-02), matching the legacy path. Filtering is template-based, so coordinate-sorted input scatters template mates and silently corrupts the both-primaries-pass logic. `open_source`/`add_source` do no generic sort-order check, so without this guard `--threads N` would feed coordinate-sorted input straight to `GroupByQueryname`.
- **`Total bases masked` parity.** All four chain step factories summed masked bases over every processed record, whereas the legacy path routes them through `retained_primary_masked_bases` (zero for a dropped template, excluding secondary/supplementary reads — fgbio's `FilterConsensusReads` semantics). This divergence was dormant pre-existing code that this cutover made reachable via `--threads N`. It affects **only the informational `Total bases masked:` log line** — never the output BAM, rejects BAM, or `--stats` file — which is why the byte-parity tests did not surface it. Fixed to mirror the legacy path exactly.

## Tests

- `test_filter_chain_matches_single_threaded` — rstest over `--threads {1,4}` × `{template, single-read}`, record-for-record vs the oracle, with exact kept-count and `@HD`-presence guards so a pass-through regression can't pass vacuously.
- `--rejects` and `--stats` parity cases (the stats case also asserts `failed_reads > 0`).
- Multi-batch case (n=1200) spanning `GroupByQueryname`'s 1000-record batch boundary.
- Chain-path regression tests for coordinate-sorted and header-less rejection.

## Test coverage note

The `Total bases masked` fix is a log-only metric with no output artifact, so there is no established, non-fragile way to assert it end-to-end in the in-process integration harness (it would require standing up the full command's pipeline setup to read the accumulator). Its correctness rests on the fix being textually identical to the legacy path (already covered for the legacy side) and on `retained_primary_masked_bases` being unit-tested in `fgumi-consensus`.

## Verification

`cargo ci-fmt && cargo ci-lint && RUSTDOCFLAGS="-D warnings" cargo ci-doc && cargo ci-test` — all green (9769 tests). No new `unsafe`.

Risk: filtering output and `Total bases masked` can change; legacy-path parity tests and explicit rejection tests pin the new output; unsafe changes: none, and the `CLAUDE.md` allowlist is unchanged; memory bounds, queue capacity, and thread/backpressure policy: none.

Routes `fgumi filter --threads N` through `execute_chain` and retains the no-`--threads` path as the parity oracle.

Adds queryname-grouped input validation for chain execution.

Aligns masked-base statistics with the legacy implementation for single reads and templates.

Adds integration coverage for output parity, filtering results, statistics, multi-batch grouping, coordinate-sorted input, and headerless input.

All CI checks pass across 9,769 tests.

### DIFF
--- a/src/lib/commands/filter.rs
+++ b/src/lib/commands/filter.rs
@@ -381,6 +381,10 @@ impl Command for Filter {
         // Validate parameter counts (1-3 values for duplex support)
         self.validate_parameters()?;
 
+        if self.threading.threads.is_some() {
+            return self.execute_chain(command_line);
+        }
+
         let timer = OperationTimer::new("Filtering consensus reads");
 
         info!("Starting Filter");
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #893, opened 2026-09-01
feat(correct): route the correct command onto the declarative chain builder

nh13:
## What

Route `fgumi correct` onto the declarative chain builder — R3.2, following the merged group/dedup/sort and the in-flight filter (#892) cutovers.

`correct --threads N` now runs through `execute_chain` → `ChainSpec::single_stage(Stage::Correct, …)` → `build_for(spec)?.run()`. The dispatch moves ahead of the reader open: `execute()` validates, checks output collisions, then dispatches — so the chain opens its own source without double-consuming stdin. The no-`--threads` `execute_single_thread_mode` path is kept unchanged as an **in-process parity oracle** (the legacy `execute_threads_mode` is retained, dead-coded, and removed in a follow-up — mirroring the sort PR A→B split).

## Correctness fixes the cutover exposed (folded in)

- **`--min-distance 0` on the chain path.** `CorrectOptions::validate` bailed on `min_distance_diff == 0`, but standalone `CorrectUmis::validate` (and fgbio) accept it. Relaxed to match — verified underflow-safe (the only `min_distance_diff - 1`, in `check_umi_distances`, is guarded by an `== 0` early return; the matching threshold uses `>= min_distance_diff`).
- **`@HD` synthesis in `ChainBuilder::new`.** Added `ensure_hd_record` (before `add_pg_record`) so header-less input gets `@HD VN:1.6 SO:unsorted` synthesized on the chain path like every legacy `execute()`. This is a **general** chain-builder fix; every currently-shipped chain consumer (group/dedup/sort/extract/filter) either rejects header-less input before output or rebuilds `@HD` itself, so their behavior is unchanged (full gate green). correct is the intended changed case.
- **UMI tag derived from `--target` in `add_correct`.** It hardcoded `SamTag::RX` for the observed-UMI tag regardless of target, so `--target barcode --threads N` read `RX` (absent) and **silently dropped every record**. Now derives from `target.sequence_tag()`, matching the adjacent original-tag line and the legacy path.

## Behavior note (intentional)

On empty input with a positive `--min-corrected`, the chain path now **errors** ("No reads were processed …"), where the legacy single-threaded path silently passes (its `0 / 0 = NaN` kept-ratio compares false against the floor — a latent bug). An empty run cannot meet a positive minimum, so the chain's explicit rejection is correct. Pinned by a test; the legacy engine's NaN-pass is retired with that engine in the follow-up.

## Tests

Chain-vs-oracle parity (reusing the house `read_bam_output` helper for full-header + record parity), `--rejects` record parity, `--metrics` TSV byte-compare, a multi-batch case using **paired query-name templates** (to exercise `GroupByQueryname` cross-input-batch carry-over as well as the output-emit boundary), plus regression tests for `--min-distance 0`, header-less `@HD` synthesis, and the empty-input rejection above.

## Review

Reviewed with a deep multi-agent gauntlet (no real bugs; the three fixes above came from pre-implementation Fable reviews). One deferred suggestion: the `--min-distance 0` test proves no-crash + chain/oracle parity but does not yet exercise the ambiguity-rejection contract with a multi-UMI whitelist — a follow-up test-strengthening item.

## Verification

`cargo ci-fmt && cargo ci-lint && RUSTDOCFLAGS="-D warnings" cargo ci-doc && cargo ci-test` — all green (9769 tests). No new `unsafe`.

Risk: corrected UMI output, headers, rejects, and metrics can change for `fgumi correct --threads N`, pinned by no-thread parity tests and chain integration tests; unsafe: none, and `CLAUDE.md` allowlist changes: none; memory bounds, queue capacity, and thread/backpressure policy: none.

Fix: Route threaded correction through `execute_chain` after validation and output-collision checks, while preserving the no-thread path as the parity oracle.

- Synthesize `@HD VN:1.6 SO:unsorted` for header-less input.
- Derive the observed tag from `--target`.
- Accept `--min-distance 0`.
- Reject empty input when positive `--min-corrected` is required.
- Add coverage for records, headers, rejects, metrics, multi-batch processing, and edge cases.

### DIFF
--- a/src/lib/commands/correct.rs
+++ b/src/lib/commands/correct.rs
@@ -664,6 +664,17 @@ impl Command for CorrectUmis {
         }
         reject_output_collisions(&outputs)?;
 
+        // --threads N: run the correct stage on the declarative chain builder. Dispatch
+        // here — after the reader-free pre-flight above, but BEFORE the timer / UMI-set
+        // load / distance check / reader below. execute_chain → add_correct re-emits the
+        // timer, banner, UMI-set load, distance check, and threading logs and opens its
+        // own source, so running them here too would double-log and pre-consume stdin.
+        // The no-`--threads` single-threaded path below stays the in-process parity oracle.
+        if self.threading.threads.is_some() {
+            return self.execute_chain(command_line);
+        }
+
+        // ---- legacy no-threads oracle path (everything below stays as-is) ----
         let timer = OperationTimer::new("Correcting UMIs");
 
         // Load UMI sequences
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #894, opened 2026-09-01
feat(simplex): route the simplex command onto the declarative chain builder

nh13:
## What

Route `fgumi simplex` onto the declarative chain builder — R3.3, the first of the consensus trio.

`simplex --threads N` (on a `consensus`-feature build) now runs through `execute_chain` → `ChainSpec::single_stage(Stage::Simplex, …)` → `build_for(spec)?.run()`. The no-`--threads` single-threaded fast path is kept unchanged as an **in-process parity oracle** (legacy `execute_threads_mode` retained, removed in a follow-up).

## Feature-gating (the consensus-trio concern)

`commands::simplex` is gated `simplex`, but the chain machinery (`StageOptionsBag.simplex`, `add_simplex`) is gated `consensus`, and `simplex` does **not** imply `consensus`. So `execute_chain` and its dispatch are `#[cfg(feature = "consensus")]`-gated, and on a non-`consensus` build the dispatch is absent and execution **falls through to the legacy threaded path** (no regression, no bail). Verified: `cargo check -p fgumi --no-default-features --features simplex` compiles, alongside the default and `--all-features` builds.

## Correctness fix the cutover exposed (folded in)

- **`check_consensus_sort_order` guard in `add_simplex`.** Both legacy simplex paths call it right after opening the reader (it accepts template-coordinate/query-grouped input and rejects coordinate-sorted / plain-queryname / unsorted). `add_simplex` never did, and `GroupByMi` has no out-of-order detection — so `--threads N` would have **silently mis-grouped** a badly-sorted input into wrong consensus calls instead of rejecting it. Added the same guard (the one `builder.rs` edit). (`add_duplex`/`add_codec` almost certainly share this gap — noted for their cutovers.)

## Dispatch placement

The dispatch moves ahead of the timer/reader: the reader-free pre-flight (`io.validate`, output-collision check, `validate_read_bounds`, the `--ref`/`--methylation-mode` bail) runs first on both paths, then the consensus-gated `--threads` dispatch, then the unchanged legacy tail. An invalid run no longer prints the timer/banner before erroring — intentional, matching dedup's ordering.

## Tests

Chain-vs-oracle parity across plain, `--rejects`, `--stats` (byte-compare), methylation mode, overlapping-consensus (`--consensus-call-overlapping-bases false`, the disabled path), `--allow-unmapped`, a multi-batch case, and a coordinate-sorted rejection proving the new sort-order guard.

## Review

Spec Fable-reviewed against the code (the sort-order guard came from that review). A deep gauntlet review is running and any findings will be addressed before this is presented for merge.

## Verification

`cargo ci-fmt && cargo ci-lint && RUSTDOCFLAGS="-D warnings" cargo ci-doc && cargo ci-test` — all green (9768 tests); simplex suite 111 passing; the three feature-check legs (`--no-default-features --features simplex`, default, `--all-features`) all compile. No new `unsafe`.

Risk: threaded consensus simplex output changes, pinned by single-threaded parity and sort-order rejection tests; `unsafe`: none, with no `CLAUDE.md` allowlist update; memory bounds, queue capacity, and thread/backpressure policy: none.

Fix: Route consensus-feature threaded simplex execution through the declarative chain builder. Preserve the legacy fallback and add consensus sort-order validation to `add_simplex`.

### DIFF
--- a/src/lib/commands/simplex.rs
+++ b/src/lib/commands/simplex.rs
@@ -377,10 +377,7 @@ impl SimplexOptions {
 
 impl Command for Simplex {
     fn execute(&self, command_line: &str) -> Result<()> {
-        // Start timing
-        let timer = OperationTimer::new("Calling simplex consensus");
-
-        // Validate inputs
+        // ---- reader-free pre-flight (runs on BOTH paths) ----
         self.io.validate()?;
         let mut outputs: Vec<(&Path, &str)> = vec![(self.io.output.as_path(), "--output")];
         if let Some(path) = &self.rejects_opts.rejects {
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #897, opened 2026-09-01
feat(clip): route the clip command onto the declarative chain builder

nh13:
## What

Route `fgumi clip` onto the declarative chain builder — R3.x. This is the clip half of the command-cutover campaign (filter #892, correct #893, simplex #894, codec #895, duplex #896).

`clip --threads N` now runs through `execute_chain` → `ChainSpec::single_stage(Stage::Clip, …)` → `build_for(spec)?.run()`. The no-`--threads` single-threaded engine (`execute_single_threaded`) is kept unchanged as the **in-process parity oracle**. The old threaded engine (`execute_threads_mode`) is **removed in this PR**: once `--threads` dispatches to the chain it has no call site, and it was never the parity oracle, so it and its now-orphan helpers (`CollectedClipMetrics`, `ClipProcessedBatch` and its `MemoryEstimate` impl + unit test) are deleted here rather than carried dead into a follow-up.

`clip` is a core, non-feature-gated command, so the dispatch is unconditional. clip writes no rejects (only `--output` and single-threaded-only `--metrics`), so there is no rejects-header-provenance decision. The dispatch sits after the reader-free pre-flight (output-collision check, input + reference existence, the clipping-option validation, and the `--metrics`/`--threads` fail-fast) but before the banner / timer / reader, so `add_clip` — which re-emits those and opens its own source — does not double-log or pre-consume stdin.

## Two real defects the cutover exposed (fixed here)

- **`require_query_grouped` was missing on the chain path.** Both legacy clip paths enforce fgbio's `Bams.requireQueryGrouped`, but the dormant `add_clip` did not — so routing `--threads` onto the chain would silently mis-clip coordinate-sorted input the legacy path rejected. Added the guard in `add_clip`, gated on `Stage::Clip` being the source stage so a future fused `group→clip` chain (upstream stage orders the records) is not wrongly rejected. `ChainSpec` exposes only `single_stage` today, so the guard always fires. Pinned by the already-parameterized `test_clip_rejects_coordinate_sorted_input` / `test_clip_rejects_headerless_input`, whose `--threads` case now runs the chain.

- **The chain clip step over-clipped reads with existing clipping.** The dormant `build_clip_process_step` reimplemented per-template clipping and carried an explicit `KNOWN DIVERGENCE — MUST be resolved in the clip wiring PR` comment: it selected the primary pair by positional index (ignoring secondary/supplementary reads, and skipping templates with ≥3 records) and applied fixed clipping with "clip N more" (`clip_*_end_of_alignment`) instead of the canonical "ensure at least N including existing clipping" semantics — over-clipping any read already carrying soft/hard clips. This surfaced immediately: the existing unit test `test_fixed_position_clip_counts_existing_clipping::case_2_multi_threaded` expects `5S10M5S` (existing 5′ clip counted, 5 new 3′ bases) but the chain produced `8S7M5S`. Fixed by exposing `ClipParams`/`ClipParams::from_clip`/`ClipParams::clip_template` as `pub(crate)` and delegating the chain step to `cap.params.clip_template(records, &clipper, None)` — the exact code the oracle runs (its docstring already declares it "the single shared implementation used by both threading paths"). Parity is now by construction, not a parallel implementation. This let the now-dead `update_mate_info_raw` (itself carrying a "resolve in the clip wiring PR" note) be removed; `clip_template` repairs mate info via the canonical `set_mate_info_raw`. The `--metrics` detailed collection is never produced under `--threads`, so the chain passes `None` and drives its atomic `overlap_clipped`/`extend_clipped` counters off `clip_template`'s returned per-template flags.

## Tests

- `test_clip_chain_matches_single_threaded` (`#[case]` threads 1/2/4): chain vs. single-threaded oracle, full normalized-header + record parity via `read_bam_output`, with non-vacuous guards (all 16 records survive; at least one output CIGAR ≠ 8M, so clipping actually ran).
- `test_clip_rejects_metrics_with_threads`: pins the reader-free `--metrics` + `--threads` fast-fail (`expect_err`, message contains "cannot be used with --threads"), so a regression can't silently drop the user's requested metrics file.
- The pre-existing threaded tests now route through the chain and pass: coordinate / header-less rejection (the guard), primary-pair all-options, supplementary mate repair, secondary-only, fragment, past-mate, upgrade-supplementary, `--check-crc` handling, stdin-once, and `test_clip_command_single_and_multi_threaded_outputs_match`.
- The fixed `test_fixed_position_clip_counts_existing_clipping::case_2_multi_threaded`.

## Verification

`cargo ci-fmt && cargo ci-lint && RUSTDOCFLAGS="-D warnings" cargo ci-doc && cargo ci-test` — all green (9772 tests). The three feature legs compile (`--no-default-features`, default, `--all-features`, all `--all-targets`). No new `unsafe`.

Risk: output changes for `clip` threaded execution, pinned by single-threaded parity tests and shared `ClipParams::clip_template`; unsafe changes: none; memory, queue, and backpressure policy changes: none.

Threaded `clip` execution now uses the declarative chain builder. The single-threaded path remains the parity oracle. Reader-free validation runs before dispatch and rejects invalid query grouping, output collisions, clipping options, and `--metrics` with `--threads`.

The chain path reuses `ClipParams::clip_template`. This preserves existing clipping behavior, prevents over-clipping, and applies canonical mate repair. The obsolete threaded engine and helper code were removed.

Integration tests cover thread-count parity, normalized headers, clipping behavior, validation failures, and existing threaded scenarios.

### DIFF
--- a/src/lib/commands/clip.rs
+++ b/src/lib/commands/clip.rs
@@ -6,43 +6,28 @@
 
 use crate::alignment_tags::regenerate_alignment_tags_raw;
 use crate::clipper::{ClippingMode, RawRecordClipper};
-use crate::grouper::TemplateGrouper;
 use crate::logging::OperationTimer;
 use crate::metrics::clip::{ClipCounts, ClippingMetricsCollection};
 use crate::metrics::writer::write_metrics as write_metrics_tsv;
-use crate::per_thread_accumulator::PerThreadAccumulator;
 use crate::reference::ReferenceReader;
 use crate::sam::SamTag;
-use crate::template::{
-    InsertSizeEnd, TemplateBatch, TemplateIterator, compute_insert_size_from_ends,
-};
-use crate::unified_pipeline::{
-    GroupKeyConfig, Grouper, MemoryEstimate, run_bam_pipeline_from_reader,
-};
+use crate::template::{InsertSizeEnd, TemplateIterator, compute_insert_size_from_ends};
 use crate::validation::validate_file_exists;
 use anyhow::Result;
 use clap::Parser;
 use fgumi_bam_io::ProgressTracker;
-use fgumi_bam_io::{
-    create_bam_reader_for_pipeline_with_opts, create_raw_bam_reader_with_opts,
-    create_raw_bam_writer,
-};
+use fgumi_bam_io::{create_raw_bam_reader_with_opts, create_raw_bam_writer};
 use fgumi_raw_bam::RawRecord;
 use log::info;
-use noodles::sam::Header;
-use std::io;
 use std::path::PathBuf;
-use std::sync::Arc;
-use std::sync::atomic::{AtomicU64, Ordering};
 
 use super::command::Command;
 use super::common::{
     BamIoOptions, CompressionOptions, QueueMemoryOptions, SchedulerOptions, ThreadingOptions,
-    build_pipeline_config,
 };
 
 /// Clips reads in a BAM file to remove overlaps
-#[derive(Parser, Debug)]
+#[derive(Parser, Debug, Clone)]
 #[command(
     name = "clip",
     about = "\x1b[38;5;173m[POST-CONSENSUS]\x1b[0m \x1b[36mClip overlapping reads in BAM files\x1b[0m",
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

Issue #773, opened 2026-08-15
extract silently drops templates from a mismatched-length FASTQ pair


nh13:
## Summary

`fgumi extract` silently drops templates from the surplus tail of a mismatched-length FASTQ pair. In a synthetic fixture where one input stream carries 30,000 more templates than the other, **29,800 of those 30,000 were absent from the output**, with no error, no warning, and a zero exit status.

## Reproduction

Build a paired FASTQ input in which the two streams have unequal record counts, and run `fgumi extract` over it. Count the emitted templates against the expected number.

Observed in a synthetic fixture: 29,800 of 30,000 surplus templates lost. The loss is **not** a function of memory settings — it reproduces identically with queue-memory gating disabled, so it is independent of the backpressure work in #766 / #764.

## Impact

Silent data loss. A truncated or mismatched FASTQ pair — which is exactly what a partial transfer, an interrupted download, or an upstream tool failure produces — yields a short output that reports success. Nothing in the exit status or the logs distinguishes it from a complete run.

The two defensible behaviors are to process the surplus, or to fail loudly. Dropping records while reporting success is neither.

## Notes

Found while adding queue-memory backpressure to the FASTQ pipeline (#766). The test added there uses an ungated run as its oracle specifically so that it pins budget-neutrality without encoding this defect as expected behavior — so this issue is not blocked by that work, and fixing it should not require touching the gate.

Build under test: `main` at `3d83a55d`.

nh13:
**Correction: the reported magnitude does not reproduce, and there are three distinct failure modes rather than one.**

The "29,800 of 30,000 surplus templates lost" figure above could not be reproduced in either code path, and should be disregarded. Measured against fresh fixtures (release binary, threads 1/4/8):

| input | actual behavior |
| --- | --- |
| gzip, R1 100,150 / R2 100,050 | exit 0, **100 records silently lost**, no warning |
| gzip, R1 130,000 / R2 100,000 | exit 0, 30,000 surplus emitted as **single-end fragments**, with 30,000 `WARN` lines |
| BGZF, R1 5,000 / R2 4,000 | **hard hang** — killed at 10 minutes, 0-record output, no deadlock-detector line |
| matched control | unaffected |

So the defect is three-headed:

[...]

PR #900, opened 2026-09-03
feat(extract): route the extract command onto the declarative chain builder

nh13:
Routes `fgumi extract --threads N` onto the declarative chain builder (Pattern A), the last of the R3 command cutovers. The no-`--threads` serial loop (`process_singlethreaded`) stays as the in-process parity oracle; the legacy `process_with_pipeline` typed-step pipeline is deleted.

The dispatch predicate changes from `is_parallel()` to `threads.is_some()`. For `Option<usize>` threads these are equivalent — the old `is_parallel()` was already true for every `Threads(n)`, `--threads 1` included — so this swaps the parallel *implementation* for the chain builder; it does not change which `--threads` values run in parallel.

## Full input parity

Extract is the only cutover whose source is FASTQ rather than BAM, and it supports three input shapes the legacy `--threads` path all parallelized: multiple files, `--interleaved` (one physical stream de-interleaved into R1/R2), and stdin. All three still parallelize, and the chain path is held byte-for-byte identical to the serial oracle:

- **The chain opens its own FASTQ readers and detects the quality encoding** inside `ChainBuilder::open_source`, through the same `detect_encoding_and_open_fastq_readers` helper the serial oracle uses. Both paths therefore open identical readers (de-interleaving and stdin sample-and-replay included) and apply the same detected encoding by construction — no reader is injected through the (Debug/Clone) `SourceSpec`.
- **`SourceSpec::InterleavedFastq`** is a new variant rather than an overload of `Fastqs`: interleaved input is one path carrying two read structures, which violates `Fastqs`' one-structure-per-path invariant (`PairedBams` is the precedent for a structurally distinct source). Its two de-interleaved halves run through the drift-bounded round-robin read topology (`ReadFastqInputs::new → ParseFastqChunks → ZipFastqRecords`), **not** the N==2 `PairRawFastq` path — the latter lets R1 outrun R2, which can overrun the de-interleaver's ~1M-record lockstep cap.
- **`PendingSource::Fastq`** now carries the opened readers, the detected encoding, and a `force_round_robin` flag; `add_extract` applies the detected encoding over the projected `ExtractOptions`.
- **`--check-crc`/`--no-check-crc`** are wired through `ExtractOptions`. Multi-file readers decode single-threaded (the pipeline parallelizes across streams); only a single physical stream (stdin/interleaved) takes the per-stream decode-thread budget, avoiding N×threads BGZF-decode over-subscription.

## Error-message parity

The chain's FASTQ out-of-sync errors are now directional ("R2 ended before R1"), matching the serial oracle so the operator learns which file ran short (the shared FASTQ steps `pair_fastq`, `parse_zip_fastq`, and `zip_fastq` all report the same shape). The extract step also preserves the full anyhow context chain when it converts to `io::Error`, so a "read name too long" root cause survives instead of being flattened to the top context.

## Deletions

`process_with_pipeline`, `make_raw_records_static`, `ExtractConfig`, `ExtractedBatch`, and `validate_template_record_count` are removed — the chain's `extract_batch` enforces the same one-record-per-read-structure invariant (issue #773), and the serial oracle never used that function. This drops extract's `unified_pipeline` imports to a single still-referenced symbol (`fastq_out_of_sync_error`).

## Suggested reading order

1. `src/lib/commands/extract.rs` — `to_extract_options`, `execute_chain`, the `execute()` dispatch, and the shared `detect_encoding_and_open_fastq_readers` / `open_fastq_input_readers` helpers (the deletions are the bulk of the diff here).
2. `src/lib/pipeline/chains/source_spec.rs` — the `InterleavedFastq` variant + constructor.
3. `src/lib/pipeline/chains/builder.rs` — `open_source` (`open_fastq_source` helper), `PendingSource::Fastq`, the `add_source` topology routing, and `add_extract`'s encoding override.
4. The `steps/source/*` error-message changes and the tests.

## Testing

Chain-vs-serial-oracle parity tests with fixed-value anchors — one for the RX/RG path and one covering the CB/CY, BC/QT, QX, `--single-tag`, and `--annotate-read-names` branches — plus `--threads 1` coverage and directional out-of-sync assertions on the shared steps. The existing multithreaded-equivalence, interleaved-equivalence, mismatched-pair, and read-name-limit tests now exercise the chain path. Full local gate green (fmt, clippy `-D warnings`, doc, `--no-default-features`/`--all-features`, and the complete test suite).

## Deferred follow-ups

- The legacy FASTQ pipeline driver (`run_fastq_pipeline` / `FastqPipelineConfig` and `test_fastq_pipeline_memory_backpressure.rs`) is now dead but lives in `src/lib/unified_pipeline/`, whose removal is a separate planned phase — left for that cleanup rather than a piecemeal deletion here.
- Per-file BGZF decode parallelism vs the legacy all-BGZF pipeline-internal decode is a perf question to benchmark, not a correctness one; this PR is gated on byte-parity like the prior sibling cutovers.
- Relocating `QualityEncoding` + the FASTQ-open helpers out of `commands::extract` into a shared source module would tidy the pipeline→commands dependency, but follows the pre-existing pattern and is better done as its own sweep.

Risk: output changes are intended to remain pinned by serial parity tests; unsafe changes: none; memory, queue, and thread/backpressure policy changes: interleaved inputs force round-robin processing.

Fix: route explicit `--threads` extraction through the declarative chain while retaining serial processing as the parity oracle.

- Add `SourceSpec::InterleavedFastq` with validation and runall support.
- Share FASTQ reader setup, quality detection, stdin replay, compression, and CRC handling.
- Preserve complete extraction error context.
- Improve directional R1/R2 and stream EOF diagnostics.
- Remove the legacy pipeline extraction implementation and unused types.
- Add integration coverage for threaded output parity, interleaved and mismatched inputs, read-name limits, CRC behavior, quality detection, and extraction option branches.

### DIFF
--- a/src/lib/commands/extract.rs
+++ b/src/lib/commands/extract.rs
@@ -26,18 +26,15 @@ use crate::fastq::FastqSet;
 use crate::fastq::ReadSetIterator;
 use crate::fastq_deinterleave::deinterleave;
 use crate::fastq_parse::strip_read_suffix;
-use crate::grouper::FastqTemplate;
 use crate::logging::OperationTimer;
 use crate::sam::SamTag;
-use crate::unified_pipeline::{
-    FastqPipelineConfig, MemoryEstimate, fastq_out_of_sync_error, run_fastq_pipeline,
-};
+use crate::unified_pipeline::fastq_out_of_sync_error;
 use crate::validation::validate_input_exists;
 use anyhow::{Context, Result, bail, ensure};
 use bstr::{BString, ByteSlice};
 use clap::Parser;
 use fgumi_bam_io::ProgressTracker;
-use fgumi_bam_io::{RawBamWriter, create_raw_bam_writer, open_output_writer};
+use fgumi_bam_io::{RawBamWriter, create_raw_bam_writer};
 use fgumi_raw_bam::UnmappedSamBuilder;
 use fgumi_raw_bam::fields::flags;
 use log::{debug, info};
@@ -66,28 +63,6 @@ use std::str::FromStr;
 
 const BUFFER_SIZE: usize = 1024 * 1024;
 const QUALITY_DETECTION_SAMPLE_SIZE: usize = 400;
-
-/// Output buffer for the parallel (`--threads N`) pipeline, which writes whole
-/// compressed BGZF blocks. Sized well above `BGZF_MAX_BLOCK_SIZE` so a block
-/// lands in the buffer instead of passing straight through it.
-const PIPELINE_OUTPUT_BUF_CAPACITY: usize = 256 * 1024;
-
-/// Pre-serialized batch of BAM records for the pipeline output type.
-///
-/// Contains raw BAM record bytes with `block_size` prefixes, ready for
-/// the compress step. Replaces `Vec<RecordBuf>` as the pipeline `P` type.
-struct ExtractedBatch {
-    /// BAM records with `block_size` prefixes, ready for compress step.
-    data: Vec<u8>,
-    /// Number of records in this batch.
-    num_records: u64,
-}
-
-impl MemoryEstimate for ExtractedBatch {
-    fn estimate_heap_size(&self) -> usize {
-        self.data.capacity()
-    }
-}
 
 /// Compression format detected from file header
 #[derive(Debug, Clone, Copy, PartialEq)]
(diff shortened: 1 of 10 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #908, opened 2026-09-04
test(dedup): fix inconsistent mate-strand flags in duplicate-group fixtures

nh13:
## Summary
The dedup test fixtures built paired reads with an internally inconsistent mate-strand flag, which made each read pair count as *two* single-mate templates instead of one. This fixes the fixtures, re-baselines the seven tests calibrated against that behavior, and corrects the metric-model docs the re-baseline exposed as stale. **No production behavior change.**

## Root cause
`create_duplicate_group_inner` built R1 with `PAIRED | FIRST_SEGMENT` but omitted `MATE_REVERSE`, while its mate R2 carried `REVERSE`. A real aligner always sets `MATE_REVERSE` on R1 when R2 is reverse; without it, R1 mis-computes its mate's unclipped 5' position, so R1 and R2 resolve to different template keys, land in different position groups, and each pair is counted as two single-mate templates.

## Fix
- Add `MATE_REVERSE` to R1 (flags 65 → 97) so the fixtures are well-formed.
- Re-baseline the 7 dedup tests whose expected values followed from the pair-split behavior. The corrected model — verified against ground-truth dedup output — is **template = read pair** (one template, two reads): the read-unit reconciliation invariants (`2*mapped_pairs + …`) are re-anchored from `total_templates` to `total_reads`, template-unit invariants added against `total_templates`, and per-library / duplication-ladder / index-threshold / pair-orphan expectations re-derived from the fixture (e.g. three identical mapped pairs → 3 templates, 1 unique + 2 duplicate, 6 reads).
- Correct the now-stale metric-model docs (`DeduplicationMetrics` in `fgumi-metrics`; `mapped_pair_reads` / `count_template_pair_orphan` comments in `dedup.rs`) that described the "each mate in its own position group / pair splits into two templates" model and asserted `2*mapped_pairs + … == total_templates` (really the read-unit identity `== total_reads`).

`fgumi dedup` already groups well-formed pairs correctly (matching fgbio and Picard `READ_PAIRS_EXAMINED`); only the fixtures were unrealistic and some doc comments described the wrong model.

## Testing
Full suite (4473 tests) green. coderabbitai-review and gauntlet both run; all findings fixed (a stale docstring, and the metric-model doc drift above).

Risk: deduplication metric output changes, pinned by corrected paired-read fixtures and updated template/read-unit assertions; unsafe changes none; memory-bound, queue-capacity, and thread/backpressure changes none.

Fix: Correct R1’s `MATE_REVERSE` flag so paired reads resolve to one template.

- Re-baseline seven deduplication tests.
- Update pair, orphan, per-library, duplication-ladder, unmapped-pair, and index-threshold expectations.
- Clarify read-unit and template-unit metric documentation.
- No production behavior changes.

### DIFF
--- a/crates/fgumi-metrics/src/dedup.rs
+++ b/crates/fgumi-metrics/src/dedup.rs
@@ -30,24 +30,28 @@ use crate::library_size::estimate_library_size;
 ///    dropped at write time, after they have been counted here.
 /// 3. **Diagnostics** — `secondary_reads`, `supplementary_reads`,
 ///    `missing_tc_tag`.
-/// 4. **Pair/orphan breakdown** — the Picard `DuplicationMetrics` split. Because
-///    `dedup` dedups each mate in its own position group, reads are classified by
-///    their SAM flags: `mapped_pairs`/`duplicate_pairs` count mapped read *pairs*
+/// 4. **Pair/orphan breakdown** — the Picard `DuplicationMetrics` split. A read
+///    pair is one *template* but two *reads*; `dedup` classifies each primary read
+///    by its SAM flags: `mapped_pairs`/`duplicate_pairs` count mapped read *pairs*
 ///    (both mates mapped; Picard `READ_PAIRS_EXAMINED`/`READ_PAIR_DUPLICATES`,
-///    already halved from the two mates to whole pairs), `mapped_orphans`/
+///    halved from the two mates to whole pairs), `mapped_orphans`/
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #901, opened 2026-09-03
bug(group): Supplementary/secondary alignments change the reported molecule count


msto:
## Summary

In the recommended `fgumi sort` → `group` workflow, the presence of supplementary/secondary alignments changes the number of molecules `fgumi group` reports, even though `group` filters those records by default.  

## Expected behavior

Records that are filtered from grouping should have no effect on the result.

## Actual behavior

Alignments are spuriously split across molecules, inflating the reported counts in comparison to fgbio. 

## Example

| input (recommended `fgumi sort` order) | molecules |
|---|---|
| `reads.bam` as-is (supplementary present) | 48 |
| supplementary/secondary removed (`-F 0x900`) | **43** (matches `fgbio GroupReadsByUmi`) |

## Reproducing

Observed by diffing `fgumi group --strategy adjacency --edits 1` on the same input BAM, with and without supplementary/secondary alignments removed

The input is client data and can't be posted here; a minimal reprex will be shared privately.

## Detail

With supplementary alignments removed, every key maps to exactly one molecule. With supplementary present, three keys are split across multiple `MI`:

```
key <contig> 5'ends=(p,   q) umi=<UMI>  -> 2 molecules
key <contig> 5'ends=(p+1, q) umi=<UMI>  -> 3 molecules
key <contig> 5'ends=(p+2, q) umi=<UMI>  -> 3 molecules
```

These reads are the same molecule by fragment ends + UMI, yet land in different families.

## Notes / ruled out

- **Order-invariance:** with supplementary removed, `group` is order-invariant — 43 whether the
  input is `fgumi sort`- or `samtools sort --template-coordinate`-ordered. The molecule count 
  becomes order-dependent when supplementary records are present (e.g. re-sorting the same records
  with samtools gives 55, not 48 or 43), suggesting the count depends on the *interleaving* of the filtered
  supplementary records. The supplementary records themselves are byte-identical.
- **Not a tag value:** independent of `ms` (differs on every record; stripping it changes nothing)
  and `TLEN` (differs by 2 on chimeric reads, but zeroing it before *and* after sort
  changes nothing).
- The migration guide advertises *"fgumi group and fgbio's GroupReadsByUmi produce identical UMI
  groupings."* That holds for **identical input** (feeding `group` fgbio's own BAM reproduces
  fgbio's histogram bit-for-bit), but the recommended fgumi workflow diverges by these +8 molecules.

## Question

[...]

PR #909, opened 2026-09-04
feat(group,dedup): add opt-in --verify for strict template-coordinate sort-order

nh13:
## Summary

Adds an opt-in `--verify` flag to `fgumi group` and `fgumi dedup` that checks the input is in strict `fgumi sort --order template-coordinate` order and **fails fast** (non-zero exit) on the first out-of-order record, before writing a wrong result. Off by default.

`fgumi group`/`dedup` already assume template-coordinate input but only validate the header's sort tags (`SO`/`GO`/`SS`), not the actual record order — so a mislabeled or mangled file is silently mis-grouped. `--verify` closes that gap for callers who want a hard guarantee.

## Design

- **One insertion point.** The check runs inline, single-pass, at the top of `RecordPositionGrouper::process_record` — the serial chokepoint shared by all four paths (`group`/`dedup` × non-chain / `--threads N` chain). It runs *before* the secondary/supplementary skip, so the whole file's canonical order is checked, including sec/supp reads that grouping later drops.
- **Reuses the sort's own key.** It calls `fgumi_sort::extract_template_key_inline` and compares with `TemplateKey::core_cmp` (the same tolerant comparison `fgumi sort --verify` uses, which ignores the `name_hash` tie-break so both fgumi- and samtools-sorted inputs pass). It keys on the `CB` cell tag — matching what `fgumi sort --order template-coordinate` emits by default (`parse_cell_tag`), using the same fixed-seed hasher — so a correctly-sorted single-cell/multi-library input is not spuriously rejected.
- **Strict by design → opt-in.** The check is stricter than grouping/dedup strictly require (they only need a template's mates adjacent by position). That strictness is why it is opt-in, and the help text says so. For a standalone check-only pass, `fgumi sort --verify` already exists; here `--verify` is a precondition **gate** that composes with `--output`.

## Testing

- **Grouper unit tests:** in-order accept, out-of-order reject (including a mid-run regression), equal-position ties, cross-reference order, and out-of-order **secondary** reads rejected in both grouper modes (pinning that the check runs before the sec/supp skip).
- **Property tests (`proptest`, 256 cases each):** `--verify` accepts a random `(tid,pos)` stream **iff** it is non-decreasing under the template key, and any key-sorted stream is always accepted — fuzzing the fold over arbitrary orders, shuffles, and ties.
- **Integration tests (both commands, single-threaded and `--threads` chain):** `--verify` accepts a correctly-sorted input and leaves output identical to the non-verify run; rejects an out-of-order input with a clear error while the same input is accepted without `--verify`.
- **Multi-batch (600-position) chain acceptance** for both commands — guards cross-worker reorder across pipeline batch boundaries.
- **CB / multi-library regression:** an input `fgumi sort` orders by cell-barcode hash (overriding library order) is accepted by `--verify`; a CB-blind check would spuriously reject it.
- **Coordinate-mislabel:** a coordinate-ordered (mate-interleaved) file under a template-coordinate header is rejected.

## Notes

- Independent of #903 (the #901 secondary/supplementary molecule-count fix) and #908 (dedup fixture mate-flag corrections); this branch is off `main`.

Risk: output changes only when `--verify` rejects out-of-order input; `unsafe` changes are none, and no `CLAUDE.md` allowlist update applies; memory bounds, queue capacity, and thread/backpressure policy changes are none.

Fix: Use `--verify` with `fgumi group` or `fgumi dedup` to fail fast on input that is not in strict template-coordinate order. Verification is disabled by default.

- Reuses the `fgumi sort --order template-coordinate` key and comparison logic.
- Checks records before secondary and supplementary reads are skipped.
- Applies to single-threaded, threaded-chain, and multi-batch processing.
- Uses the `CB` tag and fixed-seed hash for ordering.
- Covers sorted input, out-of-order input, ties, references, secondary reads, multi-library data, threaded execution, and misleading headers.

### DIFF
--- a/src/lib/commands/dedup.rs
+++ b/src/lib/commands/dedup.rs
@@ -1246,6 +1246,20 @@ pub struct MarkDuplicates {
     #[arg(short = 'n', long = "include-non-pf-reads", value_name = "true|false", default_value = "false", num_args = 0..=1, default_missing_value = "true", action = clap::ArgAction::Set, value_parser = clap::builder::BoolishValueParser::new(), hide_possible_values = true)]
     pub include_non_pf_reads: bool,
 
+    /// Verify the input is in strict template-coordinate sort order.
+    ///
+    /// Off by default. Each record is checked against the exact ordering of
+    /// `fgumi sort --order template-coordinate` as it is read; the run aborts with
+    /// a non-zero exit on the first out-of-order record. This composes with
+    /// `--output` (a precondition gate, not a check-only mode): output is written
+    /// normally when the input is correctly ordered, and may be left partial if
+    /// the check fails mid-stream. The check is intentionally STRICTER than
+    /// deduplication requires (dedup only needs a template's mates adjacent by
+    /// position), which is why it is opt-in; for a standalone check with no output
+    /// use `fgumi sort --verify`.
+    #[arg(long = "verify", value_name = "true|false", default_value = "false", num_args = 0..=1, default_missing_value = "true", action = clap::ArgAction::Set, value_parser = clap::builder::BoolishValueParser::new(), hide_possible_values = true)]
+    pub verify: bool,
+
     /// Emit templates with no mapped read (both mates unmapped) untouched instead of
     /// discarding them. Such templates carry no alignment coordinate and no duplicate
     /// signal; passing them through preserves a complete record set (like Picard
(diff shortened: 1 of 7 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #912, opened 2026-09-04
fix(correct): log the missing/wrong-length-UMI banner at error level on the chain path

nh13:
## What

`fgumi correct`'s "missing / wrong-length UMI" summary banner was logged at `error!` on the legacy single-threaded path but at `warn!` on the chain (`--threads`) path. This restores it to `error!` on the chain path so both paths agree, matching fgbio (`CorrectUmis.scala:275-280`, which logs this summary at error level).

Impact before this change: `correct --threads` users saw the banner at WARN while default (no-`--threads`) users saw ERROR — a diagnostic divergence that could let the banner slip past error-level log filters. Output/data are unaffected; this is diagnostic-level parity only.

## Change

- `src/lib/pipeline/chains/commands/correct.rs`: the five banner lines in `CorrectFinalizeHook::finalize` are now `error!` (were `warn!`), with a comment citing the fgbio parity source. The `if total_missing > 0 || total_wrong_length > 0` guard is unchanged — the banner is still only emitted when there is something to report.
- `src/lib/commands/common.rs`: extended the shared `test_log_capture` harness with `captured_with_level()` (returns `Vec<(log::Level, String)>`) so tests can assert log level, not just message text. The existing message-only `captured()` is retained unchanged for the pre-existing sort test.

## Tests

New `#[rstest]` `missing_or_wrong_length_umi_banner_level_and_presence` with two cases:
- positive (missing=2, wrong=1): asserts every banner line is at `Level::Error`;
- negative (missing=0, wrong=0): asserts no banner line is emitted at all.

The negative case pins the guard — an implementation that dropped it and emitted the banner unconditionally would fail. Full gate green: `cargo ci-test` 9955 passed / 31 skipped, plus `ci-fmt`/`ci-lint`/`ci-doc` and `--no-default-features`/`--all-features` checks.

## Context

Part of restoring chain-path diagnostic parity ahead of retiring the legacy single-threaded paths. A related pre-existing bug surfaced during review — the two legacy `correct` paths bypass `--min-corrected` on empty input (`NaN < min == false`) while the chain path and fgbio both error — is tracked as a separate follow-up, not addressed here.

Risk: command output changes none; `unsafe` changes none and no `CLAUDE.md` allowlist update applies; memory bounds, queue capacity, and thread/backpressure policy changes none.

The chain path now logs the missing or wrong-length UMI summary at `Level::Error`. The banner remains conditional on nonzero counters.

The test harness now captures log levels through `captured_with_level()`. Tests verify error-level banners and suppression when both counters are zero.

### DIFF
--- a/src/lib/commands/common.rs
+++ b/src/lib/commands/common.rs
@@ -1929,7 +1929,7 @@ pub(crate) fn require_group_input_ordering(
 pub(crate) mod test_log_capture {
     use std::sync::{Mutex, MutexGuard, Once};
 
-    static CAPTURED_LOGS: Mutex<Vec<String>> = Mutex::new(Vec::new());
+    static CAPTURED_LOGS: Mutex<Vec<(log::Level, String)>> = Mutex::new(Vec::new());
 
     /// Serializes whole capture sessions. `CAPTURED_LOGS` is process-global, so
     /// under plain `cargo test` (one process, tests running concurrently) two
@@ -1955,7 +1955,7 @@ pub(crate) mod test_log_capture {
             CAPTURED_LOGS
                 .lock()
                 .expect("captured-log lock poisoned")
-                .push(record.args().to_string());
+                .push((record.level(), record.args().to_string()));
         }
         fn flush(&self) {}
     }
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #913, opened 2026-09-04
feat(sort): thread --sort-stats diagnostics through the chain

nh13:
## What

`fgumi sort` runs entirely on the declarative chain builder, but `--sort-stats` was inert on that path — it emitted `warn!("--sort-stats is ignored by the sort chain (not yet threaded)")` and produced no diagnostics. This threads the flag through so it actually works.

## Change

- `SortOptions` gains a `sort_stats: bool` field, projected from the CLI flag by `to_sort_options()`, and wired into `SortMerge::with_sort_stats(bool)` in **both** the terminal and intermediate `add_sort` chain branches (the intermediate branch matters for a fused `runall` sort).
- The `SortMerge` merge-loop diagnostic (`Sort merge diag: stalls=… contention=… output_full=…`) is now **gated behind `--sort-stats`**. Previously it was logged **unconditionally on every sort** — the opposite bug — so this also makes an always-on diagnostic opt-in, matching the `--sort-stats` policy established in #826.
- The single-source in-memory fast path (`emit_fast_batches`) now emits a short `Sort fast-path diag:` note under `--sort-stats`, so the flag is never a silent no-op even when the sort doesn't spill (no k-way merge to report).
- The stale `warn!` shim is deleted.

### Behavior change to call out

- `--sort-stats` now produces output (was inert).
- The `Sort merge diag:` line no longer prints on every sort — it now requires `--sort-stats`. A run that previously showed it unconditionally will need the flag.

## Scope note

This threads and gates the **existing** narrow `SortMerge` diagnostic; it does **not** restore the pre-cutover owned engine's richer (~100-line) per-phase diagnostics — that would be new instrumentation for a separate PR. The `--sort-stats` help text and the `docs/src/guide/performance-tuning.md` "Sort Statistics" section are rewritten to describe the current behavior accurately (one merge-diag line on spilling sorts; one fast-path note on in-memory sorts) and drop the now-false "own engine" claim.

## Tests

- New `tests/integration/test_sort_stats_diagnostics.rs`: a spilling case asserts the merge diag is absent without `--sort-stats` and present with it (guarded by an `assert_really_spilled` check so it can't pass vacuously if memory accounting changes), plus a small-input fast-path case pinning that the fast-path note appears (and the merge diag does not).
- Updated the pre-existing `test_sort_cutover_parity.rs` assertion that had pinned the old always-on/ignored behavior, and its stale comments.

Full gate green: `cargo ci-test` 9956 passed / 31 skipped, plus `ci-fmt`/`ci-lint`/`ci-doc` and `--no-default-features`/`--all-features`.

Risk: sort diagnostics change command output, pinned by integration tests; unsafe changes: none, and CLAUDE.md allowlist updates: none; memory, queue, and thread/backpressure policy changes: none.

- Propagates `--sort-stats` through terminal and intermediate sort chains.
- Gates merge diagnostics behind the flag.
- Adds a diagnostic for single-source in-memory sorts.
- Removes the stale ignored-option warning.
- Updates documentation and integration tests for spilling and fast-path cases.

### DIFF
--- a/crates/fgumi-pipeline-io/src/sort/merge.rs
+++ b/crates/fgumi-pipeline-io/src/sort/merge.rs
@@ -596,6 +596,11 @@ pub struct SortMerge<O: MergeOutput = RecordBatchOutput> {
     /// reaches `Done`. The standalone-sort summary finalize hook reads it to
     /// log records processed/written and the spill-chunk count.
     stats_slot: Option<Arc<parking_lot::Mutex<Option<fgumi_sort::SortStats>>>>,
+    /// Whether to log the `--sort-stats` merge-loop performance diagnostic
+    /// (`Sort merge diag: ...`: stalls/contention/backpressure counters). Off
+    /// by default -- it is instrumentation for performance investigations, not
+    /// something a normal run should show. See [`Self::with_sort_stats`].
+    sort_stats: bool,
     /// Total records ingested, captured at the merge transition (the summary's
     /// "records processed").
     processed: u64,
@@ -655,6 +660,7 @@ impl<O: MergeOutput> SortMerge<O> {
             target_batch_count: target_batch_count.max(1),
             output_byte_limit,
             stats_slot: None,
+            sort_stats: false,
             processed: 0,
             chunk_count: 0,
             dbg: MergeDiag::default(),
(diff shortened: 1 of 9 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #914, opened 2026-09-04
fix: restore chain-path diagnostic parity (dedup/group index-threshold banner, copy_umi serial inert-flag warnings, pair_fastq docstring)

nh13:
## What

Restores three diagnostics the chain path dropped or got wrong versus the legacy path. All changes are log/doc-only — no record output changes. This is groundwork for retiring the legacy single-threaded paths (once removed, the chain path is the only path, so its diagnostics must match first).

## Changes

- **dedup + group index-threshold banner** (`src/lib/pipeline/chains/builder.rs`): both chain banners logged a flat `Index threshold: {value}`, which is wrong for `--strategy edit` (logs a meaningless value) and misleading for `adjacency --edits 2`. They now call the shared strategy/edits-aware `common::log_index_threshold(...)` the legacy paths use, so all four combinations agree. Both dedup and group were fixed (group's legacy path uses the same helper — sibling parity).
- **copy_umi serial-path inert-flag warning** (`src/lib/commands/copy_umi.rs`): the no-`--threads` serial path silently ignored `--scheduler`, `--deadlock-recover`, and the `--max-memory`/`--memory-reserve`/`--memory-per-thread` queue-memory flags. It now warns for each (when set to a non-default) with wording accurate to the serial context ("has no effect on the single-threaded path; run with --threads to use the pipeline"). This is a copy_umi-local helper — the shared `common::warn_unwired_pipeline_flags` (which describes the chain engine and is correct for its chain callers) is left untouched.
- **pair_fastq docstring** (`src/lib/pipeline/steps/source/pair_fastq.rs`): the `out_of_sync_error` docstring falsely claimed all detection paths "surface byte-for-byte identical diagnostics." Reworded to state accurately that the same out-of-sync condition is detected on every path but the wording differs; the byte-identical claim is scoped to this file's two call sites. Doc-only.

## Tests

Subprocess-level (spawn the real `fgumi` binary, assert on captured stderr — the in-process log-capture harness isn't reachable from the integration-test crate), via a new shared `helpers/cli.rs::run_and_capture_logs` (dedups three near-identical spawn helpers):
- dedup + group index-threshold banner: strategy/edits-aware cases including the default `adjacency --edits 1` verbatim-value branch. The negative guards assert absence of the exact flat-banner string the pre-fix code would have emitted for each case's config (not a vacuous never-matching literal), so a regression to the flat banner fails the test.
- copy_umi inert-flag warnings on the serial path: `--deadlock-recover` (serial + chain, distinct wording each), `--scheduler`, and `--max-memory`.

Full gate green: `cargo ci-test` 9965 passed / 31 skipped, plus `ci-fmt`/`ci-lint`/`ci-doc` and `--no-default-features`/`--all-features`.

Risk: grouping, consensus, sort order, corrected UMIs, and metrics output are unchanged; diagnostic changes are pinned by shared strategy-aware logging and integration tests; unsafe changes are none and CLAUDE.md updates are not applicable; memory, queue capacity, and thread/backpressure policy changes are none.

Fix: Restore diagnostic parity across chain and legacy paths.

- Add serial `copy_umi` warnings for inert scheduler, deadlock-recovery, and queue-memory options.
- Use shared index-threshold logging for deduplication and grouping.
- Correct the `pair_fastq` `out_of_sync_error` documentation.
- Add subprocess integration tests and shared CLI log-capture helpers.

### DIFF
--- a/src/lib/commands/copy_umi.rs
+++ b/src/lib/commands/copy_umi.rs
@@ -424,6 +424,17 @@ impl Command for CopyUmi {
         info!("Input: {}", self.io.input.display());
         info!("Output: {}", self.io.output.display());
 
+        // Without this, a scheduler/pipeline flag set here (e.g.
+        // `--deadlock-recover`) is silently ignored on the serial path with no
+        // diagnostic at all. This is NOT `common::warn_unwired_pipeline_flags`:
+        // that helper is chain-specific (its wording says "the chain engine ...",
+        // and it deliberately skips `--max-memory`/`--memory-reserve`/
+        // `--memory-per-thread` because the chain DOES honor them) — on this
+        // serial path there is no chain engine at all, so those queue-memory
+        // flags are inert here too and must be warned, with wording that is true
+        // for the single-threaded context.
+        self.warn_unwired_flags_on_serial_path();
+
         let totals = self.run_single_threaded(command_line)?;
         warn_and_log_copy_umi_summary(&totals);
         if let Some(path) = &self.metrics {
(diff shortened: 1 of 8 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #915, opened 2026-09-04
feat(clip): produce --metrics on the multi-threaded chain path

nh13:
## What

`fgumi clip --metrics` was a hard error when combined with `--threads` — the detailed clipping-metrics TSV was only produced by the single-threaded path. This produces the same metrics on the multi-threaded chain path, so `--metrics` and `--threads` now work together.

## Design

Follows the established retag/correct chain-metrics idioms:

- Each chain worker accumulates into a `PerThreadAccumulator<ClippingMetricsCollection>`, passed into the shared `ClipParams::clip_template` when `--metrics` is set (the `None` fast path is unchanged — no accumulator allocated).
- A **success-only** `ClipMetricsFinalizeHook` (registered on `finalize_on_success`, mirroring retag) sums every worker slot and writes the TSV, so a failed or partial run publishes no stale metrics.
- `ClippingMetricsCollection::merge` sums the raw `fragment`/`read_one`/`read_two` counters (all 15 `ClippingMetrics` fields, via `AddAssign`); `pair`/`all` are derived once in `finalize()` on the reduced total. Because the reduction is pure commutative/associative integer addition, the chain TSV is **byte-identical** to the single-threaded oracle.
- The finalize→derive→write sequence is a single shared `ClippingMetricsCollection::finalize_and_write` helper called by both the single-threaded path and the chain hook, so they can't drift.
- Both hard bails (`builder.rs`, `commands/clip.rs`) and the "single-threaded only" doc text are removed.

The single-threaded path's behavior and output are unchanged.

## Tests

- `test_clip_metrics_match_across_threading_modes` (rstest, threads 1/2/4): asserts a full-string TSV `assert_eq!` between each threaded run and the single-threaded oracle, over a fixture with fragments, plain overlapping pairs, and pairs with pre-existing soft/hard clips. The fixture is sized (8000 templates → 16 `GroupBam` batches vs 4 worker threads) so the threaded cases genuinely shard across multiple accumulator slots.
- `parallel_threads_populate_and_merge_distinct_clip_metrics_slots` (unit, `per_thread_accumulator.rs`): spawns 8 OS threads, asserts `populated_slots > 1` via `slots()`, and that the merged reduction recovers the exact combined total — deterministically closing the "merged only a subset of slots" gap the integration test can't observe directly.
- `test_clip_metrics_not_written_on_chain_failure`: forces a genuine mid-run `Pipeline::run` failure and asserts the `--metrics` file was never written (pins the success-only contract).

Full gate green: `cargo ci-test` 9958 passed / 31 skipped, plus `ci-fmt`/`ci-lint`/`ci-doc` and `--no-default-features`/`--all-features`.

Risk: `fgumi clip --metrics` changes TSV output, pinned by byte-identical single-threaded results and success-only finalization; `unsafe` changes are none, so the `CLAUDE.md` allowlist needs no update; memory bounds, queue capacity, and thread/backpressure policy changes are none.

- Adds threaded clipping metrics support.
- Merges per-thread metrics and writes them only after success.
- Reuses finalization logic across execution modes.
- Tests cross-thread merging, output equivalence, and failure handling.

### DIFF
--- a/crates/fgumi-metrics/src/clip.rs
+++ b/crates/fgumi-metrics/src/clip.rs
@@ -307,6 +307,23 @@ impl ClippingMetricsCollection {
         }
     }
 
+    /// Merges another collection's raw `fragment`/`read_one`/`read_two` counters
+    /// into `self`.
+    ///
+    /// Deliberately does NOT touch `pair`/`all` — those are derived aggregates
+    /// that [`Self::finalize`] computes from the raw counters, so callers that
+    /// accumulate per-thread collections (e.g. the chain `--threads` clip path)
+    /// must merge every collection first and call `finalize` exactly once on the
+    /// merged total. That mirrors the single-threaded oracle's call sequence
+    /// (accumulate through the whole run, then finalize once at the end), so the
+    /// reported `pair`/`all` totals are identical regardless of how work was
+    /// sharded across threads.
+    pub fn merge(&mut self, other: &ClippingMetricsCollection) {
+        self.fragment.add(&other.fragment);
+        self.read_one.add(&other.read_one);
+        self.read_two.add(&other.read_two);
+    }
+
     /// Finalizes metrics by aggregating pair and all categories
     pub fn finalize(&mut self) {
         // Aggregate ReadOne and ReadTwo into Pair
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #918, opened 2026-09-04
fix(dedup): consult the UMI-position cache by assigning UMI groups before clearing duplicate flags

nh13:
## What

`process_position_group` ran the duplicate-flag-clearing preamble — which calls `Template::records_mut()`, unconditionally invalidating `cached_umi_position` — **before** `assign_umi_groups` read the cache via `Template::cached_umi()`. So the cache was always empty by the time it was consulted, and the `Stage::Group | Stage::Dedup` UMI-position-cache gate was **inert on the dedup path** (chain and legacy): dead infrastructure that never ran as designed.

Reorder `assign_umi_groups` to run first, then clear duplicate flags. The two steps touch disjoint state — `assign_umi_groups` only reads `cached_umi()`/`r1()`/`r2()` and writes `Template::mi`, while the flag-clear loop only writes the BAM FLAG field via `fgumi_raw_bam::set_flags` — so they commute and the output is **byte-identical** to the previous ordering. Fixes both the chain and legacy dedup paths, which both funnel through `process_position_group`.

## This is a correctness fix, not a perf optimization

Benchmarked on a Mac Studio (M3 Ultra, arm64, 8 threads, CPU-seconds via `/usr/bin/time -l`, 3 reps, byte-identical output): **within noise of baseline** (18.0M mapped reads, adjacency, umis/pos=4 — baseline 37.23 vs 37.17 CPU-s). It is deliberately framed as a fix rather than a `perf` change:

For dedup the position-group key is coordinate-based (the UMI is used only *within* a group for adjacency clustering), so the Decode step has no other reason to read the UMI. Consulting the cache therefore **relocates** the UMI `find_string_tag` scan into Decode rather than **eliminating** it — total work is unchanged. The value is that existing cache infrastructure now runs as designed instead of being silently inert, and the ordering inconsistency (a gate that did nothing) is removed.

## Tests

A cache-poison discriminating unit test (RX tag preceded by a decoy `ZZ:Z` tag at a varied aux offset) plus an end-to-end parity fixture, pinning byte-identical dedup output regardless of whether the cache is consulted or the aux-rescan fallback runs. Full gate green.

Risk: output changes only for affected deduplication grouping; cache-ordering tests and end-to-end parity tests pin the result. `unsafe` changes none, and the CLAUDE.md allowlist changes none. Memory bounds, queue capacity, and thread/backpressure policy changes none.

Reorders UMI assignment before duplicate-flag clearing so cached UMI positions remain available. Enables UMI-position caching for `Group` and `Dedup`, except with `--no-umi`. Adds unit and integration coverage for cache use, auxiliary-tag offsets, UMI partitioning, and output parity.

### DIFF
--- a/src/lib/commands/dedup.rs
+++ b/src/lib/commands/dedup.rs
@@ -957,19 +957,31 @@ pub(crate) fn process_position_group(
         });
     }
 
-    // Clear existing duplicate flags before re-marking
-    let mut templates: Vec<Template> = filtered_templates
-        .into_iter()
-        .map(|mut t| {
-            for raw in t.records_mut().iter_mut() {
-                let flg = RawRecordView::new(raw.as_ref()).flags();
-                fgumi_raw_bam::set_flags(raw, flg & !DUPLICATE_FLAG);
-            }
-            t
-        })
-        .collect();
+    // Assign UMI groups before clearing duplicate flags, so this read runs while
+    // `cached_umi_position` (recorded during the parallel Decode step; see
+    // `GroupKeyConfig::with_umi_tag` and `Template::from_decoded_records`) is
+    // still valid. `assign_umi_groups` only reads record state (`cached_umi()`,
+    // and R1/R2 orientation for `get_pair_orientation`) and writes
+    // `Template::mi`; it never calls `records_mut()` or touches the BAM bytes.
+    let mut templates: Vec<Template> = filtered_templates;
     if let Err(e) = assign_umi_groups(&mut templates, assigner, raw_tag, min_umi_length, no_umi) {
         return Err(io::Error::new(io::ErrorKind::InvalidData, e));
+    }
+
+    // Clear existing duplicate flags before re-marking. This runs *after* UMI
+    // assignment above: nothing reads `cached_umi()` again below, so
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed both outside-diff-range findings from the latest review:

- `tests/integration/test_dedup_command.rs`: removed the now-redundant `create_consistent_pair_group` helper (its R1 sets `MATE_REVERSE`, making it byte-for-byte identical to `create_duplicate_group`), and switched its five `--verify` call sites to `create_duplicate_group`.
- `src/lib/pipeline/chains/builder.rs`: added an `#[rstest]` case table asserting `bam_group_key_config().umi_tag` is `Some(RX)` for a `Dedup` chain and `None` under `--no-umi`, covering the wiring the gate tests did not.

nh13:
@coderabbitai review

PR #919, opened 2026-09-04
fix(correct): error on --min-corrected with empty input on the legacy paths

nh13:
## What

The two legacy (non-chain) `fgumi correct` execution paths silently accepted `--min-corrected` on empty input. Both computed `ratio_kept = records_written as f64 / total_records as f64` and gated `if ratio_kept < min`; on empty input `total_records == 0`, so `ratio_kept` is `0.0/0.0 = NaN`, and `NaN < min` is always `false` — so `--min-corrected` did nothing. The chain path (`check_min_corrected`) and fgbio both error in this case, so the default (no-`--threads`) path diverged from both.

## Fix

- Extract `check_min_corrected` into a single shared `pub(crate)` helper (in `src/lib/commands/correct.rs`) that treats `total_records == 0` with a positive `min` as an explicit error, and have the chain module import and reuse it instead of keeping its own copy — one implementation, no divergent duplicates.
- Both legacy sites (`execute_single_thread_mode`, and the currently-`#[allow(dead_code)]` `execute_threads_mode`, fixed for sibling parity) now call the shared helper and error on empty input, matching the chain path and fgbio.

The `--metrics` write ordering is deliberately unchanged: fgbio writes the metrics TSV *before* the `--min-corrected` assert (CorrectUmis.scala:152 — "a ratio below `--min-corrected` will cause a failure but all files will still be written"), and the legacy path already matches that.

## Tests

- `test_min_corrected_empty_input_single_threaded_mode` (rstest) and `test_min_corrected_errors_on_empty_input_threads_mode` — both legacy paths error on empty input + positive `--min-corrected`, and pass with non-empty/met input; verified red pre-fix.
- `test_min_corrected_empty_input_chain_mode` (rstest) — the live `--threads` chain path errors on empty input end-to-end.

Full gate green (`cargo ci-test` 9960 passed / 31 skipped, plus fmt/lint/doc and `--no-default-features`/`--all-features`).

## Relationship to the chain-only cutover

This is a live-bug fix on the default legacy path. A separate PR in this series retires the legacy `correct` path entirely (chain-only), which will delete this code — at which point the chain path (already correct) is the only path. Landing this first fixes the live bug immediately; if the cutover lands first, this can be closed as moot.

Risk: command output changes: none; `unsafe` changes: none, and CLAUDE.md needs no update; memory bounds, queue capacity, and thread/backpressure policy changes: none.

The shared `check_min_corrected` helper now rejects empty input when `--min-corrected` is positive across legacy and chain paths. Successful output and metrics ordering remain unchanged. Tests cover all execution paths.

### DIFF
--- a/src/lib/commands/correct.rs
+++ b/src/lib/commands/correct.rs
@@ -1370,14 +1370,7 @@ impl CorrectUmis {
 
         // Check minimum correction ratio
         if let Some(min) = self.min_corrected {
-            #[allow(clippy::cast_precision_loss)]
-            let ratio_kept = records_written as f64 / total_records as f64;
-            if ratio_kept < min {
-                bail!(
-                    "Final ratio of reads kept / total was {ratio_kept:.2} (user specified minimum was {min:.2}). \
-                    This could indicate a mismatch between library preparation and the provided UMI file."
-                );
-            }
+            check_min_corrected(records_written, total_records, min)?;
         }
 
         Ok(total_records)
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #921, opened 2026-09-04
fix(correct): write --metrics only when processing completes (fgbio parity on --min-corrected failure)

nh13:
## What

On the chain `correct` path, the metrics finalize hook was on the **always-run** list, so a run that failed with a genuine pipeline processing error still left a `--metrics` TSV on disk. This moves the metrics write to a success-only hook — but preserves fgbio's exact contract for the `--min-corrected` gate.

## The fgbio contract this preserves

fgbio writes the metrics file once processing completes, *before* checking the `--min-corrected` ratio — a ratio below the floor is a failure, but **all files (including metrics) are still written**. So the correct behavior is:

- **Pipeline processing error** (I/O, malformed input): no metrics written (the run didn't complete).
- **`--min-corrected` ratio failure** (run completed, ratio below floor): metrics **are** written, then the command errors — the metrics are exactly the diagnostic that explains *why* the ratio was low.

## How

A single success-only `CorrectMetricsFinalizeHook` (registered on `finalize_on_success`) writes the `--metrics` TSV **first**, then — only when `--min-corrected` was given — runs `check_min_corrected`. The `--min-corrected` gate is removed from the always-run `CorrectFinalizeHook` (which keeps the counter reduce, summary, #912's error-level missing/wrong-length banner, and the timer). On a pipeline error, `finalize_on_success` is skipped entirely, so neither the metrics write nor the gate runs. This is the two-hook shape retag/filter already use.

A shared `sum_correct_counts` reducer (over `CorrectCounts`) is used by both the summary hook and the metrics hook's ratio gate, so the two can't diverge on what counts as "rejected."

## Tests

- `test_correct_chain_writes_metrics_on_min_corrected_ratio_failure` — real records with kept/total = 3/5 = 0.6 against a 0.9 floor: asserts the command errors with the **ratio-branch** message AND the `--metrics` TSV is written (parsed back with `UmiCorrectionMetrics::read_metrics` and asserted on content, not just existence).
- `test_correct_chain_writes_metrics_on_min_corrected_failure_empty_input` — the empty-input (0/0) branch, same "metrics still written" contract.
- `test_correct_chain_no_metrics_on_pipeline_failure` — CRC-corruption forces a genuine pipeline error → no metrics file.

Full gate green (`cargo ci-test` 9958 passed / 31 skipped, plus fmt/lint/doc and `--no-default-features`/`--all-features`).

Stacked on #912.

Risk: metrics output changes, pinned by regression tests; grouping, consensus, sort order, and corrected UMI output changes none; unsafe changes none; memory, queue, and thread/backpressure changes none.

Metrics write after successful pipeline processing and before `--min-corrected` validation. Processing failures skip metrics. Shared count reduction supports summary and ratio validation. Tests cover empty input, low corrected ratios, and corrupted BGZF input.

### DIFF
--- a/src/lib/pipeline/chains/builder.rs
+++ b/src/lib/pipeline/chains/builder.rs
@@ -1901,9 +1901,16 @@ impl<'a> ChainBuilder<'a> {
     /// `self.header` is the input header (correct preserves input order and does
     /// not replace `self.header` with a consensus header), so it is used as-is.
     ///
-    /// Registers a [`CorrectFinalizeHook`] for: per-thread accumulator reduce →
-    /// `finalize_metrics` → summary + warn banners → `--min-corrected` ratio
-    /// gate (bail!) → `timer.log_completion`.
+    /// Registers two finalize hooks. Always-run: [`CorrectFinalizeHook`]
+    /// (per-thread counter reduce → summary → warn banners →
+    /// `timer.log_completion`). Success-only: [`CorrectMetricsFinalizeHook`],
+    /// which writes the `--metrics` TSV FIRST and then, only when
+    /// `--min-corrected` was given, enforces the ratio gate. That ordering
+    /// matches fgbio: `CorrectUmis` writes its metrics before checking the
+    /// `--min-corrected` ratio, so a ratio-below-floor failure is not treated
+    /// as a processing failure and still leaves the metrics file in place. A
+    /// genuine pipeline processing failure skips `finalize_on_success`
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #922, opened 2026-09-04
refactor(extract): retire the legacy single-threaded path; the chain is the only path

nh13:
## What

Retire the legacy single-threaded `fgumi extract` path so the declarative chain builder is the **only** execution path (`extract` carried the largest serial oracle of the per-command cutovers). `execute()` keeps all its pre-flight validation and then always dispatches to the chain; the serial engine and its now-dead helpers are deleted.

Output is unchanged: the chain is a strict superset of the serial path.

Confirmed the chain already supports every extract feature before removing the serial path: interleaved (`SourceSpec::InterleavedFastq`) and two-file (`SourceSpec::Fastqs`) input, read structures, `--check-crc`/`--no-check-crc`, UMI-from-read-names, and the QX/CY/QT quality-storage flags. Diagnostics: the `Extracting UMIs` timer and `Processed records` progress line were already on the chain; the chain additionally emits `Starting Extract` + Input/Output banners and an `Extract: completed` summary (additive). The read-name-mismatch / out-of-sync rejection is now produced by the chain FASTQ source — the exact wording a `--threads` run already emitted (two `#[should_panic]` unit tests updated from "Read names do not match" to "FASTQ read name mismatch"). Nothing user-observable is lost.

**Deleted (serial-only):** `process_singlethreaded`, `make_raw_records`, `build_template_record`, serial `create_header`/`add_to_read_group`, `validate_read_names_match`, and their now-dead imports.

## Tests

- `tests/integration/test_extract_cutover_parity.rs` — pins chain output == the pre-removal serial baseline via `FGUMI_BASELINE_BIN` (records byte-identical modulo `@PG`) across two-file, interleaved, and `--check-crc`. When the env var is unset, the self-consistency oracle validates **every** produced record (sequence + `RX` + `RG`) against the input read structure — not just the first pair — so default CI genuinely pins the full output.
- `no_threads_matches_threaded_barcode_and_annotate_tags` now feeds per-position-distinct qualities and asserts exact `CY`/`QT`/`QX` byte values (not just presence).

Full gate green (`cargo ci-test` 9957 passed / 31 skipped with `FGUMI_BASELINE_BIN` set proving cutover byte-parity, plus fmt/lint/doc and `--no-default-features`/`--all-features`). Part of the per-command R6-0 legacy-path retirements; no user-facing behavior change. (No-`--threads` now runs the chain at a single worker — the intended chain-only architecture.)

Risk: extract data output changes: none; parity tests pin output. `unsafe` changes: none; CLAUDE.md needs no update. Thread policy changes: no-`--threads` uses one chain worker; queue capacity and backpressure remain unchanged.

- Removes the legacy serial extract engine and helpers.
- Routes all `extract` executions through the declarative chain.
- Preserves paired and interleaved FASTQ, read structures, CRC checks, UMI read-name handling, and quality-storage flags.
- Adds parity coverage for paired, interleaved, and BGZF inputs.
- Updates diagnostics, header documentation, and default `PL` behavior.

### DIFF
--- a/src/lib/commands/extract.rs
+++ b/src/lib/commands/extract.rs
@@ -23,18 +23,13 @@ use crate::commands::common::{
 };
 use crate::fastq::FastqSegment;
 use crate::fastq::FastqSet;
-use crate::fastq::ReadSetIterator;
 use crate::fastq_deinterleave::deinterleave;
 use crate::fastq_parse::strip_read_suffix;
-use crate::logging::OperationTimer;
 use crate::sam::SamTag;
-use crate::unified_pipeline::fastq_out_of_sync_error;
 use crate::validation::validate_input_exists;
 use anyhow::{Context, Result, bail, ensure};
 use bstr::{BString, ByteSlice};
 use clap::Parser;
-use fgumi_bam_io::ProgressTracker;
-use fgumi_bam_io::{RawBamWriter, create_raw_bam_writer};
 use fgumi_raw_bam::UnmappedSamBuilder;
 use fgumi_raw_bam::fields::flags;
 use log::{debug, info};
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #923, opened 2026-09-04
refactor(clip): retire the legacy single-threaded path; the chain is the only path

nh13:
## What

Retire the legacy single-threaded `fgumi clip` path so the declarative chain builder is the **only** execution path. `execute()` keeps its pre-flight validation and then always dispatches to the chain; `execute_single_threaded` and its now-dead imports are deleted.

Stacked on #915 (which put clip `--metrics` on the chain), so the metrics precondition is already satisfied. Output is byte-identical (modulo `@PG`).

Every diagnostic the serial path emitted is already produced by the chain (`ChainBuilder::add_clip` + `ClipFinalizeHook`/`ClipMetricsFinalizeHook`): output records, the `--metrics` TSV, the `Clip` banner + Input/Output/mode lines, the `OperationTimer`, the `require_query_grouped` guard, the summary counters, and the CRC-verify line. Nothing needed adding.

## Tests

- `tests/integration/test_clip_cutover_parity.rs` — pins chain output == the pre-removal serial baseline via `FGUMI_BASELINE_BIN` (records byte-identical modulo `@PG`, plus the `--metrics` TSV) across fragment, overlapping FR pair, pre-existing soft-clip R1, pre-existing hard-clip R2, and secondary+supplementary. When `FGUMI_BASELINE_BIN` is unset (default CI), the self-consistency oracle proves **this run actually clipped** — it asserts the output's total clipped-base footprint strictly exceeds the input's (a passthrough leaves them equal → fails) and pins that a plain 8M R1 is left clipped by `--read-one-five-prime` + `--clip-overlapping-reads`; the `--metrics` fallback asserts non-zero counts for the exercised shapes.
- Existing determinism tests retained as worker-count-invariance checks.

Full gate green (`cargo ci-test` 9961 passed / 31 skipped with `FGUMI_BASELINE_BIN` set proving byte-parity, plus fmt/lint/doc and `--no-default-features`/`--all-features`). Part of the per-command R6-0 legacy-path retirements; no user-facing behavior change (no-`--threads` now runs the chain at a single worker).

Risk: output changes because `clip` now always uses the declarative chain; parity tests and worker-count checks pin records and metrics; `unsafe` changes: none; `CLAUDE.md` allowlist update: none; memory bounds and queue capacity: none; no-`--threads` runs with one chain worker.

- Removed the legacy serial `clip` implementation.
- Preserved pre-flight validation and query-grouped input checks.
- Added parity coverage for records and metrics across clipping scenarios.
- Added chain-routing and worker-count coverage.
- Reported test gate: 9,961 passed and 31 skipped.

### DIFF
--- a/src/lib/commands/clip.rs
+++ b/src/lib/commands/clip.rs
@@ -4,20 +4,14 @@
 //! This is useful for variant calling to avoid double-counting evidence from
 //! overlapping portions of paired reads.
 
-use crate::alignment_tags::regenerate_alignment_tags_raw;
 use crate::clipper::{ClippingMode, RawRecordClipper};
-use crate::logging::OperationTimer;
 use crate::metrics::clip::{ClipCounts, ClippingMetricsCollection};
-use crate::reference::ReferenceReader;
 use crate::sam::SamTag;
-use crate::template::{InsertSizeEnd, TemplateIterator, compute_insert_size_from_ends};
+use crate::template::{InsertSizeEnd, compute_insert_size_from_ends};
 use crate::validation::validate_file_exists;
 use anyhow::Result;
 use clap::Parser;
-use fgumi_bam_io::ProgressTracker;
-use fgumi_bam_io::{create_raw_bam_reader_with_opts, create_raw_bam_writer};
 use fgumi_raw_bam::RawRecord;
-use log::info;
 use std::path::PathBuf;
 
 use super::command::Command;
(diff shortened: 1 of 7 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #924, opened 2026-09-04
refactor(filter): retire the legacy single-threaded path; the chain is the only path

nh13:
## What

Retire the legacy single-threaded `fgumi filter` path so the declarative chain builder is the **only** execution path. `filter` was a `unified_pipeline` default consumer; `execute()` keeps its pre-flight validation and then always dispatches to the chain. Deleted: the `--threads` branch, `execute_threads_mode_{template,single_read}`, `run_filter_pipeline`, `build_filter_pipeline_config`, `FilterProcessedBatchRaw`, `Filter::write_filter_stats`, `advance_progress`/`progress_heartbeat_total`, and now-dead imports (the shared engine `process_record_raw` + `check_*` + `FilterOptions::*` is kept). Stacked on #916. Output is byte-identical (modulo `@PG`).

Every diagnostic the legacy path emitted is already produced by the chain (`ChainBuilder::add_filter` + `FilterFinalizeHook`/`FilterStatsFinalizeHook`): the CRC line, `Starting Filter` banner + params, `OperationTimer`, query-grouped rejection, `@HD`/`@PG`, kept/rejected/masked summary, the `--rejects` BAM, the success-only `--stats` TSV, and the per-batch heartbeat. Nothing needed adding.

## Restoring the name_hash_only guard

Before this PR, `test_filter_chain_matches_single_threaded_with_rg_and_cb_variation` compared the chain's `name_hash_only` decode key against the legacy path's full-key decode — the contrast that proved the RG/position skip was invisible to filtering. With the legacy path gone, both sides now use `name_hash_only`, so that test only proves worker-count invariance (its docstring was corrected to say so). To keep the skip guarded in-repo (without `FGUMI_BASELINE_BIN`), a new unit test `decode::tests::name_hash_only_matches_full_key_grouping_when_rg_and_position_vary` asserts `name_hash_only` vs a full `GroupKeyConfig` yield identical grouping for records whose RG/position vary (and that the full key genuinely varies, so the check is non-vacuous).

## Tests

- `tests/integration/test_filter_cutover_parity.rs` — pins chain output == the pre-removal serial baseline via `FGUMI_BASELINE_BIN` (kept records + `--rejects` BAM byte-identical modulo `@PG`, plus the `--stats` TSV) across template, single-read, with-rejects, with-stats, single-read+rejects, single-read+stats. The corpus now includes a real paired template so the `template` case exercises template-drop (whole template rejected) distinct from single-read. The self-consistency oracle (default CI) is mode-aware and asserts the masked-base position + regenerated `NM`.

Full gate green (`cargo ci-test` 9958 passed / 31 skipped with `FGUMI_BASELINE_BIN` set proving byte-parity, plus fmt/lint/doc and `--no-default-features`/`--all-features`). Part of the per-command R6-0 legacy-path retirements; no user-facing behavior change (no-`--threads` now runs the chain at a single worker). The shared `--threads` help text is reworded generically by the retag cutover PR (not re-touched here to avoid a merge conflict).

Risk: `filter` output may change; parity tests and the optional pre-removal baseline pin records, grouping, rejects, headers, and metrics. Consensus, sort order, and corrected UMIs: none. `unsafe`: none; CLAUDE.md allowlist: unchanged. Memory bounds and queue capacity: none; thread and backpressure policy: changed to declarative-chain scheduling with one worker when `--threads` is absent.

Fix: route all `filter` executions through the declarative chain.

- Retires the legacy single-threaded filter pipeline.
- Preserves validation, filtering, rejects, headers, statistics, and shared record processing.
- Adds parity coverage for templates, reads, rejects, statistics, masking, regenerated `NM`, and grouping.
- Updates chain documentation and batching comments.
- Test gate passed: 9,958 passed and 31 skipped.

### DIFF
--- a/src/lib/commands/filter.rs
+++ b/src/lib/commands/filter.rs
@@ -17,38 +17,26 @@ use crate::consensus_filter::{
     filter_read, is_duplex_consensus, mask_bases, mask_duplex_bases,
     mask_methylation_depth_duplex_raw_with_tags, mask_methylation_depth_simplex_raw_with_tags,
     mask_strand_methylation_agreement_raw_with_ref_bases_and_tags, mean_base_quality_full_length,
-    retained_primary_masked_bases, template_passes,
 };
-use crate::grouper::{SingleRawRecordGrouper, TemplateGrouper};
-use crate::logging::OperationTimer;
 use crate::per_thread_accumulator::PerThreadAccumulator;
-use crate::read_info::LibraryIndex;
 use crate::reference::ReferenceReader;
 use crate::tag_reversal::reverse_per_base_tags_raw;
-use crate::template::TemplateBatch;
-use crate::unified_pipeline::{
-    BamPipelineConfig, BatchWeight, GroupKeyConfig, Grouper, MemoryEstimate,
-    run_bam_pipeline_from_reader, run_bam_pipeline_from_reader_with_secondary,
-};
 use crate::validation::validate_file_exists;
-use ahash::AHashMap;
 use anyhow::{Result, bail};
 use clap::Parser;
-use fgumi_bam_io::create_bam_reader_for_pipeline_with_opts;
 use fgumi_raw_bam;
 use fgumi_raw_bam::{RawRecord, RawRecordView};
 use log::info;
 use noodles::sam::Header;
-use std::io;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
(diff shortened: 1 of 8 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #925, opened 2026-09-04
refactor(correct): retire the legacy single-threaded path; the chain is the only path

nh13:
## What

Retire the legacy single-threaded `fgumi correct` execution so the declarative chain builder is the **only** path. `correct` had TWO legacy engines — the no-`--threads` serial `execute_single_thread_mode` and the `#[allow(dead_code)]` `execute_threads_mode`. `execute()` keeps its reader-free pre-flight (validate, reject_output_collisions) and then always dispatches to the chain. Deleted both engines plus the legacy-only `CorrectProcessedBatch` (+ its `MemoryEstimate`), the three dead `CorrectUmis` wrapper methods (`load_umi_sequences`/`check_umi_distances`/`finalize_metrics`), and now-dead imports (−777 lines).

Stacked on #921 (the correct `--metrics` fgbio-parity fix), which already made the chain correct-path fgbio-correct. Output is byte-identical (modulo `@PG`).

Every diagnostic the serial engine emitted is already produced by the chain (`add_correct` + finalize hooks): CRC line, `Starting correct` banner, `OperationTimer`, `Loaded N UMI sequences`, the ambiguous-pair distance warning, summary/record-count, the missing/wrong-length error banner, the `--metrics` TSV, and the `--min-corrected` gate. Nothing needed adding — **except** one banner correctness fix the cutover surfaces (below).

## Banner fix (the cutover makes the chain banner the sole banner)

The chain's missing/wrong-length error banner printed *both* detail lines whenever *either* counter was nonzero — emitting a spurious `# 0 were missing…` / `# 0 had unexpected…` line that neither the retired serial engine nor fgbio ever printed. Each detail line is now individually guarded (`if counts.missing > 0` / `if counts.wrong_length > 0`). This is the only chain-hook change (17 lines).

## Two behavior points, both correct-by-design

- **Empty input + `--min-corrected`:** the retired no-`--threads` serial engine silently *passed* (`NaN < min` is false); the chain errors, matching fgbio. New test `cutover_empty_input_min_corrected_now_errors` pins it (and, with a baseline binary, asserts the frozen legacy binary *succeeds* on the same input — a RED/GREEN discriminator documenting the fix).
- **`--min-corrected` ratio failure** still writes the `--metrics` TSV (fgbio parity, from #921) — withheld only on a genuine pipeline processing error.

## Tests

`tests/integration/test_correct_cutover_parity.rs` pins chain == pre-removal serial baseline via `FGUMI_BASELINE_BIN` (records byte-identical modulo `@PG`, plus the `--metrics` TSV) for a normal run and a `--min-corrected` failure. The self-consistency oracle (default CI) asserts the exact corrected UMI values (`AAAAAAAT`→`AAAAAAAA`, the full post-correction RX distribution, per-UMI `total_matches`), not just counts.

Full gate green (`cargo ci-test` 9961 passed / 31 skipped with `FGUMI_BASELINE_BIN` set, plus fmt/lint/doc and `--no-default-features`/`--all-features`). Part of the per-command R6-0 legacy-path retirements; no user-facing behavior change beyond the two correct-by-design points above (no-`--threads` now runs the chain at a single worker).

**Supersedes #919** (which fixes the legacy `--min-corrected` NaN bypass): this PR deletes that legacy path entirely. If #919 lands first, this rebases cleanly; if this lands first, #919 can be closed as moot.

Risk: `correct` output changes are pinned by serial-baseline and parity tests; `unsafe` changes: none, with no CLAUDE.md allowlist update; no memory-bound, queue-capacity, or backpressure changes, but no `--threads` now uses one chain worker.

- Routes all `correct` executions through the declarative chain.
- Removes both legacy execution engines and related code.
- Preserves validation and output-collision checks.
- Adds parity tests for records, corrected UMIs, rejects, metrics, headers, and `--min-corrected` failures.
- Makes empty input with `--min-corrected` fail as in fgbio.
- Omits zero-count UMI detail lines while retaining the error banner.

### DIFF
--- a/src/lib/commands/correct.rs
+++ b/src/lib/commands/correct.rs
@@ -5,11 +5,13 @@
 //! observed UMIs to match the expected set based on mismatch tolerance and minimum
 //! distance requirements.
 //!
-//! When `--rejects` is set, the threaded pipeline routes rejected records through
-//! the unified pipeline's first-class secondary output (see
-//! [`crate::unified_pipeline::run_bam_pipeline_from_reader_with_secondary`]).
-//! Rejects land in batch-input order rather than mutex-acquisition order, and the
-//! rejects BAM inherits the input header. The pattern matches `commands::filter`.
+//! Execution always runs on the declarative chain builder (via
+//! `ChainBuilder::add_correct`); `--threads` only sets the worker count, and
+//! absent it the chain runs at a single worker.
+//! When `--rejects` is set, the chain routes rejected records through its
+//! first-class secondary output branch so rejects land in batch-input order and
+//! the rejects BAM inherits the input header. The pattern matches
+//! `commands::filter`.
 //!
 //! By default (`--target umi`) the tool reads and writes the `RX` UMI tag and
 //! stores the pre-correction value in `OX`. With `--target barcode` it instead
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #931, opened 2026-09-06
feat(sort): add --check-crc / --no-check-crc opt-out (currently hard-coded to always verify)


nh13:
### Summary

`fgumi sort` is the only `BamIoOptions`-style command that does **not** expose `--check-crc` / `--no-check-crc`. It hard-codes `verify_crc: true` (always verifies, including stdin), so there is no way to opt out of BGZF CRC32 verification on a sort.

### Why it's like this (deliberate, not a bug)

When `sort` moved onto the unified chain, the ruling was to preserve parity with the pre-chain "owned" sorter, which always verified CRC (including stdin). Using the standard `BamIoOptions::effective_check_crc()` policy (verify files, skip trusted stdin) would have flipped sort's stdin default from verify to skip — a silent regression from the owned sorter. So `sort` pins `verify_crc: true` with an in-code note, and #800 explicitly deferred sort's flag to its own issue.

- `src/lib/commands/sort.rs` — `verify_crc: true` with the parity comment.
- `src/lib/commands/common.rs` — `BamIoOptions` + `effective_check_crc()` (the policy every other command uses).
- Existing coverage: `tests/integration/test_sort_cutover_parity.rs::cutover_stdin_input_detects_corrupt_crc` pins sort's always-verify-stdin behavior.

### Proposed

Expose `--check-crc` / `--no-check-crc` on `sort`, **keeping the default as always-verify (including stdin)** so `--no-check-crc` is a pure opt-out and no default behavior changes. (Adopting the file-vs-stdin default instead is a separate, deliberate policy change that would need sign-off — see the parity note above.)

*Authored by AI, reviewed by a Human.*

PR #932, opened 2026-09-06
chore(coderabbit): mark the BGZF CRC verification policy as intentional

nh13:
CodeRabbit repeatedly flags the BGZF CRC32 verification policy as a data-integrity gap — most recently on #920, where it claimed stdin "always verifies" and that `--no-check-crc` is defeated. Both readings are wrong, and the behavior is a settled design decision from the #786/#798 stack.

This adds a `path_instructions` entry over the CRC decode files (`commands/common.rs`, `pipeline/steps/source/**`, `pipeline/chains/{builder,spec}.rs`, `fgumi-bam-io/src/reader.rs`) recording that:

- The default is **verify file input / skip trusted stdin**, resolved by `BamIoOptions::effective_check_crc()` (`!is_stdin` when neither flag is set) and threaded to the chain as `ChainSpec::verify_crc`, which `BgzfDecompress::new_with_crc` honors; `--check-crc` / `--no-check-crc` override it.
- A `PipelineReaderOpts` left at its default `verify_crc: true` in `ChainBuilder::open_source` is **inert** for the chain (the chain reads its policy from `spec.verify_crc`), so it should not be flagged.
- The BAM **header block is always CRC-verified** (noodles) in `read_header_and_replay` / `open_bam_from_stdin_boxed_buf` — a deliberate tradeoff (the CRC-configurable fgumi-bgzf reader's readahead over-consumes the `TeeReader` and breaks the exact-byte header replay), documented in-code.
- `fgumi sort` intentionally pins `verify_crc: true`; its missing `--no-check-crc` opt-out is tracked separately (#931), not a defect.

Config-only; YAML validated. No code or behavior change.

Risk: command output changes: none; `unsafe`: none, and `CLAUDE.md` allowlist is unchanged; memory bounds, queue capacity, and thread/backpressure policy: none.

Fix: add CodeRabbit guidance for the intentional BGZF CRC verification policy.

- Documents file-input verification, trusted-stdin behavior, override precedence, BAM header verification, and `fgumi sort` behavior.
- Configuration-only change. No runtime behavior changes.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -415,6 +415,26 @@ reviews:
         strand-bias distributions) for bias, off-by-one, and mis-seeded RNG, and
         require its writers to propagate the FIRST error from a worker or I/O
         thread rather than a later generic channel-closed error that masks it.
+    - path: "{src/lib/commands/common.rs,src/lib/pipeline/steps/source/**/*.rs,src/lib/pipeline/chains/builder.rs,src/lib/pipeline/chains/spec.rs,crates/fgumi-bam-io/src/reader.rs}"
+      instructions: >-
+        BGZF CRC32 verification policy here is intentional and settled (PRs
+        #786/#798) — do NOT re-flag it as a data-integrity gap or a missing
+        `--no-check-crc` wiring. The default is verify-file / skip-trusted-stdin,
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #933, opened 2026-09-06
feat(sort): add --check-crc / --no-check-crc with the file-vs-stdin default

### LINKED ISSUE #931: feat(sort): add --check-crc / --no-check-crc opt-out (currently hard-coded to always verify)

nh13:
## What

Adds `--check-crc` / `--no-check-crc` to `fgumi sort`, the only `BamIoOptions`-style command that lacked them. The flags adopt the standard file-vs-stdin default that every other BAM command uses (`resolve_check_crc`): an explicit flag wins; otherwise file input is verified and stdin (`-` / `/dev/stdin`) is trusted and skipped. Every run logs a `CRC verify: on|off (<reason>)` line at startup.

Closes #931.

## Behavior change (deliberate)

`sort` previously pinned `verify_crc: true`, so it always verified BGZF CRC32 — including on stdin. This PR changes the **stdin default from verify to skip**: a freshly piped aligner stream (`bwa-mem3 … | fgumi sort -i -`) is trusted, since corruption there is an upstream bug rather than data at rest. Pass `--check-crc` to keep verifying piped input. File input is unchanged (still verified by default); `--no-check-crc` opts a file out. The BAM header block is always verified regardless (the noodles header tee), matching every other command.

## Why it needed plumbing (not just a flag)

The original deferral assumed flipping `verify_crc` would suffice, but standalone `fgumi sort` (`[Stage::Sort]` over a BAM source) decodes through `ReadBlocks → InflateToArena → fgumi_bgzf::decompress_into_slice`, which had no CRC toggle and always verified — `ChainSpec.verify_crc` was only read by `BgzfDecompress`, a path standalone sort never takes. So the flag is plumbed to the arena decode path:

- `fgumi-bgzf`: new `decompress_into_slice_with_crc(block, decompressor, out, verify_crc)`; the existing `decompress_into_slice` delegates with `verify_crc = true`.
- `fgumi-pipeline-io`: `InflateToArena` gains a `verify_crc` field + `new_with_crc`; `new` delegates, `new_worker_copy` propagates it to every parallel worker.
- chain builder: `add_sort` constructs the arena front with `InflateToArena::new_with_crc(byte_limit, spec.verify_crc)`.

Both `fgumi-bgzf` and `fgumi-pipeline-io` are published crates: no existing `pub` signature changed (new `_with_crc` variants only; `InflateToArena` gained a private field, constructed only via constructors).

`--verify` mode reads through a separate always-verifying reader and is unaffected by these flags; passing `--check-crc`/`--no-check-crc` alongside `--verify` emits a warning rather than silently doing nothing.

## Invariant

Only the CRC32 compare is gated by `verify_crc`. The exact-`BGZF_EOF` short-circuit, ISIZE bound, `out.len() == ISIZE`, stored-frame LEN/ISIZE, and exact-fill checks all stay unconditional — a `--no-check-crc` run still rejects a structurally malformed or short/over-long block.

## Testing

- `fgumi-bgzf`: rstest tables gate the CRC compare on both decode branches (stored + deflate), prove non-CRC faults still reject under `verify_crc = false`, and pin the EOF-marker contract.
- `fgumi-pipeline-io`: `InflateToArena` honors the policy on both branches, plus a worker-copy propagation test.
- `common.rs`: `resolve_check_crc` truth table + log-reason wording.
- `sort.rs`: `build_sort_chain_spec` resolves `verify_crc` across `{file, -, /dev/stdin} × {default, --check-crc, --no-check-crc}`; flag-conflict test.
- Integration (`test_sort_cutover_parity.rs`): end-to-end `{stdin, file} × {default, --check-crc, --no-check-crc}` table on a one-bit last-body-block CRC flip — rejected exactly when the policy says verify, sorted past intact when it says skip. This case table replaces the old always-verify stdin regression test and is the behavioral gate for the arena wiring (reverting the wiring fails exactly the three skip cells).

Full local gate green: `cargo ci-test`, `ci-fmt`, `ci-lint`, `ci-publish-order`, workspace doctests.

## Suggested reading order

Bottom-up, one commit per layer: `fgumi-bgzf` → `fgumi-pipeline-io` → chain builder → `common.rs` → `sort.rs` → integration test → docs.

Risk: sort stdin output behavior changes because CRC faults are skipped by default; file input remains verified, and explicit CRC flags pin the policy. `unsafe`: none. Memory bounds, queue capacity, and thread/backpressure policy: none.

- Adds `--check-crc` and `--no-check-crc` to `fgumi sort`.
- Uses the shared file-versus-stdin CRC policy.
- Propagates CRC settings through standalone sort arena decoding.
- Preserves structural BGZF validation when CRC comparison is disabled.
- Keeps existing decompression APIs compatible.
- Logs the effective CRC setting and reason.
- Keeps `--verify` behavior unchanged and warns when CRC flags are inert.
- Adds coverage for decoder behavior, worker propagation, policy resolution, conflicts, and file/stdin integration.

### DIFF
--- a/crates/fgumi-bgzf/src/lib.rs
+++ b/crates/fgumi-bgzf/src/lib.rs
@@ -19,7 +19,8 @@ pub use header::{
 pub use reader::{
     BGZF_EOF, BGZF_FOOTER_SIZE, MAX_UNCOMPRESSED_BLOCK_SIZE, RawBgzfBlock, decompress_block,
     decompress_block_into, decompress_block_into_opts, decompress_block_slice_into,
-    decompress_block_slice_into_opts, decompress_into_slice, read_raw_blocks, uncompressed_size,
+    decompress_block_slice_into_opts, decompress_into_slice, decompress_into_slice_with_crc,
+    read_raw_blocks, uncompressed_size,
 };
 // Re-export the libdeflater decompressor so downstream crates can name the type
 // required by `decompress_block_into_opts` without depending on libdeflater
--- a/crates/fgumi-bgzf/src/reader.rs
+++ b/crates/fgumi-bgzf/src/reader.rs
@@ -820,12 +820,15 @@ fn copy_stored_and_verify_slice(
     out: &mut [u8],
     expected_crc: u32,
     block_len: usize,
+    verify_crc: bool,
 ) -> io::Result<usize> {
     // `out` is caller-sized to the footer's ISIZE, so passing `out.len()` also
     // enforces LEN == ISIZE; the returned payload is then exactly `out.len()`.
+    // The LEN/ISIZE check runs unconditionally; `verify_crc` gates only the
+    // footer CRC32 compare (see `verify_decompression`).
     let payload = parse_stored_frame(compressed, out.len())?;
     out.copy_from_slice(payload);
-    verify_decompression(out, out.len(), expected_crc, block_len, true)?;
+    verify_decompression(out, out.len(), expected_crc, block_len, verify_crc)?;
     Ok(out.len())
 }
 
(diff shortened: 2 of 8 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #935, opened 2026-09-07
feat(simulate): add the aligner replay subcommand

nh13:
## What

Adds `fgumi simulate aligner`, a fake streaming aligner for benchmarking `fgumi runall`'s align stage. It is invoked as the aligner subprocess via `--aligner::command`: it drains the interleaved FASTQ on stdin (discarding it) while concurrently streaming a pre-captured aligned BAM (`--replay-bam`) verbatim to stdout. Reading and writing run on independent threads, so it never deadlocks `AlignAndMerge` the way a single-threaded lockstep replay (the shell `replay-aligner.sh` it replaces) does. It performs no real alignment — it replays alignments captured earlier (e.g. by `bwa mem`), letting a benchmark measure fgumi's own pipeline overhead without a real aligner's compute dominating every run.

## Why

The benchmark needs a drop-in for a real streaming aligner that adds as little overhead as possible. The concurrent drain/emit design is exactly what a real aligner does; the single-threaded shell replay it replaces deadlocked the align stage (the writer fills its in-flight gate feeding FASTQ and blocks, while the reader stalls waiting for aligner output the mid-drain shell can't emit).

## Details

- Streams the replay BAM byte-for-byte (header first; every record, including secondary/supplementary, in stored order), so the replay must be in input / queryname-grouped order (as `bwa mem` emits it), which is how `AlignAndMerge` regroups records into templates by queryname.
- Trailing positional tokens (the `{ref}` / `{threads}` a command template substitutes) are accepted and ignored, so the command is a drop-in for real-aligner templates. `--replay-bam` must precede any positional token — documented on the field, in `docs/simulate-cli.md`, and pinned by a test.
- The emitter locks stdout once and block-buffers it: the replay is BGZF, dense in newline bytes, so the default line-buffered stdout would split nearly every copy chunk at its last newline and roughly double the write syscall count.
- The exit paths are documented and each has a dedicated no-hang test: full emit joins the (detached) stdin drainer so the caller finishes feeding FASTQ before exit; a broken stdout pipe and an emit error both return without joining, since joining could otherwise block forever on stdin that never reaches EOF.

## Testing

- Unit tests cover the concurrent drain/emit (including a deadlock reproduction against a single-threaded write-then-read caller over real OS pipes), broken-pipe clean exit, the emit-error-without-join path, verbatim streaming, a full BAM round-trip through the project's BAM reader, and the argument-ordering constraint.
- A subprocess integration test drives the real binary through OS pipes with payloads larger than a pipe buffer.
- The integration module is gated on the `simulate` feature so a non-`simulate` build compiles it out rather than failing (matching `test_simulate_sort.rs`).
- `cargo ci-fmt`, `cargo ci-lint` (clippy pedantic), and the aligner unit + integration tests all pass.

## Scope

Ported from the `feat-runall` branch onto `main`, deliberately limited to the aligner command: the existing delegating `region_to_bin` is left untouched, the separate `simulate sort` command is not included, and only the aligner-relevant `docs/simulate-cli.md` additions are ported.

Risk: output changes only for the new `fgumi simulate aligner` command, pinned by verbatim replay of the supplied BAM; `unsafe` changes: none; memory, queue, and thread/backpressure policy: changed through concurrent FASTQ draining and block-buffered BAM output.

Adds `fgumi simulate aligner` with replay-file validation, trailing argument support, concurrent FASTQ draining, broken-pipe handling, and integration coverage. Updates the CLI documentation and command dispatch.

### DIFF
--- a/docs/simulate-cli.md
+++ b/docs/simulate-cli.md
@@ -13,8 +13,9 @@ Generate synthetic sequencing data for testing and benchmarking the fgumi pipeli
 | `fgumi simulate grouped-reads` | Template-coord sorted BAM with MI tags | Input for `simplex`/`duplex`/`codec` | Required |
 | `fgumi simulate consensus-reads` | Mapped BAM with consensus tags | Input for `filter` | Required |
 | `fgumi simulate correct-reads` | Unmapped BAM + includelist | Input for `correct` | Not used |
+| `fgumi simulate aligner` | Replays a pre-captured aligned BAM to stdout | Fake aligner for `runall --aligner::command` | Not used |
 
-**Note:** All simulate subcommands (except `correct-reads`) require `--reference` / `-r` pointing to a reference FASTA file. Positions are sampled from real chromosomes, template sequences are extracted from the reference, and BAM headers contain actual contig names and lengths. Read orientations are a 50/50 mix of F1R2 and R1F2 (strand coin flip per molecule).
+**Note:** The data-generating simulate subcommands (all except `correct-reads` and `aligner`) require `--reference` / `-r` pointing to a reference FASTA file. Positions are sampled from real chromosomes, template sequences are extracted from the reference, and BAM headers contain actual contig names and lengths. Read orientations are a 50/50 mix of F1R2 and R1F2 (strand coin flip per molecule).
 
 ---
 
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #941, opened 2026-09-08
perf(group,dedup,clip): use DrainFirstScheduler for terminal grouping chains

nh13:
## Problem

The declarative chain-builder runtime made threaded `fgumi group`, `dedup`, and `clip` slower than the pre-chain-builder release (group ~7–11%), threaded mode only. A CPU profile over a matched window showed the slowdown is **entirely wait, not compute**: on-CPU samples are identical, but wait samples are +46%, dominated by `__psynch_cvwait`. Under N pool workers the default upstream-first `ChainOrderScheduler` has every worker favour production, so they all descend on the serial grouping spine (`GroupByPosition`/`GroupBam → process → … → Serialize`), contend its `parking_lot` mutex, and park on the condvar.

## Fix

`DrainFirstScheduler` walks a worker's live steps downstream-first; combined with the pre-existing skip-on-Serial-contention rule, exactly one worker drains the serial spine while the rest fall through to feeding the parallel decompress front. It was already opted in for sort-source chains; this extends it to **terminal `group`, `dedup`, and `clip`** — all the same drain-bound shape (a serial spine that must see a template's records together, fed by parallel decompress, with light per-template work).

The Terminal-vs-Intermediate decision lives in a pure, unit-tested `grouping_stage_wants_drain_first(stage, position)` predicate (mirroring `stages_want_umi_cache`). A `group` fused into a consensus stage (Intermediate) stays on the default — it's production-bound downstream. The flag is only ever set, never cleared, so a sort-sourced `sort→group→consensus` chain still runs drain-first via `add_sort`.

**The gate is deliberately narrow.** An empirical scheduler sweep (via a hidden `--pool-scheduler` override — see the stacked follow-up PR) found drain-first modestly helps some other commands (retag ~5%, simplex/codec ~3% at 60M) but **catastrophically hurts `extract` (~5.6× slower** — its FASTQ source is the bottleneck and downstream-first starves it). So only the confirmed, output-verified strong winners are enabled here; the rest stay on the default pending their own measured decision.

## Results (60M-record input, patched vs base, same tree, warm cache, tight reps)

| threads | group | dedup | clip |
| --- | --- | --- | --- |
| 1 | ~0% (control) | ~0% | ~0% |
| 2 | −12% | −11% | −7.5% |
| 8 | −14% | −8% | −7.2% |
| 16 | −8% | −6% | — |

The threads=1 control (no contention → no difference) confirms the mechanism.

## Correctness (perf-only, output unchanged)

- `group`: `compare bams --command group` = **EQUIVALENT** (9,407,981 molecules matched), 53,516,090 records.
- `dedup`: `compare bams --command dedup` (positional) = **IDENTICAL**, 0 diffs.
- `clip`: `compare bams --command clip` (positional) = **IDENTICAL**, 0 diffs, 60,099,378 records.
- New `grouping_stage_wants_drain_first` case-matrix unit test; existing threaded group/dedup/clip determinism + command suites pass; `cargo ci-fmt`/`ci-lint` clean.

Risk: output changes are none; `unsafe` changes are none and CLAUDE.md needs no update; thread scheduling policy changes, but memory bounds and queue capacity are none.  
Fix: Apply `DrainFirstScheduler` to terminal `group`, `dedup`, and `clip` stages. Keep intermediate grouping stages on the default scheduler. Parameterized tests cover terminal and intermediate cases.

### DIFF
--- a/src/lib/pipeline/chains/builder.rs
+++ b/src/lib/pipeline/chains/builder.rs
@@ -470,14 +470,33 @@ pub struct ChainBuilder<'a> {
     /// and uses the requested thread count like every other stage.
     override_pipeline_threads: Option<usize>,
 
-    /// Set when a BAM sort source is wired through the parallel-inflate arena
-    /// front (`ReadBlocks → InflateToArena → FindBoundariesAndSort`), for any
-    /// sort order. [`Self::build`] then opts the ENTIRE sort-first pipeline —
-    /// including fused `Sort → Group/Simplex/Duplex/Codec` chains — into the
-    /// downstream-first [`DrainFirstScheduler`](crate::pipeline::core::runtime::DrainFirstScheduler), so the sort's serial
-    /// boundary/key scan overlaps the parallel inflate (drained ahead of
-    /// production) instead of starving behind it on the shared pool. Chains
-    /// without a sort source keep the default upstream-first scheduler.
+    /// Set for a chain whose drain-bound shape benefits from downstream-first
+    /// dispatch. [`Self::build`] reads it to select the
+    /// [`DrainFirstScheduler`](crate::pipeline::core::runtime::DrainFirstScheduler)
+    /// in place of the default upstream-first scheduler. Two shapes set it:
+    ///
+    /// - A BAM sort source wired through the parallel-inflate arena front
+    ///   (`ReadBlocks → InflateToArena → FindBoundariesAndSort`), for any sort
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #942, opened 2026-09-08
feat(pipeline): add hidden --pool-scheduler override for A/B benchmarking

nh13:
> #941 has merged, so this PR is rebased onto `main` and now contains a single commit: the new `--pool-scheduler` flag. (The earlier `perf(group,dedup,clip)` commit was a duplicate of what shipped in #941 and has been dropped.)

## Why

The chain engine picks the pool dispatch scheduler per chain — drain-first for terminal `group`/`dedup`/`clip` and a BAM `sort` source, upstream-first otherwise. Validating that choice, or measuring whether an as-yet-unenabled command would benefit, meant building two binaries. This adds a runtime knob to A/B the tuning on any command from one build.

## What

A hidden `--pool-scheduler <auto|drain-first|chain-order>` on the shared `SchedulerOptions` (flattened into every chain command, cloned into `ChainSpec`), so it reaches `ChainBuilder::build` with no per-command wiring:

- `auto` (default) — automatic per-chain decision; **no behavior change**.
- `drain-first` / `chain-order` — force one direction on every chain.

`build()` resolves it through a pure, unit-tested `resolve_use_drain_first(override, auto)` and emits a `warn!` when it overrides the automatic choice, so a benchmark records that a non-default scheduler ran. Value names match the `--pipeline-stats` scheduler names.

- **Hidden** (`hide = true`); distinct from the legacy, chain-unwired `--scheduler` strategy flag.
- **Coverage:** rides on `SchedulerOptions`, so it reaches extract, copy-umi, correct, align, group, simplex, duplex, codec, clip, filter, dedup, retag. It does **not** reach `sort`, `zipper`, `fastq`, `downsample` (they don't flatten `SchedulerOptions`) — notably `sort` is drain-first but not overridable via this flag. Worth a follow-up if we want full coverage.

## What it already found (empirical sweep, this flag, 60M-class inputs, t8, tight reps)

| command | Δ drain-first vs chain-order | note |
| --- | --- | --- |
| group / dedup / clip | −6 … −14% | shipped in #941 (output verified) |
| retag | **−4.9%** | helps — candidate |
| simplex / codec | **−3.0 … −3.5%** | helps (modest) — candidate |
| **extract** | **+459% (5.6× slower)** | FASTQ source starved by downstream-first — must stay chain-order |

Takeaway: drain-first is **not** universally safe and **not** reliably predictable from chain shape — most commands modestly benefit, but a source-bound command (extract) is catastrophically hurt. Per-command measurement is mandatory, which is what this flag is for. Only the confirmed strong, output-identical winners (group/dedup/clip) are enabled by default (shipped in #941).

## Testing

- `resolve_use_drain_first_matrix` — 6-case `rstest` over the override × auto matrix.
- `pool_scheduler_override_warning_matrix` — 6-case `rstest` over the warning message selection.
- End-to-end: hidden from `--help`; rejects bogus values; `group --pool-scheduler chain-order` logs "forcing upstream-first"; `retag --pool-scheduler drain-first` logs "forcing downstream-first"; plain runs log nothing.
- `cargo ci-fmt`/`ci-lint` clean; existing `SchedulerOptions` unit tests updated for the new field.

Usage: `<cmd> --pool-scheduler drain-first` vs `--pool-scheduler chain-order` (same input/build) is a clean A/B; add `--pipeline-stats` to correlate.

Risk: output changes none; unsafe changes none and no `CLAUDE.md` allowlist update applies; memory bounds, queue capacity, and thread/backpressure policy changes none.

- Adds hidden `--pool-scheduler <auto|drain-first|chain-order>` support.
- Propagates the override through `ChainSpec` to `ChainBuilder::build`.
- Preserves automatic scheduling by default.
- Logs a warning when a forced mode changes the automatic choice.
- Adds resolution, validation, help visibility, warning, and default-behavior tests.

### DIFF
--- a/src/lib/commands/common.rs
+++ b/src/lib/commands/common.rs
@@ -1051,6 +1051,29 @@ pub struct AllowUnmappedOptions {
     pub enabled: bool,
 }
 
+/// Override for the chain-builder pool dispatch scheduler.
+///
+/// The chain engine normally chooses the pool's step-walk direction per chain:
+/// drain-bound shapes (terminal `group`/`dedup`, and a BAM `sort` source) run
+/// downstream-first (`DrainFirstScheduler`), everything else runs upstream-first
+/// (`ChainOrderScheduler`). This hidden override forces one choice regardless, so
+/// the tuning can be A/B-benchmarked on any command from a single binary without
+/// a rebuild. `Auto` is the default and changes nothing.
+///
+/// The value names match the scheduler names surfaced in `--pipeline-stats`
+/// (`drain-first` / `chain-order`), so a benchmark can correlate the flag with
+/// what actually ran.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, clap::ValueEnum)]
+pub enum PoolScheduler {
+    /// Use the automatic per-chain decision (the default; no override).
+    #[default]
+    Auto,
+    /// Force downstream-first dispatch (`DrainFirstScheduler`) on every chain.
+    DrainFirst,
+    /// Force upstream-first dispatch (`ChainOrderScheduler`) on every chain.
+    ChainOrder,
+}
+
 /// Options for pipeline scheduler configuration.
 ///
 /// Controls which scheduling strategy is used for thread work assignment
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #944, opened 2026-09-09
fix(runall): wire --methylation-mode/--ref into the fused filter stage

nh13:
## Summary

`fgumi runall`'s fused filter stage silently dropped the top-level `--methylation-mode` / `--ref`, so runall's methylation-aware filter option `--filter::min-conversion-fraction` was unreachable. `build_stage_options_bag`'s `Stage::Filter` arm never threaded those flags into the filter stage the way the Simplex/Duplex arms do, so the chain builder's `filter.validate_parameters()` saw `MethylationMode::Disabled` and rejected a legitimately-set `--methylation-mode` — a fused `… → filter` chain with `--methylation-mode em-seq` failed with "requires --methylation-mode to be set" even though it was set.

## What changed

- The `Stage::Filter` arm now injects the resolved `methylation_mode` and, when methylation is requested, the top-level reference into the projected filter options before validation (mirrors the Simplex/Duplex arms). A user-supplied `--filter::ref` is preserved when no top-level `--ref` is given.
- Filter consumes `--methylation-mode` only through `--filter::min-conversion-fraction` (the one filter option whose check reads the resolved mode; `--require-strand-methylation-agreement` and `--min-methylation-depth` use the reference/methylation tags but not the mode). So the up-front dead-flag guard treats a filter chain as a methylation consumer only when that option is set — a bare `--methylation-mode` on a filter chain stays rejected as inert rather than being silently accepted as a no-op.
- Renamed the methylation-gated reference helper `consensus_reference` → `methylation_reference`, now that the filter stage uses it too.

## Tests

- Unit tests for the option-bag injection, the `--filter::ref` fallback, and both directions of the dead-flag guard (live when `--filter::min-conversion-fraction` is set; rejected-as-inert otherwise).
- End-to-end fused `consensus → filter` parity test that runs a methylation filter through `runall` and asserts record + header equivalence against the staged standalone `simplex` + `filter` chain.

Risk: grouping output: none; consensus output: yes for methylation-aware fused `runall`, pinned by staged-chain parity tests; sort order: none; corrected UMI output: none; metrics output: none. `unsafe`: none; CLAUDE.md allowlist: unchanged. Memory bounds, queue capacity, and thread/backpressure policy: none.

- Threads top-level `--methylation-mode` and `--ref` into fused filter stages.
- Preserves `--filter::ref` when no top-level reference is set.
- Rejects unused methylation options.
- Renames `consensus_reference` to `methylation_reference`.
- Adds unit and integration coverage for option wiring and fused/staged parity.

### DIFF
--- a/src/lib/commands/runall.rs
+++ b/src/lib/commands/runall.rs
@@ -937,17 +937,18 @@ impl RunAll {
         })
     }
 
-    /// Returns the reference FASTA to thread into the consensus caller's
-    /// `#[arg(skip)]` `reference` slot, gated on `--methylation-mode` being
-    /// set: `Some(self.reference.clone())` when methylation mode is
-    /// requested, `None` otherwise (even if `--ref` was supplied — e.g. to
-    /// feed an upstream align stage in the same chain).
+    /// Returns the reference FASTA to thread into a methylation-consuming
+    /// stage's `reference` slot, gated on `--methylation-mode` being set:
+    /// `Some(self.reference.clone())` when methylation mode is requested,
+    /// `None` otherwise (even if `--ref` was supplied — e.g. to feed an
+    /// upstream align stage in the same chain).
     ///
     /// Called by [`Self::build_stage_options_bag`] to populate the
-    /// `#[arg(skip)]` `reference` slot on the Simplex/Duplex consensus
-    /// options structs.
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #945, opened 2026-09-09
fix(sort): restore the === Sort Phase Timing === per-phase diagnostic

nh13:
## Summary

Restores the `=== Sort Phase Timing ===` per-phase diagnostic for `fgumi sort`, which the sort-on-chain cutover (#885) dropped. Pre-cutover, the owned `RawExternalSorter` engine logged a per-phase wall-time breakdown (read+decompress / in-memory sort / spill write / consolidation / k-way merge / write output) on every sort — the tool `CLAUDE.md`'s Benchmarking Notes call "the primary attribution tool". Standalone `fgumi sort` now runs entirely through the declarative chain and never calls that engine, so the block was gone (the owned timer is now reachable only from tests). `--sort-stats` was repurposed to gate the k-way-merge scheduling counters, which are a different signal.

The chain still *models* every phase — each is a discrete pipeline step (`ReadBlocks`/`InflateToArena` → `FindBoundariesAndSort`/`SortBuffer` → `SpillGather`/`SpillBlockCompress`/`SpillWrite` → `SortSpillDecompress` → `SortMerge` → `BgzfCompress`/`WriteBgzfFile`) — and the runtime already records each step's cumulative busy time. So rather than resurrect the retired engine, this re-derives the same breakdown as a roll-up over the end-of-run pipeline stats snapshot.

## What changed

- New `SortPhaseTimingFinalizeHook` (`src/lib/pipeline/chains/commands/sort.rs`): maps each sort step name to its phase, sums the per-step `total_run_ns` into six phase buckets, and logs the `=== Sort Phase Timing ===` block (per-phase seconds + %, a total line, and a one-line note that the percentages are shares of cumulative per-step *busy* time, since the steps run concurrently). Reads `snapshot.steps` only — never the `detached` view, which would double-count the detached writer, whose time `record()` already folds into its step's `total_run_ns`.
- `ChainBuilder::build` registers the hook only for a **sole-`[Stage::Sort]`** chain (`is_sort_terminal()`), matching how `add_sort` scopes `SortSummaryFinalizeHook`, and only when a `PipelineStats` collector is already attached and stats output was requested (`FGUMI_PIPELINE_STATS=1`; standalone `fgumi sort` does not accept `--pipeline-stats`). The default sort path attaches no collector, so it pays nothing — preserving the sort engine's zero-overhead hot path (which the external `fg-labs/mako` consumer depends on). The block is deliberately not emitted for a fused/intermediate sort (e.g. `runall`'s sort→group→…): there the terminal `BgzfCompress`/`WriteBgzfFile` sink serializes the final stage's output, not the sort's, and the intermediate `DecodeFromRecords`/`ParseBamRecords` adapters around the sort are unclassified, so a phase block there would misattribute the shared sink and drop adapter time — a faithful fused breakdown needs step-index scoping, out of scope here.
- `CLAUDE.md` Benchmarking Notes §2 updated: it described the now-dead owned-engine timer as always-on ("just `RUST_LOG=info`"). It now documents the chain-derived roll-up and its `FGUMI_PIPELINE_STATS=1` + `RUST_LOG=info` invocation, and notes the library `fgumi-sort` engine still carries its own always-on `SortPhaseTimer` for callers that drive `RawExternalSorter` directly.

Enabling it:

```sh
RUST_LOG=info FGUMI_PIPELINE_STATS=1 fgumi sort -i in.bam -o out.bam --order coordinate
# === Sort Phase Timing ===
#   read + decompress      0.001s   18.3%
#   in-memory sort         0.001s   11.3%
#   spill write            0.000s    3.8%
#   consolidation          0.000s    3.7%
#   k-way merge            0.001s    9.8%
#   write output           0.003s   53.1%
#   total                  0.005s  100.0%
#   (cumulative per-step busy time; sort steps run concurrently, so each percentage is a share of total sort work, not of wall-clock time)
```

The block is emitted just above the full `Pipeline end-of-run stats:` per-step table it rolls up from.

## Tests

- Unit tests (`commands/sort.rs`): phase bucketing across all six phases; the `SortBuffer` ingest branch; non-sort steps (fused-chain adapters) ignored; reads `steps[].total_run_ns` not `detached[].busy_ns` (no double count); the rendered block's header/phases/percentages/total; `None`/silent when no phase did work; an `#[rstest]` case table pinning every recognized sort step name (incl. the currently-unwired `CompressSpill`) to its phase.
- Integration tests (`test_sort_stats_diagnostics.rs`): a real spilling `fgumi sort` under `FGUMI_PIPELINE_STATS=1` emits the block and its dominant phases (in-memory sort, k-way merge, write output) report a **nonzero** parsed percentage (so a mis-bucketed classifier arm that zeroes a phase fails, not just a missing label); the block is absent on the default path; `--sort-stats` alone (its merge-diagnostic flag) still fires but does not emit the phase block — pinning the two diagnostics as independently gated.

Risk: adds pinned diagnostic output for standalone `fgumi sort` when `FGUMI_PIPELINE_STATS=1`; no `unsafe` changes or CLAUDE.md allowlist changes; no memory-bound, queue-capacity, or thread/backpressure changes.

Restores `=== Sort Phase Timing ===` with six phase durations, percentages, total time, and cumulative busy-time guidance. Fused chains, default sort behavior, and `--sort-stats` remain unchanged.

Adds unit and integration coverage for phase attribution, output gating, formatting, empty statistics, and spilling.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -416,18 +416,48 @@ throughout" from "parallel phase followed by a serial tail". Note that `io_in`
 collapses to near zero once the input is in page cache — a low `io_in` means the
 run was CPU-bound *on that invocation*, not that the tool does little I/O.
 
-**2. In-tree phase timers — the primary attribution tool.** `fgumi sort` already
-carries `SortPhaseTimer` (`crates/fgumi-sort/src/external.rs`): per-phase `f64`
-accumulators, a `time()` helper wrapping each span, and a `log_summary()` that
-emits `=== Sort Phase Timing ===` at `info!` (read+decompress / in-memory sort /
-spill write / consolidation / k-way merge / write output, each with a percentage).
-Prior campaigns got full baseline phase attribution from this with **zero new
-code** — just `RUST_LOG=info`.
+**2. In-tree phase timers — the primary attribution tool.** `fgumi sort` reports
+a `=== Sort Phase Timing ===` per-phase cumulative busy-time breakdown
+(read+decompress / in-memory sort / spill write / consolidation / k-way merge /
+write output, each with a percentage). Percentages are cumulative per-step
+busy-time shares — steps may run concurrently — not wall-clock proportions.
+Standalone `fgumi sort` runs on the declarative pipeline chain, where each phase
+is a discrete step and the runtime already records every step's cumulative busy
+time; the breakdown is re-derived from that end-of-run stats snapshot
+(`SortPhaseTimingFinalizeHook`,
+`src/lib/pipeline/chains/commands/sort.rs`). Enable it with **`RUST_LOG=info` plus
+`FGUMI_PIPELINE_STATS=1`** — standalone `fgumi sort` does not accept
+`--pipeline-stats` (that flag exists only on commands that flatten it, e.g. the
+runall front-end, which honor it too). The phase-timing *report* is opt-in via
+`FGUMI_PIPELINE_STATS=1`; it does not itself add a collector, because the
+standalone sort path already attaches a `PipelineStats` collector for its
+default deadlock monitor (`deadlock_timeout: 10`). Other features may collect
+statistics regardless, so the report is what is gated, not stats collection:
 
-Mirror that pattern when optimizing any other command: the phases you would name
-in a design doc are exactly the phases worth timing, it works identically on
-macOS and Linux, and it cannot be defeated by inlining or missing symbols the way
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #948, opened 2026-09-09
feat(metrics): compute consensus QC metrics inline in fused runall and standalone consensus

nh13:
## What & why

Consensus QC metrics (CS/SS family sizes, downsampling yield curves, UMI counts) currently require a separate `fgumi simplex-metrics` / `duplex-metrics` pass — a second full read of the grouped BAM — and the fused `runall` pipeline suppressed group metrics entirely in fused mode. This PR computes those metrics **inline during the single streaming pass**, matching the separate-pass tools' numeric output exactly, with **zero throughput or memory overhead when metrics are off**.

## User-facing changes

- **Standalone consensus:** new optional `fgumi simplex --metrics <prefix>` (and `duplex`, `codec`) — the same numbers as running `simplex-metrics`/`duplex-metrics` afterward, with no second BAM read. `--intervals` restricts which templates contribute. (The one file the separate `duplex-metrics` can produce that the inline path does not is `<prefix>.duplex_umi_counts.txt`, gated behind that tool's own `--duplex-umi-counts`; it has no inline equivalent.)
- **Fused pipeline:** new `runall --simplex::metrics` / `--duplex::metrics` / `--codec::metrics`, plus a convenience `runall --all-metrics <PREFIX>` that fills in every applicable metrics output (group + correct + filter + the active consensus stage) for any option not set explicitly.
- **Group metrics in fused mode:** `runall --group::*` metrics now flow through instead of being nulled.

All flags are opt-in; default behavior and output are unchanged.

## Design

Two tap points, one rule (does the metric's aggregation key span more than one parallel-dispatch unit?):

- **T1 — fused runall:** metrics accumulate in the serial per-position group-closure that already exists in the group stage (`PerThreadAccumulator` + a finalize hook). Zero cross-batch risk.
- **T2 — standalone consensus:** a dedicated `Serial` collector step consumes an ordered coordinate-group fragment stream (third output branch + reorder stage), so live memory is bounded to *O(one open coordinate group)* rather than the number of simultaneously-open groups.

Both paths share the same reducer functions, so the metric math cannot drift between them. **Zero-overhead-when-off is structural, not a runtime guard:** the metrics step is monomorphized out of the chain when no metrics flag is set (a chain-shape test asserts this on the built DAG).

## Correctness

- Numeric parity anchor: inline (T1 fused + T2 standalone) == separate-pass across {simplex, duplex, codec} × family shapes × {1, 2, 8} threads × BED/Picard intervals, plus a multi-batch/multi-worker boundary case. The simplex fixtures use real paired templates and a non-vacuity guard so the assertions cannot pass on empty output.
- Consensus output is **byte-identical** to the pre-metrics path (metrics never perturb calling).
- Full workspace test suite green; `-D warnings -W clippy::pedantic` clean; no `unsafe`; no committed fixtures.

## Performance

Benchmarked on a 28-core workstation, 16 threads, baseline = merge-base:

- **Metrics off (default path):** indistinguishable from baseline — wall −0.12% on an 89M-read group→consensus run, +0.38% on a 678M-read consensus-only run; CPU within ±0.15%; peak RSS equal-to-lower. Output byte-identical.
- **Consensus metrics on:** ≈ +12% wall/CPU on the pass, **flat memory** (the bounded-collector design holds). Obtaining the same metrics the old way costs a full single-threaded second pass over the BAM.

## Suggested reading order

1. `crates/fgumi-pipeline-core/src/metrics_collector.rs` — the generic `Serial` collector step.
2. `src/lib/inline_metrics_collector.rs` — the accumulator, fragment/collector, and the T1 adapter.
3. `src/lib/commands/shared_metrics.rs` — the shared reducer functions (the DRY anchor).
4. `src/lib/pipeline/chains/builder.rs` + `chains/commands/{group,simplex,duplex,codec}.rs` — the T1/T2 wiring and monomorphized step selection.
5. `src/lib/commands/runall.rs` — un-nulling group metrics + the `--all-metrics` aggregator.

Risk: Metrics output changes when enabled, pinned by shared reducers and byte-parity tests; grouping, consensus, sort order, and corrected UMIs: none. Unsafe changes: none; CLAUDE.md allowlist update: none. Memory bounds, queue capacity, and thread/backpressure policy: none.

Fix: Add opt-in inline metrics to standalone and fused consensus pipelines without extra BAM reads or pipeline stages.


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #949, opened 2026-09-09
test(metrics): close --all-metrics and group-metrics integration coverage gaps

nh13:
Stacked on #948 (the inline consensus metrics feature) — **base this PR on #948; review/merge that first.** Test-only, no production changes.

Adds integration coverage for metric-output paths that #948's tests exercised only for file creation, or only in the negative (stage-absent) direction:

- `--all-metrics` output **content** matches setting the equivalent per-stage flags explicitly — group's three files and simplex's three (including `simplex_yield_metrics`), byte-for-byte.
- `--all-metrics` fills in and writes the **correct** stage's metrics when a correct stage is present, content-matching an explicit `--correct::metrics`.
- `--all-metrics` fills in and writes the **filter** stage's stats when a filter stage is present, content-matching an explicit `--filter::stats`.
- Fused `--group::metrics` produces the same three group files as standalone `fgumi group --metrics` (previously only `--family-size-histogram` was content-compared).

The `--all-metrics` comparisons run two invocations of the same binary, so the fixtures use real paired templates and a data-row guard on the family-size files — the assertions cannot pass on empty output. Correct and filter are covered as separate single-stage chains (having both in one chain would pull in the align stage, which needs an external aligner and reference).

Risk: yes, metrics output handling changes; no grouping, consensus, sort-order, or corrected-UMI output changes; centralized derived paths and byte-for-byte parity tests pin metrics output; no `unsafe` changes and no CLAUDE.md allowlist update; no memory-bound, queue-capacity, or thread/backpressure changes.

Adds integration tests for `--all-metrics`, per-stage metrics, fused group metrics, and standalone `group --metrics`. Tests require non-empty data rows.

Adds collision checks for derived metrics paths in `simplex`, `duplex`, and `codec`.

### DIFF
--- a/src/lib/commands/codec.rs
+++ b/src/lib/commands/codec.rs
@@ -549,6 +549,18 @@ impl Command for Codec {
         // Validate the input exists (stdin paths are exempt — the reader
         // streams them in a single pass).
         self.io.validate()?;
+        // `--metrics <prefix>` derives the duplex metrics artifact set (codec
+        // reuses the duplex inline writer), written by
+        // `ConsensusMetricsFinalizeHook` AFTER the BAM sink. Those derived paths
+        // must join the collision check, else `--metrics result` + `-o
+        // result.family_sizes.txt` passes here and the metrics write clobbers the
+        // BAM. Reuses `duplex_metrics_paths` (the superset runall's collision
+        // guard also registers) so the two never drift. Kept in a local so
+        // `outputs` can borrow it.
+        let metrics_artifacts: Vec<std::path::PathBuf> = match &self.metrics {
+            Some(prefix) => crate::inline_metrics_collector::duplex_metrics_paths(prefix),
+            None => Vec::new(),
+        };
         let mut outputs: Vec<(&Path, &str)> = vec![(self.io.output.as_path(), "--output")];
         if let Some(path) = &self.rejects_opts.rejects {
             outputs.push((path.as_path(), "--rejects"));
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed the outside-diff finding on \`simplex.rs\` (reject collisions with all \`--metrics\` artifacts): the \`simplex\`/\`duplex\`/\`codec\` pre-flights now derive the consensus-metrics artifact paths (\`<prefix>.family_sizes.txt\`, \`<prefix>.umi_counts.txt\`, \`<prefix>.simplex_yield_metrics.txt\` for simplex; the duplex set for duplex/codec) and feed them to \`reject_output_collisions\` alongside \`--output\`/\`--rejects\`/\`--stats\`, so a \`--metrics\` prefix that would clobber another output is rejected before \`ConsensusMetricsFinalizeHook\` writes anything. The suffixes now come from a single source of truth (\`METRICS_SUFFIX_*\` in \`inline_metrics_collector.rs\`) shared with the writers, and regression tests cover all three commands plus the derivation.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai full review

nh13:
@coderabbitai review

PR #950, opened 2026-09-09
test: consolidate the byte-identity file-compare idiom into a shared helper

nh13:
Pure test-infra hygiene. The cutover-parity tests each reimplemented the byte-identity file-compare idiom `assert_eq!(read_to_string(a), read_to_string(b))` inline — 7 sites across 5 files (dedup, group, clip, retag, correct). This consolidates them into a single `assert_text_files_eq(actual, expected, label)` in the shared `tests/integration/helpers/assertions.rs`, with unit tests for the identical / divergent / missing-file cases.

Behavior-preserving: same files, same `actual`/`expected` ordering, same per-site labels — the existing cutover assertions are unchanged and the full suite stays green. The helper deliberately does **not** guard for non-empty content: whether an output file must carry data rows is specific to the metric being compared, so any such check stays with its caller.

Risk: command output changes: none; `unsafe` changes: none, and `CLAUDE.md` allowlist changes: none; memory bounds, queue capacity, and thread/backpressure policy changes: none.

Added `assert_text_files_eq` for byte-identical text-file assertions. Added tests for identical, divergent, and missing files. Replaced seven inline comparisons in cutover-parity tests while preserving file order, labels, assertion behavior, and caller-specific non-empty checks.

### DIFF
--- a/tests/integration/helpers/assertions.rs
+++ b/tests/integration/helpers/assertions.rs
@@ -299,11 +299,62 @@ pub fn assert_bam_sorted(bam: &std::path::Path, order: &str, key_types: Option<&
     );
 }
 
+/// Asserts two text files exist and are byte-for-byte identical, panicking
+/// with `label` and the offending path named on either a read error or a
+/// content mismatch.
+///
+/// This is the shared home for the `assert_eq!(read_to_string(a),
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #955, opened 2026-09-13
feat(pipeline): park idle pool workers to fix the thread-oversubscription cliff

nh13:
## Summary

Fixes the thread-oversubscription **cliff** in the chain-builder pool: running a command with more worker threads than it has parallel work no longer *slows it down*.

The pool is replicated round-robin polling with no producer→consumer wakeup — an idle worker exponential-backoff *sleeps* then re-polls. When there are more workers than parallel work, the surplus churn poll→sleep→poll, and that churn actively slows the productive workers. On `fgumi correct` at 32 threads this was a hard regression: **~6.5s wall vs ~1.9s at 16 threads** (pool 92% idle, 208s of aggregate idle churn), because a single serial grouping stage feeds the parallel work and nothing else can run.

## Result

`fgumi correct`, 10M templates, c8g.8xlarge (32 vCPU):

| threads | before | after |
|---------|--------|-------|
| 16 | 2.03s | 1.90s |
| 32 | **6.48s** (1.9× vs 1 thread) | **1.91s** (6.4×) |

The cliff is gone — throughput is now flat past the useful-parallelism point instead of collapsing. Aggregate idle dropped 208s → 35s (real parking, not spin-churn). Output is unchanged.

## Approach (two layers)

**L0 — remove the contention, take the cheap wins**
- Test-and-test-and-set on the `Serial` step mutex: `parking_lot::try_lock` writes (invalidates) the holder's cache line even on failure, so a surplus worker probing a held ceiling step stole that line every idle pass. Probe `is_locked()` (a shared load) first.
- Raise the pool backoff floor 1µs → 20µs so a stray-success reset doesn't drop a surplus worker onto a per-microsecond re-poll ramp.

**L2 — park the surplus (the real fix)**
- `PoolEventCount`: a Vyukov event-count over `parking_lot` Mutex/Condvar. Idle workers arm (register + SeqCst fence + snapshot), re-poll once, then block only if still empty; producers publish work then `notify_one`. When nobody is parked, `notify_one` is one fence + one read-shared load — no store, no lock, no wake. Correctness is the C++20 SeqCst-fence rule (arm fence and notify fence are totally ordered → no interleaving parks on a poppable item).
- Notify seam is the driver's dispatch outcome: `Progress` → `notify_one`, `Finished` → `notify_all`. Using the outcome (not the queue push site) captures the reorder must-accept stash without threading the parker through every queue.
- Concurrency ceiling gates `notify_one` on `awake = n_pool − waiters < ceiling` so the per-`Progress` cascade can't re-wake the whole surplus. Default = ungated; a controller can lower it later. `notify_all` (cancel/close/`Finished`) is never gated.
- **Pinned workers** (Exclusive/sticky owners, Serial affinity targets) keep the old sleep-backoff — a shared `notify_one` can't target a specific worker. **Detached drivers** notify but don't wait. **Cancel/error** wake all parked workers via a late-bound `Weak<PoolEventCount>` on `PipelineSignal`. Only wired when `n_threads > 1`; single-thread/fused paths are behaviour-identical.

**L1 (per-command)** — pin the `correct` chain's serial `GroupByQueryname` to one worker (optional per-instance affinity, default unchanged for every other chain) so surplus workers `Skip` it instead of thrashing its mutex.

## Testing

- Full workspace suite green (9988 tests). New `PoolEventCount` unit tests incl. a no-lock-when-idle proof, a moved-generation test, and a 2000-iteration lost-wakeup stress; a TTAS test proving a held Serial step reports `Contention` without running.
- Benchmarked on 32-vCPU Graviton: `correct` cliff gone; `extract` K=2 (push-heavy — the always-on notify canary) flat 16→32; `sort` (exercises the detached-driver notifier + pinned paths) scales cleanly with no regression.

## Notes

- Stacked on #951 (thread/bandwidth telemetry); base branch `nh/pipeline-thread-telemetry`.
- This stops the *cliff*. It does not raise `correct`'s ceiling — a single-threaded `GroupByQueryname` (~1.5s, runs end-to-end) still caps scaling past ~16 threads. Parallelizing that stage is separate follow-up work.
- The concurrency-ceiling **controller** (feedback from edge depth) is deferred: the ungated ceiling was sufficient in benchmarks (the per-`Progress` cascade churn did not hurt).

Risk: output change none; unsafe change none, and no CLAUDE.md allowlist update is required; memory-bound and queue-capacity changes none, but worker backpressure changes through parking and wakeup limits.

Fix: Park idle workers, reduce serial-step lock contention, and pin `GroupByQueryname` in the `correct` chain.

- Add `PoolEventCount` for coordinated worker parking and wakeups.
- Handle progress, completion, cancellation, errors, pinned workers, and detached drivers.
- Increase the minimum backoff and add test-and-test-and-set locking.
- Preserve output and pass 9,988 workspace tests.
- Improve reported 32-thread `fgumi correct` runtime from 6.48s to 1.91s.


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #952, opened 2026-09-12
feat(pipeline): add K-way parallel FASTQ decoding with StepK

nh13:
> **Stacked on #951** (`nh/pipeline-thread-telemetry`). Review/merge that first; this PR's diff is only the K-way FASTQ decode work on top of it.

## Summary

Adds **K-way parallel FASTQ decode** to the declarative chain engine via a new homogeneous K-input `StepK` abstraction, replacing the old 2-input FASTQ zip.

### `StepK` engine (`crates/fgumi-pipeline-core`)
- New **K-input `StepK`** trait alongside `Step`/`Step2`: `StepCtxK`, `KInputHandles`, `append_step_k`, and the `TypedStepK` erased adapter (its `unsafe` cached-handle dispatch mirrors `TypedStep2` exactly and is documented in the CLAUDE.md allowlist).

### Parallel BGZF FASTQ decode front (`src/lib/pipeline/steps/source`)
- Per stream: `ReadFastqBlocks` (raw block read) → `FastqDecompress` (pooled inflate) → `FindFastqBoundaries` (4-line record-seam framing) feeding a unified K-stream lockstep join `ZipRawFastqK`/`WrapRawFastq1` → `ParseAndZipFastqN`.
- Replaces the old 2-input `pair_fastq.rs`/`zip_fastq.rs`/`parse_fastq.rs` (deleted); the record-level zip is lifted verbatim into `zip_streams` — byte-identical positional join with count/name-concordance checks (chain-level parity test asserts it against the fused reader).
- `--interleaved` now uses the same unified path; R1/R2 divergence is bounded by per-stream output-queue backpressure (`per_step_byte_limit`), well under the de-interleaver's `MAX_PENDING_RECORDS` cap.

### Review fixes (CodeRabbit-style + gauntlet, applied in this session)
- `--async-reader` now honored on the BGZF split path (was a silent no-op for bgzip inputs).
- `KInputHandles` made opaque over a single `Vec` (closes a `>= 2`-invariant bypass; drops dead inline-array complexity).
- FASTQ decode errors name the failing stream (`R{n}`; K-way desync names all present streams).
- Record-identity assertion added to the 3-input BGZF-split integration test; out-of-sync **direction** pinned in `zip_streams`; dead tag-literal allowlist entries trimmed.

## Reading order
1. `crates/fgumi-pipeline-core/src/{step.rs, builder.rs, erased.rs, handles.rs}` — the `StepK` abstraction + typed adapter + `KInputHandles`.
[...]

PR #956, opened 2026-09-13
fix(bam-io): return an error instead of panicking on >65,535 libraries

nh13:
LibraryIndex::from_header assigned each distinct @RG library a sequential u16 index via `names.len().try_into().expect(...)`, which panics once a header carries more than 65,535 distinct libraries. Every consensus-metrics, grouping, and dedup path that builds a LibraryIndex from an untrusted input header could therefore crash the process rather than surface a diagnosable error. CodeRabbit flagged this on #952 as a pre-existing, out-of-scope issue; this PR addresses it against `main`.

## What changed

- `LibraryIndex::from_header` is now fallible (`-> anyhow::Result<Self>`): the vacant-entry arm maps the `u16` overflow to an error naming the distinct-library limit, replacing the panicking `.expect(...)`. The `# Panics` doc becomes `# Errors`.
- `Result` is threaded through the three builder seams that construct a `LibraryIndex` from a header — `bam_group_key_config`, `source_group_key_config`, and `finish_consensus_tail` — plus the direct group/simplex/duplex/codec/dedup metrics sites and `process_templates_from_bam`.
- Test-only construction (tiny fixtures that cannot overflow) uses `.expect`.

## Tests

- New regression test: a header with `u16::MAX + 1` distinct libraries errors rather than panics.
- New boundary test: exactly `u16::MAX` distinct libraries still builds.
- Full suite green (4621 passed), `cargo fmt --check` and `clippy --all-features --all-targets` clean.

Risk: command output changes: none; `unsafe` changes: none, and `CLAUDE.md` allowlist changes: none; memory bounds, queue capacity, and thread/backpressure policy changes: none.

`LibraryIndex::from_header` now returns an error instead of panicking when a header exceeds 65,535 distinct libraries. The error propagates through grouping, consensus, codec, deduplication, and metrics paths.

Tests cover overflow and the supported 65,535-library boundary.

### DIFF
--- a/crates/fgumi-bam-io/src/grouping.rs
+++ b/crates/fgumi-bam-io/src/grouping.rs
@@ -648,7 +648,7 @@ mod tests {
                 .expect("read group builds");
             header = header.add_read_group(bstr::BString::from(id), rg);
         }
-        LibraryIndex::from_header(&header.build())
+        LibraryIndex::from_header(&header.build()).expect("builds")
     }
 
     /// Build one mapped mate of a pair carrying `RG`, `CB` and `MC`.
--- a/crates/fgumi-bam-io/src/library.rs
+++ b/crates/fgumi-bam-io/src/library.rs
@@ -99,11 +99,13 @@ impl LibraryIndex {
     /// Each unique library name gets a sequential index starting from 0.
     /// Index 0 is reserved for "unknown" library.
     ///
-    /// # Panics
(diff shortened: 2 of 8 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #958, opened 2026-09-13
perf(correct): resolve the template UMI from the primary read

nh13:
## Summary

Resolve a template's UMI in `fgumi correct` from the **primary read** — the way `group` and `dedup` already do — instead of scanning every record's aux data and cross-checking that they all agree.

> Stacked on #957 (`nh/parallel-queryname-grouping`) — review/merge that first. This PR targets that branch; the diff shown here is this PR's one commit.

## What changed

`correct` previously resolved the UMI via `extract_and_validate_template_umi_raw`: it scanned the UMI tag on every record in the template and errored if they disagreed. `group` and `dedup` instead read the UMI from the primary read alone (`Template::cached_umi()` when the decode-time cache is present, else one scan of the primary read's aux), because all reads of a well-formed template share one UMI by construction (`fgumi extract` sets it identically on both mates).

The new `extract_template_umi` makes `correct` consistent with them, dropping the per-record cross-check.

## Behavior change

`correct` no longer errors on a template whose reads carry **mismatched** UMIs — it uses the primary read's UMI, exactly as `group`/`dedup` already do.

- **Well-formed input is byte-identical** — the primary read's UMI equals the value the old cross-check returned.
- A non-`Z`-typed RX on the **primary** read is still rejected.
- fgbio's `CorrectUmis` has never validated per-template UMI consistency either: it corrects each read independently and explicitly supports multiple UMIs per template. The removed check was a **fgumi-only invention** (present since fgumi's initial commit `bc001456`) with no fgbio precedent — so dropping it moves `correct` toward fgbio's behavior, not away from it.

Tests updated to pin the trust-primary semantics (mismatched or non-primary non-`Z` UMIs resolve to the primary rather than erroring).

## Benchmark (c8g.8xlarge, 32 vCPU Graviton4, 20M-record RX input, vs base `0ba1fc6e`)

Content-identical output (`fgumi compare bams --command correct`: 14,030,750 records, **0 diffs**), and neutral-to-slightly-faster at every thread count (min of 3 reps):

| threads | base | this PR |
|--------:|-----:|-----:|
| 1  | 20.33s | 20.27s |
| 8  | 3.48s | 3.15s |
| 16 | 2.15s | 2.06s |
| 32 | 1.58s | 1.46s |

The win is small — the UMI lookup is a minor part of `correct` (whitelist edit-distance correction dominates) — so this lands primarily as a **consistency + simplification** change (correct now matches group/dedup and fgbio), with a modest throughput bonus and no regression.

## Note on scope (why this is smaller than it started)

This PR originally also preserved the decode-time UMI cache through the queryname groupers so `correct` could skip the aux scan entirely. Benchmarking showed that part was **net-negative**: preserving the cache double-decoded record flags (`primary_read_cached_umi` pre-pass + the constructor's own categorization) and tripled the per-record buffer size (`DecodedRecord` vs `RawRecord`) in the grouping hot loop, costing more than the single aux scan it saved. That commit was dropped; only the `correct`-side simplification — which the benchmark shows is a genuine small win — remains.

## Verification

`cargo ci-fmt`, `cargo ci-lint` (pedantic, `-D warnings`), `cargo ci-test` (**10202 passed, 31 skipped**) all green; output content-identical to baseline as above.

Risk: corrected UMIs change only for mismatched templates, using the primary read; grouping, consensus, sort order, and metrics: none; unsafe: none, with no CLAUDE.md allowlist change; memory bounds, queue capacity, and thread/backpressure policy: none.

Fix: `fgumi correct` now uses the assembled template’s primary-read UMI and ignores secondary-read mismatches and type errors. Non-`Z` primary UMIs remain rejected. Tests cover the updated behavior and preserve output for well-formed input.


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #959, opened 2026-09-13
fix(extract): honor --check-crc/--no-check-crc on the BGZF FASTQ split decode

nh13:
## Summary

The all-file BGZF parallel-decode split for FASTQ input (`build_bgzf_fastq_split` → `FastqDecompress`) reads its per-block CRC32 policy from `ChainSpec.verify_crc`, but `Extract::execute_chain` hardcoded that field to `false`. As a result `--check-crc` and the file-input default silently skipped CRC verification on that path, accepting corrupted FASTQ blocks. This path shipped to `main` with the StepK FASTQ decode work (via #952, merged into #951).

CodeRabbit flagged this on #951 (`builder.rs:889`), but the thread was resolved as "stale / symbols removed" — the symbols do exist on `main`, and the bug is real.

## Fix

Resolve `verify_crc` through the shared `resolve_check_crc` policy (`--check-crc` wins, `--no-check-crc` disables, file inputs verify by default). The spec construction is extracted into a testable `Extract::build_extract_chain_spec`, mirroring `Sort::build_sort_chain_spec`.

## Why the existing test missed it

`test_bgzf_fastq_honors_check_crc` uses a small input that is read in full during quality-encoding detection (`sample_detection_quals`), which opens its own policy-honoring reader and rejects the corrupted trailing block before the split decoder ever runs. The new `split_decoder_honors_check_crc_past_detection_window` places the corruption past the detection window so only the full split decode reaches it, exercising the decoder's own policy (verified failing before the fix, passing after).

## Also

Aligns the `parse_and_zip_two_streams_preserves_ordinal` doc with what it actually covers (a second CodeRabbit finding on #951 resolved as stale — also still valid on `main`): it exercises the parse/zip primitives, not the step; step wiring is covered by the chain tests.

## Verification

`cargo ci-fmt`, `cargo ci-lint`, `cargo ci-tag-literals`, `cargo ci-publish-order`, and the affected tests all pass locally.

Risk: command output changes: none; `unsafe` changes: none, so no CLAUDE.md allowlist update is required; memory bounds, queue capacity, and thread/backpressure policy: none.

Fix: Resolve `ChainSpec.verify_crc` through `resolve_check_crc`. `--check-crc` enables verification, `--no-check-crc` disables it, and file inputs verify by default.

- Extracted chain-spec construction into `Extract::build_extract_chain_spec`.
- Added tests for CRC policy resolution and corruption beyond the quality-detection window.
- Updated the `parse_and_zip_two_streams_preserves_ordinal` documentation.


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #961, opened 2026-09-13
perf(filter): decode-free BAM fast path for single-record filter

nh13:
First of three stacked PRs extending the decode-free path (from the `fgumi fastq` no-key work) to the per-record-transform stages. This one converts **single-record `fgumi filter`** and lands the shared seam the later `copy_umi`/`retag` PRs reuse.

## What and why

A single-record filter (not `--filter-by-template`) reads only the raw record — it never touches a `GroupKey` — yet its source decode was building an owned `Vec<DecodedRecord>` (one heap allocation per record) and computing a `name_hash` that is immediately discarded. On a BAM source this now takes the borrowed `RecordBatch` fast path (`ParseBamRecords`) instead of `DecodeRecords`, eliminating the per-record heap allocation and the dead key computation. The per-record copy count is unchanged; the win is allocation-elimination (the `mi_*` malloc/free slice from the earlier profiling), so it is framed as a mechanism, not a benchmarked percentage — a representative benchmark is a follow-up.

Output is **byte-identical**: the discarded key never reached the encoder, and the per-record transform is the same `process_record_raw_call` on both paths.

## How

- **`for_each_raw_record`** (`commands/mod.rs`): iterates a `RecordBatch`'s borrowed record byte-ranges into one reused scratch `RawRecord` (cleared, not truncated, each record), calling `validate_record_for_decode` first — the fast path drops `DecodeRecords`, which otherwise ran that guard, so without it a truncated record would panic the unchecked field accessors instead of erroring cleanly.
- **`filter_one_record`** (`commands/filter.rs`): the single per-record body — transform, masked-bases tally, keep/reject routing — shared by all four single-record filter builders (decoded/raw × no-rejects/with-rejects), so the owned and borrowed paths cannot drift.
- **Builder** (`builder.rs`): `ChainTailKind::RecordBatch`, `first_stage_uses_raw_record_fast_path` (single-record filter only; `--filter-by-template` keeps the `DecodedRecordBatch` path its `GroupByQueryname` grouper needs, with a defensive bail if that pairing is ever reached), a `ParseBamRecords` BAM preamble, and `add_filter` selecting the raw builders on the `RecordBatch` tail. SAM source keeps the `DecodedRecordBatch` path unchanged.

## Scope

Out of scope (later PRs / deliberate): the SAM-source fast path, `--filter-by-template`, and `copy_umi`/`retag`.

## Tests

`cargo ci-test`: 10200 passed, 0 failed. New coverage: the `for_each_raw_record` seam (scratch reuse without stale tail; fail-closed on a truncated record; raw-vs-decoded framing/routing byte-parity under a length-changing mutation), the fast-path selection predicate, and a single-record `--rejects` chain-vs-single-worker parity test exercising the with-rejects fast path end-to-end. The existing filter integration + cutover-parity suites (single-record BAM, no-rejects) stay green.

Reviewed with a CodeRabbit-style pass and a multi-agent gauntlet (10 findings, all addressed): the doc-placement fix, the `filter_one_record` de-duplication, the scratch-reuse doc correction, and the with-rejects test.

Risk: output changes none; `unsafe` changes none and CLAUDE.md allowlist changes none; memory bounds, queue capacity, and thread/backpressure policy changes none. Fix: add a validated raw BAM fast path with scratch-buffer reuse while preserving decoded paths where required.

- Route eligible single-record BAM filters through borrowed `RecordBatch` data.
- Preserve decoded paths for SAM and template filtering.
- Share filtering, framing, metrics, and reject handling.
- Add tests for validation, scratch reuse, parity, routing, rejects, and multi-worker behavior.


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #964, opened 2026-09-14
perf(zipper): fold per-record PG/AS/XS aux scans into single-pass walks

nh13:
## Summary

Follow-up to #955-era `rebuild_with` work (stacked on #962). Profiling `fgumi zipper` at 1 thread — the realistic streaming setting (`bwa-mem3 | fgumi zipper | fgumi sort`), where zipper runs a single-thread tight loop — showed the merge is CPU-bound and serial, with `find_tag_position` still ~14% of CPU *after* #962 removed the per-tag copy-loop scans. That 14% is now dominated by fixed per-record single-tag lookups that scan the aux from offset 0:

- `find_tag_type(PG)` — the `has_pg` probe, once per mapped record. Aligner output carries no per-record `PG`, so this is a full aux scan that always fails.
- `normalize_int_tag_to_smallest_signed` for `AS` and `XS` — each a `find` scan plus a `remove` scan, four scans per record.

## What changed

**`perf(zipper)`** — two single-walk helpers replace those scans, each reusing one scratch buffer across records:
- `copy_unmapped_tags_single_pass` copies the mapped record's tags, resolves `PG` precedence (keep the mapped read's own `PG`, drop the unmapped one) in the same walk, and appends the copy set — no separate `has_pg` scan, and no per-record allocation.
- `normalize_as_xs_single_pass` normalizes `AS` then `XS` in one walk instead of four scans.

**`refactor(raw-bam)`** — `extract_int_value` bundled the integer-decode ladder with aux-slice positioning, so a caller holding a tag's value bytes (from an `AuxTagsIter`/`TagEntry` walk) couldn't reuse it. Split out `decode_int_value(type_byte, value_bytes)`; `extract_int_value` now slices and delegates. Removes the now single-use `tag_value_bytes`.

## Correctness: byte-identical

On a 10.9M-record CODEC dataset (extract → bwa-mem3 → zipper), output is byte-identical to the pre-change binary:
- strict `samtools view` md5 (fields **+ tags + order**): identical.
- `fgumi compare bams`: **0 content diffs across 10,941,868 records**.

All 146 zipper/merge/align tests pass; 806 `fgumi-raw-bam` tests pass. New tests cover `decode_int_value` (per type, non-integer types, short slices); the existing AS/XS-normalization and `has_pg` (`pg_already_on_mapped_read_blocks_the_unmapped_pg_copy`) tests exercise the new helpers.

## Performance

Standalone `fgumi zipper`, 1 thread, page-cache warm, 10.9M records:
- `find_tag_position` drops from **~14% → ~8%** of CPU (samply, `threadCPUDelta`-weighted).
- The command is faster in every measured rep; ~5–9% less CPU (the exact figure is noisy on an interactive laptop — this host's wall time is unreliable under load, so the number is CPU-seconds and directional). A quiet dedicated host would tighten it.

This is a CPU-reduction change, which is what matters for zipper's place in the pipeline: every CPU-second it doesn't burn goes to the aligner (upstream) and sort (downstream) sharing the box.

## Reading order

1. `crates/fgumi-raw-bam/src/tags.rs` — `decode_int_value` + `extract_int_value` delegation.
2. `src/lib/commands/zipper.rs` — `copy_unmapped_tags_single_pass`, `normalize_as_xs_single_pass`, and their wiring into `merge_raw_with`.

## Base

Stacked on #962 (`nh/raw-tags-rebuild`) — it builds on that PR's `merge_raw_with` structure. Retarget to `main` once #962 merges.

Risk: zipper auxiliary-tag output can change, pinned by byte-identical zipper/merge/align tests; unsafe changes: none, so no `CLAUDE.md` allowlist update is needed; memory, queue, and thread/backpressure policy changes: none.

- Optimize `PG`, `AS`, and `XS` handling with single-pass walks and reusable scratch buffers.
- Preserve `PG` precedence and duplicate-tag last-wins behavior.
- Add shared checked integer decoding through `decode_int_value`.
- Add tests for tag updates, normalization, ordering, malformed tails, and no-op cases.
- Report approximately 5–9% lower CPU usage.


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #966, opened 2026-09-14
perf: parallel gather for the in-memory sort fast path

nh13:
## What

`fgumi sort`'s in-memory fast path gathered the whole sorted chunk into output blocks **serially** on one detached coordination thread — a flat ~1.2 s cost that did not scale with `--threads` and floored the sort's wall clock while the worker pool sat mostly idle. This PR fans that gather across a bounded, step-owned rayon pool sized to the phase-2 (`merge`) thread budget.

Only the **fast path** (a single already-sorted in-memory chunk, no k-way merge) is touched. The spill / k-way-merge path is deliberately untouched — it is a genuinely serial loser-tree walk whose winner order cannot be produced out of order.

## How

1. **Plan block boundaries up front, off the cold arena.** New `record_len` accessors (`InMemoryChunk` / `TemplateMemChunk` / `MemoryChunkErased`) read only the per-record `len` index. `plan_fast_path_blocks` reproduces the serial count/byte-cap split byte-for-byte from lengths alone. A `MergeBatchBuilder::FRAME_OVERHEAD_PER_RECORD` const (`4` for the framed `BlockBuilder`, `0` for `RecordBatchBuilder`) keeps the byte-cap plan in lockstep with each builder's `total_bytes()` growth.
2. **Gather windows in parallel, emit in strict ascending dense ordinal order.** `SortMerge` is `Detached`, so its output edge has no reorder stage (the by-ordinal reassembly is downstream at `BgzfCompress` and needs a dense, gap-free stream). Gathered-but-unpushed blocks ride the existing held-slot backpressure idiom plus a small `fast_pending` queue, drained in order at the top of the next dispatch.
3. **Bounded + gated.** The window is `fast_path_threads * 4` blocks. Gated on `FAST_PATH_PARALLEL_MIN_RECORDS = 64Ki` so small chunks keep the cheaper serial gather; a degenerate single-block plan falls through to serial rather than spinning up a pool. The pool is step-owned and bounded to `num_phase2_threads` so it never oversubscribes past `--threads`.

## Results (EC2 c8g.8xlarge, 32 vCPU Graviton4, coordinate sort, 20 M records)

| threads | wall before | wall after | speedup |
|--------:|------------:|-----------:|--------:|
| 16      | 1.75 s      | 0.92 s     | ~1.9×   |
| 32      | 1.84 s      | 0.93 s     | ~2.0×   |

Total CPU-seconds is roughly flat — the win is converting serial gather time into parallel gather time. 16 threads is the throughput sweet spot (21.9 M rec/s).

## Correctness

Output is **byte-for-byte identical** to the serial gather. Verified with `fgumi compare bams --command sort` (record-content + order) across coordinate / template-coordinate / queryname at 16 threads, and against the spill regime (fast path not taken) — all IDENTICAL. Parity unit tests assert the parallel gather emits identical dense ordinals and framed bytes as the serial gather across count-cap, byte-cap, small-multi-block, partial-tail, and a deterministic **backpressure** regime (small output queue forcing sustained mid-window reject/resume) for **both** the `BlockOutput` and `RecordBatchOutput` builders.

## Gates

`cargo ci-fmt`, `cargo ci-lint`, `cargo ci-doc` clean; `cargo ci-test` — 10219 passed, 31 skipped.

Risk: sort output stays byte-identical through serial-equivalent block planning and ordered emission; unsafe code: none added, and the CLAUDE.md allowlist remains unchanged; memory and backpressure policy: bounded parallel gathering uses the merge-thread budget.

Fix: parallelize only large, multi-block in-memory gathers. Keep small inputs serial. Keep spill and k-way merge serial.

- Add `record_len` accessors for in-memory chunks.
- Add per-record framing overhead to `MergeBatchBuilder`.
- Add configurable fast-path thread limits.
- Pass phase-2 thread limits from pipeline builders.
- Add parity tests for bytes, framing, ordinals, boundaries, and backpressure.
- Reported tests and benchmarks pass, with approximately 1.9–2.0× speedup on the stated large coordinate-sort benchmark.


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #967, opened 2026-09-15
perf(extract): reduce per-record allocation (zero-copy segment views + reused Phred scratch)

nh13:
## Summary

`fgumi extract`'s per-record path allocated and copied more than it needed to: every FASTQ segment's bases/qualities were `to_vec()`-copied out of the source record into an owned `FastqSet`, the per-stream sets were merged (`combine_readsets`), and the Phred+33 conversion allocated a fresh `Vec` per record. This removes those allocations with a borrowed segment-view hot path and a reused conversion buffer. Output is byte-for-byte identical — this is a pure allocation-reduction change, no output bytes move.

Two commits:

1. **Zero-copy segment views** — new crate-internal `FastqSegmentView` / `FastqSetView` hold `&[u8]` spans into the source record instead of owned copies. `extract_batch` builds one `FastqSetView` directly from each source record's `sequence()`/`quality()` slices (segments appended R1-then-R2 in input order, combined header from the first record — matching `combine_readsets` semantics), and the shared record builder `make_raw_records_from_view` consumes it. The owned `FastqSet` / `make_raw_records_from_fastq_set` path is retained `#[cfg(test)]`-only and now routes through the same view core (`FastqSetView::from_owned`), so the owned and zero-copy paths share one record builder and cannot diverge.
2. **Reused Phred scratch buffer** — `to_standard_numeric_into(quals, &mut out)` reuses one scratch buffer across the per-record loop instead of allocating a fresh `Vec` per emitted record.

## Results

EC2 c8g.8xlarge (32 vCPU Graviton4), 40M read pairs (gzip R1/R2, `8M143T`/`151T`), 16 threads, cache warmed, warmup + 4 interleaved before/after runs:

| binary          | user+sys (s) | delta            |
|-----------------|-------------:|------------------|
| main (baseline) | ~204.8       | —                |
| this branch     | ~195.6       | **−9.2 (−4.5%)** |

Wall clock is flat because this input is gzip (single-threaded inflate) + output-compression bound — the win is in CPU-seconds, not wall clock. That is the metric that matters: `extract` typically runs piped alongside the aligner, so −9.2 CPU-seconds is handed straight back to alignment.

## Correctness

Byte-for-byte identical output, validated two ways:

- **Unit** — new `view_path_matches_owned_path_byte_for_byte` (rstest, 4 cases: single-end, paired-end, UMI-only, full structure with cell + sample + molecular barcodes under all optional tags) asserts the view path's `RawRecord` bytes equal the owned path's, and pins that the combined header is sourced from the first record (distinct R1/R2 names).
- **Guard parity** — new `segment_into_matches_owned_guard_classification` (rstest, 6 cases) pins that `FastqSetView::segment_into` and `FastqSet::from_record_with_structure` classify every error/skip input identically (length mismatch, over-long fixed, too-few-bail, too-few-skip, ok), so a future edit to one guard cannot silently diverge them.
- Existing `no_threads_matches_threaded*` cross-worker byte-identity tests unchanged.
- **EC2** — `fgumi compare bams --command extract` between main's output and this branch's = IDENTICAL across all 80,015,824 records (core fields + tag values).

## Gates

`cargo ci-fmt`, `cargo ci-lint`, `cargo ci-doc`, `cargo ci-doctest` clean; `cargo ci-test` — 10269 passed, 31 skipped.

## Reading order

Start with commit 1: `src/lib/fastq.rs` (the view types + `segment_into`), then `src/lib/commands/extract.rs` (`make_raw_records_from_view`, the shared core), then `src/lib/pipeline/steps/extract.rs` (`extract_batch` builds the view directly). Commit 2 (scratch buffer) is a small, self-contained follow-on.

Risk: command output changes none; byte-parity tests and large-record validation pin output. Unsafe changes none; CLAUDE.md needs no allowlist update. Memory bounds, queue capacity, and backpressure policy changes none.

Fix: use borrowed FASTQ views and a reusable Phred buffer to reduce per-record allocations.

- Production extraction uses `FastqSetView`.
- The owned path remains available for tests through shared record construction.
- Segmentation and error behavior remain unchanged.
- Tests cover single-end, paired-end, tagged inputs, and guard parity.
- Reported CPU time improved by approximately 4.5%.


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #969, opened 2026-09-16
perf(filter): single-pass scalar consensus-tag extraction (-9% CPU)

nh13:
## Summary

The read-level consensus filter walked a record's aux block ~9 times per record, each `find_*_tag` restarting at offset 0:

- `is_duplex_consensus` reads `aD` + `bD`
- `filter_read` reads `cD` + `cE`
- `filter_duplex_read` re-reads `cD` + `cE`, then `aD`/`aM`/`bD`/`bM`/`aE`/`bE`

`aD`/`bD` were scanned up to 3× per record, so the cost was `O(tags · aux_len)` per record. Profiling the duplex pipeline showed `find_tag_position` as a hot spot (8.35% self-time in standalone `filter`).

This introduces `ConsensusScalarTags::from_aux()` — a single `O(aux_len)` walk that collects all eight scalar consensus tags in one pass, mirroring the existing `extract_aux_string_tags` pattern (and in the same spirit as #964's single-pass zipper walks). New allocation-free `filter_read_tags` / `filter_duplex_read_tags` operate on the struct; the AB/BA tier comparison is factored into a shared `duplex_tier_result()` so the per-tag and single-pass paths cannot drift. The `filter` command hot path builds the struct once per record and routes classification + both filter decisions through it.

## Performance (measured, c8g.8xlarge / 32 vCPU Graviton4, t16)

Standalone `fgumi filter` over a 9,998,652-read duplex consensus BAM, 3 reps:

| | wall | CPU |
|---|---|---|
| base | 6.11 s | 89.2 s |
| this PR | 5.59 s | 80.9 s |
| **delta** | **−8.5%** | **−9.3%** |

`perf` self-time for `find_tag_position`: **8.35% → 3.36%** (~60% of the tag-scan cost removed; the residual is the per-base *array* tags in `mask_duplex_bases`, intentionally untouched).

## Correctness

Byte-identical to the per-tag path — output unchanged (kept 9,998,652 every run):

- `consensus_scalar_tags_match_find` asserts the single-pass struct + `_tags` filters equal the per-tag `find_*` results across 7 tag combinations (simplex, full duplex, `aM`/`bM` fallback, missing pairs, absent-`cD`, interleaved/out-of-order tags).
- 91 filter unit + 48 filter integration + the command-level `check_filters_raw` tests pass; clippy clean.

## Notes

The same single-pass technique does **not** apply to the duplex/simplex/codec callers: those hot paths read exactly one tag per record (MI in strand partitioning, RX per source read), so there is no multi-tag re-scan to collapse. This PR is scoped to `filter`, where the redundancy actually was.

Risk: output changes — none; `unsafe` changes — none, with no CLAUDE.md allowlist update required; memory, queue, and thread/backpressure policy changes — none.

Fix: parse the eight scalar consensus tags once per record and reuse them for classification and filtering.

- Added `ConsensusScalarTags::from_aux()` with matching first-occurrence, presence, and malformed-aux semantics.
- Added allocation-free `filter_read_tags()` and `filter_duplex_read_tags()`.
- Reused parsed tags in the `filter` command.
- Added shared duplex tier evaluation.
- Added parity tests for missing, malformed, duplicated, and out-of-order tags.
- Reported benchmarks show 8.5% lower wall time and 9.3% lower CPU time.
- Reported unit, integration, command-level, and clippy checks pass.


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #971, opened 2026-09-17
perf(zipper): reduce serial-merge CPU and allocations at 1-2 threads

nh13:
## Summary

Optimizes `fgumi zipper`'s serial merge — the path it runs in production, at 1–2 threads between the aligner and the sorter, where both CPU efficiency and wall time matter. Three changes, all byte-identical to the previous output.

**Risk verdict:** (1) Output — none; byte-identical across commands, verified with `fgumi compare bams` (IDENTICAL) and a per-record on-disk-integer-width check over 60,116,876 records. (2) `unsafe` — none added. (3) Memory / thread / backpressure policy — unchanged; a single aux-rebuild scratch buffer is now reused across templates rather than reallocated per record, so steady-state memory stays a function of record size, not input size.

## What changed

1. **Fold AS/XS normalization into the tag-copy rebuild.** The old path normalized the AS/XS alignment-score tags to the smallest signed width in a dedicated per-record pass that ran roughly four linear tag scans plus two aux memmoves and two appends for every mapped record. A new `RawTagsEditor::rebuild_with_int_normalized` performs that re-encoding during the single aux rebuild the tag-copy step already does, so it costs no extra scan or memmove. It is byte-identical to `rebuild_with` followed by `normalize_int_tag_to_smallest_signed` per tag — including first-key-occurrence semantics on degenerate duplicate-key aux — and is covered by oracle and property tests.

2. **Reuse the aux-rebuild buffer.** A scratch `Vec` is threaded through `process_raw` and `merge_one_template_with` so its allocation is reused across templates on the serial merge thread and across the align-and-merge / ZipperMerge pipeline loops.

3. **Stringify the mate CIGAR without per-op allocations.** `Template::fix_mate_info` set the `MC` tag via `cigar_to_string_from_raw`, which allocates an intermediate `Vec<u32>` of ops plus one `String` per CIGAR op. The new `cigar_to_bytes_into` formats directly into a caller-provided buffer with no per-op allocation. `fix_mate_info` is shared infrastructure, so group, consensus and dedup benefit too.

## Performance

Synthetic 60.1M-record set (chr17), at the production thread counts:

| threads | wall | CPU |
| --- | --- | --- |
| 2 | −18% | −9% |
| 1 | −3% | −3% |

zipper is I/O-codec-bound at these thread counts (input BGZF decompression dominates), so the merge-side win is single-digit at one thread and larger at two, where the serial merge thread is the wall-gating stage.

## Validation

- `fgumi compare bams`: IDENTICAL (0 content diffs, 60,116,876 records) against the pre-change binary.
- Per-record check: 0 field / tag-value / on-disk-width diffs across all records.
- Full workspace suite green: clippy `-D warnings -W pedantic`, fmt, nextest plus doctests.
- Reviewed with a CodeRabbit-style pass and a multi-lens adversarial review; findings applied — duplicate-key first-occurrence parity, empty-adds and spill/edge-case test coverage, scratch threading to the pipeline paths, a debug-build coverage assert, and doc accuracy.

## Follow-on (not in this PR)

Scaling zipper beyond ~2 threads is a separate change: decompose the merge into a chain pipeline (parallel decode of the two inputs, a serial zip step, a parallel per-template merge step, parallel compress) so higher `--threads` fan out without raising per-record CPU. Left out deliberately — it trades CPU for wall and only helps the 4+ thread regime.

Risk: No changes to grouping, consensus, sort order, corrected UMIs, or metrics output; byte-identical output was verified across 60,116,876 records. No `unsafe` changes; no `CLAUDE.md` allowlist update. No memory-bound, queue-capacity, thread, or backpressure policy changes.

Fix: Reuse zipper scratch buffers and format mate CIGAR values without intermediate allocations.

- Fuse AS/XS normalization with auxiliary-tag rebuilding.
- Reuse scratch storage across templates and pipeline batches.
- Add and re-export `cigar_to_bytes_into`.
- Preserve MC-tag behavior and malformed or truncated CIGAR handling.
- Add tests for normalization, malformed tags, duplicate tags, overflow, and truncated CIGARs.

Validation reports no on-disk differences and a passing workspace test suite. At two threads, wall time decreased 18% and CPU usage decreased 9%. At one thread, both decreased 3%.


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #973, opened 2026-09-18
perf(consensus): gated multi-base fast path for simplex consensus

nh13:
Adds a second consensus fast path, `ConsensusBaseBuilder::try_multibase_fast_path`, for positions with two or more observed bases — the case the existing unanimous fast path defers and that otherwise falls through to the full log-sum-exp `call_full`. `call_full` is untouched and remains the fgbio-faithful oracle; the new path is inserted into `call()` between `try_unanimous_fast_path` and `call_full`.

## What it does

It computes the posterior the max-factored way, from the **same** `likelihoods` array `call_full` consumes:

- `r = sum over k != max of e^(L[k] - L[max])` (same-sign exponents; unobserved lanes share one exponent)
- `ln_consensus_error = ln(r) - log1p(r)`

replacing `call_full`'s `ln_sum_exp_array` (three sequential `log1pexp` calls, each a branch ladder) plus `ln_not` with at most three scalar `exp`s, one `ln`, and one `log1p`.

## Why it is safe (same-Phred gate)

The max-factored reordering equals `call_full` in exact arithmetic but is not bit-identical to its min-first `log1pexp` fold, so the two can round to different integer Phreds — empirically only at deep (`|L|` >= ~1600), near-zero-gap pileups where `call_full` itself loses a bit to `fl(l + S)` cancellation. Rather than reproduce that, the path returns `Some` **only when the Phred is stable across a magnitude-scaled error band** `+/- delta` around `ln_final_error` (`delta = UNANIMOUS_MARGIN_HEADROOM * u * max_magnitude / log1p(r)`; the `1/log1p(r)` factor captures `ln_not`'s error amplification for confident calls). Near a bracket boundary it returns `None` and defers to the `call_full` oracle. This mirrors the discipline of the existing unanimous fast path, which is likewise conditionally exact and defers to `call_full`.

Output is therefore **byte-identical** to before: verified with `fgumi compare bams` == IDENTICAL, 0 content diffs over all 20,000,000 consensus records of a 60M-read run, and the existing simplex/duplex/codec integration tests (which pin consensus output to fgbio) pass unchanged.

## Measurement

-7.3% on the SimplexConsensus step and -5.2% end-to-end at one thread on a 60M-read, 1%-error grouped BAM (Graviton4). The unanimous fast path (a table lookup) is untouched, so high-quality / unanimous positions are unaffected; this speeds up only the multi-base positions that previously ran `call_full`.

## Testing

- `test_multibase_fast_path_matches_call_full_exhaustive`: every reachable position up to depth 4 across the pre-UMI x post-UMI x quality grid must return exactly what `call_full` returns.
- `test_multibase_fast_path_matches_call_full_deep`: contiguous deep scans (to depth ~2600, past the first divergence at ~1606) of unanimous and genuine 2/3/4-base pileups, across the default and extreme rates, moderate and low qualities, and all four winner lanes — asserting the gate defers wherever it cannot prove the Phred.
- `test_multibase_fast_path_is_taken_on_shallow_positions`: liveness — counts how many shallow multi-base positions the fast path actually handles and asserts a floor, so a regression to always-deferring is caught even though the parity tests would still pass.

Risk: consensus output is pinned by the stability gate and `call_full` fallback; no `unsafe` changes or CLAUDE.md allowlist update; no memory, queue, thread, or backpressure changes.

Fix: add the gated `try_multibase_fast_path` with exhaustive parity, deep-scan, and liveness tests. Reported validation shows byte-identical output across 20 million records and a 7.3% `SimplexConsensus` performance improvement.


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #976, opened 2026-09-18
perf(decode): fold UMI-position cache into the group-key aux scan

nh13:
## What

The decode step scanned each record's aux block **twice** on the group/dedup/consensus hot path: once in `compute_group_key_from_raw` (via `extract_aux_string_tags`) for the RG/cell/MC group key, then again in `cache_umi_position` (via `find_string_tag_position`) for the UMI value position (#334). The single-pass extractor already accepts a `umi_tag` and returns the UMI position, so the second walk is redundant.

This folds the UMI-position capture into the same aux scan that builds the key:

- New `compute_group_key_and_umi_from_raw` / `key_and_umi_for_mode` pass the UMI tag into the one `extract_aux_string_tags` call and return the record-relative `(offset, len)`.
- The BAM (`DecodeRecords`, `DecodeFromRecords`) and SAM (`ParseSamChunk`) decode paths use these via a new `apply_cached_umi` helper, and fall back to the standalone `cache_umi_position` scan only when the key was built without reading aux data (`KeyMode::None`/`NameHashOnly`, or a name-only key path). `compute_group_key_from_raw` / `key_for_mode` are kept as thin wrappers.

Because the fused pipeline decodes once, this also trims `runall`'s single decode, not just the standalone commands.

## Tag-resolution semantics (byte-identity scope)

The UMI is now resolved by the same `extract_aux_string_tags` pass as RG/cell/MC, so the cached UMI is consistent with the key's own tag resolution. For **spec-legal** records (each aux tag present at most once, SAM §1.5) this yields the identical position the standalone `find_string_tag_position` scan produced. The two deliberately differ only on **malformed** duplicate/type-shadowed tags — the same already-accepted divergence the `MC` tag carries via `validate_mc_tag` (`src/lib/grouper.rs`): `extract_aux_string_tags` skips a non-`Z` entry and takes a later `Z` copy, whereas `find_tag_position` resolves the first id match and rejects it if non-`Z`. The relaxation makes the cached value agree with the value the grouping key actually uses.

## Verification

- **Byte-identical output** on well-formed input: `fgumi compare bams` reports `Content diffs: 0` / IDENTICAL for **both group and dedup** over a real 60M-record benchmark BAM.
- Existing SAM-vs-BAM decode parity, group/dedup MI-determinism (threads 1/4), and `runall` staged-vs-fused tests pass unchanged; new parity tests assert the folded capture equals the standalone scan on the primary, single-end, and `tc`-stamped secondary/supplementary paths, plus the `apply_cached_umi` fallback branch.
- `cargo clippy --all-targets --all-features -- -D warnings -W clippy::pedantic`, `cargo fmt`, and rustdoc all clean.

## Measurement

At `t=1` on a 60M-read, 1%-error grouped BAM (Graviton4), CPU-seconds, 2 reps:

| command | before | after | Δ |
|---|---|---|---|
| group | 126.1 | 123.6 | **−2.0%** |
| dedup | 139.6 | 137.6 | **−1.4%** |

The removed work is the per-record `cache_umi_position` aux walk; it is part of the shared decode path, so simplex/duplex/codec and `runall`'s single decode benefit as well.

Risk: output changes — none for well-formed input, pinned by parity tests and existing tag-resolution behavior; unsafe changes — none, so no `CLAUDE.md` allowlist update is needed; memory bounds, queue capacity, and thread/backpressure policy changes — none.

Fix: capture the UMI position during the existing grouping-key auxiliary-tag scan and reuse it during BAM and SAM decoding.

- Added and re-exported `key_and_umi_for_mode`.
- Added fallback scanning when the key mode does not capture auxiliary data.
- Preserved key-only APIs and existing grouping fallbacks.
- Covered primary, paired, and `tc` secondary/supplementary records.
- Added parity tests for standalone UMI scanning and fallback behavior.


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

